PR #21, opened 2015-04-14
seems like the proper variable name here is {{site.data.kong_latest}}

ahmadnassri:
or, can change the plugin to output to `{{site.latest}}`, but that felt wrong.


ahmadnassri:
@rainum @thibaultCha

PR #28, opened 2015-04-17
[docs] add osx to installation methods

thibaultCha:
:warning: Not yet ready.

- [x] OSX logo
- [ ] .pkg instructions
- [x] Homebrew instructions
- [ ] script instructions

This just present each installation method. Although I am not sure we want to repeat steps 2 and 3 every time...

screenshot 2015-04-17 03 30 57


ahmadnassri:
is this mavericks specific setup?

should indicate such if that's the case, in both logo + description

thibaultCha:
> is this mavericks specific setup?

no

> do you want to support older versions of osx?

we do, I don't know how far

thibaultCha:
@thefosk I need infos on where to find the .pkg. Will it be uploaded on GH releases?

@Nijikokun Same for the OS X install script

Thanks!

PR #31, opened 2015-04-20
[docs] cli reference

thibaultCha:


app/docs/0.1.1beta-2/cli.md
```diff
@@ -0,0 +1,132 @@
+---
+title: CLI Reference
+---
+
+# CLI Reference
+
+Kong comes with a ***Command Line Interface*** (refered to as "CLI") which lets you perform operations such as starting and stopping a node. Each command is run in the context of a single node, since Kong as no cluster awareness yet.
```
ahmadnassri: `(refered to as "CLI")` => `*(CLI)*`

app/docs/0.1.1beta-2/cli.md
```diff
@@ -0,0 +1,132 @@
...
+
+Almost every CLI command needs access to your configuration file in order to be aware of what is your node's NGiNX working directory, (known as the "prefix path" for those familiar with NGiNX), and referenced as `nginx_working_dir` in your Kong configuration file.
```
ahmadnassri: `NGiNX` is always branded as `NGINX` on their website, lets keep to that.

thibaultCha: I've seen other places in the website where it is "NGiNX", so I was only trying to be consistent. This needs to be updated everywhere if still present.

app/docs/0.1.1beta-2/cli.md
```diff
@@ -0,0 +1,132 @@
...
+
+**Note:** If you haven't already, we recommand you to read the [configuration guide][configuration-guide].
+
+---
+
+## kong
+
+The CLI's entry point is the `kong` command. Print its help message by invoking:
```
ahmadnassri:
```
The CLI's entry point is the `kong` command. 

View the usage info by typing:
``

File: app/docs/0.1.1beta-2/cli.md [hunk 4]
```diff
@@ -0,0 +1,132 @@
...
+$ kong --help
+```
+
+Like many other CLI tools, the second parameter will be the command you want to perform:
+
+```bash
+$ kong <command>
+```
+
+To print the help message of any command, execute:
```
ahmadnassri: `usage info`

Nijikokun:
> ("CLI")

Wouldn't require quotes, but :+1:

app/docs/0.1.1beta-2/cli.md
```diff
@@ -0,0 +1,132 @@
...
+
+```bash
+$ kong restart [-c path_to_config]
+```
+
+---
+
+## reload
+
+Very handy, this command will allow you to change your configuration at runtime, without any downtime. It takes advantage of NGiNX's [reload](http://wiki.nginx.org/CommandLine#Loading_a_New_Configuration_Using_Signals) signal to spawn new workers with the new configuration, while killing the old ones.
```
Nijikokun: - Not sure about the `very handy` statement, is it required?
- Avoid repetitive usage of `This command`, instead do something like so:

  Reloads the NGINX configuration at runtime and avoids potential downtime by leveraging the NGINX [reload](http://wiki.nginx.org/CommandLine#Loading_a_New_Configuration_Using_Signals) signal.

thibaultCha:
Everything updated

app/docs/0.1.1beta-2/cli.md
```diff
@@ -0,0 +1,132 @@
...
+
+## stop
+
+This command executes a fast shutdown of Kong. It is actually a wrapper aroung the NGiNX `stop` signal:
+
+```bash
+$ kong stop
+```
+
+In order to stop an instance, the CLI needs to know what is the working directory of your process. If you started Kong with a certain configuration file, you must stop it with the same one. Thus, the stop command also accepts a configuration option:
```
Nijikokun: - Being direct might be better here:

  To stop a specific instance, the CLI requires knowledge of the instances working directory. You can do this referencing the configuration file you used to start your Kong instance:

app/docs/0.1.1beta-2/cli.md
```diff
@@ -0,0 +1,132 @@
...
+Kong is now running and listening on two ports, which are by default:
+
+- `8000`, that will be used to process the API requests.
+- `8001`, called admin port, provides the Kong's internal RESTful API that you can use to operate Kong, and should be private and firewalled.
+
+---
+
+## stop
+
+This command executes a fast shutdown of Kong. It is actually a wrapper aroung the NGiNX `stop` signal:
```
Nijikokun: - More direct approach might be better here

  Terminates a Kong instance by firing the NGINX `stop` signal:

app/docs/0.1.1beta-2/cli.md
```diff
@@ -0,0 +1,132 @@
...
+Kong is now running and listening on two ports, which are by default:
+
+- `8000`, that will be used to process the API requests.
+- `8001`, called admin port which provides the Kong's internal RESTful API that you can use to operate Kong.
+
+---
+
+## stop
+
+This command executes a fast shutdown of Kong. It is actually a wrapper aroung the NGINX `stop` signal:
```
Nijikokun: - More direct approach might be better here

  Terminates a Kong instance by firing the NGINX `stop` signal:

app/docs/0.1.1beta-2/cli.md
```diff
@@ -0,0 +1,132 @@
...
+
+```bash
+$ kong <command> --help
+```
+
+---
+
+## start
+
+The first command you are likely to execute will be `start`. As you would expect, it starts a Kong node on your current machine using a configuration file.
```
Nijikokun: - Being more direct would be better, CLI reference is less of a guide and more of a reference.

  Initializes a Kong (instance / node), optionally a configuration file may be passed:

  ```bash
  $ kong start [options]
  ```

thibaultCha: Well if it's just this then better only rely on the usage info of each command then, no need for a web version.

thibaultCha:
pushed some updates

PR #45, opened 2015-04-25
[feature] configuration reference

thibaultCha:
Here is the Kong configuration reference. I tried to be as inspired as possible for the intro text but well, it wasn't that efficient...

Also I am not sure about the side menu in `docs.html`.

@Nijikokun Please review, rephrase, reword :)

@thefosk I still don't know what `nginx_plus_status` really does. Could you please update the PR or explain here.

**Depends on Mashape/kong#162 being merged.**

### DIFF
--- a/app/_layouts/docs.html
+++ b/app/_layouts/docs.html
@@ -33,6 +33,13 @@ header_caption: Learn how Kong empowers your APIs
     </nav>
 
     <nav>
+      <h5>Configuration</h5>
+      <ul>
+        <li><a href="/docs/{{page.kong_version}}/configuration">Configuration reference</a></li>
+      </ul>
+    </nav>
+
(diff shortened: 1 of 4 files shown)


montanaflynn:
Can we fix #48 in this PR?

Nijikokun:
Alright, on reviewing it, and yes @montanaflynn this will fix #48 

For the time being @thibaultCha `nginx_plus_status` does this if you were wondering: http://getkong.org/plugins/nginx-plus-monitoring/

thibaultCha:
Yeah I know it's for the plugin, but I don't get why it's needed exactly.

Nijikokun:
> No this is correct. Kong.yml is not

My apologies, I was going by kong.yml since there was no example

PR #67, opened 2015-05-08
Changed the needed `name` parameter in filelog pg

therebelrobot:
Looks like it was copied from the tcplog page.


thibaultCha:
thanks

Issue #12, opened 2015-04-08
docs.html layout should be versioned


thibaultCha:
[_layouts/docs.html](https://github.com/Mashape/getkong.org/blob/master/app/_layouts/docs.html) is what is on the left side of a documentation page. Not every version to document will have the same articles (in fact, the 2 versions we currently have already differ, resulting in a few 404s), and in the future there might be a lot more to it.

To dynamically generate it, we can either have a list of all articles in [kong_versions.yml](https://github.com/Mashape/getkong.org/blob/master/app/_data/kong_versions.yml), or we could dynamically generate that list via the [versions.rb](https://github.com/Mashape/getkong.org/blob/master/app/_jekyll_plugins/versions.rb) plugin. One is more flexible, the other is less to maintain.

Then `docs.html` simply becomes:

```html
<aside class="page-navigation">
  {% for section in page.doc_link %}
  <nav>
    <h5>{{section.title}}</h5>
    <ul>
      {% for article in section.articles %}
        <li href="/docs/{{page.kong_version}}/{{article.link}}">{{article.name}}</li>
      {% endfor %}      
    </ul>
  </nav>
  {% endfor %}
</aside>
```

Which solution do we like better: YAML or automated?

thibaultCha:
@ahmadnassri @montanaflynn @thefosk @Nijikokun

ahmadnassri:
best to keep in `kong_versions.yml`

reasons:

1. we limit to "supported" versions only (think enterprise LTS)
2. jekyll plugins make updating the site manual, whereas you can just leave it for gh-pages

Nijikokun:
> jekyll plugins make updating the site manual, whereas you can just leave it for gh-pages

could just build on travis

ahmadnassri:
> its already manual...

if that's the case already, then no change, other point still stands to greater value I think

Nijikokun:
> if that's the case already, then no change, other point still stands to greater value I think

We could have have a list of "non-supported" versions for that, easier to maintain that list than a list of every version **plus** their pages.

ahmadnassri:
> We could have have a list of "non-supported" versions for that, easier to maintain that list than a list of every version plus their pages.

wouldn't you just want to publish ONLY the supported pages on the website? 

scenario:
when people bookmark an old version docs url, they come back and its 404, which is *good* as that's feedback for them to know its no longer supported, and start looking at new docs, realize update is needed.

Nijikokun:
[...]

PR #68, opened 2015-05-11
Docs versioning

rainum:
#12

### DIFF
--- a/gulpfile.js
+++ b/gulpfile.js
@@ -39,7 +39,7 @@ gulp.task('styles', function () {
   return gulp.src('app/_assets/stylesheets/index.less')
     .pipe($.plumber())
     .pipe($.less())
-    .pipe($.uncss({ html: glob.sync('dist/**/*.html') }))
+    //.pipe($.uncss({ html: glob.sync('dist/**/*.html') }))
     .pipe($.autoprefixer())
     .pipe($.minifyCss())
     .pipe($.rename('styles.css'))
--- a/app/_assets/javascripts/dropdown.js
+++ b/app/_assets/javascripts/dropdown.js
@@ -1,0 +1,165 @@
+/* ========================================================================
+ * Bootstrap: dropdown.js v3.3.4
+ * http://getbootstrap.com/javascript/#dropdowns
+ * ========================================================================
+ * Copyright 2011-2015 Twitter, Inc.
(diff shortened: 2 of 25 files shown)


gulpfile.js
```diff
@@ -39,7 +39,7 @@ gulp.task('styles', function () {
   return gulp.src('app/_assets/stylesheets/index.less')
     .pipe($.plumber())
     .pipe($.less())
-    .pipe($.uncss({ html: glob.sync('dist/**/*.html') }))
+    //.pipe($.uncss({ html: glob.sync('dist/**/*.html') }))
```
thibaultCha: This is very important, why removing it?

orliesaurus: https://github.com/Mashape/getkong.org/commit/14e2f69d643434f11e57cd3a8456ad11b612f354

thibaultCha:
```
options = {
  ignore: ['#added_at_runtime', /test\-[0-9]+/],
};
```

thefosk:
@rainum can we close this?

Nijikokun:
This removes the `latest` alias that we use to refer to the latest documentation.

Pinging @montanaflynn and @rainum you guys can discuss this further as @rainum removed the functionality.

thibaultCha:
It shouldn't, according to the code it's added by the Ruby plugin.

Nijikokun:
> It shouldn't, according to the code it's added by the Ruby plugin.

Ah I missed this section. Looks good otherwise.

thibaultCha:
Well except for the step back removing uncss.

Nijikokun:
> Well except for the step back removing uncss.

It's a small thing, just need @rainum to add ignore option for these classes.

thibaultCha:
It actually doesn't work. I see no dropdown even with 2 versions and the `latest` redirection is gone, there is only the one on `/docs/`

Nijikokun:
> It actually doesn't work.

I was fearful of this, when I first implemented the `alias` functionality I was going to do it automatically in this file but was afraid it would not work. *sigh*

montanaflynn:
Wait, what's going on? Don't deploy this if it breaks all the `/latest/` links!

thibaultCha:
It doesn't because they were removed from 0.2.0-2 but they are still in 0.2.1.

PR #84, opened 2015-05-27
Cleanup

ahmadnassri:
- remove unused pages & css 60c3d44, b3c1c65, 1093753
- use dropdown.js from bootstrap as bower dependency  16645c2
- package.json updates 493c857
- cleaner gitignore b70a384
- layout is not needed when not using front matter, use null value otherwise b1bf3d6, a6c0e04
- install bundler in instructions d76a012
- cleaner cache busting variable  be45e7c
- clean alert corners 42db146
- apply standard formatting for javascript  8af5151
- update company name in license 9259e3f
- move share.png  7c3cf48
- better leverage jekyll config defaults + fixed sitemaps 8040f12
- enhanced 3rd party javascript widgets 96490bf
- changes in master from @rainum 647ae5d, fdc40a6


rainum:
looks great :+1: Good work, @ahmadnassri!

gulpfile.js
```diff
@@ -20,13 +23,12 @@ var jekyllConfig = jekyllConfigs.LOCAL
...
   content: 'app/**/*.{markdown,md,html,txt,yml,yaml}',
   styles: 'app/_assets/stylesheets/**/*.{less,css}',
-  js: 'app/_assets/javascripts/**/*.js',
+  js: ['app/_assets/javascripts/**/*.js', 'bower_components/bootstrap/js/dropdown.js'],
   images: 'app/_assets/images/**/*',
   fonts: 'bower_components/font-awesome/fonts/**/*.*'
 }
 
 gulp.task('styles', function () {
-  var glob = require('glob')
```
thibaultCha: Why changing the style of an established project just to make it to your own sauce? It was like that, it is nicer because those are local dependencies, respect it and move on.

app/sitemap.txt
```diff
@@ -0,0 +1,7 @@
+---
```
thibaultCha: It's useless to have both hence why it was removed.

app/sitemap.xml
```diff
@@ -1,14 +1,15 @@
...
 ---
 <?xml version="1.0" encoding="utf-8"?>
 <urlset xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://www.sitemaps.org/schemas/sitemap/0.9 http://www.sitemaps.org/schemas/sitemap/0.9/sitemap.xsd" xmlns="http://www.sitemaps.org/schemas/sitemap/0.9">
-{% for page in site.pages %}{% if page.sitemap != null and page.sitemap != false %}
+{% for page in site.pages %}{% if page.sitemap != false %}
   <url>
     <loc>{{ site.url }}{{ page.url }}</loc>
     <lastmod>{{ site.time | date_to_xmlschema}}</lastmod>
     <changefreq>weekly</changefreq>
+    <priority>1.0</priority>
```
thibaultCha: Same I think this is useless. All the pages are at the same priority.

gulpfile.js
ahmadnassri: not to fit my sauce, it looked weird, like it was done after original changes (like `connect` below) so I thought it was a mistake ... easily reverted.

app/sitemap.txt
ahmadnassri: I've found that some secondary search engines are more successful with `txt` version (Bing, Yahoo, etc ...) be good to keep both.

app/sitemap.xml
ahmadnassri: correct, I added it because I wanted to add a lower priority for older doc versions, didn't get the logic working quite right yet ... but added the priority tag for now.

rainum:
Guys, I think we should merge it, because we need cache busting asap, what do you think?
@ahmadnassri @thibaultCha

ahmadnassri:
I'm fine with it on my end, unless @thibaultcha *really* wants those `require` back where they were ... which is no big deal, I can do it later today

ahmadnassri:
the sitemap prioritizing will take more time and perhaps a custom plugin, but it's already this way now so no harm in merging

app/sitemap.xml
thibaultCha: Docs priority in #90

PR #92, opened 2015-06-13
update rate limintg

sinzone:
shorted discription


montanaflynn:
"Incoming requests" is redundant

> Block incoming requests whose body is greater than a specific size

could just be 

> Block requests whose body is greater than a specific size

which fits in two lines: 

screen shot 2015-06-14 at 9 33 24 pm

ahmadnassri:
fixed in 2312f7773a6223e3927c789d0405bcefcd20fd18

PR #93, opened 2015-06-13
Revert "update rate limintg"

ahmadnassri:
Reverts Mashape/getkong.org#92


sinzone:
wrong. too long.

ahmadnassri:
the language was wrong.

PR #95, opened 2015-06-16
(Fix admin) api's response - previous node omitted

ajayk:


thibaultCha:
Sorry, I already updated it, as well as in the old versions of the documentation

PR #119, opened 2015-08-13
major.minor.x

ahmadnassri:
feat(versioning): use major.minor.x for groupping patch versions, decreasing the amount of work needed for releasing a patch


jekyll.yml
```diff
@@ -2,9 +2,6 @@
 name: Kong - Open-Source API and Microservice Management Layer
 url: https://getkong.org
 share_url: http://getkong.org
-repo: https://github.com/Mashape/kong
-docrepo: https://github.com/Mashape/getkong.org
```
ahmadnassri: this wasn't used anywhere

jekyll.yml
```diff
@@ -2,9 +2,6 @@
 name: Kong - Open-Source API and Microservice Management Layer
 url: https://getkong.org
 share_url: http://getkong.org
-repo: https://github.com/Mashape/kong
```
ahmadnassri: was only used once

jekyll.yml
```diff
@@ -2,9 +2,6 @@
...
-docrepo: https://github.com/Mashape/getkong.org
-documentation: docs
```
ahmadnassri: seemed redundant for this to be a variable.

app/_jekyll_plugins/versions.rb
```diff
@@ -11,7 +11,7 @@ def generate(site)
       # Add a `version` property to every versioned page
       site.pages.each do |page|
         parts = Pathname(page.path).each_filename.to_a
-        if parts[0] == site.config["documentation"]
+        if parts[0] == "docs"
```
thibaultCha: was nicer to have it as a variable

jekyll.yml
thibaultCha: still nice to have as a variable

jekyll.yml
thibaultCha: was used before, could be used again

jekyll.yml
thibaultCha: might be renamed one day

thibaultCha:
Ok for the .x renaming

Not ok for removing the variables (except the unused one eventually, but I feel like there should be a link on the website pointing to this repo, so it shouldn't be removed but rather used). Those were all are variables in case of quick renaming or moving of the project.

thibaultCha:
By the way releasing a patch version will still require an update of the `luarocks_version` property anyways.

app/_jekyll_plugins/versions.rb
ahmadnassri: when I actually removed the variable from the `yml` file it broke, not exactly sure why, so I brought it back there, but forgot about this spot.

jekyll.yml
ahmadnassri: we can bring it back when needed

jekyll.yml
ahmadnassri: and risk breaking all our SEO? unlikely ...

ahmadnassri:
> By the way releasing a patch version will still require an update of the luarocks_version property anyways.

yep, this just eliminates the need to redo the whole folder...

ahmadnassri:
> but I feel like there should be a link on the website pointing to this repo, so it shouldn't be removed but rather used

I had to double check to make sure we do link to the repo somewhere ... and we do (phew!) but seems like the variable itself was never used...

so yeah, i do agree, we should keep the variable, and update any existing reference to use it.

ahmadnassri:
@thibaultCha cleaned up a bit, please take another look

PR #120, opened 2015-08-13
fix(gulp) only require cloudflare in its own task

thibaultCha:
Otherwise simply running `gulp` for local work would throw an error
because of the missing environment variables, even if they are only
required when deploying.


ahmadnassri:
very good! :+1:

PR #148, opened 2015-10-20
Added clock skew details for hmac plugin

shashiranjan84:
added details about `Clock Skew` in HMAC plugin doc

### DIFF
--- a/app/plugins/hmac-authentication.md
+++ b/app/plugins/hmac-authentication.md
@@ -130,6 +130,12 @@ the order they appear.
 3. If value is not the last value then append an ASCII newline `\n`.
   The string MUST NOT include a trailing ASCII newline.
 
+### Clock Skew
+
+The HMAC Authentication plugin also implements a clock skew check [as described in the specification](https://tools.ietf.org/html/draft-cavage-http-signatures-00#section-3.4) to prevent replay attacks. By default, a minimum lag of 300s in either direction (past/future) is allowed. Any request with a higher or lower date value will be rejected. The length of the clock skew can be edited through the plugin's configuration by setting `clock_skew` property (`config.clock_skew` POST parameters).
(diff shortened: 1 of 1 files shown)


app/plugins/hmac-authentication.md
```diff
@@ -129,6 +129,12 @@ the order they appear.
 
 3. If value is not the last value then append an ASCII newline `\n`.
   The string MUST NOT include a trailing ASCII newline.
+  
+### Clock Skew
+
+Plugin also implements the clock-skew to prevent Replay Attack. The server and 
+requesting machine should be synchronized with NTP and a valid date in GMT format should 
+be sent using either `X-Date` or `Date` header.
```
ahmadnassri: you can also implement `clock-skew` from the specification to prevent replay attacks. 

The server and requesting machine should be synchronized with NTP and a valid date in GMT format should be sent using either `X-Date` or `Date` headers.

ahmadnassri:
@shashiranjan84 provided a better worded paragraph

thibaultCha:
I like @ahmadnassri's paragraph better too.

shashiranjan84:
I have already updated it based on my conversation with him. I can't use his exact lines as its not correct. Clock skew is not something user can implement, its already implemented as a mandatory check.

ahmadnassri:
but you still use the wording "Plugin also ..." it just sounds wrong.

thibaultCha:
"Plugin also implements" => weird english

By default a minimum lag of `300s` allowed ===> missing an *is*? weird english too 

Plugin configuration ===> plugin's configuration

 by setting the property `config.clock_skew` ==> by setting the `config.clock_skew` proeprty

either `X-Date` or `Date` header => either `X-Date` or `Date` headers

PR #150, opened 2015-10-26
fix(layout) Remove the gh-star bait button

thibaultCha:
Finally remove this bait button, is 3000 stars enough?

### DIFF
--- a/app/_includes/nav.html
+++ b/app/_includes/nav.html
@@ -1,7 +1,6 @@
 <div class="navbar">
   <div class="container">
     <a href="/" class="navbar-brand"><img src="/assets/images/branding.svg" alt="Kong" width="99" height="32"></a>
-    <iframe class="header-star" src="https://ghbtns.com/github-btn.html?user=Mashape&amp;repo=kong&amp;type=star"frameborder="0" scrolling="0" width="55" height="20"></iframe>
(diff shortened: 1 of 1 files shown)


ahmadnassri:
@sinzone ?

sinzone:
not yet. Devs use it as quick bookmark to view it later when they have time. it's very useful. 

![] (http://cl.ly/image/2c0x3Q3a3Z2m/Image%202015-10-26%20at%2010.45.06%20PM.png)

thibaultCha:
The issue here is the it is next to the actual logo and screams "CLICK ME".

The screenshot you posted is a good example to follow (unlike Kong): it's non intrusive at the bottom. Which we have on the website but this removes the bait button from the header. 

On a side note, it doesn't actually star the repo, it is just a link to it.

sinzone:
yes that's what i say.. won't auto follow. 

i agree that next to logo is weird, but till we have a new design/found a prominent place (that one in the example is right after install "main call to action" in the center) than don't remove it.

has to be somewhere in this area ![] (http://cl.ly/image/1P0P0H314109/Image%202015-10-26%20at%2011.50.09%20PM.png). do you have ideas?

sinzone:
can be in the yellow notifo area perhaps

ahmadnassri:
closing, since removal is not the desired result here.

PR #151, opened 2015-10-26
fix(layout) footer overlapping content since

thibaultCha:
Fix for a66946e9d181b9caf08d2f5a84de46ca7db3d120 making the footer overlap content.


ahmadnassri:
we removed the banner, so this wasn't needed

thibaultCha:
Now it’s future proof.

PR #171, opened 2015-12-02
typo fix, and retrieve plugin updated to be a GET request

Tieske:
a small typo, and an inconsistency.

Updated based on doc inconsistency, without checking the code! So __please check carefully before merging__.

The "Retrieve Plugin" was listed as a POST, was missing a header and included a "Request body" block.

### DIFF
--- a/app/docs/0.5.x/admin-api.md
+++ b/app/docs/0.5.x/admin-api.md
@@ -343,7 +343,7 @@ HTTP 204 NO CONTENT
 
 ## Consumer Object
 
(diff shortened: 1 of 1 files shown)


thibaultCha:
Yep, that sounds about right, nice

thibaultCha:
(this should also be fixed for 0.4, 0.3, and 0.2 versions. As well as updating the docs_nav versions in `app/_data` since the order of the endpoints was changed here. I'll take care of that.)

thibaultCha:
Correction: this was only necessary on 0.5.x, so we're good for the other versions.

ahmadnassri:
deployed.

PR #180, opened 2015-12-10
slightly better way to end that sentence

ahmadnassri:


thibaultCha:
Mhhhh nope, that sentence was here to explain that it can be a rocks living in it's own repo and installed by Luarocks iirc.

ahmadnassri:
yeah, that part i didn't change, just added cuz it seemed like the sentence was unfinished.

PR #183, opened 2015-12-14
typo, and removed Lua 5.1 dependency. 

Tieske:
See PR https://github.com/Mashape/kong/pull/799

### DIFF
--- a/app/install/source.md
+++ b/app/install/source.md
@@ -18,8 +18,6 @@ redirect_from: /install/compile/
 
     Install [Lua v{{lua_version}}](http://www.lua.org/versions.html#5.1)
 
-    Install [Luarocks v{{luarocks_version}}](https://github.com/keplerproject/luarocks/wiki/Download)
-
     Install [Dnsmasq](http://www.thekelleys.org.uk/dnsmasq/)
 
     Install [OpenSSL](https://www.openssl.org/)
@@ -35,6 +33,8 @@ redirect_from: /install/compile/
     ```bash
     $ ./configure --with-pcre-jit --with-ipv6 --with-http_realip_module --with-http_ssl_module --with-http_stub_status_module
     ```
+
+    Install [Luarocks v{{luarocks_version}}](https://github.com/keplerproject/luarocks/wiki/Download) (note: configure to run with the Lua/LuaJit interpreter included with OpenResty/Nginx)
 
     Some of the dependencies may be available in your favorite package manager.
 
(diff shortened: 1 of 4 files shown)


app/install/ubuntu.md
```diff
@@ -27,7 +27,7 @@ Start by downloading the corresponding package for your configuration:
 
     ```bash
     $ sudo apt-get update
-    $ sudo apt-get install netcat lua5.1 openssl libpcre3 dnsmasq
```
thibaultCha: We have to make sure that the OpenResty bundle puts LuaJIT in the $PATH of the system it is installed on, for all ours distributions.

Tieske: I don't think Openresty does it. But the Kong debian package does.

Just installed it on Ubuntu 14, while leaving `lua5.1` out. By installing it using
````
sudo dpkg --ignore-depends=lua5.1 -i kong-0.5.4.trusty_all.deb
````
Everything worked out of the box.

Anyway, I would appreciate it if someone else could look into this. I have a Windows background and I'm not familiar enough (yet) with this package stuff. For now it would take too much time for me to be efficient.

Comment (thibaultCha) [inline app/install/ubuntu.md]: @thefosk can confirm that all the packages have it. I'll check homebrew.

File: app/install/source.md [hunk 2]
```diff
@@ -36,6 +34,8 @@ redirect_from: /install/compile/
     $ ./configure --with-pcre-jit --with-ipv6 --with-http_realip_module --with-http_ssl_module --with-http_stub_status_module
     ```
 
+    Install [Luarocks v{{luarocks_version}}](https://github.com/keplerproject/luarocks/wiki/Download) (note: configure to run with the Lua/LuaJit interpreter included with OpenResty/Nginx)
```
thibaultCha: Best to say that LuaJIT 2.1 is needed (in the $PATH), and that when compiling OpenResty, it is possible to point to that LuaJIT. Same when compiling Luarocks.

thibaultCha: I'll update the source instructions too

Tieske: OpenResty includes LuaJit 2.0.4. I think we need stronger wording to encourage people to use the OpenResty included one, to prevent potential bugs or test failures from running on different executables.

Tieske: just committed the LuaJit reference with the changed wording

thibaultCha:
I applied this in a 0.6.0 branch created from master.

The workflow for the website is to make all the documentation changes under the previous folder version (here, `0.5.x`), and open a PR. Any changes for the upcoming releases are made on that new branch, and because we make changes in the previous folder, we clearly see what are the changes. Then, just before merging, we rename the folder to the new version (will be `0.6.x` here), grab the original `0.5.x`, The diff suddenly becomes much bigger, but that's fine, we just merge and deploy.

PR #188, opened 2015-12-17
[docs/response-transformer] new config added

shashiranjan84:
Updated doc to add details about append and replace

### DIFF
--- a/app/plugins/response-transformer.md
+++ b/app/plugins/response-transformer.md
@@ -8,11 +8,17 @@ breadcrumbs:
 nav:
   - label: Getting Started
     items:
+      - label: Installation
       - label: Configuration
+      - label: Order of execution
+      - label: Examples
 ---
 
 Transform the response sent by the upstream server on the fly on Kong, before returning the response to the client.
 
+<div class="alert alert-warning">
+  <strong>Note on transforming bodies:</strong> Be aware of the performamce of transformations on the response body. In order to parse and modify a JSON body, the plugin needs to retain it in memory, which might cause pressure on the worker's Lua VM when dealing with large bodies (several MBs). Because of Nginx's internals, the `Content-Length` header will not be set when transforming a response body.
+</div>
 ----
 
 ## Configuration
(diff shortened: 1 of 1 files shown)


thibaultCha:
The distinction between those 4 different directive is not described. It should be very clear that `add` will "do this if that", `append` will "do that is this" etc etc. (forgot exactly which keywords do what in details).

shashiranjan84:
i have updated the description like add adds new header, append adds new or append etc...

thibaultCha:
Not clearly enough though. It should clearly state "and not if"/"if and only if" etc...

shashiranjan84:
Updated description to make it look obvious.

thibaultCha:
Btw this plugin config would be better as config.headers.add rather than config.add.headers. So that headers, JSON are stored each in their own table instead.

app/plugins/response-transformer.md
```diff
@@ -9,6 +9,7 @@ breadcrumbs:
 
 Transform the response sent by the upstream server on the fly on Kong, before returning the response to the client.
 
+Caution: To transform the JSON body, plugin buffers the whole response body in memory instead of streaming it synchronously to the client. Also technically its not possible to the set `content-length` with new body size so it is removed form response headers if any of JSON body transformer configs are set. It may give unwanted results for responses with huge body.
```
thibaultCha: The proper format for notes is an `alert alert-warning` div with a `Note:` prefix. We use it in a lot of places in the website:

```html
<div class="alert alert-warning">
  <strong>Note on transforming bodies:</strong> Be weary of performing transformations on the responses bodies. In order to parse and modify a JSON body, the plugin needs to retain it in memory, which might cause pressure on the worker's Lua VM when dealing with large bodies (several MBs). Because of Nginx's internals, the `Content-Length` header will not be set when transforming a response body.
</div>
```

shashiranjan84:
Updated the doc.

shashiranjan84:
Ready for review.

PR #199, opened 2016-02-17
docs(about) add "Why Lua" section

sonicaghi:
@ahmadnassri @thibaultCha


ahmadnassri:
I think this needs to be re-thought.

sinzone:
keep it open till you have a better solution.

PR #202, opened 2016-02-24
Updated example code to reflect body attribute

afajem:
The example code for this plugin appears to still be referring to the `.form` config attribute. The references to `.form` attribute in the example has been changed to `.body`.


thibaultCha:
Thanks!

PR #207, opened 2016-03-04
Heroku installation revs for Kong 0.7.0

mars:
This is a rev to simply & update the Heroku install instructions, since we released a Kong 0.7.0 buildpack for Heroku.

@ahmadnassri I opened this a few days ago, but it somehow was closed. Here it is again.

Also, it seems that the Heroku install page is now missing, so maybe merging this will put it back :8ball:

### DIFF
--- a/app/install/heroku.md
+++ b/app/install/heroku.md
@@ -64,37 +64,15 @@ $ curl http://localhost:8001/status
 
 1. **Configuration**:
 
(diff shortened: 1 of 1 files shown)


thibaultCha:
Great, thanks! 

> Also, it seems that the Heroku install page is now missing, so maybe merging this will put it back

Yeah, some weird stuff going on there, but it's back up! :)

PR #213, opened 2016-04-15
No latest directory was created

NeckBeardPrince:
Clicking on "Quickly learn how to use Kong with the 5-minute Quickstart." in a how to install Kong page causes an error because it's looking for /docs/latest but there is no /latest created.

https://getkong.org/install/vagrant/

fyi not at all a frontend guy..


ahmadnassri:
good catch! thanks for reporting.

PR #224, opened 2016-05-11
Remove unnecessary ".lua" extension

jmdacruz:
The module name should not have the ".lua" extension:

```
"kong.plugins.<plugin_name>.daos.lua"
```
should be:
```
"kong.plugins.<plugin_name>.daos"
```


thibaultCha:
Correct, thanks!

PR #231, opened 2016-06-02
Adds Rail Europe company to about/who

theetrain:
This updates https://getkong.org/about/who/ to include Rail Europe.

image


sinzone:
oh @theetrain we forgot Expedia. add Expedia.

ahmadnassri:
@theetrain 👍

PR #243, opened 2016-06-11
feat(community): update meetup list

theetrain:
This brings the list to 24 meetups around the world. Right now they're listed alphabetically, but we may want to consider grouping them by continent in the future.

image


ahmadnassri:
:shipit:

PR #244, opened 2016-06-22
--url is mandatory since URL is not the last value

gomes:
--url is mandatory since URL is not the last value


ahmadnassri:
@gomes I can repeat any of the API calls without using `--url` given the same argument order ... curious what version of curl and what environment are you running in?

gomes:
curl --version
curl 7.43.0 (x86_64-apple-darwin15.0) libcurl/7.43.0 SecureTransport zlib/1.2.5

But right know I've tested again and probably the issue was in Kong, not with command, because it works, I'm sorry. I will close this.

PR #259, opened 2016-07-25
Removed Producthunt button from footer

darshilpatel:
Before
68747470733a2f2f636c2e6c792f314931563234334e3164306f2f496d616765253230323031362d30372d32322532306174253230332e31382e3432253230504d2e706e67

After
capture

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -2,3 +2,5 @@ source 'https://rubygems.org'
 gem 'jekyll', '2.4'
 gem 'jekyll-redirect-from'
 gem 'rouge', '1.9'
+gem 'redcarpet'
+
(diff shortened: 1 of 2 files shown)


Gemfile
```diff
@@ -2,3 +2,5 @@ source 'https://rubygems.org'
 gem 'jekyll', '2.4'
 gem 'jekyll-redirect-from'
 gem 'rouge', '1.9'
+gem 'redcarpet'
```
thibaultCha: I don't think this change belongs to the scope of this PR?

darshilpatel: Good catch. Fixed it

thibaultCha:
Thanks!

PR #263, opened 2016-08-12
release: 0.9

thibaultcha:
* new 0.9.x docs folder
* added 0.9 as latest version in meta YAML files
* new CLI reference 
* new configuration reference
* add new 'tracking' generator to the correlation-id plugin


thibaultcha:
Addressed all comments

PR #269, opened 2016-08-26
Fixed broken example

elruwen:
Hi!

When creating a plugin using the documentation I found the following error.

Cheers
Ruwen


thibaultcha:
Thanks

PR #272, opened 2016-08-31
Fix "Create a Customer" example. Add response data to other examples.

LyleScott:
I've added the "-d ''" parameter to the _Create a Customer_ curl example. If this is not done, the endpoint returns a bad response.

I've also added the response data to other endpoint examples for that page since they were on some of the other examples. The more examples, the better IMHO.


thibaultcha:
Thanks

PR #283, opened 2016-09-28
updated kong description

sinzone:


app/index.html
```diff
@@ -1,6 +1,6 @@
 ---
 id: homepage
-title_override: Kong - Open-Source Microservice and API Management Layer
+title_override: Kong - The Open-Source API Gateway and Microservice Management Layer
```
thibaultcha: `s/Microservice/Microservices`?

thibaultcha: Also Open-Source should probably be "open-source" (like below)?

app/index.html
```diff
@@ -9,7 +9,7 @@
     <div class="row">
       <div class="one-half column">
         <h2><a href="/plugins">Secure, Manage &amp; Extend your&nbsp;APIs and Microservices</a></h2>
-        <p><a href="/plugins">The open-source management layer for APIs, delivering high performance and reliability.</a></p>
+        <p><a href="/plugins">The open-source API Gateway and Microservices Management Layer, delivering high performance and reliability.</a></p>
```
theetrain: I'd say capitalise 'Open-Source', same as you do in the title.

thibaultcha: works for me too

sinzone: ok

app/index.html
thibaultcha: The `The` actually takes too much space in the limited space offered in the tab title imo, I'd rather have people see "Open-source API Mgmt..." instead of "The open-source API..."

thibaultcha:
Manually applied on master with minor modifications.

PR #286, opened 2016-09-30
kong-vagrant uses postgres, not cassandra

moredip:
From looking at the kong-vagrant repo it looks like it uses Postgres as a backing store, not Cassandra. 

https://github.com/Mashape/kong-vagrant/blob/master/provision.sh#L15


thibaultcha:
Thanks

PR #302, opened 2017-01-12
Update keyauth behaviour

bogdangrigg:
This is the behavior I'm experiencing currently - hopefully I didn't miss a previous step.


Tieske:
are you aware you updated the docs for the `0.3.x` version of Kong?

bogdangrigg:
Hi Tieske

Nope - I just updated master. What's the correct branch to do this on?

However, the same problem exists in the version you have online at https://getkong.org/docs/0.9.x/getting-started/enabling-plugins/

Also, getting a 401 is just the behavior I'm currently experiencing (from a fresh install). Not sure if I missed something & it should actually be a 403.

bogdangrigg:
Very good point. I'll check and update the specific versions (I see now 0.10.x is the latest).

PR #315, opened 2017-02-22
fix(docs): add curl flag

theetrain:
Add location flag to docker.


thibaultcha:
Could you elaborate on the why?

theetrain:
@thibaultcha:

@john-paul-keenan had noticed an error when using the curl command WITHOUT `-L`:
```
Empty reply from server
```

john-paul-keenan:
I updated docke (now using Version 1.13.1 (15353)) and the command works without the -L flag. BUT, it also works with it so I think this is still good for some of our users

PR #319, opened 2017-02-28
fix(docs): bring back version dropdown

theetrain:
Bring back the missing version dropdown menu.

screen shot 2017-02-27 at 7 25 08 pm

### DIFF
--- a/app/_data/kong_versions.yml
+++ b/app/_data/kong_versions.yml
@@ -77,18 +77,6 @@
   lua_doc: true
 -
   release: "0.9.x"
-  version: "0.9.5"
-  luarocks_version: "0.9.5-0"
-  dependencies:
-    luajit: "2.1.0-alpha"
-    luarocks: "2.3.0"
-    cassandra: "2.2.x"
-    postgres: "9.4+"
-    openresty: "1.9.15.1"
-    serf: "0.7"
-  lua_doc: true
--
-  release: "0.9.x"
   version: "0.9.9"
   luarocks_version: "0.9.8-0"
   dependencies:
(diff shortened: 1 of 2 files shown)


gulpfile.js
```diff
@@ -23,7 +23,7 @@ var sources = {
   content: 'app/**/*.{markdown,md,html,txt,yml,yaml}',
   styles: paths.assets + 'stylesheets/**/*',
   js: [
-    paths.assets + 'javascripts/**/*.js',
```
thibaultcha: I don't think this should be changed. It also doesn't appear to be in the scope of this PR?

gulpfile.js
```diff
@@ -45,8 +45,7 @@ gulp.task('styles', function () {
 gulp.task('javascripts', function () {
   return gulp.src(sources.js)
     .pipe($.plumber())
-    .pipe($.sourcemaps.init())
-    .pipe($.sourcemaps.write('maps'))
```
thibaultcha: ditto

app/_data/kong_versions.yml
```diff
@@ -77,18 +77,6 @@
...
-    luarocks: "2.3.0"
-    cassandra: "2.2.x"
-    postgres: "9.4+"
-    openresty: "1.9.15.1"
-    serf: "0.7"
-  lua_doc: true
--
-  release: "0.9.x"
   version: "0.9.9"
   luarocks_version: "0.9.8-0"
```
thibaultcha: This should be updated to 0.9.9-0

nijikokun: :+1:

nijikokun: :+1: agreed

PR #328, opened 2017-03-10
docs(install) add Mesos + Kubernetes instructions

shashiranjan84:


thibaultcha:
Re-worked with @shashiranjan84. We now need a 3rd commit to properly align the new boxes on the `/install` page.

Note: reminder to remove Heroku.

PR #329, opened 2017-03-10
Fix/install page

theetrain:
- Overhauled the 'install' grid appearance. Items on the last row are now centred.
- The .md pages were updated to pass echint.

screen shot 2017-03-10 at 5 32 20 pm


thibaultcha:
@theetrain Thank you!

PR #338, opened 2017-03-15
fix(nav): add new enterprise link

theetrain:
- fix newrelic script not being loaded
- fix enterprise link not going to enterprise page
- remove gelato and galileo links from nav

Branch name is a misnomer.


app/_includes/head.html
```diff
@@ -26,7 +26,8 @@
   <meta name="google-site-verification" content="CrU3zp02dNKTe8NSAipL4NCPkrIjDXG8fViTZ-MIzP4"/>
 
   <!-- New Relic -->
-  <script src="/assets/newrelic.js"></script>
```
thibaultcha: Seems like this should have been a separate commit. Any reason why it wasn't loaded though? Just curious!

theetrain: In this `/dist` directory, newrelic.js is nowhere to be found since our gulp task only copies the concatenated app.js. Since NewRelic is meant to be loaded before all other dependencies, I found Jekyll `include` to be the preferred approach and keep it inline.

thibaultcha: Okay, thanks

thibaultcha:
@coopr Good to merge?

coopr:
LGTM, but really, I'm the wrong guy to review PRs.

Yes, I'm totally happy with this being deployed to prod - if you are happy too.

thibaultcha:
The review is because this changes the header on the website to remove the links to Galileo and Gelato, hence why :)

thibaultcha:
(Which is in alignment with our objectives, but just making sure it isn't too soon to deploy).

coopr:
Oh, yeah, that, thanks @thibaultcha - sure, :shipit:

PR #342, opened 2017-03-16
doc(plugins) sync logging plugin docs

p0pr0ck5:
* Missing latency field in TCP logging example
* Proper upstream_url in several logging examples
* Remove undefined authenticated_entity keys

### DIFF
--- a/app/plugins/file-log.md
+++ b/app/plugins/file-log.md
@@ -79,16 +79,24 @@ Every request will be logged separately in a JSON object separated by a new line
     },
     "authenticated_entity": {
         "consumer_id": "80f74eef-31b8-45d5-c525-ae532297ea8e",
-        "created_at":	1437643103000,
         "id": "eaa330c0-4cff-47f5-c79e-b2e4f355207e",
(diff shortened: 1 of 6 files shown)


thibaultcha:
Sweet, thanks!

It all looks good except one thing: since 0.10.0, final slashes are disallowed in the `upstream_url` field of an API. Mind removing those we have in there?

thibaultcha:
Actually, it seems the entire definition of API Objects is still for 0.9.x 😅 

Now that we have so many more fields in the API definition, this might warrant an update of the basic serializer in Kong to remove the undesired one (ex: created_at is not in the `consumer` field, but still the the `api` field...).

Side note: those logging serializers were a nice idea that wasn't entirely followed through (ability to configure more or less verbose serializers for all logging plugins), but logging plugins have became less and less of a priority... Just thought you might be curious about those.

p0pr0ck5:
Nice catch. Re-did the `api` key for each logging example based on https://getkong.org/docs/0.10.x/admin-api/#api-object

thibaultcha:
Thanks!

PR #350, opened 2017-03-21
change hosts response value

rominirani:
The response value for hosts attribute in the Add Your API section had
an incorrect value of example.org. It is now changed to example.com to
reflect the rest of the document.


thibaultcha:
Thank you!

PR #364, opened 2017-03-28
Noted the name change on config.origin

john-paul-keenan:
This was formerly config.origin and was updated to config.origins so I wanted to point that out


thibaultcha:
Fixed with 421a454

PR #370, opened 2017-04-04
added tag to the kong docker command

john-paul-keenan:
I think it is best practice to use a tag when calling an image. Using latest will save us the step of adding in the number each time. This may also avoid confusion for people who already have an older image of Kong cached.


thibaultcha:
Good point! Wouldn't it make sense to include the version tag though? Thinking of this page being versioned someday, if ever...

thibaultcha:
Better including the version tag once the `/install` page will be versioned. Thanks for the patch!

thibaultcha:
Forgot to change the commit message after merging a few good ones from @p0pr0ck5, butthis should have been `docs(install) add version tag to docker image`.

PR #372, opened 2017-04-04
docs(plugins) OAuth 2.0 Introspection plugin

thefosk:
OAuth 2.0 Introspection plugin docs.


app/_includes/head.html
```diff
@@ -49,7 +49,6 @@
   <link rel="dns-prefetch" href="//cdn.segment.com">
   <link rel="dns-prefetch" href="//sidecar.gitter.im">
   <link rel="dns-prefetch" href="//buttons.github.io">
-  ##preload_assets##
```
thefosk: @theetrain I had to remove this because it was showing up in my body - thoughts? It was added with 8ba0ecd626177c10ae15e3414b672f1c5032b7b4

theetrain: @thefosk that token should be replaced by a string on `npm start` https://github.com/Mashape/getkong.org/blob/master/gulpfile.js#L153

You shouldn't remove it. Was it showing up after gulp?

thefosk: @theetrain yep, it was showing in the body after running `npm start` (which in turn runs `gulp`)

theetrain: @thefosk, I pulled your branch and restored the token - I can see one situation where you may experience this: if you save a page, then `gulp jekyll` runs and browser-sync refreshes the page before `gulp html` finishes:

```bash
[21:23:32] Finished 'jekyll' after 3.93 s
[21:23:32] Starting 'html'...
[ ▶ ] Reloading Browsers...              <--- page refresh
[21:23:43] all files 11.92 MB
[21:23:43] Finished 'html' after 11 s .  <-- ##preload_assets## gets replaced, no auto refresh
```
If you manually refresh after `html` finishes, the token should disappear. Please try again and let me know if that works.

thefosk: @theetrain you were right - I have restored the directive.

app/docs/0.10.x/plugin-development/distribution.md
```diff
@@ -224,22 +224,22 @@ Then, you should see the following log for each plugin being loaded:
...
 
 3. To remove the plugin thoroughly, delete the plugin-related files from
-   each of the Kong nodes. Make sure to have completed step 2, including
-   restarting/reloading Kong, before deleting the files. If you used LuaRocks
-   to install the plugin, you can do `luarocks remove <plugin-name>` to remove
-   it.
+  each of the Kong nodes. Make sure to have completed step 2, including
+  restarting/reloading Kong, before deleting the files. If you used LuaRocks
+  to install the plugin, you can do `luarocks remove <plugin-name>` to remove
+  it.
```
thibaultcha: Also please restore those changes as they are not necessary anymore.

PR #374, opened 2017-04-05
fix(distribution): pass echint

theetrain:
No content changes.

### DIFF
--- a/app/docs/0.10.x/plugin-development/distribution.md
+++ b/app/docs/0.10.x/plugin-development/distribution.md
@@ -224,22 +224,22 @@ Then, you should see the following log for each plugin being loaded:
 There are three steps to completely remove a plugin.
 
 1. Remove the plugin from your Kong api configuration. Make sure that it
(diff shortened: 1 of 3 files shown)


app/docs/0.10.x/plugin-development/distribution.md
```diff
@@ -224,22 +224,22 @@ Then, you should see the following log for each plugin being loaded:
...
 
 3. To remove the plugin thoroughly, delete the plugin-related files from
-   each of the Kong nodes. Make sure to have completed step 2, including
-   restarting/reloading Kong, before deleting the files. If you used LuaRocks
-   to install the plugin, you can do `luarocks remove <plugin-name>` to remove
-   it.
+  each of the Kong nodes. Make sure to have completed step 2, including
+  restarting/reloading Kong, before deleting the files. If you used LuaRocks
+  to install the plugin, you can do `luarocks remove <plugin-name>` to remove
+  it.
```
thibaultcha: Sweet! Could we remove this hunk from the patch now that the linter is ignoring such errors?

PR #378, opened 2017-04-06
fix(footer): restore subscribe form

theetrain:
- Bring back subscribe form in footer area:

image


app/_includes/footer.html
```diff
@@ -1,3 +1,7 @@
+{% if page.hide_subscribe == null %}
```
theetrain: Note, every page displays this form. The only page with `hide_subscribe: true` is the defunct enterprise page.

PR #381, opened 2017-04-10
docs(log) logging plugins now also log the consumer

thefosk:
Goes in hand with https://github.com/Mashape/kong/pull/2367

### DIFF
--- a/app/plugins/file-log.md
+++ b/app/plugins/file-log.md
@@ -98,13 +98,18 @@ Every request will be logged separately in a JSON object separated by a new line
         "upstream_send_timeout": 60000,
         "upstream_url": "http://httpbin.org"
     },
+    "consumer": {
+        "username": "demo",
+        "created_at": 1491847011000,
+        "id": "35b03bfc-7a5b-4a23-a594-aa350c585fa8"
+    },
     "latencies": {
         "proxy": 1430,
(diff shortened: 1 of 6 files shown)


app/plugins/file-log.md
```diff
@@ -113,12 +118,14 @@ A few considerations on the above JSON object:
...
 * `api` contains Kong properties about the specific API requested
-* `authenticated_entity` contains Kong properties about the authenticated consumer (if an authentication plugin has been enabled)
+* `authenticated_entity` contains Kong properties about the authenticated credential (if an authentication plugin has been enabled)
+* `consumer` contains the authenticated Consumer (if an authentication plugin has been enabled)
 * `latencies` contains some data about the latencies involved:
   * `proxy` is the time it took for the final service to process the request
   * `kong` is the internal Kong latency that it took to run all the plugins
   * `request` is the time elapsed between the first bytes were read from the client and after the last bytes were sent to the client. Useful for detecting slow clients.
-
+* `client_ip` contains the original client IP address
```
thibaultcha: the "original" client IP address is not enough details imo, people need to know where that value is coming from, to better understand how to use it and if they can rely on it or not at all. This should, in the future because this is not the scope of this PR, explicitly state the actual Nginx variable it is based on.

PR #393, opened 2017-04-21
docs(rate-limiting) hide_client_headers option

p0pr0ck5:


Tieske:
Another mention of this new flag would be nice under the heading `Headers sent to the client`

p0pr0ck5:
@Tieske done!

PR #405, opened 2017-04-27
Update faq.md

seils:
Change 'should' to negative form.


thibaultcha:
Merged, thank you for the fix!

PR #408, opened 2017-04-28
docs(log) http-log authentication and retries logging

Tieske:
Deals with Kong issue https://github.com/Mashape/kong/pull/2429: logging retries and failure info
and https://github.com/Mashape/kong/pull/2430 for adding a basic auth header with the http log plugin

This documents the additional changes in https://github.com/Mashape/kong/pull/2467.

### DIFF
--- a/app/plugins/file-log.md
+++ b/app/plugins/file-log.md
@@ -77,9 +77,21 @@ Every request will be logged separately in a JSON object separated by a new line
             "access-control-allow-origin": "*"
         }
     },
+    "tries": [
+        {
+            "state": "next",
+            "code": 502,
+            "ip": "127.0.0.1",
+            "port": 8000
+        },
+        {
+            "ip": "127.0.0.1",
+            "port": 8000
+        }
+    ],
     "authenticated_entity": {
         "consumer_id": "80f74eef-31b8-45d5-c525-ae532297ea8e",
-        "id": "eaa330c0-4cff-47f5-c79e-b2e4f355207e",
+        "id": "eaa330c0-4cff-47f5-c79e-b2e4f355207e"
     },
     "api": {
         "created_at": 1488830759000,
@@ -112,6 +124,7 @@ A few considerations on the above JSON object:
 
 * `request` contains properties about the request sent by the client
 * `response` contains properties about the response sent to the client
+* `tries` contains the list of (re)tries (successes and failures) made by the load balancer for this request
 * `api` contains Kong properties about the specific API requested
 * `authenticated_entity` contains Kong properties about the authenticated consumer (if an authentication plugin has been enabled)
 * `latencies` contains some data about the latencies involved:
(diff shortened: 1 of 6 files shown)


app/plugins/file-log.md
```diff
@@ -77,9 +77,21 @@ Every request will be logged separately in a JSON object separated by a new line
             "access-control-allow-origin": "*"
         }
     },
+    "tries": [
```
p0pr0ck5: this is valid for the file-log plugin?

app/plugins/http-log.md
```diff
@@ -40,11 +40,15 @@ form parameter                  | default | description
 ---                             | ---     | ---
 `name`                          |         | The name of the plugin to use, in this case: `http-log`
 `consumer_id`<br>*optional*     |         | The CONSUMER ID that this plugin configuration will target. This value can only be used if [authentication has been enabled][faq-authentication] so that the system can identify the user making the request.
-`config.http_endpoint`          |         | The HTTP endpoint (including the protocol to use) where to send the data to
+`config.http_endpoint`          |         | The HTTP endpoint (including the protocol to use) where to send the data to.
```
theetrain: > The HTTP endpoint (including the protocol) `to which the data will be sent.`

app/plugins/file-log.md
thibaultcha: Curious: why not? It should use the same serializer as the other ones, so yes. Maybe your understanding of the file-log plugin is that it doesn't log requests/response info?

app/plugins/http-log.md
```diff
@@ -40,11 +40,15 @@ form parameter                  | default | description
...
 `config.method`<br>*optional*   | `POST`  | An optional method used to send data to the http server, other supported values are PUT, PATCH
 `config.timeout`<br>*optional*  | `10000` | An optional timeout in milliseconds when sending data to the upstream server
 `config.keepalive`<br>*optional*| `60000` | An optional value in milliseconds that defines for how long an idle connection will live before being closed
 
+**NOTE:** If the `config.http_endpoint` contains a username and password, then
```
thibaultcha: I think we should make clear "where":
```
contains a username and password in the authentication section of the URL.
Ex: http://bob:password@example.com/logs
```

app/plugins/file-log.md
```diff
@@ -112,6 +124,7 @@ A few considerations on the above JSON object:
 
 * `request` contains properties about the request sent by the client
 * `response` contains properties about the response sent to the client
+* `tries` contains the information of the successive (re)tries, the upstream address and failure information
```
thibaultcha: I think a bit more context here wouldn't hurt:
```
contains the list of (re)tries (successes and failures) made by the load balancer for this request
```

app/plugins/file-log.md
p0pr0ck5: Yep, misread this. Thought this had to do with the log plugin retrying, not the retries in the serialized data. My bad!

PR #410, opened 2017-04-28
docs(file-log) added the `reopen` parameter

Tieske:


app/plugins/file-log.md
```diff
@@ -17,6 +17,10 @@ nav:
 
 Append request and response data to a log file on disk.
 
+It is not recommended to use this plugin in production, it would be better to
+use another logging plugin, for example `syslog`, in those cases (see the
+`reopen` parameter below).
```
p0pr0ck5: Let's have an explanation (at least of brief one) of why this is the case.

app/plugins/file-log.md
```diff
@@ -38,6 +43,7 @@ form parameter                | default | description
 `name`                        |         | The name of the plugin to use, in this case: `file-log`
 `consumer_id`<br>*optional*   |         | The CONSUMER ID that this plugin configuration will target. This value can only be used if [authentication has been enabled][faq-authentication] so that the system can identify the user making the request.
 `config.path`                 |         | The file path of the output log file. The plugin will create the file if it doesn't exist yet. Make sure Kong has write permissions to this file.
+`reopen`                      | `false` | Determines whether the log file is closed and reopened on every request. If the file is not reopened, and has been removed/rotated, the plugin will keep writing to the stale file descriptor, and hence loose information.
```
p0pr0ck5: 'lose' not 'loose' (i would fix and merge this myself, but the above comment needs a bit more fleshing out i think? i can handle it if you like).

app/plugins/file-log.md
Tieske: updated!

thibaultcha:
Merged with a note about which version starts introducing this parameter.

PR #411, opened 2017-04-28
Feat/terminate

Tieske:


thibaultcha:
Seems like this needs some rebasing on `0.10.2`

Tieske:
yep

PR #414, opened 2017-05-01
[cli] added the miss path to defaults for the check command

codebien:


Tieske:
thx!

app/docs/0.10.x/cli.md
```diff
@@ -46,7 +46,7 @@ Usage: kong check <conf>
 
 Check the validity of a given Kong configuration file.
 
-<conf> (default /etc/kong.conf) configuration file
```
thibaultcha: This old path should be removed as well, it's not what the command will currently do.

Tieske: @thibaultcha not sure what you mean?

PR #415, opened 2017-05-10
Broken self link, and misc markdown fixes

mdaniel:
This journey started when I noticed that (ironically) the [this website is Open-Source](https://getkong.org/docs/0.10.x/proxy/#conclusion) link is incorrectly marked-down(?). Then I noticed the markdown link right below it was also bogus, which led me to look for other incorrect markdown `grep -Ero '\[[a-z0-9-]{5,}\]' .` ... and here we are.

I tested them locally, and everything looked like I expected, so I hope it works when deployed, also.


thibaultcha:
That's great, thank you!

PR #419, opened 2017-05-15
api plugin update and delete docs should match Kong routes

lostphilosopher:
Currently (0.10.2) Kong does not support UPDATE or DELETE by name. 

See also: https://github.com/Mashape/kong/issues/2527 - if this gets fixed this PR won't be necessary.


thibaultcha:
Thanks for bringing this to our attention, taken care of this by reverting the original commit who introduced this change instead. This change will be re-introduced when releasing Kong 0.11 and its docs.

PR #425, opened 2017-05-24
feat(styles): remove bottom margin removal from leading paragraph

nijikokun:
Fixes the leading paragraph for the Plugins page by removing the margin removal declaration.

screen shot 2017-05-24 at 3 10 19 pm


coopr:
LGTM! Thanks @niji!

PR #428, opened 2017-06-01
docs(utils) document updates random_string return

p0pr0ck5:


nijikokun:
Looks good to me, would suggest using `The random string (modified ...` to `A modified ...`

p0pr0ck5:
@nijikokun thanks, done.

app/docs/0.10.x/lua-reference/modules/kong.tools.utils.html
```diff
@@ -349,7 +349,7 @@ <h2 class="section-header "><a name="Functions">Functions</a></h2>
     <h5>Returns:</h5>
     <ul>
       <li>
-        string  The random string (a uuid without hyphens)
+        string  The random string (a modified Base64-encoded psuedorandom data generated via CSPRNG).
```
thibaultcha: "psuedorandom"

Tieske:
unless I'm mistaken this is generated from the source code using Ldoc, so the proper place to update would be in the source code and then regenerate.

p0pr0ck5:
ah thanks @Tieske :)

Tieske:
@p0pr0ck5 can the branch be deleted?

PR #433, opened 2017-06-06
fix(travis): upgrade node version

theetrain:
Node 4 was giving our builds issues when Yarn ran. Now it ought to work.


thibaultcha:
Sweet, thanks!

PR #434, opened 2017-06-07
List JWT credentials

therealgambo:
Fixes: https://github.com/Mashape/getkong.org/issues/423


thibaultcha:
Manually merged, thank you!

PR #435, opened 2017-06-07
docs(rate-limiting) Redis Sentinel support

coopr:
Add mention of Redis Sentinel support in Kong Enterprise


thibaultcha:
:+1:

thibaultcha:
We could make this more visible with an HTML note that is more visible, with a yellow background (some examples are visible on the website).

Long term: we could also consider implementing a custom style for "EE-only" notes across the documentation. Ex: document a feature and have an "EE-only" note/box that says a better version of said feature is included in EE.

thefosk:
Highlighted in the box.

PR #436, opened 2017-06-12
docs(plugins) redis sentinel support

thefosk:


thibaultcha:
I feel like the content of #435 is better than that of this PR, but this PR uses the notes style as I suggested in #435 to make this more visible.

Might want to synchronize with @coopr here?

thefosk:
I have missed #435 - this is a duplicate.

PR #446, opened 2017-06-29
Add documentation for created_at parameter in Create or Update API

tanelso2:
https://github.com/Mashape/kong/issues/1324 mentions this, but no
documentation was ever written.
I'm not sure when created_at started to be required, put the above
issue was reported on 0.8.x and I ran into it today on 0.10.x, so I
updated the docs from 0.8.x to 0.10.x

### DIFF
--- a/app/docs/0.10.x/admin-api.md
+++ b/app/docs/0.10.x/admin-api.md
@@ -512,11 +512,16 @@ HTTP 200 OK
 
 <div class="endpoint put">/apis/</div>
 
+Attributes | Description
+---:| ---
+`id`<br>**semi-required** | The unique identifier of the API to update (only required if updating)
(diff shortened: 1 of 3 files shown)


p0pr0ck5:
@tanelso2 thanks! Shouldn't this apply to all objects that can be updated via PUT, not just APIs?

tanelso2:
@p0pr0ck5 Sure thing. Wasn't aware that it applied to other objects, I've only run into it on APIs. Will get to that soon.

Sorry for the slow response, I've been on vacation

coopr:
Closing as stale - this docs improvement may still be needed however - if it is, and you, fine reader, are able to refresh this PR, or open a new one, please do so!

PR #450, opened 2017-07-13
docs(aws-lambda) add a reminding note for URI-encoded secrets in payloads

thibaultcha:
Follow-up doc improvement for Mashape/kong#2651


p0pr0ck5:
It would be a better improvement IMO to update all relevant entries where this encoding could occur (e.g. auth keys/tokens, where 'random' strings can be provided). BTW this is exactly why the new `random_string` only uses alphanumeric chars ;)

thibaultcha:
Yeah, this is a start.

thibaultcha:
(`random_string` has always returned alphanumeric chars only)

p0pr0ck5:
(sorry, had meant the transformation of b64: https://github.com/Mashape/kong/blob/release/0.11/kong/tools/utils.lua#L211).

Issue #451, opened 2017-07-13
Use --data-urlencode where appropriate


p0pr0ck5:
Apply the concept in #450 to all examples.

levfishbluefish:
@p0pr0ck5 - Can this issue be closed?

PR #454, opened 2017-07-14
docs(migrations) manual execution of migrations

Tieske:


app/docs/0.11.x/getting-started/quickstart.md
```diff
@@ -20,7 +20,8 @@ supports PostgreSQL and Cassandra).
     Issue the following command to [start][CLI] Kong:
 
     ```bash
```
kikito: The same in all the other changes -two separate command prompts instead of one with a comment.

Tieske: I thought about that, but especially for a "quick start" I really focussed on keeping the page as short as possible, such that page on a first glance keeps the impression of very short and minimal steps.

app/docs/0.11.x/plugin-development/index.md
```diff
@@ -8,16 +8,30 @@ chapter: 1
...
-
-Hence, Kong is a Lua application designed to load and execute Lua modules (which we more commonly refer to as "*plugins*") and provides an entire development environment for them, including database abstraction, migrations, helpers and more...
-
-Your plugin will consist of Lua modules which will be loaded and executed by Kong, and it will benefit from two APIs:
-
-- [lua-nginx-module API][lua-nginx-module]: allows interaction with Nginx itself, such as, for example, retrieving the request/response, or accessing Nginx's shared memory zone.
-- **Kong's plugin environment**: allows interaction with the datastore holding your configurations (APIs, Consumers, Plugins...) and various helpers allowing the interaction of plugins between each other. This is the environment that will be described by this guide.
+Before going further, it is necessary to briefly explain how Kong is built,
```
p0pr0ck5: is this just a line length change? if so, can we leave it out of this PR; it's unrelated to change scope and can be handled in a separate patch

app/docs/0.11.x/getting-started/quickstart.md
p0pr0ck5: I'm inclined to agree with @kikito here. The throwaway comment "(only first time)" is fairly ambiguous, giving no context to its use or value. We should be more explicitly about what's going on; at the least let's detail it a bit more elsewhere in the docs and link to it from these quickstart guides.

app/docs/0.11.x/plugin-development/index.md
Tieske: I know, but the point here is that if we don't change those things as we go, we don't do it at all. Hence I'm in favor of keeping those in. It's not a code file after all.

p0pr0ck5: This logic seems faulty. If it needs to be changed and we need a reminder, we can make an issue (like #451), or just put up a separate PR since it's already on your mind. Making this change here makes future archaeology efforts difficult; it's a bad pattern to get into.

Tieske: yes, we could do that, and then they'll end up in the bottom half of this page https://github.com/Mashape/getkong.org/issues (all from 2015)

app/docs/0.11.x/getting-started/quickstart.md
Tieske: updated!

app/docs/0.11.x/plugin-development/index.md
Tieske: reverted!

PR #456, opened 2017-07-19
docs(0.10.x) Add warning about case-sensitivity

OlzhasAlexandrov:
There are issues with case-sensitivity of Kong when dealing with AWS ELBs and basically situations, when URL is given with capital case letters.
In other words, having URLs in uppercase and then calling Kong APIs using those specified URLs in either `hosts` or `upstream_url` results in an error, since Kong thinks there is no API due to URL "differences"

### DIFF
--- a/app/docs/0.10.x/getting-started/adding-your-api.md
+++ b/app/docs/0.10.x/getting-started/adding-your-api.md
@@ -20,6 +20,8 @@ helpful for learning how Kong proxies your API requests.
 Kong exposes a [RESTful Admin API][API] on port `:8001` for managing the
 configuration of your Kong instance or cluster.
 
+Kong is case-sensitive; therefore, make sure to have your hosts and upstream_url in lowercase.
+
 1. ### Add your API using the Admin API
 
     Issue the following cURL request to add your first API ([Mockbin][mockbin])


OlzhasAlexandrov:
@p0pr0ck5 Sure, I've amended the commit; apologies for my ignorance:)

Tieske:
just wondering: should this be documented, or be fixed by forcing everything to lowercase?

OlzhasAlexandrov:
@Tieske if `hosts` and `upstream_url` accepted only Domain Name System (DNS), then it would be great to force the lowercase.
However, there might be situations when `hosts` and `upstream_url` will contain URL, not only the DNS part.

**How about forcing it by default and still allowing us to add a flag to disable lowercase transformation?**

Tieske:
yes, sorry, I meant only hosts obviously

p0pr0ck5:
We should clarify the meaning of this. To say that "Kong" is case sensitive is a fairly sweeping and ambiguous term; could we update this patchset more explicitly note the sensitivity, as discussed here?

OlzhasAlexandrov:
@p0pr0ck5 Could you please send the changes that you believe would explain the potential error better?

Regarding Kong being case-sensitive, you are welcome to reword it as you would like. I just wanted to bring up this issue.

Please let me know, if I could do anything else.

kikito:
Hi @OlzhasAlexandrov , sorry it took us to much time to get back to you on this.

I interpret what @p0pr0ck5 as follows: "kong is case sensitive" is inaccurate because it isn't in all cases (for example on headers). A more concise phrase would be something along the lines of "When matching a request against an api's `hosts` field, the match is case sensitive".

Ideally this should also be added to the Admin API section, and not only on the "getting started" guide.

I've created a new issue for fixing this problem (#586) and will close your PR. Please feel free to send a new PR with the proposed changes.

Again, thanks for taking the time to send a PR in the first place. The generated discussion was helpful.

PR #459, opened 2017-07-24
docs(configuration) serving website and API

salazar:
Instructions on how to setup a website with the same server
name and port as the API.

### DIFF
--- a/app/docs/0.10.x/configuration.md
+++ b/app/docs/0.10.x/configuration.md
@@ -12,6 +12,7 @@ title: Configuration Reference
 - [Custom Nginx configuration & embedding Kong](#custom-nginx-configuration-embedding-kong)
   - [Custom Nginx configuration](#custom-nginx-configuration)
   - [Embedding Kong](#embedding-kong)
+  - [Serving both a website and your APIs from Kong](#serving-both-a-website-and-your-apis-from-kong)
 - [Properties reference](#properties-reference)
   - [General section](#general-section)
   - [Nginx section](#nginx-section)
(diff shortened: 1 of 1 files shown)


app/docs/0.10.x/configuration.md
```diff
@@ -12,6 +12,7 @@ title: Configuration Reference
 - [Custom Nginx configuration & embedding Kong](#custom-nginx-configuration-embedding-kong)
   - [Custom Nginx configuration](#custom-nginx-configuration)
   - [Embedding Kong](#embedding-kong)
+  - [Serving API and Website](#serving-api-and-website)
```
thibaultcha: We should use a slightly more descriptive title here, like:

"Serving both a website and your APIs from Kong"

app/docs/0.10.x/configuration.md
```diff
@@ -203,6 +204,38 @@ http {
 }
 ```
 
+#### Serving API and Website
+
+A common use case for this is to serve a website and API from the
```
thibaultcha: "for this" implies some context but no such context exist yet because this is the beginning of this chapter.

Maybe better to start like so:
```
A common use case for API providers is to make Kong serve both a website and
the APIs themselves over the Proxy port (`80` or `443` in production). For example:
`https://my-api.com` (website) and `https://my-api.com/api/v1` (API).

To achieve this, we cannot satisfy ourselves with declaring a new HTTP server like the
previous example, and must use a different setup. A good solution is to use a custom
Nginx configuration template which adds a new `location` block serving the website
alongside the Kong Proxy `location` block.

Example:
```

app/docs/0.10.x/configuration.md
```diff
@@ -203,6 +204,38 @@ http {
...
+
+pid pids/nginx.pid;                      # this setting is mandatory
+error_log logs/error.log ${{LOG_LEVEL}}; # can be set by kong.conf
+
+events {}
+
+http {
+  # inline Kong configuration template here
+  ...
+  location /website {
```
thibaultcha: The intent from API providers is usually to serve the website from root (`/`), and the API from some arbitrary path (`/api/v1`).

To better reflect the example, I would advise showing the inlined Kong configuration file (to some extent, only the relevant part, by leaving the boilerplate ngx_lua directives in `...` placeholders), and highlight how the Kong Proxy handler `location /` has been changed to something like `location /api/v1`. Then, we add a comment over this line, saying something like:

```nginx
# here, we declare our custom location serving our website (or API portal)
# which we can optimize for serving static assets
location /api/v1 {
    root /var/www/my-api.com/html;
    index index.htm index.html;
    ...
}
```

app/docs/0.10.x/configuration.md
coopr: I suggested this super-short sub-title to @salazar  - I now see that it is a little too short :)

salazar:
[Website and API, same server name + port](https://trello.com/c/EgbZecNc/31-website-and-api-same-server-name-port)

salazar:
Thanks for the points @thibaultcha. Please review again.

app/docs/0.10.x/configuration.md
```diff
@@ -12,6 +12,7 @@ title: Configuration Reference
 - [Custom Nginx configuration & embedding Kong](#custom-nginx-configuration-embedding-kong)
   - [Custom Nginx configuration](#custom-nginx-configuration)
   - [Embedding Kong](#embedding-kong)
+  - [Serving both a wesite and your APIs from Kong](#serving-both-a-website-and-your-apis-from-kong)
```
thibaultcha: typo: website ;)

app/docs/0.10.x/configuration.md
```diff
@@ -203,6 +204,66 @@ http {
 }
 ```
 
+#### Serving both a website and your APIs from Kong
+
+A common use case for API providers is to make Kong serve both a website
+and the APIs themselves over the Proxy port &mdash; `80` or `443` in
```
thibaultcha: Is `&mdash;` properly parsed by the Markdown engine? Didn't run this myself

app/docs/0.10.x/configuration.md
```diff
@@ -203,6 +204,66 @@ http {
...
+  charset UTF-8;
+
+  # any contents until Kong's Proxy server block
+  ...
+
+  # Kong's Proxy server block
+  server {
+    server_name kong;
+
+    # any contents until the location / block
```
thibaultcha: Those are helpful to keep track of where we're at :+1:

app/docs/0.10.x/configuration.md
```diff
@@ -203,6 +204,66 @@ http {
...
+    }
+
+    # Kong's Proxy location / has been changed to /api/v1
+    location /api/v1 {
+      set $upstream_host nil;
+      set $upstream_scheme nil;
+      set $upstream_uri nil;
+
+      # Any remaining configuration for the Proxy location
+      ...
```
thibaultcha: Maybe closing the above `location /api/v1` block would be preferable beneath that `...`.

app/docs/0.10.x/configuration.md
```diff
@@ -203,6 +204,66 @@ http {
...
+  }
+
+}
```
thibaultcha: Here, maybe a comment saying:
```
# Kong's Admin API server block goes beneath
...
```

app/docs/0.10.x/configuration.md
salazar: Yes, it is! I'm running it locally.

app/docs/0.10.x/configuration.md
salazar: Fixed!

PR #461, opened 2017-07-27
Clarification of the headers.append doc in request-transformer

Tigraine:
Resulted from the discussion here: https://github.com/Mashape/kong/issues/2733


Tieske:
@Tigraine thx!

PR #463, opened 2017-07-30
Updated link to point to the correct doc section

UkiahSmith:


Tieske:
@UkiahSmith @p0pr0ck5 Keep in mind that this is against the 0.10 docs, but 0.11 branch was already created as a copy of those, so it needs to be done against that version as well.

p0pr0ck5:
@Tieske would `master` be re-merged into the release branch as part of the release process?

Tieske:
the versions are not branches, they start as a simple new directory and the old version being copied in there

PR #467, opened 2017-08-02
docs(install) update OpenResty version and flags

hishamhm:
* Bump OpenResty version to the latest release
* Remove --without-luajit-lua52, matching setup
  currently used in Travis.

### DIFF
--- a/app/_data/kong_versions.yml
+++ b/app/_data/kong_versions.yml
@@ -99,6 +99,6 @@
     luarocks: "2.4.2"
     cassandra: "3.x.x"
     postgres: "9.4+"
-    openresty: "1.11.2.2"
+    openresty: "1.11.2.4"
     serf: "0.7.0"
   lua_doc: true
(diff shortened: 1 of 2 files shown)


hishamhm:
A few questions: (@thibaultcha?)

1. Does this need to be updated anywhere else? This is the only place I could find
2. I noticed that [app/_data/kong_versions.yml](https://github.com/Mashape/getkong.org/blob/master/app/_data/kong_versions.yml#L91) still mentions Serf in the 0.11 section, and grepping the 0.11.0 branch I see it mentioned in the [Heroku page](https://github.com/Mashape/getkong.org/blob/master/app/install/heroku.md) as well. Can these be removed? (If so, in a separate commit?)

thibaultcha:
@hishamhm 

> Does this need to be updated anywhere else? This is the only place I could find

This should be the only place indeed, along with our `.travis.yml` in the Kong repo, and an environment variable in the distribution scripts. So nothing else user-facing!

> I noticed that app/_data/kong_versions.yml still mentions Serf in the 0.11 section, and grepping the 0.11.0 branch I see it mentioned in the Heroku page as well. Can these be removed? (If so, in a separate commit?)

Both the Heroku page and the `serf` attribute of `kong_versions.yml` can be removed. Those are leftovers 😅 A separate commit would be ideal!

Thanks!

PR #472, opened 2017-08-04
docs(enable-plugins) fix response in example

hishamhm:
Make the documented response match the behavior observed when running the commands with Kong 0.11.0rc2.

### DIFF
--- a/app/docs/0.11.x/getting-started/enabling-plugins.md
+++ b/app/docs/0.11.x/getting-started/enabling-plugins.md
@@ -52,14 +52,14 @@ from unauthorized use.
     ```
 
     Since you did not specify the required `apikey` header or parameter, the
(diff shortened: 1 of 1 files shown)


hishamhm:
@thibaultcha Any preferences on scope/commit msg?

thibaultcha:
I think previous commits touching the quickstart guide are scope `docs(quickstart)`!

hishamhm:
> Isn't this the case with previous versions as well?

Not sure.  I assume at some point in the past that was the correct error message. But Installing the other three versions and re-running the sequence of commands to get to this error would take more time than I was expecting to spend with this edit... I'll commit this one for 0.11 this time, and I'll keep previous versions in mind in case something more critical shows up (for the YAML change I checked what version the format changed and patched previous versions already).

PR #473, opened 2017-08-04
docs(configuration) update to use `kong prepare`

hishamhm:
Update documentation to refer to `kong prepare` instead of `kong compile`, matching the feature set for 0.11. I also revised the entire section for customized Nginx setups, based on my experience learning my way around it, hopefully making it clearer.

### DIFF
--- a/app/docs/0.11.x/configuration.md
+++ b/app/docs/0.11.x/configuration.md
@@ -9,9 +9,9 @@ title: Configuration Reference
 - [Configuration loading](#configuration-loading)
 - [Verifying your configuration](#verifying-your-configuration)
 - [Environment variables](#environment-variables)
-- [Custom Nginx configuration & embedding Kong](#custom-nginx-configuration-embedding-kong)
+- [Customized Nginx setup](#customized-nginx-setup)
   - [Custom Nginx configuration](#custom-nginx-configuration)
-  - [Embedding Kong](#embedding-kong)
+  - [Starting Kong via Nginx](#starting-kong-via-nginx)
 - [Properties reference](#properties-reference)
   - [General section](#general-section)
   - [Nginx section](#nginx-section)
@@ -103,7 +103,7 @@ $ export KONG_LOG_LEVEL=error
 
 [Back to TOC](#table-of-contents)
 
-### Custom Nginx configuration & embedding Kong
+### Customized Nginx setup
 
 Tweaking the Nginx configuration is an essential part of setting up your Kong
 instances since it allows you to optimize its performance for your
@@ -120,9 +120,9 @@ directory, moments before starting Nginx.
 The default template can be found at:
 https://github.com/Mashape/kong/tree/master/kong/templates.  It is split in
 two Nginx configuration files: `nginx.lua` and `nginx_kong.lua`. The former is
-minimalistic and includes the later, which contains everything Kong requires to
-run. Moments before starting Nginx, those two files are copied into the prefix
-directory, which looks like so:
+minimalistic and includes the latter, which contains everything Kong requires to
+run. When `kong start` runs, right before starting Nginx, it copies these
+two files into the prefix directory, which looks like so:
 
 ```
 /usr/local/kong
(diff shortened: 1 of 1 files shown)


app/docs/0.11.x/configuration.md
```diff
@@ -201,19 +206,45 @@ http {
...
+environments, to ensure that the Nginx master process is the foreground
+process in the container, instead of the `kong` CLI command.
+
+To launch Kong from within Nginx, you need an Nginx configuration file
+which includes the Kong Nginx sub-configuration using the `include` directive
 (similar to the examples of the previous section). However, you will need the
-final configuration and not the template. For this, use the `compile` command,
-which outputs a fully-compiled Nginx sub-configuration to `stdout`:
+final configuration and not the template. To prepare this configuration file,
+use the `kong prepare` command, which generates `nginx.conf` and
```
thibaultcha: This command also creates logs files, default self-generated SSL certificates, and the compiled `.kong_env` file. No need to mention the last one, but we should probably mention the others.

We can then say that once `kong compile` is run, all env variables and config files have been compiled and no further tweaking can occur. One can only run `nginx` with the running prefix (and config) all ready. Just hoping this might clarify the process for users who wonder how al of this plays together...

app/docs/0.11.x/configuration.md
```diff
@@ -201,19 +206,45 @@ http {
...
+you can include the generated `nginx-kong.conf` to your own `nginx.conf`
+instead.
+
+Alternatively, if you created your own template with your
+[custom Nginx configuration](#custom-nginx-configuration) you can also
+use the `--nginx-conf` flag with `kong prepare`:
 
-# now start OpenResty with a configuration that "includes" nginx-kong.conf
-$ nginx -c /usr/local/openresty/conf/nginx.conf
+``` bash
```
thibaultcha: (Didn't run this branch myself but just double checking: this space before `bash` is accepted by the Markdown parser and will not produce some strange quirk, will it?)

hishamhm: No quirks, I'm running the branch here on localhost:3000 and everything looks fine.

thibaultcha: :+1:

app/docs/0.11.x/configuration.md
hishamhm: Thanks for the feedback!

> This command also creates logs files, default self-generated SSL certificates

Only if they don't exist already, right? Re-running `kong prepare` does not seem to overwrite the files in `logs/` or `ssl/`.

> We can then say that once kong compile is run, all env variables and config files have been compiled and no further tweaking can occur. 

(You mean `kong prepare`, right?)

As I understand it, no further tweaking of `kong.conf` or `custom_nginx.template` takes effect after `kong prepare`, but the user may further tweak the generated `nginx.conf`, right? I alluded to this in later in the text. But I can make this clearer saying explicitly which files are and which ones aren't available for tweaking after `kong prepare`.

thibaultcha: > Only if they don't exist already, right?

Yes indeed!

> (You mean kong prepare, right?)

Damn... Yes.

> As I understand it, no further tweaking of kong.conf or custom_nginx.template takes effect after kong prepare, but the user may further tweak the generated nginx.conf, right? I alluded to this in later in the text. But I can make this clearer saying explicitly which files are and which ones aren't available for tweaking after kong prepare.

Yes, one can still tweak the `nginx.conf` and `nginx-kong.conf` files, but other settings that go into the Lua VM (like the PostgreSQL host, or other similar values) cannot be modified anymore. (Well, technically they can, since they are in `.kong_env`)

hishamhm: Added an extra bit mentioning logs and ssl, and a summary at the end of the section with a tl;dr about the relationship of the various files.

hishamhm: At this point, to make it clearer, I'd have to add a diagram 😃

thibaultcha: Haha, no need, this is great!

hishamhm:
Thank you for the review! Merging it in, then!

PR #474, opened 2017-08-07
docs(plugin-dev) use new config file format

hishamhm:
Remove lingering reference to the YAML config file format,
updating it to the custom kong.conf format.
The syntax-highlighting marker in Markdown remains tagged as
`yaml` only to gray out the comment.


thibaultcha:
(Part of this could already be merged against master and published today too)

thibaultcha:
(Also, let's avoid merge commits! Rebasing or squashing are encouraged - squashing of a single commit even lets you edit the scope with no additional efforts before the merge :wink:)

hishamhm:
If I use "Squash and merge" here the commit will pick up the title of the PR (which I can edit here), is that correct?

thibaultcha:
Quite! Sounds good :+1: (you can also edit the commit right before confirming the squash)

PR #475, opened 2017-08-09
docs(cors) fix origin value

WALL-E:
Two pages have the same origin if the protocol, port (if one is specified), and host are the same for both pages.

ref: https://tools.ietf.org/html/rfc6454#section-4


thibaultcha:
Correct, thank you!

PR #479, opened 2017-08-09
docs(configuration) update Nginx templates to Kong 0.11.0

thibaultcha:


thibaultcha:
(Let's try to avoid merge commits when not necessary in the future)

PR #481, opened 2017-08-10
(docs) proxy example explanation error

shakedel:
Either I did not understand the documentation or there is a minor error


thibaultcha:
Manually merged with some edits to ensure linting passes, thanks!

PR #490, opened 2017-08-15
feat(install) update download locations

salazar:
WIP

* Update download locations to point to Bintray (yay);
* Update package names - `kong` is now `kong-community-edition`
* Add Debian 9 and Ubuntu Zesty;
* Remove OSX `.pkg` from the intall page - use Homebrew instead.

### DIFF
--- a/app/install/centos.md
+++ b/app/install/centos.md
@@ -11,15 +11,15 @@ breadcrumbs:
 
 Start by downloading the corresponding package for your configuration:
 
-- [CentOS 6]({{ site.links.download }}/{{site.data.kong_latest.version}}/kong-{{site.data.kong_latest.version}}.el6.noarch.rpm)
-- [CentOS 7]({{ site.links.download }}/{{site.data.kong_latest.version}}/kong-{{site.data.kong_latest.version}}.el7.noarch.rpm)
+- [CentOS 6]({{ site.links.download }}/kong-community-edition-rpm/download_file?file_path=dists/kong-community-edition-{{site.data.kong_latest.version}}.el6.noarch.rpm)
+- [CentOS 7]({{ site.links.download }}/kong-community-edition-rpm/download_file?file_path=dists/kong-community-edition-{{site.data.kong_latest.version}}.el7.noarch.rpm)
 
 ### YUM Repositories
 
-You can also install Kong by using the following YUM repositories and following the Bintray instructions:
+You can also install Kong via YUM; follow the instructions on the "Set Me Up"
+section on the page below.
 
-- [CentOS 6 YUM](https://bintray.com/mashape/kong-rpm-el6-{{site.data.kong_latest.release}})
-- [CentOS 7 YUM](https://bintray.com/mashape/kong-rpm-el7-{{site.data.kong_latest.release}})
+- [RPM Repository](https://bintray.com/kong/kong-community-edition-rpm)
 
 ----
 
@@ -31,7 +31,7 @@ You can also install Kong by using the following YUM repositories and following 
 
     ```bash
     $ sudo yum install epel-release
-    $ sudo yum install kong-{{site.data.kong_latest.version}}.*.noarch.rpm --nogpgcheck
+    $ sudo yum install kong-community-edition-{{site.data.kong_latest.version}}.*.noarch.rpm --nogpgcheck
     ```
 
 2. **Prepare your database**
--- a/app/install/debian.md
+++ b/app/install/debian.md
@@ -11,15 +11,17 @@ breadcrumbs:
 
 Start by downloading the corresponding package for your configuration:
 
-- [7 Wheezy]({{ site.links.download }}/{{site.data.kong_latest.version}}/kong-{{site.data.kong_latest.version}}.wheezy_all.deb)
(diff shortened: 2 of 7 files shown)


salazar:
@thibaultcha reopened it, based on `0.11.0`. It's WIP to avoid getting it merged before I solve an issue I'm working on. Will do tomorrow morning; in the meanwhile, I'd appreciate a second pair of eyes on where else we need to update those locations! : )

app/install/centos.md
```diff
@@ -11,15 +11,15 @@ breadcrumbs:
...
 
 ### YUM Repositories
 
-You can also install Kong by using the following YUM repositories and following the Bintray instructions:
+You can also install Kong via YUM; follow the instructions on the "Set Me Up"
+section on the page below.
 
-- [CentOS 6 YUM](https://bintray.com/mashape/kong-rpm-el6-{{site.data.kong_latest.release}})
-- [CentOS 7 YUM](https://bintray.com/mashape/kong-rpm-el7-{{site.data.kong_latest.release}})
+- [RPM Repository](https://bintray.com/kong/kong-community-edition-rpm)
```
thibaultcha: * Concerning the YUM and APT repositories, a better "call to action" would be to directly inline the repositories sources and instructions. 
* Granted, the scope of this PR is not to provide inline repositories setup instructions, only to update the links - just a thought!
* Additionally, the instructions specified by bintray don't include the obligatory `apt-get update` step, or is that step not needed? Or are we counting on the below section: `If you are downloading the package, execute: sudo apt-get-update`? I feel like the "Install Kong" section could be split between manual download of the package, and download via the YUM/APT repositories. Not in this PR's scope either.

thibaultcha:
Proper commit scope: `feat(install)` - it's not a fix if it wasn't broken ;) It's just a necessary update due to our internal decisions :)

app/install/debian.md
```diff
@@ -11,15 +11,16 @@ breadcrumbs:
...
-- [8 Jessie]({{ site.links.download }}/{{site.data.kong_latest.version}}/kong-{{site.data.kong_latest.version}}.jessie_all.deb)
+- [7 Wheezy]({{ site.links.download }}/kong-community-edition-deb/download_file?file_path=dists/kong-community-edition-{{site.data.kong_latest.version}}.wheezy_all.deb)
+- [8 Jessie]({{ site.links.download }}/kong-community-edition-deb/download_file?file_path=dists/kong-community-edition-{{site.data.kong_latest.version}}.jessie_all.deb)
 
 ### APT Repositories
 
-You can also install Kong by using the following APT repositories and following the Bintray instructions:
+You can also install Kong via APT; follow the instructions on the "Set Me Up"
+section on the page below, setting  *distribution* to the appropriate value
+(e.g., `wheezy`) and *components* to `main`.
```
salazar: This is a controversial topic; some say that it depends. I prefer to go with what's intuitive; "e.g." means "for instance" and a comma should follow in that case. Same goes for "i.e.", which would translate to something like "that is". After a quick Google search, I'd prefer to keep the commas.

app/install/centos.md
salazar: Those are all valid points; will work on them in a different PR. Works?

salazar:
Thanks for the commit message type/scope explanation; I sometimes struggle to find the proper one ; ) It sounds weird to call it a "feature" change.

app/install/debian.md
thibaultcha: I never realized this was controversial! (Never saw a comma being used until now) I did some googling around as well and you are right, it is quite controversial. I'm ok with one or the other, but our docs should be consistent in its usage, that's all! :)

thibaultcha:
Labelled as "ready but on hold" until we have the packages as discussed!

app/install/osx.md
```diff
@@ -18,9 +17,10 @@ breadcrumbs:
 
 1. **Install Kong**
 
-    Use the `.pkg` installer or the Homebrew formula. For the Homebrew formula, follow the instructions described in the repository.
-
-    **Note**: After downloading the `.pkg` installer, you will have to **right click**, select "Open" and authorize it.
+    ```
```
thibaultcha: Shall we include a short sentence to give some additional context here?
```
Use the [Homebrew](https://brew.sh/) package manager to add Kong as a tap and install it:

...
```

salazar: Good point. Done.

app/install/osx.md
```diff
@@ -9,18 +9,20 @@ breadcrumbs:
...
 
-- [.pkg Installer]({{ site.links.download }}/{{site.data.kong_latest.version}}/kong-{{site.data.kong_latest.version}}.osx.pkg)
 - [Homebrew Formula]({{ site.repos.homebrew }})
 
 ----
 
 ### Installation
 
-1. **Install Kong**
+Use the [Homebrew](https://brew.sh/) package manager to add Kong as a tap and install it:
```
thibaultcha: Shouldn't this move right above the code snippet and under "Install Kong"? Seems a bit out of context there no?

salazar: Wouldn't say out of context; before the actual installation steps to make sure the user is familiar with it before jumping into it?

salazar: (either way works for me)

salazar: fixed

PR #491, opened 2017-08-16
docs(plugin-dev) new 0.11.0 cache invalidation mechanism

thibaultcha:


app/docs/0.11.x/plugin-development/custom-logic.md
```diff
@@ -171,18 +185,59 @@ return CustomHandler
...
+loggly                    |
+syslog                    |
+statsd                    |
+runscope                  |
+http-log                  |
+request-termination       |
+tcp-log                   |
+galileo                   |
+correlation-id            |
+file-log                  |
```
thibaultcha: need to add those priorities now that @hishamhm submitted #2812 as we discussed!

hishamhm: want me to send the reordered list?

app/docs/0.11.x/plugin-development/entities-cache.md
```diff
@@ -15,195 +15,275 @@ chapter: 7
...
 ```
-local cache = require "kong.tools.database_cache"
+local singletons = require "kong.singletons"
+local cache = singletons.cache
 ```
 
 There are 2 levels of cache:
 
-1. Lua memory cache (local to an nginx worker)
+1. L1: Lua memory cache (local to an nginx worker)
```
hishamhm: style suggestion: 
```
1. L1: Lua memory cache - local to an nginx worker
```

app/docs/0.11.x/plugin-development/entities-cache.md
```diff
@@ -15,195 +15,275 @@ chapter: 7
...
    This can hold any type of Lua value.
-2. Shared memory cache - SHM (local to an nginx node, but shared between all workers)
-   This can only hold scalar values, and hence requires (de)serialization.
+2. L2: Shared memory cache - SHM (local to an nginx node, but shared between all
+   workers). This can only hold scalar values, and hence requires
```
hishamhm: style suggestion:

```
2. L2: Shared memory cache (SHM) - local to an nginx node, but shared between all
   workers. This can only hold scalar values, and hence requires
```

app/docs/0.11.x/plugin-development/entities-cache.md
```diff
@@ -15,195 +15,275 @@ chapter: 7
...
-| `value, ... = cache.get_or_set(key, ttl, function, ...)` | This is a utility method that retrieves an object with the specified key, but if the object is `nil` then the passed function will be executed instead, whose return value will be used to store the object at the specified key. This effectively makes sure that the object is only loaded from the datastore one time, since every other invocation will load the object from the in-memory cache.
+Function name                                 | Description
+----------------------------------------------|---------------------------
+`value, err = cache:get(key, opts?, cb, ...)` | Retrieves the value from the closest cache level. If no cache (L1 nor L2) has the value, invokes `cb` in protected mode. `cb` must return one (and only one) value that will be cached. It *can* throw errors, as those will be caught and properly logged by Kong, at the `ngx.ERR` level. This function **does** cache negative results (`nil`). As such, one must rely on its second argument `err` when checking for errors.
```
hishamhm: What is `opts?`?

app/docs/0.11.x/plugin-development/entities-cache.md
thibaultcha: Sounds good to me. This documentation is way too light tbh, compared to the mlcache documentation that this `kong.cache` module actually deserves... See https://github.com/thibaultcha/lua-resty-mlcache/tree/docs/readme :D

app/docs/0.11.x/plugin-development/entities-cache.md
```diff
@@ -15,195 +15,275 @@ chapter: 7
...
+`ttl, err, value = cache:probe(key)`          | Checks if a value is cached in the L2 cache. If it is, returns its remaining TTL. It not, returns `nil`. The value being cached can also be a negative caching. The third return value is the value being cached itself.
+`cache:invalidate_local(key)`                       | Evicts a value from the node's cache (L1 and L2).
+`cache:invalidate(key)`                             | Evicts a value from the node's cache **and** propagates the eviction events to all other nodes in the cluster.
+`cache:purge()`                                     | Evicts **all** values from the node's cache (L1 and L2).
```
hishamhm: If I am reading the documentation correctly, from the programmer API perspective, the existence of two levels of cache is transparent, isn't it? When data is in L1 it is always in L2 as well, and operations on the `cache` object always operate on both levels, or on L2 directly. In other words, apart from speed one can't "manipulate" L1 explicitly. 

(For instance, one can't keep non-serializable data in L1, right?)

If that's the case and the levels are transparent, it would be best to keep the explanation more abstract in the API specification, referring only to the `node's cache` rather than specifying levels. (The least we specify, the more we have room to change things later.) The explanation about L1 and L2 could then be restricted to a section discussing the implementation, without tying it to the programmer-visible API.

app/docs/0.11.x/plugin-development/entities-cache.md
```diff
@@ -15,195 +15,275 @@ chapter: 7
...
   return responses.send_HTTP_FORBIDDEN("Invalid authentication credentials")
 end
+
+-- set an upstream header if the credential exists and is valid
+ngx.req.set_header("X-API-Key", credential.apikey)
 ```
 
-By doing so it doesn't matter how many requests the client makes with that particular api-key, after the first request every lookup will be done in-memory without querying the datastore.
+With the above mechanism in place, once a Consumer has made a request with
+his or her API key, the cache will be considered warm and subsequent requests
```
hishamhm: Style suggestion: `his or her` → `their` ( https://en.wikipedia.org/wiki/Singular_they )

app/docs/0.11.x/plugin-development/entities-cache.md
```diff
@@ -15,195 +15,275 @@ chapter: 7
...
+won't result in a database query.
 
 #### Updating or deleting a custom entity
 
-Every time a cached custom entity is updated or deleted on the datastore, for example using the Admin API, it creates an inconsistency between the data in the datastore, and the data cached in-memory in the Kong node. To avoid this inconsistency, we need to delete the cached entity from the in-memory store and force Kong to request it again from the datastore. In order to do so we must implement an invalidation hook.
+Every time a cached custom entity is updated or deleted in the datastore (i.e.
+using the Admin API), it creates an inconsistency between the data in
+the datastore, and the data cached in the Kong nodes memory. To avoid this
```
hishamhm: `in the Kong nodes' memory`, or `in the memories of Kong nodes`, I think?

app/docs/0.11.x/plugin-development/entities-cache.md
```diff
@@ -15,195 +15,275 @@ chapter: 7
...
 
 ---
 
-### Invalidating custom entities
+### Cache invalidation for your entities
 
-Every time an entity is being created/updated/deleted in the datastore, Kong notifies the datastore operation across all the nodes telling what command has been executed and what entity has been affected by it. This happens for APIs, Plugins and Consumers, but also for custom entities.
+If you wish that your cached entities be invalidated upon a CRUD operation
+rather than having to wait for them to reach their TTL, you have to follow
+a few steps. This process can ba automated since 0.11.0 for most entities,
```
hishamhm: typo: `can be`

app/docs/0.11.x/plugin-development/entities-cache.md
```diff
@@ -15,195 +15,275 @@ chapter: 7
...
+but manually subscribe to some CRUD events might be required to invalidate
```
hishamhm: `subscribe` → `subscribing`

app/docs/0.11.x/plugin-development/entities-cache.md
```diff
@@ -15,195 +15,275 @@ chapter: 7
...
+some entities with ore complex relationships.
```
hishamhm: typo: `ore` → `more`

app/docs/0.11.x/plugin-development/entities-cache.md
```diff
@@ -15,195 +15,275 @@ chapter: 7
...
+end
+
+-- do something with the credential
 ```
 
-In the example above the plugin is listening to the `ENTITY_UPDATED` and `ENTITY_DELETED` events and responding by invoking the appropriate function. The `message_t` table contains the event properties:
+If the `cache_key` is generated like so and specified in an entity's schema,
+cache invalidation will be an automatic process: every CRUD operation that
+affects this API key will be make Kong generate the affected `cache_key`, and
+send broadcast it to all of the other nodes on the cluster so they can evict
```
hishamhm: `send broadcast` → `broadcast`

app/docs/0.11.x/plugin-development/custom-logic.md
thibaultcha: @hishamhm Thanks!

app/docs/0.11.x/plugin-development/entities-cache.md
thibaultcha: I'm afraid this format is really not appropriate to document `opts` :( So I left it out... It's documented in the upcoming release of lua-resty-mlcache for now (internal or advanced documentation maybe). I think ideally once we have some more time, we can improve this document to instead point to a more complete reference of the available API...

app/docs/0.11.x/plugin-development/entities-cache.md
thibaultcha: Yeah that's true - I mostly updated this doc to the new API, mentions of the caching levels exited before, and I'm also biased because of the lua-resty-mlcache required technical documentation 😅

thibaultcha:
@hishamhm Thanks! All updated, including the plugins ordering table!

PR #499, opened 2017-08-21
fix(docs): Fix example in entities cache: Use dao find_all method instead of find_by_keys

GuillaumeSmaha:
In DAO docs, the method `find_by_keys` doesn't exist since version 0.8.x.

### DIFF
--- a/app/docs/0.10.x/plugin-development/entities-cache.md
+++ b/app/docs/0.10.x/plugin-development/entities-cache.md
@@ -73,7 +73,7 @@ Bringing back our authentication plugin example, to lookup a credential with a s
 local function load_entity_key(api_key)
   -- IMPORTANT: the callback is executed inside a lock, hence we cannot terminate
   -- a request here, we MUST always return.
-  local apikeys, err = dao.apikeys:find_by_keys({key = api_key}) -- Lookup in the datastore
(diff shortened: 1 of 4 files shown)


app/docs/0.10.x/plugin-development/entities-cache.md
```diff
@@ -73,7 +73,7 @@ Bringing back our authentication plugin example, to lookup a credential with a s
 local function load_entity_key(api_key)
   -- IMPORTANT: the callback is executed inside a lock, hence we cannot terminate
   -- a request here, we MUST always return.
-  local apikeys, err = dao.apikeys:find_by_keys({key = api_key}) -- Lookup in the datastore
+  local apikeys, err = dao.apikeys:find({key = api_key}) -- Lookup in the datastore
```
thibaultcha: The method that replaced `find_by_keys` is `find_all`. `find()` expects only the primary key of an entity and returns a single value, not an array of rows. Sorry!

GuillaumeSmaha: Yes, that right !
I my case, I had a primary key.
I will update to use `find_all`

GuillaumeSmaha:
@thibaultcha Nice catch ! 
I update with `find_all`
Thanks

thibaultcha:
Thank you!

PR #500, opened 2017-08-22
docs(rate-limiting) add hide_client_headers reference

xiii:
There is a parameter introduced with https://github.com/Mashape/kong/commit/209872eb800215144d35d39dff353c4348b07874 called **hide_client_headers** that is missing from the rate-limiting plugin documentation.


thibaultcha:
@xiii Thanks! Two things though:

1. Can we add this attribute to the `response-ratelimiting` plugin as well? It has been added to it by the same commit you referenced.
2. Could we move this attribute above the redis options in this list of attributes? I believe it could use some more visibility.

Thanks!

xiii:
cheers @thibaultcha ! updated after your comments, let me know if there is anything more.

thibaultcha:
@xiii Thanks!

PR #517, opened 2017-09-11
added explanation on targets 

mvanholsteijn:
As discussed in issue [#2876](https://github.com/Mashape/kong/issues/2876), the concept of deleting target objects was not quite clear. 

In kong 0.11 the /upstreams/<id>/targets/active API gives a 'method not allowed'. So I removed it from the documentation too.

### DIFF
--- a/app/docs/0.11.x/admin-api.md
+++ b/app/docs/0.11.x/admin-api.md
@@ -1711,6 +1711,54 @@ HTTP 200 OK
 
 ---
 
+### List active targets
+
+Retrieve a list of active targets (targets whose most recent weight is not 0)
+for a given upstream.
+
+<div class="alert alert-warning">
+  <strong>Note:</strong> This endpoint is only available with Kong 0.10.1+
+</div>
+
+### Endpoint
+
+<div class="endpoint get">/upstreams/{name or id}/targets/active/</div>
+
+Attributes | Description
+---:| ---
+`name or id`<br>**required** | The unique identifier **or** the name of the upstream for which to list the targets.
+
+#### Response
+
+```
+HTTP 200 OK
+```
+
+```json
+{
+    "total": 2,
+    "data": [
+        {
+            "created_at": 1485524883980,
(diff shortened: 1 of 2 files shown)


app/docs/0.11.x/admin-api.md
```diff
@@ -1705,53 +1711,6 @@ HTTP 200 OK
...
-Retrieve a list of active targets (targets whose most recent weight is not 0)
-for a given upstream.
-
-<div class="alert alert-warning">
-  <strong>Note:</strong> This endpoint is only available with Kong 0.10.1+
-</div>
-
-### Endpoint
-
-<div class="endpoint get">/upstreams/{name or id}/targets/active</div>
```
thibaultcha: This endpoint seems to be mis-represented. It needs a trailing slash like so:
```
/upstreams/{name or id}/targets/active/
```

The changes proposed in this PR look healthy to me (making the `created_at` field's meaning explicit - but instead of removing this endpoint, we should probably simply add the trailing slash it needs.

It used not to require a trailing slash, but recent changes in LuaJIT and Lapis made it so that it is now required. If the endpoint was declared without a trailing slash in the Kong codebase, it could - once again - not require a trailing slash. Again, this is a breaking change introduced by recent changes in LuaJIT and Lapis.

thibaultcha:
Thanks for the suggestion - I believe the intent is clearly aligned our needs here, and that the `active` endpoint needs to be taken care of differently :)

mvanholsteijn:
Oke, I have restored the /targets/active API documentation, and included the trailing /

thibaultcha:
Thank you!

app/docs/0.11.x/admin-api.md
```diff
@@ -1711,6 +1711,54 @@ HTTP 200 OK
 
 ---
 
+### List active targets
+
+Retrieve a list of active targets (targets whose most recent weight is not 0)
+for a given upstream.
+
+<div class="alert alert-warning">
```
Tieske: Since this is the tree for 0.11 docs, this warning can be omitted here

PR #518, opened 2017-09-12
fix(openwhisk) fix wrong host in example

shashiranjan84:
fix https://github.com/Mashape/kong/issues/2886.


thibaultcha:
@shashiranjan84 Could we keep `example.com` as the example value (like we started doing in most parts of the documentation) instead? Sorry. Should just be a few search and replace!

shashiranjan84:
Done @thibaultcha.

thibaultcha:
Thanks!

PR #522, opened 2017-09-22
fix(plugins) precedence description

Tieske:
@coopr ping


thibaultcha:
Rewritten for clarify and applying to other versions of the same doc. Thanks!

PR #528, opened 2017-10-13
fix(docs) mockbin

coopr:


thibaultcha:
Thanks! Manually merged with proper commit message and the same fix for the 0.10.x version of the docs!

PR #541, opened 2017-10-31
feat(balancer) adds consistent hashing description

Tieske:
related to PR https://github.com/Kong/kong/pull/2875


app/docs/0.12.x/loadbalancing.md
```diff
@@ -179,7 +193,67 @@ __Exception__: When a DNS record has `ttl=0`, the hostname will be added
...
+The ring-balancer is designed to work both with a single node as well as in a cluster.
+For the weighted-round-robin algorithm there isn't much difference, but when using
+the hash based algorithm it is important that all nodes build the exact same
+ring-balancer to make sure they all work identical. To do this the balancer
+must be build completely deterministic. As such DNS can not be used as the
+balancers might/will slowly diverge because the DNS ttl has only second precision
+and renewal is determined by when a name is actually requested.
+
+So when using the hashing approach in a Kong cluster, add `target` entities only
+by their IP address, and never by name.
```
coopr: Is this enforced "in the code" or is this advised only in the documentation? If the latter, I think it may be insufficient to keep users out of this situation - but I'm not clear what a better solution it.

app/docs/0.12.x/loadbalancing.md
```diff
@@ -87,10 +76,32 @@ The DNS resolver will start resolving the following record types in order:
...
+- Whenever the DNS record is refreshed a list is generated to handle the
+weighting properly. Try to keep the weights as multiples of each other to keep
+the algorithm performant, e.g., 2 weights of 17 and 31 would result in a structure 
+with 527 entries, whereas weights 16 and 32 (or their smallest relative 
+counterparts 1 and 2) would result in a structure with merely 3 entries,
+especially with a very small (or even 0) `ttl` value.
+
+- Some nameservers do not return all entries (due to UDP packet size) in those
+cases (for example Consul returns a maximum of 3) Kong will only use the ones
+provided and hence be limited. To mitigate this use another nameserver, use IP
```
coopr: "...a given Kong node will only use the few upstream service instances provided by the nameserver. In this scenario, it is possible that the pool of upstream instances will be loaded inconsistently, because Kong is effectively unaware of some of the instances, due to the limited information provided by the nameserver."

app/docs/0.12.x/loadbalancing.md
Tieske: I reworked the docs, with an explicit "caveats" section also for the balancer, I think that the casual reader will be more inclined to read that explcitly

app/docs/0.12.x/loadbalancing.md
```diff
@@ -87,10 +76,36 @@ The DNS resolver will start resolving the following record types in order:
...
+This order is configurable through the [`dns_order` configuration property][dns-order-config].
+
+[Back to TOC](#table-of-contents)
+
+#### **DNS caveats**
+
+- Whenever the DNS record is refreshed a list is generated to handle the
+weighting properly. Try to keep the weights as multiples of each other to keep
+the algorithm performant, e.g., 2 weights of 17 and 31 would result in a structure 
+with 527 entries, whereas weights 16 and 32 (or their smallest relative
```
thibaultcha: Please, for the sake of us, configure your editor to strip those trailing whitespaces. They make everyone else's diffs noisier than they should be and just bring conflict. See: https://github.com/Kong/kong/pull/3083

app/docs/0.12.x/loadbalancing.md
```diff
@@ -87,10 +76,36 @@ The DNS resolver will start resolving the following record types in order:
...
+
+- Whenever the DNS record is refreshed a list is generated to handle the
+weighting properly. Try to keep the weights as multiples of each other to keep
+the algorithm performant, e.g., 2 weights of 17 and 31 would result in a structure 
+with 527 entries, whereas weights 16 and 32 (or their smallest relative 
+counterparts 1 and 2) would result in a structure with merely 3 entries,
+especially with a very small (or even 0) `ttl` value.
+
+- Some nameservers do not return all entries (due to UDP packet size) in those
+cases (for example Consul returns a maximum of 3) a given Kong node will only
```
thibaultcha: Slightly confusing here: which cases is "those cases" referring to?

app/docs/0.12.x/loadbalancing.md
```diff
@@ -146,6 +160,9 @@ the initial setup only features 2 targets.
 The tradeoff here is that the higher the number of slots, the better the random 
 distribution, but the more expensive the changes are (add/removing targets)
 
+Detailed information on adding and manipulating
+upstreams is available in the `upstream` section of the [Admin API reference][upstream-object-reference].
```
thibaultcha: it seems like the below change is trying to limit the umber of columns (laudable!), but this one was left untouched :(

app/docs/0.12.x/loadbalancing.md
```diff
@@ -179,7 +197,70 @@ __Exception__: When a DNS record has `ttl=0`, the hostname will be added
...
+[Back to TOC](#table-of-contents)
+
+#### **Balancing algorithms**
+
+By default a ring-balancer will use a weighted-round-robin scheme. The alternative
+would be to use the hash-based algorithm. The input for the hash can be either
+`none`, `consumer`, `ip`, or `header`. When set to `none` the
+weighted-round-robin scheme will be used, and hashing will be disabled.
+
+There two options, a primary and a fallback in case the primary fails (eg. if the
```
thibaultcha: typo: There _are_? Maybe we should not use a comma here.

thibaultcha: Also, we consistently use `(e.g., ...)` in our docs, not `eg.`.

app/docs/0.12.x/loadbalancing.md
```diff
@@ -179,7 +197,70 @@ __Exception__: When a DNS record has `ttl=0`, the hostname will be added
...
+primary is set to `consumer`, but no consumer is authenticated)
+
+The different hashing options:
+
+- `none`: Do not use hashing, but use weighted-round-robin instead (default)
```
thibaultcha: missing a dot at the end of this bullet point (the dot is present in every other bullet point).

app/docs/0.12.x/loadbalancing.md
```diff
@@ -179,7 +197,70 @@ __Exception__: When a DNS record has `ttl=0`, the hostname will be added
...
+
+[Back to TOC](#table-of-contents)
+
+#### **Balancing caveats**
+
+The ring-balancer is designed to work both with a single node as well as in a cluster.
+For the weighted-round-robin algorithm there isn't much difference, but when using
+the hash based algorithm it is important that all nodes build the exact same
+ring-balancer to make sure they all work identical. To do this the balancer
+must be build completely deterministic.
```
thibaultcha: "must be build completely deterministic" does not sound like proper English? It also sounds redundant. Alternatives could be:
- must be built in a deterministic fashion
- must be built in a deterministic way
- must be built deterministically

app/docs/0.12.x/loadbalancing.md
hishamhm: @thibaultcha: just a side note, in this repo's [.editorconfig](https://github.com/Kong/getkong.org/blob/master/.editorconfig), `trim_trailing_whitespace` is explicitly set to `false` for `.md` files. (IIRC trailing whitespaces _are_ significant for Markdown, e.g. when marking code blocks via indentation and those contain indented blank lines.) 

So the best advice here is that everyone should use [editorconfig](https://editorconfig.org)-enabled editors!

thibaultcha: This is true but not the reason why `trim_trailing_whitespace` is disabled in this repository. Added in #374, the goal of the PR was to produce an editor config that would validate the current state of this source, nothing more.
Afaik, there are no significant trailing whitespaces in this entire website. Being its biggest contributor, and always trimming those, I never encountered any issue with it.
Once again, we are creating a toxic feedback loop over ridiculous problems.

Tieske:
@thibaultcha comments addressed.

PR #543, opened 2017-11-02
EE forward proxy

travispamaral:


travispamaral:
[(2) Create plugins hub entry for EE-only Forward Proxy plugin](https://trello.com/c/0pQLAEs6/244-2-create-plugins-hub-entry-for-ee-only-forward-proxy-plugin)

p0pr0ck5:
LGTM

PR #545, opened 2017-11-03
Add security note about handling private keys+HTTP

pduldig-at-tw:
Added a warning about disclosure of private keys over HTTP interface for Certificate API

### DIFF
--- a/app/docs/0.11.x/admin-api.md
+++ b/app/docs/0.11.x/admin-api.md
@@ -1078,6 +1078,8 @@ HTTP 201 Created
     "created_at": 1485521710265
 }
 ```
+#### Security Note
+Warning: Posting the certificate and private key to an HTTP endpoint may lead to inadvertent disclosure of the private key. Ensure all requests involving certificates are performed over https only.
 
 ---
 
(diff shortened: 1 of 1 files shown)


p0pr0ck5:
I wonder if there should be some generic discussion about this here as well: https://getkong.org/docs/0.11.x/secure-admin-api/, either in addition to, or as a replacement, for this, as well as https://getkong.org/docs/0.11.x/admin-api/, which notes explicitly "This API is designed for internal use and provides full control over Kong".

For completeness sake, I hypothetically would say this should probably be extended to include other sensitive data, like redis credentials in the rate limiting plugins. That's a bit of a straw man though; it seems more appropriate to write some generic notes about TLS traffic. Various use cases and threat models will consider various portions of API and object configuration sensitive, and while I certainly agree something like private keys is indeed sensitive, we should be careful in making such assumptions about other response elements. Thus instead I think a general call-out in a central location in the document would be appropriate (and some such discussion does exist, albeit without explicit reference to TLS traffic) rather than attempting to cherry-pick specific portions in a doc. Thoughts?

thibaultcha:
> Thus instead I think a general call-out in a central location in the document would be appropriate

:+1:

kikito:
I have created a new issue for doing what @p0pr0ck5 proposed, and will close this PR. Thanks @pduldig-at-tw for taking the time to send the PR which ultimately generated the discussion!

PR #548, opened 2017-11-08
docs(admin) clarify PUT endpoints behavior

thibaultcha:
As a follow-up to:

Kong/kong#3007 (PR)
Kong/kong#2774 (issue)

This commit adds clarifications as to the current behavior of PUT
endpoints on the Admin API. As much as we acknowledge that their
behavior is not compliant with the HTTP specifications, we do wish to
clarify their behavior to avoid confusion as of today, until a different
implementation sees the light of day.

### DIFF
--- a/app/docs/0.11.x/admin-api.md
+++ b/app/docs/0.11.x/admin-api.md
@@ -411,7 +411,12 @@ HTTP 200 OK
 
 {{ page.api_body }}
 
-The body needs an `id` parameter to trigger an update on an existing entity.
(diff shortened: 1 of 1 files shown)


thibaultcha:
We could also add this note to previous versions of the Admin API docs, if said note resonates with everybody here.

kikito:
Merging this - adding it to previous docs is ok but could be done as part of a separate PR (or if someone raises the issue on the previous versions of kong)

PR #549, opened 2017-11-09
Added note about 100 max query parameter limitation of the admin api …

mikob:
…when configuring plugins using an array

See discussion here: https://github.com/Kong/kong/issues/3017


kikito:
Thanks for taking the time to send this PR and apologies for the time it took us to review and merge it!

PR #551, opened 2017-11-20
docs(jwt) add clarification for consumer parameter in jwt

kikito:
Related with https://github.com/Kong/kong/issues/2898


app/plugins/jwt.md
```diff
@@ -59,7 +59,7 @@ form parameter                          | default | description
 `config.claims_to_verify`<br>*optional* |         | A list of registered claims (according to [RFC 7519][rfc-jwt]) that Kong can verify as well. Accepted values: `exp`, `nbf`.
 `config.key_claim_name`<br>*optional*   | `iss`   | The name of the claim in which the `key` identifying the secret **must** be passed.
 `config.secret_is_base64`<br>*optional* | `false` | If true, the plugin assumes the credential's `secret` to be base64 encoded. You will need to create a base64 encoded secret for your Consumer, and sign your JWT with the original secret.
-`config.anonymous`<br>*optional*        | ``      | An optional string (consumer uuid) value to use as an "anonymous" consumer if authentication fails. If empty (default), the request will fail with an authentication failure `4xx`.
+`config.anonymous`<br>*optional*        | ``      | An optional string (consumer uuid) value to use as an "anonymous" consumer if authentication fails. If empty (default), the request will fail with an authentication failure `4xx`. Please note that it must be the internal kong UUID, not the custom user ID.
```
thibaultcha: Formatting seems a bit odd to me. Could we update to:
```
Please note that this value must refer to the Consumer `id` attribute which is internal to Kong, and **not** its `custom_id`.
```

This way we referred to the attribute names, we did not assume that `custom_id` could or could not be a UUID, and we capitalized "Kong". What do you think?

kikito: Thanks for reviewing! I have reformatted the phrase as you mentioned and also added it to other `anonymous` params in other plugins (only in 0.11)

thibaultcha:
Thank you!!

> (only in 0.11)

Yes, plugins docs isn't versioned yet anyways, so this property is visible to all versions :/

PR #555, opened 2017-11-22
docs(hmac-auth) add 2 new endpoints to list hmac-auths

hbagdi:
Two new endpoints have been added to the hmac-auth plugin:
* `/hmac-auths/` to paginate through all hmac-auth credentials
* `/hmac-auths/:credential_key_or_id/consumer` to retrieve the
Consumer associated with a credential

See https://github.com/Kong/kong/pull/3009


thibaultcha:
Manually merged as well, thank you very much!

PR #556, opened 2017-11-22
docs(jwt) add 2 new endpoints to list JWTs

hbagdi:
Two new endpoints have been added to the JWT plugin:

* `/jwts/` to paginate through all JWT credentials
* `/jwts/:jwt_key_or_id/consumer` to retrieve the Consumer
associated with a JWT

See https://github.com/Kong/kong/pull/3003

### DIFF
--- a/app/plugins/jwt.md
+++ b/app/plugins/jwt.md
@@ -24,6 +24,8 @@ nav:
       - label: Generate public/private keys
       - label: Using the JWT plugin with Auth0
       - label: Upstream Headers
+      - label: Paginate through the JWT credentials
+      - label: Retrieve the Consumer associated with a JWT credential
 ---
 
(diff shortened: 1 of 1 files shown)


thibaultcha:
Manually merged as well! We are only left with the ACL updates - not sure if we will include them in 0.11.2 yet because the PR to Kong is more recent and hasn't been reviewed yet iirc. We will definitely try to get it in as well!

Thank you so much for your help with this documentation - greatly appreciated!

hbagdi:
@thibaultcha Thanks for being so prompt with all these PRs.

I understand that the release date is very close and we might not have bandwidth to accommodate new changes.
But yes like you said, let's try to get the ACL PR in. 
If there are major obstacles then we can put it off for a later release.

I'm waiting for this release to use to use the ACL and keys list endpoints.
In hindsight, I should have sent out ACL before the other ones.

PR #558, opened 2017-11-27
docs(admin-api quickstart) set http_if_terminated false by default

WALL-E:
Make the documented response match the behavior in 0.11.0

ref: [kong/dao/schemas/apis.lua#L206](https://github.com/Kong/kong/blob/f2ee2da36f4d381fdda9ceedf5339dc1d1421d4c/kong/dao/schemas/apis.lua#L206)


coopr:
This was noted in our Changelog https://github.com/Kong/kong/blob/master/CHANGELOG.md#core-3 but not changed in our docs - thanks for contribution @WALL-E !

thibaultcha:
@WALL-E Awesome, thanks!

PR #562, opened 2017-12-08
Updated docs to include rename functionality.

UkiahSmith:


hishamhm:
@UkiahSmith while we're at it, shouldn't it include `config.rename.body` as well?

PR #576, opened 2018-01-24
feat(plugins): Add new EE Canary Release page

travispamaral:


thibaultcha:
You need to add this plugin to `app/_data/plugins.yml`.

PR #577, opened 2018-01-24
docs(key-auth) add DELETE method

zckrs:


thibaultcha:
Thanks, merged with minor edits

PR #583, opened 2018-02-15
release: 0.13.0

thibaultcha:


app/docs/0.13.x/auth.md
```diff
@@ -0,0 +1,210 @@
...
+The easiest way to think about consumers is to map them one-on-one to users. Yet, to Kong this does not matter.
+The core principle for consumers is that you can attach plugins to them, and hence customize request behaviour.
+So you might have mobile apps, and define one consumer for each app, or version of it. Or have a consumer per
+platform, e.g. an android consumer, an iOS consumer, etc.
+
+It is an opaque concept to Kong and hence they are called "consumers" and not "users".
+
+## Anonymous Access
+
+Prior to Kong 0.10.x, a given API could be configured to allow **only** authenticated access (by applying an
```
coopr: Wondering if we should stop mentioning "prior to Kong 0.10.x" - now that we are at 0.13, 0.9 is a long time ago.

app/docs/0.13.x/auth.md
```diff
@@ -0,0 +1,210 @@
...
+[auth plugin][plugins]) or **only** anonymous access - it was not possible to have a given API allow some 
+users to be authenticated and others to access anonymously.
+
+Kong 0.10.x adds the ability to configure a given API to allow **both** authenticated **and** anonymous access.
```
coopr: Same here - "Kong 0.10.x adds..." - would be better to remove that and just state "Kong has the ability to configure a given API..."

app/docs/0.13.x/auth.md
```diff
@@ -0,0 +1,210 @@
...
+      },
+      ...
+    }
+    ```
+    
+    It shows the request was succesful, but anonymous.
+
+## Multiple Authentication
+
+Kong 0.10.x extends the ability to apply multiple authentication plugins for a given API, allowing
```
coopr: Same here - better to remove references to old versions.

PR #589, opened 2018-02-20
docs(index.html) replace api by route+service

kikito:


app/index.html
```diff
@@ -85,12 +85,16 @@ <h2>Get Started in 1 Minute</h2>
       <div class="one-half column">
         <ol>
           <li>
-            <h4>Add your API on Kong</h4>
-            <p>After <a href="/install">installing</a> and starting Kong, use the Admin API on port 8001 to add a new API. Kong will route every incoming request with the specified public DNS to the associated target URL.<br/><br/></p>
+            <h4>Register a Service in Kong</h4>
+            <p>After <a href="/install">installing</a> and starting Kong, use the Admin API on port 8001 to add a new Service. Services represent upstream servers containing an API.</p>
```
hishamhm: suggestion: containing an API -> exposing an API

(Also, I don't know, but "Register" sounds a bit strange to me... I'm not sure if it's because the word has some other overtones (e.g. "register this software"), or because "registering a service" sounds like more work than "adding a service". Anyway, feel free to ignore.)

app/index.html
```diff
@@ -85,12 +85,16 @@ <h2>Get Started in 1 Minute</h2>
...
           </li>
           <li>
-            <h4>Add Plugins on the API</h4>
-            <p>Then add extra functionality by using <a href="/plugins">Kong Plugins</a>. You can also create your own plugins.</p>
+            <h4>Add a Route to the Service</h4>
+            <p>Once you have a Service, add one (or many!) Route to it. Routes control how requests are transmitted to Services, as well as some basic transformations.</p>
```
hishamhm: > ` one (or many!) Route `

I like the touch there, but "many Route" reads bad (the programmer in me gets irked at `one (or many!) Routes` but I think that's the correct form). Might be worth double-checking.

app/index.html
thibaultcha: How about "Configure a Service in Kong" then?

thibaultcha:
Pushed some updates - let's merge as soon as this is ready before tomorrow :)

PR #590, opened 2018-02-20
docs(0.13.x) add Admin API Services and Routes [BIS]

kikito:
This is a continuation of #587 with a few things I noticed were missing on the Admin API docs:

* It says `Service` most of the time it said `API` before (not always).
* Ditto for `api_id` and `service_id`.
* Precedence of routing rules has changed (especially since we now have route_id in addition to service_id
* Some paths have changed. `/api/<id>/plugins` is not available in "service mode" (because plugins can't be nested inside services, only routes can for now). So I have changed it to just `/plugins` (and then you add the `service_id` on the json body)

### DIFF
--- a/app/docs/0.13.x/admin-api.md
+++ b/app/docs/0.13.x/admin-api.md
@@ -98,7 +98,7 @@ target_body: |
 upstream_body: |
     Attributes | Description
     ---:| ---
-    `name` | This is a hostname like name that can be referenced in an `upstream_url` field of an `api`.
+    `name` | This is a hostname like name that can be referenced in an `url` field of a `Service`.
     `slots`<br>*optional* | The number of slots in the loadbalancer algorithm (`10`-`65536`, defaults to `1000`).
     `hash_on`<br>*optional* | What to use as hashing input: `none`, `consumer`, `ip`, or `header` (defaults to `none` resulting in a weighted-round-robin scheme).
     `hash_fallback`<br>*optional* | What to use as hashing input if the primary `hash_on` does not return a hash (eg. header is missing, or no consumer identified): `none`, `consumer`, `ip`, or `header` (defaults to `none`).
@@ -134,6 +134,8 @@ snis_body: |
     ---:| ---
     `name` | The SNI name to associate with the given certificate.
     `ssl_certificate_id` | The `id` (a UUID) of the certificate with which to associate the SNI hostname.
+
+
 
 ---
 
@@ -698,7 +700,7 @@ of `hosts`, `uris`, and `methods`. Kong will proxy all requests to the API to th
 
 <div class="endpoint post">/apis/</div>
 
-#### request body
+#### Request body
 
 {{ page.api_body }}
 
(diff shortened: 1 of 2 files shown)


app/docs/0.13.x/admin-api.md
```diff
@@ -1257,64 +1258,8 @@ Attributes | Description
...
-              "minute": 300,
-              "hour": 20000
-          },
-          "enabled": true,
-          "created_at": 1422386585
-      }
-    ],
-    "next": "http://localhost:8001/plugins?size=2&offset=4d924084-1adb-40a5-c042-63b19db421d1"
-}
-```
```
thibaultcha: It seems like we should keep this example response but update it?

app/docs/0.13.x/admin-api.md
```diff
@@ -1257,64 +1258,8 @@ Attributes | Description
...
-
-#### Request Querystring Parameters
-
-Attributes | Description
----:| ---
-`id`<br>*optional* | A filter on the list based on the `id` field.
-`name`<br>*optional* | A filter on the list based on the `name` field.
-`api_id`<br>*optional* | A filter on the list based on the `api_id` field.
+`service_id`<br>*optional* | A filter on the list based on the `service_id` field.
+`route_id`<br>*optional* | A filter on the list based on the `route_id` field.
```
thibaultcha: it seems like we removed `consumer_id` here

app/docs/0.13.x/admin-api.md
```diff
@@ -1257,64 +1258,8 @@ Attributes | Description
...
-          "created_at": 1422386585
-      }
-    ],
-    "next": "http://localhost:8001/plugins?size=2&offset=4d924084-1adb-40a5-c042-63b19db421d1"
-}
-```
-
----
-
-### List Plugins per API
```
thibaultcha: Makes me think: we should add "List plugins per Route" and "List plugins per Service"

app/docs/0.13.x/admin-api.md
thibaultcha: Hmm I see - you want to remove "deprecated" `/apis/.../` endpoints?

Also: all changes made to the sections of this doc should be also reflected in the YAML sidebar

app/docs/0.13.x/admin-api.md
```diff
@@ -1885,8 +1824,8 @@ HTTP 204 No Content
 
 The upstream object represents a virtual hostname and can be used to loadbalance
 incoming requests over multiple services (targets). So for example an upstream
-named `service.v1.xyz` with an API object created with an `upstream_url=https://service.v1.xyz/some/path`.
-Requests for this API would be proxied to the targets defined within the upstream.
+named `service.v1.xyz` with an Service object created with an `url=https://service.v1.xyz/some/path`.
```
thibaultcha: "a Service"

thibaultcha: We also need to refer to a Service's `host` here (and not `url`) since it is the only field that matters for linking the Upstream with a Service.

app/docs/0.13.x/admin-api.md
kikito: I removed the "List Plugins per API" section. The git diff makes it look as if I had removed the previous one (because the responses are very similar in both cases)

app/docs/0.13.x/admin-api.md
kikito: Maybe in the future when we have a `/service/xx/plugins` section. For now, I have just removed this section.

app/docs/0.13.x/admin-api.md
kikito: > Hmm I see - you want to remove "deprecated" /apis/.../ endpoints?

Indeed

> all changes made to the sections of this doc should be also reflected in the YAML sidebar

Done!

PR #598, opened 2018-02-21
docs(0.13.x/admin-api) add node_id to `/` endpoint

hishamhm:
Includes the commit for 0.12.x as well (should it?)

### DIFF
--- a/app/docs/0.12.x/admin-api.md
+++ b/app/docs/0.12.x/admin-api.md
@@ -139,6 +139,7 @@ HTTP 200 OK
 ```json
 {
     "hostname": "",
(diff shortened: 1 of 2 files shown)


app/docs/0.12.x/admin-api.md
```diff
@@ -156,6 +157,8 @@ HTTP 200 OK
 }
 ```
 
+* `node_id`: A UUID representing the running Kong node. This UUID is randomly generated when Kong starts, so the node will have a different `node_id` each
+  time it is restarted.
```
thibaultcha: When merging the 0.13.x branch into master, we will re-use the 0.12.x directory (thus bringing changes from the master branch) - so this hunk can be discarded imho - or kept, doesn't matter

PR #600, opened 2018-02-22
Order of releases in the docs dropdown

travispamaral:
This PR adds reversed to the loop of releases in the docs dropdown

Before:

before

After:

after


travispamaral:
[(1) Put newest versions at the top of version drop down](https://trello.com/c/Oo9lmok4/495-1-put-newest-versions-at-the-top-of-version-drop-down)

PR #606, opened 2018-02-28
docs(*) 0.12.2 and 0.13.0rc1 releases

hishamhm:


thibaultcha:
Seems like this is obsolete now (not sure why we did not close it earlier)

PR #608, opened 2018-03-01
docs(plugins) update the plugin layout to generate "Getting Started"

kikito:
Since this change modifies the plugins layout, all plugins will have to be adapted to using it (if this is not desired, we can use two different plugin layouts).

There are 4 steps to transform a plugin:

1. Move the "top description" out of the "content" and into a field called `description` in the yml front of the file.
2. Add a section called `params` to the yml front. It needs the following fields:
  * `name` (the name of the plugin, i.e. `key-auth`).
  * `api_id` true if this plugin can be attached to APIs, false if it doesnt
  * `service_id` true if this plugin can be attached to services, false otherwise
  * `route_id` ditto for routes
  * `config`: This will be a list of n fields, with 4 properties each. Usually
    you can get this list by reformatting the "markdown table" at the end of 
    the legacy "description" section of the page you are transforming.
    * `name`: The name of the field (like `hide_credentials`)
    * `required`: true or false. If required, they will be included on the getting
       started examples. If false, it will be marked as optional on the field table.
    * `default`: The default value, if any.
    * `description`: Text. Notice that yaml can be picky with special characters
       inside descriptions. If you are having trouble, use `|` and start the 
       description indented under the `description: |`, as shown on this PR. 
  * `extra` accepts a markdown field which will be rendered after the table of 
    fields (see example on this PR)
3. Remove the `Getting Started` section from the `nav` item in the yml front.
4. Remove the "Terminology" and "Configuration" sections from the content. Those sections are now generated automatically.

Here's a screenshot of the generated top - everything up to "Usage" is generated from the yml:

auth

### DIFF
--- a/app/_layouts/plugin.html
+++ b/app/_layouts/plugin.html
@@ -5,13 +5,127 @@ id: page-plugin
 header_btn_text: Report Bug
 header_btn_href: mailto:support@mashape.com?subject={{ page.header_title }} Plugin Bug
 breadcrumbs: null
+
 ---
+
+{% capture config_required_fields %}{%
+  for field in page.params.config %}{%
+  if field.required %} \
+    --data "{{ field.name }}={{ field.default }}"
+{% endif %}{% endfor %}{% endcapture %}
+
+{% if page.params.api_id %}
+{% capture plugin_config_for_api %}
+Configuring on top of an [API](/docs/latest/admin-api/#api-object) by executing the following request on your Kong server:
+
+```bash
+$ curl -X POST http://kong:8001/apis/{api}/plugins \
+    --data "name={{page.params.name}}" {{ config_required_fields }}
+```
+
+* `api`: either id or name of the API that this plugin configuration will target.
+{% endcapture %}
+{% endif %}
+
+{% if page.params.service_id %}
+{% capture plugin_config_for_service %}
+Configure on top of a [Service](/docs/latest/admin-api/#service-object) by executing the following request on your Kong server:
+
+```bash
+$ curl -X POST http://kong:8001/services/{service}/plugins \
+    --data "name={{page.params.name}}" {{ config_required_fields }}
+```
+
+* `service`: the id or name of the Service that this plugin configuration will target.
+{% endcapture %}
+{% endif %}
+
+{% if page.params.route_id %}
+{% capture plugin_config_for_route %}
+Configure on top of a [Route](/docs/latest/admin-api/#Route-object) with:
+
+```bash
+$ curl -X POST http://kong:8001/routes/{route_id}/plugins \
+    --data "name={{page.params.name}}" {{ config_required_fields }}
+```
+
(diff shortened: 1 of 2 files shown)


kikito:
@Tieske good point! I will do those changes

kikito:
@Tieske made the change and some small additions - updated the screenshot as well.

app/_layouts/plugin.html
```diff
@@ -43,6 +157,53 @@ <h5>{{ nav.label }}</h5>
...
+          {% if page.params.service_id %}
+            {{ plugin_config_for_service | markdownify }}
+          {% endif %}
+          {% if page.params.route_id %}
+            {{ plugin_config_for_route | markdownify }}
+          {% endif %}
+          {% if page.params.consumer_id %}
+            {{ plugin_config_for_consumer | markdownify }}
+          {% endif %}
+          <p>All plugins can be configured using the <code>/plugins</code> endpoint. A plugin which is not associated to any api, service, route or consumer is considered "global", and will be run on every request.</p>
```
Tieske: "and will be run on every request": I'm not a native speaker but I think it should be either "will run on" or "will be executed on"

Tieske:
content wise (besides the nitpick above) it's ok, but (opinions ahead!) I don't like the design of the "Configuration" section with the three consecutive "Configure on top of ..." sections, I think they should visually standout in some way.

kikito:
@Tieske I replaced "run" by "executed".

I too think that it gets kind of repetitive. One option would be using some kind of "tab system". But for that I would need help from the front guys. We could make it a tab-based interface:

``` plain
+----------+-----------+-------------------+
|   API    |  Service  |  Route   |Consumer|
+----------+-----------+-------------------+
|                                          |
| curl --data ...                          |
|                                          |
|                                          |
|                                          |
|                                          |
|                                          |
|                                          |
+------------------------------------------+
```
Or maybe have a `select` at the top and update the curls:

``` plain
+-----------+
|   API   |V|
+------------------------------------------+
|                                          |
| curl --data ...                          |
|                                          |
|                                          |
|                                          |
|                                          |
|                                          |
|                                          |
+------------------------------------------+
```

app/_layouts/plugin.html
thibaultcha: `run ran run` :D

thibaultcha:
@kikito Tab system is a great idea (especially since combination of any of these (Route/Service/Consumer) is possible. Let's keep it handy for later maybe?

thibaultcha:
This looks great to me :+1: Let's use it!

thibaultcha:
I think we may need a link to the explanation for plugins precedence though:

https://github.com/Kong/getkong.org/blob/master/app/docs/0.12.x/admin-api.md#precedence

kikito:
@thibaultcha I am merging this as it is and then I'll make the precedence addition on another branch where I have already made some other changes to the layout.

PR #609, opened 2018-03-06
ADDED: Humor.

HypertextPP:
I'm probably not the first person to call consumers "Kongsumers" but oh well.


Tieske:
After internal feedback, we have come to the Kongclusion that it would probably not help prevent the Kongfusion. So we will close this now.

If you are Kongfident it adds additional value anyway, please persuade us by sending some 🍌 . And we'll be glad to reopen the PR.

PR #612, opened 2018-03-12
chore(*) update mashape to kong

hbagdi:


thibaultcha:
Manually merged with minor modifications (License year while we're at it, + changed scope to `chore(license)`, since `*` means "multiple areas were touched in this commit"). Thanks!

PR #618, opened 2018-03-16
docs(proxy) add `X-` prefix to Kong specific headers

hbagdi:

PR #620, opened 2018-03-16
docs(proxy) add `X-` prefix to Kong specific headers

hbagdi:


app/docs/0.13.x/proxy.md
```diff
@@ -878,10 +878,10 @@ following headers will be added by Kong and the full set of headers be sent to
 the client:
 
 - `Via: kong/x.x.x`, where `x.x.x` is the Kong version in use
-- `Kong-Proxy-Latency: <latency>`, where `latency` is the time in milliseconds
+- `X-Kong-Proxy-Latency: <latency>`, where `latency` is the time in milliseconds
   between Kong receiving the request from the client and sending the request to
   your upstream service.
-- `Kong-Upstream-Latency: <latency>`, where `latency` is the time in
+- `X-Kong-Upstream-Latency: <latency>`, where `latency` is the time in
```
thibaultcha: @hbagdi I think this version of the patch is missing the updates for the previous docs versions?

hbagdi: @thibaultcha I updated the previous version in #618 which was against master and this one was for `0.13.x` only
Do you want to put all of those in here?

thibaultcha: @hbagdi Got it, sorry about that

thibaultcha:
Thanks!

PR #624, opened 2018-03-16
Adding guidance around hide_credentials

aaronhmiller:
Myself and a customer were confused about the nuances of hide_credentials, requiring an escalation. Added some doc that should help clarify this in the future.

### DIFF
--- a/app/plugins/basic-authentication.md
+++ b/app/plugins/basic-authentication.md
@@ -53,7 +53,7 @@ blacklist groups of users.
 form parameter                             | default | description
 ---                                        | ---     | ---
 `name`                                     |         | The name of the plugin to use, in this case: `basic-auth`
-`config.hide_credentials`<br>*optional*    | `false` | An optional boolean value telling the plugin to hide the credential to the upstream API server. It will be removed by Kong before proxying the request
+`config.hide_credentials`<br>*optional*    | `false` | An optional boolean value telling the plugin to hide the credentials from the upstream API service. If true, credentials will be removed by Kong before proxying the request, implying that a direct request of the upstream does not require basic auth. The auth burden is therefore moved into the plugin. If the upstream still requires basic auth, set `config.hide_credentials=false`
 `config.anonymous`<br>*optional*           | ``      | An optional string (consumer uuid) value to use as an "anonymous" consumer if authentication fails. If empty (default), the request will fail with an authentication failure `4xx`. Please note that this value must refer to the Consumer `id` attribute which is internal to Kong, and **not** its `custom_id`.
 
 ----
@@ -61,6 +61,8 @@ form parameter                             | default | description
 ## Usage
 
 In order to use the plugin, you first need to create a consumer to associate one or more credentials to. The Consumer represents a developer using the final upstream service.
+
+A note about `hide_credentials`: Setting it to `false` (the default) implies that the upstream service requires basic authentication. Setting it to `true` puts auth into Kong. Use cURL commands directly to the upstream to determine how to set this value.
 
 ### Create a Consumer
 


app/plugins/basic-authentication.md
```diff
@@ -53,7 +53,7 @@ blacklist groups of users.
 form parameter                             | default | description
 ---                                        | ---     | ---
 `name`                                     |         | The name of the plugin to use, in this case: `basic-auth`
-`config.hide_credentials`<br>*optional*    | `false` | An optional boolean value telling the plugin to hide the credential to the upstream API server. It will be removed by Kong before proxying the request
+`config.hide_credentials`<br>*optional*    | `false` | An optional boolean value telling the plugin to hide the credentials from the upstream API service. If true, credentials will be removed by Kong before proxying the request, implying that a direct request of the upstream does not require basic auth. The auth burden is therefore moved into the plugin. If the upstream still requires basic auth, set `config.hide_credentials=false`
```
thibaultcha: > the upstream API service

This is confusing. We make a particular effort in these docs to use the same terminology all the time (at least, when we catch such inconsistencies...).

We usually refer to the upstream as "the upstream service", or "your upstream service". 

> The auth burden is therefore moved into the plugin. If the upstream still requires basic auth, set `config.hide_credentials=false`

This part is confusing and seems unnecessary? The whole sentence is also really long.

How about:
```
An optional boolean value telling the plugin to hide the credentials from
the upstream service. If `true`, the plugin will strip the credentials from the
request (i.e., the `Authorization` header) before proxying it.
```

The fact that the "burden" is moved from the upstream service to Kong is... obvious? It is the point of this plugin.

app/plugins/basic-authentication.md
```diff
@@ -62,6 +62,8 @@ form parameter                             | default | description
 
 In order to use the plugin, you first need to create a consumer to associate one or more credentials to. The Consumer represents a developer using the final upstream service.
 
+A note about `hide_credentials`: Setting it to `false` (the default) implies that the upstream service requires basic authentication. Setting it to `true` puts auth into Kong. Use cURL commands directly to the upstream to determine how to set this value.
```
thibaultcha: Ditto, better remove this paragraph, it seems to only add confusion. Suddenly I am left wondering whether I should still have authentication in my upstream service, or not, or if Kong even solves this problem at all because this doc seems to suggest I should still have basic-auth upstream?

Additionally, the passing the credentials from Kong to the upstream service is unlikely to work, unless the upstream service shares Kong's Consumers and registered (or vice-versa). Really, this is just very confusing imho.

app/plugins/basic-authentication.md
aaronhmiller: Agreed. Fewer, more accurate words more clear. Thank you.

app/plugins/basic-authentication.md
aaronhmiller: I've removed it. However, the crux of what is confusing is that we allow hide_credentials to be set at all. It seems that the only reason to setup the plugin at all is if you do not have auth on the upstream. If you do (the hide_credentials=false setting) you shouldn't use the plugin in the first place. This also brings into question the default value, which given our discussion, should default to true.

thibaultcha: I definitely agree, `hide_credential` is a confusing (and maybe unnecessary) configuration value... Historically, authentication plugins would do nothing to strip away the credentials they matched (transparent proxy, etc...) until some users got confused or experienced strange behavior from their upstream trying to validate those Kong-specific credentials, and the possibility of stripping them away was added with this parameter, without changing the default to preserve backwards-compatibility (and because not much was given to it either, tbh). Nowadays it indeed seems like `true` should be the default albeit breaking the transparent proxy paradigm, but adhering to the API Gateway one.

thibaultcha:
Looks like this was moved to #625.

aaronhmiller:
Not sure if I'm following best Git practice, but I put PRs on 0.13.x-plugins and master branches because it looks like due to routes & services, merging between them will be very difficult.

thibaultcha:
Maybe, but it will be the job of whoever is doing the merge. Editing the same file in both branches will simple make this person's job more complicated since it will most likely create a conflict.

aaronhmiller:
I think I understand. I was trying to follow some of your earlier advice about having to apply changes to two branches, but I mistook that as being the responsibility of the committer. However, that seems to be the job of the merger/maintainer instead. As always, thank you for the patience and education.

PR #625, opened 2018-03-16
Added clarfication around hide_credentials

aaronhmiller:
Caused an escalation with a prospect and we'd all benefit with some extra detail about the setting.


thibaultcha:
Other authentication plugins have such a property, I think we should update their reference as well or else this is inconsistent.

aaronhmiller:
Good point. I'll hunt them down and make similar edit suggestions.

thibaultcha:
Many of those have been taken car of by now, including this plugin. So closing this for now!

PR #630, opened 2018-03-19
docs(0.13.x) use /services/:service/plugins in examples

hishamhm:
Use `/services/:service/plugins` whenever possible in examples, instead of `/plugins`, since the latter requires the service ID to be used, and the former can use service name.

(Includes bonus unrelated commit with a backport from @thibaultcha's fix of the health endpoint from `docs/0.12.x` to `docs/0.13.x`.)

### DIFF
--- a/app/docs/0.13.x/clustering.md
+++ b/app/docs/0.13.x/clustering.md
@@ -121,12 +121,16 @@ On node `A`, we add a Service and a Route:
 ```bash
 # node A
 $ curl -X POST http://127.0.0.1:8001/services \
-    --data "name=example" \
+    --data "name=example-service" \
     --data "url=http://example.com"
 
-$ curl -X POST http://127.0.0.1:8001/services/example/routes \
-    --data "paths[]=example.com"
+$ curl -X POST http://127.0.0.1:8001/services/example-service/routes \
+    --data "paths[]=/example"
 ```
+
+(Note that we used `/services/example-service/routes` as a shortcut: we
+could have used the `/routes` endpoint instead, but then we would need to
+pass `service_id` as an argument, with the UUID of the new service.)
 
 A request to the Proxy port of both node `A` and `B` will cache this Service, and
 the fact that no plugin is configured on it:
@@ -149,9 +153,8 @@ Now, say we add a plugin to this Service via node `A`'s Admin API:
 
 ```bash
 # node A
-$ curl -X POST http://127.0.0.1:8001/plugins \
+$ curl -X POST http://127.0.0.1:8001/services/example-service/plugins \
     --data "name=example-plugin"
-    --data "service_id=<example_service_id>"
 ```
 
 Because this request was issued via node `A`'s Admin API, node `A` will locally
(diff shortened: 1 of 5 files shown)


app/docs/0.13.x/clustering.md
```diff
@@ -149,9 +149,8 @@ Now, say we add a plugin to this Service via node `A`'s Admin API:
 
 ```bash
 # node A
-$ curl -X POST http://127.0.0.1:8001/plugins \
+$ curl -X POST http://127.0.0.1:8001/services/example/plugins \
```
thibaultcha: I am only noticing now that is example is skewed.

1. The above Route creation warrants a comment saying something like "we could use /routes but /services/:service/routes is a shortcut
2. It creates `paths[]=example.com` which is incorrect
3. "example" as a name is confusing, because it conflicts in this endpoint with the previous `:8000/example` requests. I like "example-service" better as used in other occurrences in this patch

app/docs/0.13.x/plugin-development/plugin-configuration.md
```diff
@@ -23,9 +23,8 @@ Your plugin's configuration is being verified against your schema when a user is
 For example, a user performs the following request:
 
 ```bash
-$ curl -X POST http://kong:8001/plugins/ \
+$ curl -X POST http://kong:8001/services/my-service/plugins/ \
```
thibaultcha: Seems like we should use `/services/<service-name-or-id>/plugins` here to stay consistent (and also self-document the "by-id" endpoint

app/docs/0.13.x/plugin-development/plugin-configuration.md
```diff
@@ -180,9 +179,8 @@ return {
 Such a configuration will allow a user to post the configuration to your plugin as follows:
 
 ```bash
-$ curl -X POST http://kong:8001/plugins \
+$ curl -X POST http://kong:8001/services/<the-service-name>/plugins \
```
thibaultcha: ditto, maybe `<service-name-or-id>` as well

app/docs/0.13.x/clustering.md
hishamhm: I used `example` because that was the name of the Service being created above. Can change both.

app/docs/0.13.x/plugin-development/plugin-configuration.md
hishamhm: I tried to be consistent with `my-custom-plugin` in the line below it. Your suggestion works too

hishamhm:
@thibaultcha please re-review!

app/docs/0.13.x/clustering.md
```diff
@@ -121,13 +121,17 @@ On node `A`, we add a Service and a Route:
...
 
-$ curl -X POST http://127.0.0.1:8001/services/example/routes \
-    --data "paths[]=example.com"
+$ curl -X POST http://127.0.0.1:8001/services/example-service/routes \
+    --data "paths[]=/example"
 ```
 
+(Note that we used `/services/example-service/routes` as a shortcut: we
+could have used the `/routes` endpoint instead, but then we would need to
+pass `service_id` as an argument, with the UUID of the new service.)
```
thibaultcha: :+1:

thibaultcha:
Merged!

PR #622, opened 2018-03-16
Convert plugins to the new yaml-based format

hishamhm:

PR #631, opened 2018-03-19
Updating Plugin Docs

UkiahSmith:
@hishamhm 's branch `plugin-yml` needs to be merged before this branch. 

For these plugins: 

ip-restriction
bot-detection
request-termination
request-size-limiting
correlation-id


app/_layouts/plugin-ee.html
```diff
@@ -0,0 +1,45 @@
+---
+layout: default
+sitemap: true
+id: page-plugin
+header_btn_text: Report Bug
+header_btn_href: mailto:support@mashape.com?subject={{ page.header_title }} Plugin Bug
```
darrenjennings: Are we still support@mashape?

thibaultcha: This needs to be updated and I am doing so locally already as I am merging #622

thibaultcha:
@UkiahSmith The plugin-yml branch has been merged in 0.13.x. Will you please rebase your branch on it? Also, could we include both #633 and all subsequent PRs about plugins updates in the same PR (maybe this one) to avoid spreading over and complicating the review? Thanks!

UkiahSmith:
@thibaultcha Changes from #633 are now in this branch/PR, as well as rebased off of 0.13.x branch.

thibaultcha:
@UkiahSmith I just pushed a commit to improve the verbiage around the changes. Our intent is now to to present Service as _the_ 1st class citizen (alongside Route, on the second step of the pedestal) in Kong. All of the verbiage and examples should always refer to Services first, and be based on Services and/or Routes (when it makes sense, some other plugins might be for Routes only).

Additionally, when referring to APIs we should now point out that they are a deprecated entity (maybe not every time we refer to it, but at least once per document).

So I made those two changes which will squash in yours.

One thing that isn't done yet and that I noticed while going through the changes is that the examples were not updated:

- hmac-auth has a complete example showing how to configure an API and proxy requests through it with the hmac-auth plugin. Those examples need to be updated to be built around a Service/Route couple
- jwt also has such examples
- maybe others, I stopped looking after a while
- some plugins have a simple "example call over the `8000` port", and those have `:8000/{api path}` in their URL. They also need to be updated (I made one such update as "guidance" but there are others).

Will you be able to address the above leftover items?

hishamhm:
I made another round of reviewing, squashed the commits, and merged into the `0.13.x` branch. Thank you, @UkiahSmith!

thibaultcha:
Thanks @UkiahSmith @hishamhm !

PR #637, opened 2018-03-20
docs(ee/docker) fix broken dev portal link

darrenjennings:


aaronhmiller:
late to the party, but thank you! had a prospect notice the issue and nice y'all are on the mofo.

PR #638, opened 2018-03-20
OIDC plugin documentation import

rainest:
Adding the OIDC plugin documentation from the plugin repository. Stripped obsolete content regarding installation (plugin is bundled in current versions) and separate plugins (all functionality is now handled by the unified plugin).


darrenjennings:
We should link to this documentation or redirect to this documentation... and remove some of the duplicated content from https://getkong.org/plugins/ee-openid-connect/ such as "capable providers". @coopr do you have a preference on what we do? I like the idea of the landing page that has some short context as to what OIDC _is_.

coopr:
Removing duplicated content is a great idea. I'm cool with either linking the OIDC plugin tile directly to the docs - or you could put a summary on the plugin page, and the remaining details on the docs page. I'm ambivalent.

rainest:
@coopr does note that the content under /docs/enterprise/ is versioned, so having it separated is useful. OIDC is a bit odd since the plugin is technically versioned separately from EE itself, but I think documenting whatever version is bundled with a given EE version is fine.

coopr:
@rainest - indeed @bungle is correct - like many EE plugins (and, for that matter, CE plugins), plugins have their own versions, that move at their own speeds - but each version of Kong EE and CE has a fixed and known set of plugin versions that are bundled with it.

PR #650, opened 2018-03-28
docs(hmac-auth) update restrictions

shashiranjan84:
### Summary

Inform users about request-line format and affect of request parameter change on HMAC
signature validation.


app/plugins/hmac-authentication.md
```diff
@@ -72,6 +72,10 @@ params:
 In order to use the plugin, you first need to create a Consumer to associate
 one or more credentials to.
 
+Note: Please make sure that all the request parameters being used by client to
```
thibaultcha: Looks like this sentence has some some grammatical issues? We should maybe explain _why_ as well.
```
Because the HMAC signature is generated by the client, you should make sure that Kong does not
update or remove any signed request parameter before this plugin runs.
```

Or something like that...

app/plugins/hmac-authentication.md
```diff
@@ -142,7 +146,7 @@ the order they appear.
   ASCII space ` `.
 
 2. If the header name is `request-line` then append the HTTP
-  request line, otherwise append the header value.
+  request line(must be in ASCII format), otherwise append the header value.
```
thibaultcha: missing a space

also, (in ASCII format) reads better in parenthesis

shashiranjan84:
@thibaultcha thanks for review, made necessary changes.

PR #652, opened 2018-03-28
updates property from uris to paths for updating routes

augwa:
uris isn't a valid property for routes, it should be paths.


thibaultcha:
Hi, thank you for the contribution, but this was already fixed by #649, which pre-dates this PR. Closing this then

PR #679, opened 2018-04-17
doc(plugins) show acl explanation only on authentication plugins

kikito:


hishamhm:
I thought you were going to add an extra boolean to non-ACL auth plugins to enable that paragraph. Didn't want to over-DRY it?

kinman:
Or, introduce another configuration if and only if the next authentication plugin fails to include the ACL paragraph.

kikito:
> I thought you were going to add an extra boolean to non-ACL auth plugins to enable that paragraph. Didn't want to over-DRY it?

I think this was on the limit. I didn't feel strongly about any of the two options, so I decided to use what we had.

coopr:
Does this need to be added to EE auth plugin pages too?

PR #690, opened 2018-04-25
fix(ee/plugins) update forward proxy icon

darrenjennings:
Update icon to existing icon that is not cut off.


darrenjennings:
[fix icon for forward proxy plugin](https://trello.com/c/zIqttr3f/875-fix-icon-for-forward-proxy-plugin)

PR #692, opened 2018-04-26
Document enabled flag for plugins

### LINKED ISSUE #691: Document plugin enabled flag
- `getkong.org` url(s):
admin-api/#add-plugin and similar
all plugin pages

### Summary

The plugin schema has an enabled flag that has been present for at least 3 years, but is not mentioned in documentation for whatever reason (although it is shown in example responses, we do not indicate that it is user-settable or what its purpose is).

As one may expect, enabled is a boolean that determines whether the plugin is applied. It defaults to `true`. Setting it to `false` can be useful for configuration that only needs to be active temporarily, e.g. a maintenance page or a rate limit that is only enforced during peak hours.

rainest:
### Summary

Add `enabled` to plugin pages and admin API documentation. 

### Issues resolved

Fix #691 

### Checklist:
- [x] [Commit message & atomicity](https://github.com/Kong/getkong.org/blob/master/CONTRIBUTING.md#commit-atomicity) checked
- [x] Documentation 
- [x] Spellchecked my updates
- [x] Ready to be merged

### DIFF
--- a/app/docs/0.10.x/admin-api.md
+++ b/app/docs/0.10.x/admin-api.md
@@ -30,6 +30,7 @@ plugin_configuration_body: |
     `name` | The name of the Plugin that's going to be added. Currently the Plugin must be installed in every Kong instance separately.
     `consumer_id`<br>*optional* | The unique identifier of the consumer that overrides the existing settings for this specific consumer on incoming requests.
     `config.{property}` | The configuration properties for the Plugin which can be found on the plugins documentation page in the [Plugin Gallery](/plugins).
+    `enabled` | Whether the plugin is applied. Default: `true`.
 
 target_body: |
     Attributes | Description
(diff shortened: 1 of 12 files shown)


Tieske:
@rainest I think the distinction [here](https://github.com/Kong/kong/issues/1279#issuecomment-275367327) also applies to the plugins. Hence we should document whether it falls through or not.

rainest:
Not in quite the same way (since they wouldn't reject the request outright), but matching does pass through to other configurations with lower precedence. Updated the admin API doc precedence section on versions where it's present (0.9 onward).

app/docs/0.12.x/admin-api.md
```diff
@@ -1,4 +1,6 @@
----
+--- Note that if config B is
+disabled, config A will apply to requests that would have otherwise matched
+config B.
```
thibaultcha: This looks like a mistake?

app/docs/0.10.x/admin-api.md
```diff
@@ -759,7 +760,9 @@ The order of precedence is, from highest to lowest:
 configurations): for an API (Plugin config A), and for a Consumer (Plugin
 config B), then requests authenticating this Consumer will run Plugin config B
 and ignore A (2.). However, requests that do not authenticate this Consumer
-will fallback to running Plugin config A (3.).
+will fallback to running Plugin config A (3.). Note that if config B is
+disabled, config A will apply to requests that would have otherwise matched
```
thibaultcha: I think we'd better mention _how_ a plugin can be "disabled", or else the reader will be confused. Something like:
```
if config B is disabled (its `enabled` flag is set to `false`), then config A will apply to requests...
```

PR #696, opened 2018-05-05
docs(plugin-dev) fix ports used while running tests

hbagdi:


app/docs/0.10.x/plugin-development/tests.md
```diff
@@ -84,7 +84,9 @@ describe("my plugin", function()
 end)
 ```
 
-> Reminder: With the test Kong configuration file, Kong is running with its proxy listening on port 8100 and Admin API on port 8101.
+> Reminder: With the test Kong configuration file, Kong is running with
+its proxy listening on port 9000 (HTTP), 9443 (HTTPS)
```
kikito: @hbagdi have the ports retroactively changed in older versions of Kong (like 0.10.x)?

hbagdi: I'm not sure when the ports were changed. I tried using git history but couldn't figure that out (yet).

thibaultcha: The intent in 0.10 was to run it on :9000 already, so LGTM https://github.com/Kong/kong/blob/0.10.0/spec/kong_tests.conf#L3

PR #697, opened 2018-05-07
ubuntu.md

aaronhmiller:
created Ubuntu EE Install Guide

### Summary

Added Ubuntu install guide for Kong Enterprise Edition

### Full changelog

* Added Ubuntu install guide
* Added step-by-step guidance

### Issues resolved

Fix Missing Ubuntu install guide & step-by-step guidance

### Checklist:
- [x] [Commit message & atomicity](https://github.com/Kong/getkong.org/blob/master/CONTRIBUTING.md#commit-atomicity) checked
- [x] Documentation
- [x] Spellchecked my updates
- [x] Ready to be merged

### DIFF
--- a/app/docs/enterprise/0.31-x/installation/ubuntu.md
+++ b/app/docs/enterprise/0.31-x/installation/ubuntu.md
@@ -31,17 +31,11 @@ $ exit
 
 ## Finish Kong Enterprise setup steps
 
-### Add contents of your license file
 ```bash
+# Add contents of your license file (copy & paste)
 $ sudo vi /etc/kong/license.json
-```
-
-### Copy kong.conf.default to create kong.conf
-```bash
 $ sudo cp /etc/kong/kong.conf.default /etc/kong/kong.conf
-```
-### Uncomment and add 'kong' to pg_password line
-```bash
+# Uncomment and add 'kong' to pg_password line
 $ sudo vi /etc/kong/kong.conf
 ```
 
@@ -51,7 +45,7 @@ $ kong migrations up
 $ sudo /usr/local/bin/kong start
 ```
 
-### Setup HTTPie to make commands easier (optional)
+### (Optional) Setup HTTPie to make commands easier
 ```bash
 $ sudo apt install httpie
 ```
--- a/app/docs/enterprise/0.31-x/installation/ubuntu.md
+++ b/app/docs/enterprise/0.31-x/installation/ubuntu.md
@@ -64,6 +64,8 @@ $ curl -i -X POST \
   --data 'name=demo' \
   --data 'uris=/' \
   --data 'upstream_url=http://httpbin.org'
+$ curl -i -X GET \
+  --url http://localhost:8000/ip
   
 # httpie syntax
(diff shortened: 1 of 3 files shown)


app/docs/enterprise/0.31-x/installation/ubuntu.md
```diff
@@ -0,0 +1,78 @@
...
+
+Download the EE binary for Ubuntu (or Debian) from Bintray
+SFTP the file to your Ubuntu server
+
+```bash
+$ sudo apt-get update
+$ sudo apt-get install openssl libpcre3 procps perl
+$ sudo dpkg -i kong-enterprise-edition-0.31-1.zesty.all.deb
+```
+NOTE: EE file may differ in last step above
```
shashiranjan84: We support other distributions of Ubuntu too, I think it would be worth mentioning them?

app/docs/enterprise/0.31-x/installation/ubuntu.md
```diff
@@ -0,0 +1,78 @@
...
+$ sudo vi /etc/kong/kong.conf
+```
+
+### Run migrations and start kong
+```bash
+$ kong migrations up
+$ sudo /usr/local/bin/kong start
+```
+
+### Setup HTTPie to make commands easier
```
shashiranjan84: Not required, just mention that for this you used `httpie` but user can use any http client like curl

aaronhmiller:
Thank you! I incorporated your suggestions and in the process, also found that I didn't update right hand menu.

aaronhmiller:
Ready for merging IMHO

app/docs/enterprise/0.31-x/installation/ubuntu.md
```diff
@@ -0,0 +1,83 @@
...
+$ sudo vi /etc/kong/kong.conf
+```
+
+### Run migrations and start kong
+```bash
+$ kong migrations up
+$ sudo /usr/local/bin/kong start
+```
+
+### (Optional) Setup HTTPie to make commands easier
```
shashiranjan84: I think we can drop this Httpie, given now you using curl example?

aaronhmiller: We could, but I like gently pushing people to use a "better curl" and that's HTTPie.  ⚡️

thibaultcha: For the sake of simplicity, better stick to single and explicit examples (explicit would be `curl`). or the sake of portability (as well as explicitness), all of our examples are given with `curl`. And for the sake of consistency, we should probably avoid giving examples with multiple tools in some places, and examples with a single tool elsewhere. It also makes the examples longer and harder to follow...

aaronhmiller: Fair enough. I mention `vi` instead of other editors for the same reason. Making it more about the action than the tool. Simple & clean is why I like HTTPie, but same logic justifies simplifying the guidance w/ curl.

PR #698, opened 2018-05-09
doc(api) Synch the certs admin API with Kong  branch

kikito:
This PR documents changes that are currently happening on the `next` branch in Kong, in which we moved SSL certificates from the old DAO to the new one.

Summary of changes:

* `ssl_certificate_id` field renamed to `certificate_id`
* Slight change in the way the `snis` pseudo attribute works on `/certificates` (previously it could be a string, now it must be an array)
* `PUT` changes in both `/certificates` and `/snis`
* snis have now a primary key `id` (a UUID). The `name` field is indexed but not the primary key any more.
* because of the previous change, a lot of endpoint now accept `name or id` instead of just `name`.

### DIFF
--- a/app/docs/0.14.x/admin-api.md
+++ b/app/docs/0.14.x/admin-api.md
@@ -1,0 +1,2591 @@
+---
+title: Admin API
+
+api_body: |
+    Attribute | Description
+    ---:| ---
+    `name`                        | The API name.
+    `hosts`<br>*semi-optional*    | A comma-separated list of domain names that point to your API. For example: `example.com`. At least one of `hosts`, `uris`, or `methods` should be specified.
+    `uris`<br>*semi-optional*     | A comma-separated list of URIs prefixes that point to your API. For example: `/my-path`. At least one of `hosts`, `uris`, or `methods` should be specified.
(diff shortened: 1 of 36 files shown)


thibaultcha:
This needs to be opened against a `release/ce-0.14` branch that should be based off master.

app/docs/0.13.x/admin-api.md
```diff
@@ -1620,18 +1620,30 @@ HTTP 200 OK
...
 
-The behavior of `PUT` endpoints is the following: if the request payload **does
-not** contain an entity's primary key (`id` for Certificates), the entity will
-be created with the given payload. If the request payload **does** contain an
-entity's primary key, the payload will "replace" the entity specified by the
-given primary key. If the primary key is **not** that of an existing entity, `404
-NOT FOUND` will be returned.
+If the certificate for the provided id exists, or an SNI with the provided name exists,
+this request will update said certificate using its payload. If the payload includes an `snis`
+pseudo-attribute, then the list of associated snis associated with the certificate will also
```
thibaultcha: this repeats `associated`

PR #699, opened 2018-05-09
docs(plugins) update jwt typo servies

kikito:
Consumers (and most of their associated Admin API) have been moved to the new DAO in `next` by updating the PUT action to use `/consumers/uuid-or-name` instead of `/consumers` and documenting the change in implementation.


kikito:
Apologies, it seems I created a PR for the wrong branch, closing this one.

PR #700, opened 2018-05-09
doc(api) document new PUT for services, routes & consumers

kikito:
`next` includes new PUT methods.

SSL certificates have their own PR where they have other changes besides PUT.

### DIFF
--- a/app/docs/0.13.x/admin-api.md
+++ b/app/docs/0.13.x/admin-api.md
@@ -425,6 +425,42 @@ HTTP 200 OK
 
 ---
 
+### Update Or create Service
+
+#### Endpoint
+
+<div class="endpoint put">/services/{name or id}</div>
+
+Attributes | Description
+---:| ---
+`name or id`<br>**required** | The `id` **or** the `name` attribute of the Service to update.
+
+#### Request Body
+
+{{ page.service_body }}
+
+Inserts (or replaces) the Service under the requested resource with the definition
+specified in the body. The Service will be identified via the `name or id` attribute.
+
+When the `name or id` attribute has the structure of an UUID, the Service being inserted/replaced
+will be identified by its `id`. Otherwise it will be identified by its `name`.
+
+When creating a new Service, if an `id` is not specified (neither in the URL or the body) then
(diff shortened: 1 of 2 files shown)


app/docs/0.13.x/admin-api.md
```diff
@@ -425,6 +425,35 @@ HTTP 200 OK
...
+#### Response
+
+```
+HTTP 201 Created or HTTP 200 OK
+```
+
+See POST and PATCH responses.
+
+---
+
```
thibaultcha: As usual, those new endpoints need to be added to the YAML definition of the sidebar links at https://github.com/Kong/getkong.org/blob/master/app/_data/docs_nav_0.13.x.yml

app/docs/0.13.x/admin-api.md
```diff
@@ -425,6 +425,35 @@ HTTP 200 OK
...
+`name or id`<br>**required** | The `id` **or** the `name` attribute of the Service to update.
+
+#### Request Body
+
+{{ page.service_body }}
+
+If a Service with the given name or id exists, the request updates it with the payload.
+If it doesn't exist, a new one will be created using the payload and the given id or name.
+
+Notice that passing a name on the url and a different name on the payload is not allowed.
```
thibaultcha: I don't think the phrasing is the proper verbiage to use with `PUT`. The description should hint a firmer logic and hint the PUT specifications. Maybe something along the lines of:
```
Inserts (or replaces) the Service under the requested resource with the definition
specified in the body.

A Service is identified by its `id` or `name`, and either of these can be specified as
the resource identifier in the endpoint.
```

Can you confirm that if I do not specify an `id` (neither in the URL nor in the body), one will be generated?

Also, no explanation is given about when the resource identifier will be interpreted as `id`, vs as `name`.

Additionally, adding a body example could help clarifying a few things, such as whether or not the `name` or `id` fields should be present in the payload as well as the body or not (likely not, but an example will make things clearer).

thibaultcha:
This should be opened against a `release/0.13.2` branch based off master I believe.

PR #718, opened 2018-05-23
docs(ee/*) .32 fixes, RL advanced, table styling, architecture doc

darrenjennings:
* fix link to rate limiting (RL) advanced in 0.32 nav
* fix table styling on RL and break word on configs in second column
* add architecture guide doc to 0.32 from 0.31

image

### DIFF
--- a/app/docs/enterprise/0.32-x/plugins/rate-limiting-advanced.md
+++ b/app/docs/enterprise/0.32-x/plugins/rate-limiting-advanced.md
@@ -31,7 +31,7 @@ $ curl -i -X POST http://kong:8001/plugins \
 |`config.limit`|| one of more request per window to apply
 |`config.window_size`||One more more window sizes to apply (defined in seconds).
 |`config.identifier` | `consumer` | How to define the rate limit key. Can be `ip`, `credential`, or `consumer`.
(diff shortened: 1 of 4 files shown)


app/docs/enterprise/0.32-x/plugins/rate-limiting-advanced.md
```diff
@@ -31,7 +31,7 @@ $ curl -i -X POST http://kong:8001/plugins \
 |`config.limit`|| one of more request per window to apply
 |`config.window_size`||One more more window sizes to apply (defined in seconds).
 |`config.identifier` | `consumer` | How to define the rate limit key. Can be `ip`, `credential`, or `consumer`.
-|`config.dictionary_name`| `kong_rate_limiting_counters` | The shared dictionary where counters will be stored until the next sync cycle.
+|`config.dictionary_name`| <div style="max-width: 160px;">`kong_rate_limiting_counters`</div> | The shared dictionary where counters will be stored until the next sync cycle.
```
gszr: Is it possible to avoid style code in here? It's unclean and will mean more work when we need to update the plugin's documentation.

darrenjennings: @gszr I do not like it either but it is a patch that helps isolate these types of column widths until we can come up with better table UI.

PR #727, opened 2018-05-29
docs(api) remove API object

kikito:


thibaultcha:
Let's also do the link from the plugins docs to the API entity sections of the 0.13.x version as we discussed. We can probably come up with a much better phrasing for the "Enabling the plugin for an API" section in the plugins template, that lets us include this link in a way that makes it obvious it is redirecting to the 0.13.x version of the Admin API docs...

kikito:
@thibaultcha done!

app/_layouts/plugin.html
```diff
@@ -107,6 +121,9 @@
     {% if page.params.consumer_id %}
       <tr><td><code>consumer_id</code></td><td></td><td>The id of the Consumer which this plugin will target.</td></tr>
     {% endif %}
+    {% if page.params.api_id %}
+      <tr><td><code>api_id</code></td><td></td><td>The id of the API which this plugin will target. <strong>Not available since Kong 0.14.0.</strong></td></tr>
```
thibaultcha: APIs will not be dropped in Kong 0.14.0, only the documentation on the Admin API

thibaultcha:
Manually merged (in both the release and master branches)!

PR #729, opened 2018-05-29
docs(configuration) document headers field

hbagdi:
With CE 0.14.0, a new `headers` configuration option is
introduced and `server_tokens` and `latency_tokens` are removed.

See Kong/kong#3300

Signed-off-by: Harry Bagdi <harrybagdi@gmail.com>

### DIFF
--- a/app/docs/0.14.x/configuration.md
+++ b/app/docs/0.14.x/configuration.md
@@ -607,21 +607,32 @@ Default: `60`
 
 ---
 
-##### **server_tokens**
+##### **headers**
 
-Enables or disables emitting Kong version on error pages and in the `Server`
-or `Via` (in case the request was proxied) response header field.
+Comma-separated list of headers which should be injected by Kong.
 
-Default: `on`
+Each value in the list can be one of the response headers below, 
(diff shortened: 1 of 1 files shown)


app/docs/0.14.x/configuration.md
```diff
@@ -607,21 +607,29 @@ Default: `60`
...
-Enables or disables emitting Kong latency information in the `X-Kong-Proxy-Latency`
-and `X-Kong-Upstream-Latency` response header fields.
-
-Default: `on`
+##### **headers**
+
+Comma-separated list of headers which should be injected by Kong.
+
+Values can be:
+  response header fields.
```
kikito: This phrase could be rephrased to be more exact, and its formatting is a bit weird. This is my suggestion:

> Each value on the list can be one of the response header names below, or one of the two accepted aliases(`server_tokens` or `latency_tokens`).

Also, please include a blank line between this paragraph and the list following it. Some markdown libraries don't parse that as a list without it.

thibaultcha:
Merged, thanks!

PR #733, opened 2018-06-04
docs(conf) removed deprecated config directives

Tieske:


thibaultcha:
Good catch, thanks!

thibaultcha:
@Tieske Merged too quickly. This does not replace the description of other configuration directives that _mention_ those properties, like `admin_ssl_cert`, etc... Mind submitting an additional PR for those? These descriptions can be taken from the `kong.conf.default` file.

PR #744, opened 2018-06-15
Serverless Functions

nijikokun:
Adds documentation for the Serverless Function plugins.

### DIFF
--- a/app/plugins/serverless-functions.md
+++ b/app/plugins/serverless-functions.md
@@ -77,7 +77,7 @@ different priority in the plugin chain.
     -- Terminate request early if our custom authentication header
     -- does not exist
     if not custom_auth then
-      return kong.exit(401, "Invalid Credentials")
+      return kong.response.exit(401, "Invalid Credentials")
     end
 
     -- Remove custom authentication header from request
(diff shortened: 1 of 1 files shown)


app/plugins/serverless-functions.md
```diff
@@ -0,0 +1,157 @@
...
+  config:
+    - name: functions
+      required: true
+      default: "[]"
+      value_in_examples: "[]"
+      description: Array of stringified Lua code to be cached and run in sequence during access phase.
+
+---
+
+> **Note**: Serverless plugins are available in Enterprise as of version 0.32
```
coopr: This clause can be removed.

app/plugins/serverless-functions.md
```diff
@@ -0,0 +1,157 @@
...
+In short, serverless functions give you the full capabilities of a custom plugin
+in the access phase without ever redeploying / restarting Kong. 
+
+----
+
+### Notes
+
+#### Fake Upstreams
+
+Since the [Service][service-url] entity requires defining an upstream you may
```
coopr: Same comment as before - I think this needs to be "must" instead of "may". If "may" is in fact correct, I suggest clarifying what happens if you don't have a fake upstream and terminate the request.

nijikokun: It's `may`, in this context

app/plugins/serverless-functions.md
nijikokun: Good catch

PR #750, opened 2018-06-21
fix(*) Update URL to notification widget

travispamaral:
The notification banner URL was never changed from Mashape and was causing a 404


thibaultcha:
LGTM, although could probably be removed nowadays - this was especially useful at a time where we had multiple websites to publish announcements to.

PR #751, opened 2018-06-21
docs(repo-link) add repo-link header button

bungle:
### Summary

Adds three `header` parameters:
- `page.header_btn_repo_href` — link to source repo
- `header_btn_repo_text` — text for the button (by default it is `Source Code`)
- `page.header_btn_repo_target ` — the target attribute value (not used, but as an option)

Also adds `page.header_btn_repo_href` to following plugins:
- `azure-functions`
- `openwhisk`
- `zipkin`

The `page.header_btn_repo_href` for `serverless-functions` is added to respective PRs:
- https://github.com/Kong/docs.konghq.com/pull/744

We need to add it to `prometheus` once such PR exists.

### DIFF
--- a/app/_includes/header.html
+++ b/app/_includes/header.html
@@ -29,9 +29,18 @@
         {% endif %}
       </div>
 
-      {% if page.header_btn_text %}
+      {% if page.header_btn_text or page.header_btn_repo_href %}
         <div class="page-header-right">
-          <a class="page-header-btn" href="{{ page.header_btn_href | liquify }}" target="{{ page.header_btn_target }}">{{ page.header_btn_text | liquify }}</a>
+          {% if page.header_btn_repo_href %}
+            {% if header_btn_repo_text %}
+              <a class="page-header-btn" href="{{ page.header_btn_repo_href | liquify }}" target="{{ page.header_btn_repo_target }}">{{ page.header_btn_repo_text | liquify }}</a>
+            {% else %}
+              <a class="page-header-btn page-header-btn-repo" href="{{ page.header_btn_repo_href | liquify }}" target="{{ page.header_btn_repo_target }}">Source&nbsp;Code</a>
+            {% endif %}
+          {% endif %}
+          {% if page.header_btn_text %}
+            <a class="page-header-btn" href="{{ page.header_btn_href | liquify }}" target="{{ page.header_btn_target }}">{{ page.header_btn_text | liquify }}</a>
+          {% endif %}
         </div>
       {% endif %}
 
--- a/app/_includes/header.html
+++ b/app/_includes/header.html
@@ -29,17 +29,12 @@
         {% endif %}
       </div>
 
-      {% if page.header_btn_text or page.header_btn_repo_href %}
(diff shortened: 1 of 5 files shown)


hishamhm:
> We need to add it to prometheus once such PR exists.

And also to `functions`. Should we hold this PR until we get those plugins in and then add all the buttons all at once, or should we merge these now and then remember to ask to add the buttons in the other plugin PRs?

coopr:
My opinion is that we should add 'em as we go - better to have some buttons now, and some later, than none now, and more later.

bungle:
Already added to Functions PR:
https://github.com/Kong/docs.konghq.com/pull/744/commits/40cfef23c697eece5ac076a7597c68350707a222

bungle:
@coopr what is your take, should we also link CE plugins even when they are in Kong repo? E.g. a direct link to: e.g. here: https://github.com/Kong/kong/tree/master/kong/plugins/acl

bungle:
@hishamhm this is based to `release/ce-0.14.0` atm. So even if we merge this nothing will go public. Also it is harmless to add those new properties in other places even if this is not merged. They are there, but will not get used.

coopr:
@bungle I don't have a strong opinion for linking to Github when the plugin is still in the Kong repo - I could see reasons to do it, and reasons not to do it. I'll leave that one up to you - I don't think there is a "wrong" answer :)

app/_includes/header.html
```diff
@@ -29,9 +29,18 @@ <h1>{{ page.header_title }}</h1>
...
 
-      {% if page.header_btn_text %}
+      {% if page.header_btn_text or page.header_btn_repo_href %}
         <div class="page-header-right">
-          <a class="page-header-btn" href="{{ page.header_btn_href | liquify }}" target="{{ page.header_btn_target }}">{{ page.header_btn_text | liquify }}</a>
+          {% if page.header_btn_repo_href %}
+            {% if header_btn_repo_text %}
+              <a class="page-header-btn" href="{{ page.header_btn_repo_href | liquify }}" target="{{ page.header_btn_repo_target }}">{{ page.header_btn_repo_text | liquify }}</a>
+            {% else %}
+              <a class="page-header-btn page-header-btn-repo" href="{{ page.header_btn_repo_href | liquify }}" target="{{ page.header_btn_repo_target }}">GitHub</a>
```
thibaultcha: Should this say "Source Code" instead of GitHub? "GitHub" seems vague (where on GitHub does this button link to?) and the icon already communicates that it will lead there anyway.

thibaultcha:
Let's merge this as soon as it is ready and add buttons to other plugins gradually. Plugins in the Kong source repository can be added in their own commit, and upcoming plugin pages (Functions/Prometheus) can be added directly with the button.

app/_includes/header.html
bungle: I can change it to `Source Code`. I don't feel strongly on this.

thibaultcha: Let's do it then.

bungle: Oh, forgot to reply, it is done.

PR #768, opened 2018-06-26
docs(ce/configuration) Injecting Nginx directives

hishamhm:
Documentation of the new method for injecting Nginx directives introduced in
Kong CE 0.14.

### DIFF
--- a/app/docs/0.14.x/configuration.md
+++ b/app/docs/0.14.x/configuration.md
@@ -9,8 +9,11 @@ title: Configuration Reference
 - [Configuration loading](#configuration-loading)
 - [Verifying your configuration](#verifying-your-configuration)
 - [Environment variables](#environment-variables)
-- [Custom Nginx configuration & embedding Kong](#custom-nginx-configuration-embedding-kong)
-  - [Custom Nginx configuration](#custom-nginx-configuration)
+- [Injecting Nginx directives](#injecting-nginx-directives)
+  - [Injecting individual Nginx directives](#injecting-individual-nginx-directives)
+  - [Including files via injected Nginx directives](#including-files-via-injected-nginx-directives)
+- [Custom Nginx templates & embedding Kong](#custom-nginx-templates-embedding-kong)
+  - [Custom Nginx templates](#custom-nginx-templates)
   - [Embedding Kong in OpenResty](#embedding-kong-in-openresty)
   - [Serving both a website and your APIs from Kong](#serving-both-a-website-and-your-apis-from-kong)
 - [Properties reference](#properties-reference)
@@ -32,7 +35,7 @@ $ cp /etc/kong/kong.conf.default /etc/kong/kong.conf
 ```
 
 Kong will operate with default settings should all the values in your
-configuration be commented-out. Upon starting, Kong looks for several
+configuration be commented out. Upon starting, Kong looks for several
 default locations that might contain a configuration file:
 
 ```
@@ -49,7 +52,7 @@ $ kong start --conf /path/to/kong.conf
 
 The configuration format is straightforward: simply uncomment any property
 (comments are defined by the `#` character) and modify it to your needs.
-Booleans values can be specified as `on/off` or `true`/`false` for convenience.
+Boolean values can be specified as `on`/`off` or `true`/`false` for convenience.
 
 [Back to TOC](#table-of-contents)
 
@@ -87,16 +90,16 @@ environment variables of the same name. This allows you to fully configure Kong
 via environment variables, which is very convenient for container-based
 infrastructures, for example.
 
-All environment variables prefixed with `KONG_`, capitalized and bearing the
-same name as a setting will override it.
+To override a setting using an environment variable, declare an environment
+variable with the name of the setting, prefixed with `KONG_` and capitalized.
 
-Example:
+For example:
 
 ```
 log_level = debug # in kong.conf
 ```
 
-Can be overridden with:
+can be overridden with:
 
 ```
 $ export KONG_LOG_LEVEL=error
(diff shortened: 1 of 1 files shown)


app/docs/0.14.x/configuration.md
```diff
@@ -104,13 +107,138 @@ $ export KONG_LOG_LEVEL=error
...
+* Entries prefixed with `nginx_http_` will be injected to the overall `http`
+block directive.
+
+* Entries prefixed with `nginx_proxy_` will be injected to the `server` block
+directive handling Kong's proxy ports.
+
+* Entries prefixed with `nginx_admin_` will be injected to the `server` block
+directive handling Kong's Admin API ports.
+
+For example, if you the following line to your `kong.conf` file:
```
thibaultcha: -> if you _add_?

app/docs/0.14.x/configuration.md
```diff
@@ -104,13 +107,138 @@ $ export KONG_LOG_LEVEL=error
...
+
+```
+nginx_proxy_add_header=Header-Name header-value
+```
+
+This will add the following directive to the proxy `server` block of Kong's
```
thibaultcha: `s/This/it`? For the continuity with the previous sentence

app/docs/0.14.x/configuration.md
```diff
@@ -104,13 +107,138 @@ $ export KONG_LOG_LEVEL=error
...
+Nginx configuration:
+
+```
+    add_header Header-Name header-value;
```
thibaultcha: Since adding headers is something we encourage users to do in the Lua-land (via plugins), a better example might be to inject some Nginx-specific directive. A typical example is [large_client_header_buffers](large_client_header_buffers).

app/docs/0.14.x/configuration.md
```diff
@@ -104,13 +107,138 @@ $ export KONG_LOG_LEVEL=error
...
+```
+
+Like any other entry in `kong.conf`, these directives can also be specified
+using [environment variables](#environment-variables) as shown above. For
```
thibaultcha: Maybe remove "as shown above"? As shown below would be more appropriate but we already introduce the example with "For example,"

app/docs/0.14.x/configuration.md
```diff
@@ -104,13 +107,138 @@ $ export KONG_LOG_LEVEL=error
...
+example, if you declare an environment variable like this:
+
+```
+$ export KONG_NGINX_HTTP_LUA_SHARED_DICT="custom_cache 64k"
```
thibaultcha: Maybe let's keep shm injection on the side (until we support injecting multiple ones, but mostly because plugins should be able to inject their own in a more "native" fashion). The same example as the one given above (updated to large_client_header_buffers) maybe? Or another "typical" usage, such as another buffer size, or connection-reuse setting?

app/docs/0.14.x/configuration.md
```diff
@@ -104,13 +107,138 @@ $ export KONG_LOG_LEVEL=error
...
+nginx_http_include /path/to/your/my-server.kong.conf
+```
+
+Or, alternatively, configuring it via an environment variable:
+
+```
+$ export KONG_NGINX_HTTP_INCLUDE="/path/to/your/my-server.kong.conf"
+```
+
+Now, when you start Kong the `server` section from that file will be added to
```
thibaultcha: comma after "Kong" maybe?

app/docs/0.14.x/configuration.md
```diff
@@ -104,13 +107,138 @@ $ export KONG_LOG_LEVEL=error
...
+```
+
+You can make the Kong node serve this port by adding the following
+entry to your `kong.conf` file:
+
+```
+nginx_http_include /path/to/your/my-server.kong.conf
+```
+
+Or, alternatively, configuring it via an environment variable:
```
thibaultcha: s/configuring/configure maybe?

app/docs/0.14.x/configuration.md
```diff
@@ -104,13 +107,138 @@ $ export KONG_LOG_LEVEL=error
...
+your `kong.conf` file (or the value of the `-p` flag of `kong start` if you
+used it to override the prefix when starting Kong).
+
+### Custom Nginx templates & embedding Kong
+
+For the vast majority of use-cases, using the Nginx directive injection system
+explained above should be sufficient for customizing the behavior of Kong's
+Nginx instance. This way, you can manage the configuration and tuning of
+your Kong node from a single `kong.conf` file (and optionally your own
+included files), without having to deal with Kong's configuration template.
```
thibaultcha: with "custom Nginx configuration templates" (Kong's configuration is simple enough and we should try avoiding possible confusion...)

app/docs/0.14.x/configuration.md
```diff
@@ -104,13 +107,138 @@ $ export KONG_LOG_LEVEL=error
...
+
+There are two scenarios in which you may want to make use of Kong's
```
thibaultcha: ditto (custom Nginx config)

app/docs/0.14.x/configuration.md
```diff
@@ -128,12 +256,15 @@ two files into the prefix directory, which looks like so:
...
 
-If you wish to include other `server` blocks in your Nginx configuration, or if
-you must tweak global settings that are not exposed by the Kong configuration,
-here is a suggestion:
+If you must tweak global settings that are defined by Kong and not exposed by
+the Kong configuration, here is a suggestion, or even customize the Kong Nginx
+sub-configuration file to, eventually, add other Lua handlers or customize the
+`lua_*` directives, you can inline the contents of the `nginx_kong.lua`
+configuration template into a custom template file (in this example called
+`custom_nginx.template`) like this:
```
thibaultcha: > If you must tweak global settings that are defined by Kong and not exposed by the Kong configuration

This should probably be updated to reflect the new "rare conditions" that warrant a custom template (that you perfectly described above)

> here is a suggestion, or even customize the Kong Nginx sub-configuration file to, eventually, add other Lua handler

Something looks wrong here?

app/docs/0.14.x/configuration.md
hishamhm: Good point. I stole the examples from Harry's PR :)

app/docs/0.14.x/configuration.md
hishamhm: > Something looks wrong here?

Good catch, an old sentence fragment was incorrectly pasted.

PR #770, opened 2018-06-27
Adding Prometheus documentation

UkiahSmith:
- Added docs from github with simple tweaks
- Added icon


coopr:
Icon should be added via this process https://github.com/Kong/docs.konghq.com/blob/master/CONTRIBUTING.md#contributing-images-videos-etc, then it can be incorporated into docs.konghq.com by referencing its URL.

app/plugins/prometheus.md
```diff
@@ -0,0 +1,164 @@
...
+title: Plugins - Prometheus
+header_title: Prometheus
+header_icon: /assets/images/icons/plugins/prometheus.png
+breadcrumbs:
+  Plugins: /plugins
+nav:
+  - label: Usage
+    items:
+      - label: Log Format
+      - label: Kong Process Errors
```
coopr: What does this refer to?

app/plugins/prometheus.md
```diff
@@ -0,0 +1,164 @@
+---
+id: page-plugin
+title: Plugins - Prometheus
+header_title: Prometheus
+header_icon: /assets/images/icons/plugins/prometheus.png
```
coopr: Should point to a URL of the image.

app/plugins/prometheus.md
```diff
@@ -0,0 +1,164 @@
...
+
+#### Enable the plugin
+
+```bash
+$ curl http://localhost:8001/plugins name=prometheus
+```
+
+#### Reading metrics into Prometheus
+
+Metrics are availble on the admin API at `/metrics` endpoint:
```
coopr: Kong entities like "Admin API" are capitalized.

coopr: You may want to note that this response is only an example - different configurations of Kong will result in different response content. Or if that is obvious, no change needed.

app/plugins/prometheus.md
```diff
@@ -0,0 +1,164 @@
...
+kong_nginx_http_current_connections{state="total"} 8
+kong_nginx_http_current_connections{state="waiting"} 0
+kong_nginx_http_current_connections{state="writing"} 1
+# HELP kong_nginx_metric_errors_total Number of nginx-lua-prometheus errors
+# TYPE kong_nginx_metric_errors_total counter
+kong_nginx_metric_errors_total 0
+
+```
+
+In case Admin API of Kong is protected behind a firewall or requires
```
coopr: Seems like a word or two is missing from this sentence - "In case the Admin API of Kong..." or "In the case Kong's Admin API is protected..." would be better. Maybe try checking in Grammarly or Hemmingway.

app/plugins/prometheus.md
```diff
@@ -0,0 +1,164 @@
...
+authentication, you have a couple of options:
+
+
+1. Kong Enterprise users can protect the admin `/metrics` endpoint with an [RBAC user](/enterprise/0.32-x/setting-up-admin-api-rbac) that the Prometheus servers uses to access the metric data.
```
coopr: This should link to /latest/, not to a specific version of Kong EE.

app/plugins/prometheus.md
```diff
@@ -0,0 +1,164 @@
...
+    ```
+    curl -X POST http://localhost:8001/services -d name=prometheusEndpoint -d url=http://localhost:8001/metrics
+    curl -X POST http://localhost:8001/services/prometheusEndpoint/routes -d paths[]=/metrics
+    curl -X POST http://localhost:8001/services/prometheusEndpoint/plugins -d name=ip-restriction -d config.whitelist=10.0.0.0/24
+    ```
+
+3. This plugin has the capability to serve the content on a
+different port using a custom server block in Kong's NGINX template.
+
+    Please note that this requires a custom NGINX template for Kong.
```
coopr: Link to the docs that describe using a custom NGINX template with Kong.

coopr:
Thanks @UkiahSmith great tune-ups!

app/plugins/prometheus.md
```diff
@@ -0,0 +1,171 @@
...
+header_title: Prometheus
+header_icon: https://konghq.com/wp-content/uploads/2018/06/prometheus.png
+breadcrumbs:
+  Plugins: /plugins
+nav:
+  - label: Usage
+    items:
+      - label: Log Format
+description: |
+  This plugin exposes metrics in [Prometheus Exposition format](https://github.com/prometheus/docs/blob/master/content/docs/instrumenting/exposition_formats.md). It allows you to integrate with your existing monitoring and alerting system.
```
hbagdi: Can we change this to something like:
 ```
Expose metrics related to Kong and the upstream services in 
[Prometheus](https://prometheus.io/docs/introduction/overview/) Exposition format 
which can be scrapped by a Prometheus Server.
```

This is to keep the description aligned with how other plugins do it and the link to Prometheus website is more reliable than the Github link.

File: app/plugins/prometheus.md [hunk 8]
```diff
@@ -0,0 +1,171 @@
...
+
+params:
+  name: prometheus
```
hbagdi: This plugin can be applied on specific APIs/routes/services to collect stats related to specific routes/services only.
So could we update this to:
```yaml
params:
  name: prometheus
  api_id: true
  service_id: true
  route_id: true
```

app/plugins/prometheus.md
```diff
@@ -0,0 +1,171 @@
...
+---
+
+
+#### Enable the plugin
+
+```bash
+$ curl http://localhost:8001/plugins name=prometheus
+```
+
+#### Reading metrics into Prometheus
```
hbagdi: Can we add the bit about the metrics which are available in this plugin right now?
For reference, we can use https://github.com/kong/kong-plugin-prometheus#available-metrics

app/plugins/prometheus.md
```diff
@@ -0,0 +1,171 @@
...
+need to be setup to require authentication, here are a couple of options to
+allow access to the `/metrics` endpoint to Prometheus.
+
+
+1. Kong Enterprise users can protect the admin `/metrics` endpoint with an
+   [RBAC user](/enterprise/latest/setting-up-admin-api-rbac) that the
+   Prometheus servers uses to access the metric data. Access through any
+   firewalls would also need to be configured.
+
+2. You can proxy the Admin API through Kong itself, then you can create a route
```
hbagdi: This assumes that the proxy port is running on the same node as admin port.

It is possible to disable proxy (data plane) or admin (control plane) in Kong on certain nodes as needed.

Can we mention this along the lines:
`If all Kong nodes are listening for admin and proxy ports, then it is possible to create a route to /metrics endpoint.`

app/plugins/prometheus.md
```diff
@@ -0,0 +1,171 @@
...
+kong_nginx_metric_errors_total 0
+
+```
+
+In most configurations the Kong Admin API will be behind a firewall or would
+need to be setup to require authentication, here are a couple of options to
+allow access to the `/metrics` endpoint to Prometheus.
+
+
+1. Kong Enterprise users can protect the admin `/metrics` endpoint with an
```
hbagdi: Since this is specific to Kong Enterprise and RBAC, can we move this to the end of this list?

UkiahSmith: I strongly disagree with moving this to the bottom. It is important that we use opportunities like these to let people know that there is an Enterprise version of Kong, and that by referencing it first, show that it is important to us and our customers.

app/plugins/prometheus.md
UkiahSmith: I removed the example, and expanded the explanation. This configuration is too big to be a single bullet point, better to link to the relevant documentation.

app/plugins/prometheus.md
UkiahSmith: Done

app/plugins/prometheus.md
UkiahSmith: Done

app/plugins/prometheus.md
UkiahSmith: Done

thibaultcha:
It seems to me like this PR should be opened against the `release/ce-0.14.0` branch, shouldn't it?

app/plugins/prometheus.md
```diff
@@ -0,0 +1,183 @@
...
+id: page-plugin
+title: Plugins - Prometheus
+header_title: Prometheus
+header_icon: https://konghq.com/wp-content/uploads/2018/06/prometheus.png
+breadcrumbs:
+  Plugins: /plugins
+nav:
+  - label: Usage
+    items:
+      - label: Log Format
```
hbagdi: This label doesn't make sense here since this plugin doesn't log anything.

Instead, we could use `Metrics`.

app/plugins/prometheus.md
```diff
@@ -0,0 +1,183 @@
...
+
+### Reading metrics into Prometheus
+
+Metrics are availble on the Admin API at `http://localhost:8001/metrics`
+endpoint. Note the URL to the Admin API will be specific to your installation.
+
+
+#### Available metrics
+
+- *Status codes*: HTTP status codes returned by upstream services.
```
hbagdi: Can we do **bolds** instead of *Italics* here?

app/plugins/prometheus.md
```diff
@@ -0,0 +1,183 @@
...
+
+- *Status codes*: HTTP status codes returned by upstream services. 
+  These are available per service and across all services.
+- *Latencies Histograms*: Latency as measured at Kong:
+   - *Request*: Total request latency 
+   - *Kong*: Time taken for Kong to route, authenticate and run all plugins for a request
+   - *Upstream*: Time taken by the upstream to respond to the request.
+- *Bandwidth*: Total Bandwidth (egress/ingress) flowing through Kong.
+  This metrics is availble per service and across all services.
+- *DB reachability*: Can the Kong node reach it's Database or not (Guage 0/1).
```
hbagdi: This description can be improved to:
```
A guage type with a value of 0 or 1, representing if DB can be reached by a Kong node or not.
```

app/plugins/prometheus.md
```diff
@@ -0,0 +1,183 @@
...
+endpoint. Note the URL to the Admin API will be specific to your installation.
+
+
+#### Available metrics
+
+- *Status codes*: HTTP status codes returned by upstream services. 
+  These are available per service and across all services.
+- *Latencies Histograms*: Latency as measured at Kong:
+   - *Request*: Total request latency 
+   - *Kong*: Time taken for Kong to route, authenticate and run all plugins for a request
```
hbagdi: Can we make this grammatically better and more generic? How about something like:
```
Time taken by Kong to route and run all plugins for requests
```

app/plugins/prometheus.md
```diff
@@ -0,0 +1,183 @@
...
+Metrics are availble on the Admin API at `http://localhost:8001/metrics`
+endpoint. Note the URL to the Admin API will be specific to your installation.
+
+
+#### Available metrics
+
+- *Status codes*: HTTP status codes returned by upstream services. 
+  These are available per service and across all services.
+- *Latencies Histograms*: Latency as measured at Kong:
+   - *Request*: Total request latency
```
hbagdi: Can we make this more descriptive? Like:
```
Total time taken by Kong and Upstream services to service requests.
```

app/plugins/prometheus.md
```diff
@@ -0,0 +1,183 @@
...
+  This metrics is availble per service and across all services.
+- *DB reachability*: Can the Kong node reach it's Database or not (Guage 0/1).
+- *Connections*: Various NGINX connection metrics like active, reading, writing,
+  accepted connections.
+
+Here is an example of output you could expect from the `/metrics` endpoint:
+
+```bash
+$ curl http://localhost:8001/metrics
+root@vagrant-ubuntu-trusty-64:~# curl -D - http://localhost:8001/metrics
```
hbagdi: It seems like this line can go away since the line above this has the prompt and command.

app/plugins/prometheus.md
```diff
@@ -0,0 +1,183 @@
...
+kong_nginx_http_current_connections{state="waiting"} 0
+kong_nginx_http_current_connections{state="writing"} 1
+# HELP kong_nginx_metric_errors_total Number of nginx-lua-prometheus errors
+# TYPE kong_nginx_metric_errors_total counter
+kong_nginx_metric_errors_total 0
+
+```
+
+
+### Securing the Prometheus plugin
```
hbagdi: The title seems to mismatch the content under it.

Could we do "Setting up access to metrics endpoint"?

app/plugins/prometheus.md
```diff
@@ -0,0 +1,183 @@
...
+#### Available metrics
+
+- *Status codes*: HTTP status codes returned by upstream services. 
+  These are available per service and across all services.
+- *Latencies Histograms*: Latency as measured at Kong:
+   - *Request*: Total request latency 
+   - *Kong*: Time taken for Kong to route, authenticate and run all plugins for a request
+   - *Upstream*: Time taken by the upstream to respond to the request.
+- *Bandwidth*: Total Bandwidth (egress/ingress) flowing through Kong.
+  This metrics is availble per service and across all services.
```
hbagdi: Grammar and Typo.
`This metric is available per service and as a sum across all services`

app/plugins/prometheus.md
```diff
@@ -0,0 +1,183 @@
...
+  api_id: true
+  service_id: true
+  route_id: true
+
+---
+
+### Reading metrics into Prometheus
+
+Metrics are availble on the Admin API at `http://localhost:8001/metrics`
+endpoint. Note the URL to the Admin API will be specific to your installation.
```
hbagdi: It's important to mention the following bit here:
```
This plugin records and exposes metrics at node-level.
Prometheus server will need to discover all Kong nodes via your service discovery mechanism, scrape each of the Kong nodes.
```

UkiahSmith: How would this work for a node that is proxy only?  It sounds like a proxy only node would not have the admin /metrics available at all.

app/plugins/prometheus.md
UkiahSmith: Done.

app/plugins/prometheus.md
UkiahSmith: Updated.

app/plugins/prometheus.md
UkiahSmith: Done.

app/plugins/prometheus.md
UkiahSmith: Done.

app/plugins/prometheus.md
UkiahSmith: Done.

app/plugins/prometheus.md
UkiahSmith: Done.

app/plugins/prometheus.md
UkiahSmith: Done.

app/plugins/prometheus.md
UkiahSmith: Turns out this has to be the same text as the headings. It is used as a link in the sidebar.

app/plugins/prometheus.md
hbagdi: Right, so in that case, one can either use a custom NGINX template or use NGINX directive injection mechanism (shipped with 0.14) to expose these metrics on a separate port.

coopr: Howsabout suggesting that in the doc? Seems like a helpful tip!

UkiahSmith: @hbagdi @coopr  I cleaned up the explanation and hopefully make it easier to understand for someone reading it for the first time.

thibaultcha:
Taking care of getting this into the release branch (with proper rebasing). If needed, we can adjust the docs and do all sorts of tweakings later as well :+1: Just dropping a note that this plugin will need to be added to the gallery at https://konghq.com/plugins/ as well.

thibaultcha:
Cherry-picked into #702. Thanks @UkiahSmith!

PR #776, opened 2018-07-05
fix(serverless) address code snippet issues

nijikokun:
- Escape comma in code sample
- Prefer form-data when submitting files
- Cleanup head-matter content


nijikokun:
@hbagdi mind reviewing this so we can get it out?

PR #792, opened 2018-07-12
docs(ee/plugins) add ldap-auth-advanced plugin docs

darrenjennings:


app/plugins/ldap-authentication-advanced.md
```diff
@@ -0,0 +1,23 @@
+---
+id: page-plugin
+title: Plugins - LDAP Authentication Advanced
+layout: plugin-ee
+header_title: LDAP Authentication Advanced
+header_icon: https://konghq.com/wp-content/uploads/2018/07/EE-ldap.png
+breadcrumbs:
+  Plugins: /plugins
+description: |
+  The LDAP Authentication Advanced plugin for Kong Enterprise Edition is an extended version of the Community Edition LDAP Authentication plugin, with enhanced configuration options and features.
```
gszr: 80 chars per line

app/plugins/ldap-authentication-advanced.md
```diff
@@ -0,0 +1,23 @@
...
+
+---
+
+## Enterprise Features
+
+#### Consumer Mapping
+
+When authenticating via Kong, it is sometimes advantageous to apply plugin behavior at the consumer level (e.g. rate-limiting)
```
gszr: ditto

gszr: Maybe "When authenticating via Kong, it is sometimes advantageous to apply plugin behavior at the consumer level (e.g., rate-limiting); as such, the following options are provided:"?

app/plugins/ldap-authentication-advanced.md
darrenjennings: I don't believe that this is a constraint of this repo so it should not be enforced ATM. However, I'm open to this being the case, it is just hard because if linting does not warn, we will need repo maintainers to enforce... which will be hard due to the nature of this repo not currently belonging to any one team.

gszr: I do think historically getkong.org maintainers enforced this - not 100% sure. Either way, IMHO it's a matter of being a "good citizen" and trying to make the life of people with smaller screens (e.g., me) easier. (Not everyone has or likes bigger screens - or, even if they do, reading lines with 200 chars.)

darrenjennings: I'm trying to be a good citizen I promise 🤞VSCode line wraps only on markdown files by default, which is the IDE I use. I will break these lines in favor of not being IDElitist.

kinman:
Superseded by #802 which uses the same documentation structure as other enterprise plugins.

PR #793, opened 2018-07-12
Clarify difference between GUI key and RBAC token

rainest:
### Summary

Call out the auto-generated RBAC token when creating an admin

### Checklist:
- [x] Documentation 
- [x] Spellchecked my updates
- [x] Ready to be merged 

As far as I know, this distinction will actually go away in the future, but for now it's fairly easy to auto-generate an RBAC user and ignore their token when creating an admin/try to use the key auth credential as an RBAC token unsuccessfully.

### DIFF
--- a/app/enterprise/0.33-x/admin-gui/configuration/authentication.md
+++ b/app/enterprise/0.33-x/admin-gui/configuration/authentication.md
@@ -48,6 +48,8 @@ HTTP/1.1 201 Created
 }
 ```
 
+Save the generated `user_token` (`a3cebf9e-820d-4543-b760-2b6986e3bb9d` in the example above) for use later. It must be sent in the `Kong-Admin-Token` header when RBAC is enabled if using a client (e.g. cURL or HTTPie) other than the Admin GUI.
+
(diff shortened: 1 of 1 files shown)


app/enterprise/0.33-x/admin-gui/configuration/authentication.md
```diff
@@ -48,6 +48,8 @@ HTTP/1.1 201 Created
 }
 ```
 
+Save the generated `user_token` (`a3cebf9e-820d-4543-b760-2b6986e3bb9d` in the example above) for use later. It must be sent in the `Kong-Admin-Token` header when RBAC is enabled if using a client (e.g. cURL or HTTPie) other than the Admin GUI.
```
darrenjennings: Might want to add caveat that if you are using a client that is able to access `:8001` (non-proxied admin api) then it is not required.

darrenjennings: I am referring to the credential. My comment was was meant to be on the entire PR so referring to where he says they are distinct.

PR #797, opened 2018-07-19
Update admin-api.md

shivasheeshyadav:
Fixed the URL which directs to secure-admin-api


coopr:
Thanks for your contribution! 

This change, however, won't fix the issue - the broken link(s - there are more than one) are due to errors in the link URLs at the end of the document. 

The links starting on line 2398 need `/docs` removed from them - can you fix those up?

PR #804, opened 2018-07-27
Docs(logging) reference

coopr:
**NOTE**: Please read the CONTRIBUTING.md guidelines before submitting your patch,
and ensure you followed them all:

https://github.com/Kong/docs.konghq.com/blob/master/CONTRIBUTING.md

### Summary

Adding logging reference

### Checklist:

- [ ] [Commit message & atomicity](https://github.com/Kong/docs.konghq.com/blob/master/CONTRIBUTING.md#commit-atomicity) checked
- [ ] Documentation 
- [ ] Spellchecked my updates
- [ ] Ready to be merged

### DIFF
--- a/app/0.14.x/logging.md
+++ b/app/0.14.x/logging.md
@@ -4,7 +4,7 @@ title: Logging Reference
 
 # Removing Certain Elements From Your Kong Logs
 
-With new regulations surrounding protecting private data like GDPR, there is a chance you may need to change your logging habits. If you use Kong as your API Gateway, this can be done in a single location to take effect on all of your APIs. This guide will walk you through one approach to accomplishing this, but there are always different approaches for different needs. Please not, theese changes will effect the output of the NGINX access logs. This will not have any effect om Kong's logging plugins.
+With new regulations surrounding protecting private data like GDPR, there is a chance you may need to change your logging habits. If you use Kong as your API Gateway, this can be done in a single location to take effect on all of your APIs. This guide will walk you through one approach to accomplishing this, but there are always different approaches for different needs. Please note, these changes will effect the output of the NGINX access logs. This will not have any effect on Kong's logging plugins.
 
 For this example, let’s say you want to remove any instances of an email address from your kong logs. The emails addresses may come through in different ways, for example something like `/apiname/v2/verify/alice@example.com` or `/v3/verify?alice@example.com`. In order to keep these from being added to the logs, we will need to use a custom NGINX template.
 
--- a/app/0.14.x/logging.md
+++ b/app/0.14.x/logging.md
@@ -4,7 +4,7 @@ title: Logging Reference
 
 # Removing Certain Elements From Your Kong Logs
 
-With new regulations surrounding protecting private data like GDPR, there is a chance you may need to change your logging habits. If you use Kong as your API Gateway, this can be done in a single location to take effect on all of your APIs. This guide will walk you through one approach to accomplishing this, but there are always different approaches for different needs.
+With new regulations surrounding protecting private data like GDPR, there is a chance you may need to change your logging habits. If you use Kong as your API Gateway, this can be done in a single location to take effect on all of your APIs. This guide will walk you through one approach to accomplishing this, but there are always different approaches for different needs. Please not, theese changes will effect the output of the NGINX access logs. This will not have any effect om Kong's logging plugins.
 
 For this example, let’s say you want to remove any instances of an email address from your kong logs. The emails addresses may come through in different ways, for example something like `/apiname/v2/verify/alice@example.com` or `/v3/verify?alice@example.com`. In order to keep these from being added to the logs, we will need to use a custom NGINX template.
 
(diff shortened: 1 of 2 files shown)


app/0.14.x/logging.md
```diff
@@ -0,0 +1,125 @@
...
+title: Logging Reference
+---
+
+# Removing Certain Elements From Your Kong Logs
+
+With new regulations surrounding protecting private data like GDPR, there is a chance you may need to change your logging habits. If you use Kong as your API Gateway, this can be done in a single location to take effect on all of your APIs. This guide will walk you through one approach to accomplishing this, but there are always different approaches for different needs.
+
+For this example, let’s say you want to remove any instances of an email address from your kong logs. The emails addresses may come through in different ways, for example something like `/apiname/v2/verify/alice@example.com` or `/v3/verify?alice@example.com`. In order to keep these from being added to the logs, we will need to use a custom NGINX template.
+
+To start using a custom NGINX template, first get a copy of our template. This can be found [https://docs.konghq.com/0.13.x/configuration/#custom-nginx-configuration](https://docs.konghq.com/0.13.x/configuration/#custom-nginx-configuration) or copied from below
```
coopr: Use the correct URL formatting to point to the "matching" Kong version of this doc - you'll see that used in other docs.

coopr: I would suggest you only point to the other doc, and don't duplicate content here - doing so makes maintenance harder in the future.

app/0.14.x/logging.md
```diff
@@ -0,0 +1,125 @@
...
+        server_name custom_server;
+ 
+        location / {
+          ... # etc
+        }
+    }
+}
+```
+
+In order to control what is placed in the logs, we will be using the NGINX map module in our template. For more detailed information abut using the map directive, please see this guide . This will create a new variable whose value depends on values of one or more of the source variables specified in the first parameter. The format is:
```
coopr: "this guide" should link to something?

app/0.14.x/logging.md
john-paul-keenan: I wantted to refer to the version the most up to date Kong EE was built on top of.

If I'm reading this, I'd rather not need to click to another page just to get the template and then need to come back, so I'd prefer to leave it in.

coopr: The document you are authoring is for Kong CE 0.14 - the link here points to Kong CE 0.13 docs.  Instead, this link should use the special URL format that causes it to consistently link to the "matching" version of the doc. Please correct the URL.

app/0.14.x/logging.md
john-paul-keenan: yup, it sure should. Fixed in https://github.com/Kong/docs.konghq.com/pull/820

app/0.14.x/logging.md
john-paul-keenan: The custom NGINX template is different in Kong CE .13 and .14. This guide was primarily written for enterprise customers. Because Kong EE is currently built on .13, I do not want to link to something which is not accurate. 

Maybe we should put this doc under the enterprise section of the site instead? That makes more sense to me.

coopr: If this doc refers to things you can do with Kong CE 0.13 (and thus things you can do with Kong EE 0.33), then it should probably be in the Kong CE 0.13 docs. 

If it also describes things you can do in Kong CE 0.14, then it should _also_ be in the Kong CE 0.14 docs - with necessary adjustments to describe how to do those things in CE 0.14 that differ from how you'd do them in 0.13.

Feel free to make this PR only for Kong CE 0.13 docs (and later do a PR to add a modified version to Kong CE 0.14 docs), or feel free to expand this PR to add this valuable doc to both CE 0.13 and CE 0.14 docs. 

Personally, I'd prefer to keep this in the CE docs section, as documenting it only in the EE docs implies that only EE users can benefit from the advice you give herein - and, it is my understanding that your advice applies to both CE and EE users.

john-paul-keenan: Yup, makes sense to me. I made the changes to the link to use latest. Later on, I may make an EE version, but for now, this will help the more people so I'm happy with that.

p0pr0ck5:
This doc needs a stronger note that this reflects the output of nginx access logs, _*not*_ kong logging plugins

john-paul-keenan:
Thanks @p0pr0ck5 I added in some clarification. LMK if you think it's more clear or needs additional comments

coopr:
@john-paul-keenan is this ready to merge, or does it need more review?

john-paul-keenan:
This is ready to merge in my opinion.

coopr:
@john-paul-keenan I am now removing the "WiP" from the PR title, and the wip label, thanks!

Tieske:
It has a reference to the nginx logs, but for a "logging reference" page, I'd expect overall coverage of the logging subject. Since that's to big a scope, maybe add just a generic intro explaining the different types of logs to expect with a Kong system, and then pointing to individual logging plugins for details, whilst leaving the system logging in this doc.

coopr:
@Tieske is this good enough to deploy as-is, or _must_ the changes you suggested be made before it goes live?

PR #814, opened 2018-08-07
fix(formatting): Reformat Rate-limiting table

catfishbluefish:
### Summary

Tables with long parameter names overflowing content container - increase maximum page width to allow for wider tables

### Full changelog

*  Increased container max width from 1020px to 1100px

### Issues resolved

Fix JPK rate-limiting table overflow bug

###Before:
rate limiting advanced plugin - v0 33-x _ kong - open-source api management and microservice management 2018-08-07 14-25-05

###After
screen shot 2018-08-07 at 3 22 10 pm

### DIFF
--- a/app/_assets/stylesheets/base.less
+++ b/app/_assets/stylesheets/base.less
@@ -13,7 +13,7 @@ body {
   -webkit-font-smoothing: antialiased;
 
   .container {
-    max-width: 1020px;
+    max-width: 1100px;
   }
 
   @media (min-width: @navbar-collapse-point) {
--- a/app/_assets/stylesheets/base.less
+++ b/app/_assets/stylesheets/base.less
@@ -13,7 +13,7 @@ body {
   -webkit-font-smoothing: antialiased;
 
   .container {
-    max-width: 1100px;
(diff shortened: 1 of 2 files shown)


catfishbluefish:
Reformatted table: 

screen shot 2018-08-07 at 3 39 24 pm

app/_assets/stylesheets/base.less
```diff
@@ -13,7 +13,7 @@ body {
   -webkit-font-smoothing: antialiased;
 
   .container {
-    max-width: 1020px;
```
thibaultcha: With table reformatting (yay!), do we still need to increase this width? This seems like a pretty major change to the website, so if we can get away without doing it, that'd be great. Otherwise, I'm afraid we'd need the approval of someone from the design team before going ahead with this change.

catfishbluefish: Yeah I agree - I thought it might help make sure this doesn't happen on any other pages but it doesn't look like it is an issue with the reformatting.

coopr:
https://github.com/Kong/docs.konghq.com/issues/835

PR #844, opened 2018-09-04
feat(docs) add the dns_valid_ttl configuration option

Tieske:
Belongs to https://github.com/Kong/kong/pull/3730

The code is in the `master` branch, but it's not clear to me whether it will end up in a `0.14.2` or it will be `0.15`.


Tieske:
iirc this is auto generated now. It is already on the site at least. closing

PR #858, opened 2018-09-17
fix(anchor-links): Remove duplicate anchor links

leevye:
The docs site uses an extension that will automatically make any header into an anchor link. 

If there more than one header with the same text (aka multiple "Endpoint" headers) on the same page, the anchor link will only point to the first one

To avoid this, and improve SEO by not having too many headers, I have changed every page where "Endpoint" and "Response" appear multiple times to bolded words vs headers

**Endpoint** vs

### DIFF
--- a/app/plugins/aws-lambda.md
+++ b/app/plugins/aws-lambda.md
@@ -128,7 +128,7 @@ When the plugin is added to an API entity (which is deprecated as of 0.13.0),
 it is the `upsream_url` property which must be specified and resolvable as well
 (but ignored).
 
-**Response** plugins
+#### Response plugins
 
 There is a known limitation in the system that prevents some response plugins
 from being executed. We are planning to remove this limitation in the future.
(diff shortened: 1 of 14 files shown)


app/plugins/aws-lambda.md
```diff
@@ -128,7 +128,7 @@ When the plugin is added to an API entity (which is deprecated as of 0.13.0),
 it is the `upsream_url` property which must be specified and resolvable as well
 (but ignored).
 
-#### Response plugins
+**Response** plugins
```
coopr: This update needs to be made in the new `app/_hub/kong-inc/aws-lambda/index.md` file.

leevye: Actually #### Response plugins should be fine - I will actually reverse this one

leevye:
I searched for every instance of ####Endpoint and ####Response and swapped them, so we should be GTG

coopr:
Still some `#### Endpoint` in the docs - https://github.com/Kong/docs.konghq.com/search?q=%23%23%23%23+Endpoint&unscoped_q=%23%23%23%23+Endpoint

Issue #856, opened 2018-09-17
Cookie "dialog" is overlapped by "Kong Inc"


coopr:
https://docs.konghq.com/hub/

### Summary

The string "Kong Inc." overlaps the cookie warning.

### Steps To Reproduce

1. Visit https://docs.konghq.com/hub/ (in an incognito window, to force the cookie "dialog" to appear in the bottom right)
2. Scroll up and down
3. Notice that "Kong Inc." is "above" the cookie dialog

### Additional Details

Appears in both Chrome and Firefox on Mac OSX

![cookie and kong](https://user-images.githubusercontent.com/94904/45644351-6a76aa80-ba72-11e8-8814-423c44786759.gif)

leevye:
https://github.com/Kong/docs.konghq.com/pull/862

PR #862, opened 2018-09-17
fix(cookie-policy): Ensure cookie-policy is on top layer

leevye:
Moves plugin publisher down from Z-Index 1000 to 5

screen shot 2018-09-17 at 1 33 43 pm


leevye:
Solves Issue #856

PR #866, opened 2018-09-17
fix(wallarm): Changed Wallarm folder capitalization

leevye:
Some issues were caused by the wallarm folder being capitalized, need a special git config setting locally to let git notice capitalization changes


coopr:
Pulling and running path renaming doesn't consistently work for me - so I'm approving this based solely on code review.

leevye:
@coopr you need to change your git settings:
`git config core.ignorecase false` otherwise git ignores changes in capitalization

PR #882, opened 2018-09-19
misc(development) sharing my docker development environment

hutchic:
per the updated readme this should work for anyone that has Docker installed by running:

```
make develop
```


README.md
```diff
@@ -4,33 +4,13 @@ This repository is the source code for [Kong](https://github.com/Kong/kong)'s we
...
-npm start
-```
-
-## Deploying
-
-This will deploy to GitHub pages:
-
->
-```bash
-npm run deploy
```
coopr: I'm not sure we should remove these instructions - they work for many users - maybe you can add your Docker instructions as _an_ option, rather than making it the _only_ choice?

hutchic: mostly for simplicity but I can leave both

npm, ruby, rvm, bundler, grunt is quite the list of prerequisites and gotta get the versions right :(

coopr:
Pulled, ran locally, worked well!

PR #891, opened 2018-09-21
misc(ci) various improvmenets related to testing and developing

hutchic:
- the deploy command has *NOT* been tested
- replaced the docker-compose with Makefile commands
- moved local development / test commands into the Makefile
- fixed permissions when using docker
- updated the README.md

before:
```
make develop
Ctrl ^C
ls -la
drwxr-xr-x 884 root root 36864 Sep 21 20:17 node_modules
```

after:
```
make docker-run
Ctrl ^C
ls -la
drwxr-xr-x 884 ubuntu ubuntu  36864 Sep 21 20:17 node_modules
```


coopr:
@jeveleth did you test the deploy command, perchance? If not, let’s get together today and test it please.

coopr:
I've tested the non-Docker `make run` and `make test` commands - they work as expected.

coopr:
Tested:

- Without Docker:
  - `make run` = works as expected
  - `make test` = works as expected
  - `make deploy` = works as expected
- With Docker:
  - `make docker-run` = didn't work, got an error, tried again with `sudo make docker-run`, which seems to serve the docs site from localhost:3000, but results in broken images on http://localhost:3000/hub/ 

That ^^ should just work, ideally without sudo - @hutchic can you have another look?

Makefile
```diff
@@ -1,6 +1,39 @@
-develop: clean
-	docker-compose up --build
+.PHONY: clean install run test deploy docker-run docker-test docker-bash
+
+DOCKER_COMMAND=docker run -it --rm --user=`id -u`:`id -g` -v $$PWD:/srv/jekyll kongdocs
```
jeveleth: If I remove the ```--user=`id -u`:`id -g``` flag, I'm able to run `make docker-run`. Otherwise, I get permissions errors.

hutchic: that will fix it for `make docker-run` but it leaves files on your OS that are owned by root. So if you
```
make docker-run
make run
```
it's full of sadness

A possible solution would be to configure npm to install `node_modules` somewhere not shared with the host OS

jeveleth:
In local testing, I'm still running into the issue of docker hanging at compilation.
```npm WARN lifecycle docs.konghq.com@1.0.0~preinstall: cannot run in wd %s %s (wd=%s) docs.konghq.com@1.0.0 bundle install /srv/jekyll
  ⚠ spawn /srv/jekyll/node_modules/gifsicle/vendor/gifsicle ENOENT
  ⚠ gifsicle pre-build test failed
  ℹ compiling from source
  ✔ gifsicle built successfully
  ⚠ spawn /srv/jekyll/node_modules/jpegtran-bin/vendor/jpegtran ENOENT
  ⚠ jpegtran pre-build test failed
  ℹ compiling from source
  ✔ jpegtran built successfully
  ⚠ spawn /srv/jekyll/node_modules/optipng-bin/vendor/optipng ENOENT
  ⚠ optipng pre-build test failed
  ℹ compiling from source
  ✔ optipng built successfully
```
This has been the case for about five minutes before I had to kill docker, as it was whaling on my CPU.

PR #900, opened 2018-09-25
docs(hub) add optum response-size-limiting plugin

jeremyjpj0916:
### Summary
Adding our latest plugin following Kong Hub process.

### Full changelog

* Added: 
      kong-response-size-limiting/index.md
      optum_kong-response-size-limiting.png

### Checklist:
- [x] Commit message & atomicity
- [x] Spellchecked my updates
- [x] Ready to be merged


coopr:
Pulled, ran locally, made one tiny adjustment, LGTM! 

@thefosk please review.

jeremyjpj0916:
Woot! I see it 👍  , One thing @coopr , look on the page: 

https://docs.konghq.com/hub/optum/kong-response-size-limiting/

See Enterprise compatible versions not listing anything on the page?

https://github.com/Kong/docs.konghq.com/commit/d892f43b21256b8defa250ccf351bdacd7a1b6b4

The commit itself has 

"
    compatible:
      - 0.34-x
"
Something wrong with the underlying docs as code logic to not display the Enterprise edition that is cross compatible with 0.14.x?

coopr:
Hi @jeremyjpj0916 ! Comments below...

> One thing @coopr , look on the page:
> 
> https://docs.konghq.com/hub/optum/kong-response-size-limiting/
> 
> See Enterprise compatible versions not listing anything on the page?
> 
> d892f43
> 
> The commit itself has
> 
> ```
> compatible:
> - 0.34-x
> ```
> 
> Something wrong with the underlying docs as code logic to not display the Enterprise edition that is cross compatible with 0.14.x?

The indication of compatibility with Kong Enterprise 0.34 won't display until 0.34 is released and added to https://github.com/Kong/docs.konghq.com/blob/master/app/_data/kong_versions.yml - the compatibility table in Kong Hub listings depends on releases in that file.

Note too that it is not yet announced whether Kong Enterprise will be based on Kong 0.13.x or 0.14.x. 

> Another thing Hub may want to consider would be giving a way to drop the

> "Enabling the plugin on an API" section when the docs mention only being compatible with 0.13.x - 0.14.x or one or the other because the only reasons really it would be incompatible would be around the PDK or the fact it does not support API resources(which I don't think mine can, so it causes the docs to be a bit inaccurate including that)."

Sure - you _should_ be able to cause this to not appear by setting the compatibility with the API entity to `false` - your listing currently contains:

```
params:
  name: kong-response-size-limiting
  api_id: true
  service_id: true
  route_id: true
  consumer_id: true
```

(I don't know why those have `_id` on the end - doesn't make much sense - hopefully we'll fix that someday :)

> Another challenge I see is getting all these plugins to list compatibility with Kong 1.0. I intend to make sure we have full compatibility with Kong 1.0 with all plugins when the time comes.

Sure, compatibility testing would be ideal, and I get it, such testing isn't always quick and easy. Is there anything we can do to make this better for you?

PR #922, opened 2018-10-04
Docs/loadbalancing dns clarifications

coopr:
### Summary

Clarify DNS and loadbalancing configuration - inspired by https://discuss.konghq.com/t/seeking-additional-details-around-dns-resolution/1997/2 

### Checklist:

- [ ] Documentation N/A
- [x] Spellchecked my updates
- [ ] Ready to be merged 

Not quite ready to be merged - first needs comments from @Tieske or others, then needs to be applied to earlier versions of Kong's docs.

### DIFF
--- a/app/0.14.x/configuration.md
+++ b/app/0.14.x/configuration.md
@@ -1035,7 +1035,6 @@ record types. The `LAST` type means the
 type of the last successful lookup (for the
 specified name). The format is a (case
 insensitive) comma separated list.
-Remove any record types that Kong should ignore. 
 
 Default: `LAST,SRV,A,CNAME`
 
--- a/app/0.14.x/configuration.md
+++ b/app/0.14.x/configuration.md
@@ -1035,6 +1035,7 @@ record types. The `LAST` type means the
 type of the last successful lookup (for the
 specified name). The format is a (case
 insensitive) comma separated list.
+Remove any record types that Kong should ignore. 
 
 Default: `LAST,SRV,A,CNAME`
 
--- a/app/0.14.x/loadbalancing.md
+++ b/app/0.14.x/loadbalancing.md
@@ -49,6 +49,9 @@ Because there is no `weight` information, all entries will be treated as equally
 weighted in the load balancer, and the balancer will do a straight forward
 round-robin.
 
+Note that A records do not carry port information, so if you need Kong to route
+requests to specific port numbers, be sure to [configure Kong to ignore A records][dns-order-config].  
+
 [Back to TOC](#table-of-contents)
 
 #### **SRV records**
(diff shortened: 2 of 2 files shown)


app/0.14.x/loadbalancing.md
```diff
@@ -49,6 +49,9 @@ Because there is no `weight` information, all entries will be treated as equally
 weighted in the load balancer, and the balancer will do a straight forward
 round-robin.
 
+Note that A records do not carry port information, so if you need Kong to route
+requests to specific port numbers, be sure to [configure Kong to ignore A records][dns-order-config].  
+
```
ionosphere80: DNS A records map names to IP addresses. DNS SRV records map service port numbers to names. Thus, in environments with services using ephemeral IP addresses and port numbers, Kong must obtain both DNS A and SRV records for routing purposes. In this case, I recommend additional research to determine the effects of ignoring DNS A records.

app/0.14.x/loadbalancing.md
```diff
@@ -91,14 +94,16 @@ with 527 entries, whereas weights 16 and 32 (or their smallest relative
 counterparts 1 and 2) would result in a structure with merely 3 entries,
 especially with a very small (or even 0) `ttl` value.
 
-- Some nameservers do not return all entries (due to UDP packet size) in those
-cases (for example Consul returns a maximum of 3) a given Kong node will only
+- Some nameservers do not return all entries due to UDP packet size
+(for example Consul returns a maximum of 3 DNS entries). In those
```
ionosphere80: The Consul DNS server (and possibly other servers) returns a maximum of three values for a particular record via conventional UDP query to prevent returning a truncation flag. It returns all values for a particular record via EDNS UDP query or TCP query. Conventional DNS clients/libraries (without support for EDNS) such as `lua-resty-dns` in Kong only revert to a TCP query if the UDP response includes a truncation flag. Thus, Kong never reverts to a TCP query in environments using Consul DNS and may not consistently route to all instances of a service if more than three instances exist.

app/0.14.x/configuration.md
```diff
@@ -1035,6 +1035,7 @@ record types. The `LAST` type means the
 type of the last successful lookup (for the
 specified name). The format is a (case
 insensitive) comma separated list.
+Remove any record types that Kong should ignore.
```
ionosphere80: Consider mentioning that once Kong reverts to the `LAST` record type for a particular service, it no longer performs DNS queries for the service regardless of the TTL.

coopr:
@Tieske I'd suggest you make the next round of edits to this PR - you should have the permission to add commits to this branch. Thanks!

app/0.14.x/configuration.md
Tieske: I don't think that comment is accurate. Anyway this is the short description, so let's keep it short.

app/0.14.x/loadbalancing.md
Tieske: This comment should be removed. Because often lookups are chained, eg. a name points to and SRV, which contains more names, which point to A records. In this case Kong will preserve the last port found, so it will take the port number from the SRV and the IP address from the A.

app/0.14.x/loadbalancing.md
Tieske: I think the problem here is with Consul, which tries to be smart and prevents returning the truncation flag. It shouldn't do this, or at least have the behaviour configurable.

ionosphere80: From the Consul documentation...

> When the DNS query is performed using UDP, Consul will truncate the results without setting the truncate bit. This is to prevent a redundant lookup over TCP that generates additional load. If the lookup is done over TCP, the results are not truncated.

However, one can configure Consul to use the truncate flag via the `enable_truncate` option. I would still prefer a way to configure Kong to always use TCP queries if the resolver library cannot support EDNS.

app/0.14.x/loadbalancing.md
ionosphere80: Does the Kong resolver handle linking SRV records to A records or must the `dns_order` option include `SRV,A` to use both of them?

app/0.14.x/configuration.md
coopr: @Tieske can you clarify what the effect is of removing a record type from this list? I see you removed the sentence I added here.

app/0.14.x/configuration.md
```diff
@@ -1035,7 +1035,6 @@ record types. The `LAST` type means the
 type of the last successful lookup (for the
 specified name). The format is a (case
 insensitive) comma separated list.
-Remove any record types that Kong should ignore.
```
coopr: @Tieske can you clarify why this comment was removed? 

If you want Kong to ignore certain record types, can you remove them from this list? If not, then is it a requirement that all 4 record types remain present in this list?

Tieske: <opinion> too detailed </opinion>

PR #927, opened 2018-10-08
docs(hub) fix up optum response-size-limiting plugin

jeremyjpj0916:
### Summary

Set api_id: false to disable documentation around setting response size limiting to API resources as plugin does not support the deprecated resource.

Thx @coopr for the insight in our discussions over here:
https://github.com/Kong/docs.konghq.com/pull/900


coopr:
Thank _you_ @jeremyjpj0916 !

Issue #890, opened 2018-09-21
Remove "underline text on rollover" effect from Kong Hub cards


coopr:
When visiting https://docs.konghq.com/hub/, rolling over a card causes the text on that card to be underlined. This is distracting and unnecessary.

caseydierking:
I’ll take a look at this!

coopr:
Whoops, this got fixed but the issue didn't get closed.

PR #928, opened 2018-10-08
fix(hub): remove <a> tag underline for hub cards on hover

Jlawlzz:


coopr:
Pulled and ran locally - LGTM - :shipit:

coopr:
Fixes #890

PR #956, opened 2018-10-19
doc: Specify that kong host names are case-sensitive

### LINKED ISSUE #586: Specify that kong host names are case-sensitive.
As illustrated in https://github.com/Kong/getkong.org/pull/456, users might be surprised when they create an API for "EXAMPLE.COM" and find out that it is not matched by a request for "example.com". We should indicate that the `host` param in APIs (and Routes and Services) is case-sensitive.

cgomezmendez:
### Summary
Specify that kong host names are case-sensitive

### Full changelog

* Edited getting started to add about host name case sensitive
* Edited admin to add about host name case sensitive

### Issues resolved

Fix #586

### Checklist:

- [x] Documentation 
- [x] Spellchecked my updates
- [x] Ready to be merged

### DIFF
--- a/app/0.10.x/admin-api.md
+++ b/app/0.10.x/admin-api.md
@@ -1914,3 +1914,6 @@ HTTP 204 No Content
 [clustering]: /{{page.kong_version}}/clustering
 [cli]: /{{page.kong_version}}/cli
 [secure-admin-api]: /{{page.kong_version}}/secure-admin-api
+
+
+Notice that when matching a request against an api's hosts field, the match is case sensitive.
--- a/app/0.10.x/getting-started/adding-your-api.md
+++ b/app/0.10.x/getting-started/adding-your-api.md
@@ -20,6 +20,8 @@ helpful for learning how Kong proxies your API requests.
 Kong exposes a [RESTful Admin API][API] on port `:8001` for managing the
 configuration of your Kong instance or cluster.
 
+When matching a request against an api's hosts field, the match is case sensitive.
+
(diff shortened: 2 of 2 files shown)


app/0.10.x/admin-api.md
```diff
@@ -1914,3 +1914,6 @@ HTTP 204 No Content
 [clustering]: /{{page.kong_version}}/clustering
 [cli]: /{{page.kong_version}}/cli
 [secure-admin-api]: /{{page.kong_version}}/secure-admin-api
+
+
+Notice that when matching a request against an api's hosts field, the match is case sensitive.
```
coopr: While this statement may be true, it is in an incorrect place in the docs - please move it to the correct location.

Additionally, please add this clarifying note to the other version's of Kong's documentation that need it (this change is only on Kong 0.10 docs - it likely applies to other versions of Kong)

coopr: Line 8 would likely be a better place to add the note.

app/0.10.x/getting-started/adding-your-api.md
```diff
@@ -20,6 +20,8 @@ helpful for learning how Kong proxies your API requests.
 Kong exposes a [RESTful Admin API][API] on port `:8001` for managing the
 configuration of your Kong instance or cluster.
 
+When matching a request against an api's hosts field, the match is case sensitive.
```
coopr: This note is in the wrong place. Please move it to a more correct location.

PR #957, opened 2018-10-19
fix(rbac) display tables correctly

kidd:
Adding needed blank line to display tables as such.


coopr:
Great fix, thanks @kidd

PR #978, opened 2018-10-23
Replace depricated gulp-util

prog1dev:
[`gulp-util`](https://www.npmjs.com/package/gulp-util) has been deprecated. Continuing to use this dependency may prevent the use version 4 of Gulp.

See:
- https://medium.com/gulpjs/gulp-util-ca3b1f9f9ac5
- https://github.com/gulpjs/gulp-util/issues/143

### DIFF
--- a/gulpfile.js
+++ b/gulpfile.js
@@ -2,7 +2,7 @@
 
 var browserSync = require('browser-sync').create()
 var childProcess = require('child_process')
-var gutil = require('gulp-util')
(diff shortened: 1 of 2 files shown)


coopr:
Pulled, ran locally, seems to work - @prog1dev are there any particular steps I should take to test and confirm this change?

prog1dev:
@coopr Its a pretty straight forward change based on links in desc so everything should be ok

PR #998, opened 2018-10-26
Add currently active version to version dropdown

josiahdahl:
### Summary

Add the currently selected version of the documentation to the version selection dropdown so that users can see what version they're viewing easily.

**Documentation Example**
image

**Plugin Example**
image

**Lua Reference Example**
image

I couldn't find a way that looked good to change the width to make the button and dropdown the same size, so that's left off of this PR.

### Full changelog

* Show currently selected documentation version in version dropdown

### Issues resolved

Fix #886 

### Checklist:

- [X] [Commit message & atomicity](https://github.com/Kong/docs.konghq.com/blob/master/CONTRIBUTING.md#commit-atomicity) checked
- [ ] Documentation N/A
- [ ] Spellchecked my updates N/A
- [X] Ready to be merged

PR #1005, opened 2018-10-30
Add better version switching experience

### LINKED ISSUE #413: [docs] Redirect to the same page when change version
When change version not redirect to the index version page, instead redirect to the same page.

vividh:
### Summary
Currently, switching to a version from dropdown always takes to the home page of the docs. This PR adds functionality to retain the context of the current article in the selected version.

### Full changelog
* Augmented `versions` custom jekyll plugin to track an article's timeline through all versions
* Updated `docs.html` layout to include article's slug in version dropdown
* Added `(ce/ee)_docs_version_timeline.yml` to configure version timeline for articles that were created, renamed, deleted etc.

### Issues resolved
Fix #413 

### Checklist:
- [x] Documentation
- [x] Spellchecked my updates
- [x] Ready to be merged

### Configuration required before merge ready
By default, the plugin uses the current article slug when switching context to a different version, but that would not work if articles were renamed or created after the first version.

Following is a list of articles for which configuration has to be provided in [app/_data/docs_version_timeline.yml](https://github.com/Kong/docs.konghq.com/pull/1005/files#diff-d19e8e634ce1b71f7bf6d07e78b3b7c8) before the PR can be merged. *Keep adding to the list any articles that require configuration for this feature.*

Community Edition docs:
- [x] Getting Started - Configuring a Service (Previously *Adding Your API*)
- [x] Clustering Reference
- [x] Authentication Reference
- [x] Health Checks and Circuit Breakers Reference
- [x] Loadbalancing reference
- [x] Logging Reference
- [x] Network & Firewall
- [x] Proxy Reference
- [x] Securing the Admin API
- [x] Plugin Development Guide
- [x] Plugin Development Kit

Enterprise Edition docs:
- [x] Authenticating the Developer Portal (0.31-x)
- [x] How to install Kong Enterprise and PostgreSQL onto Ubuntu (0.32-x)
- [x] Rate Limiting Advanced Plugin (0.31-x)
- [x] RBAC API (0.32-x) / Section - RBAC (0.33-x)
- [x] Section - Workspaces (0.33-x)
- [x] Kong Admin GUI (0.32-x) / Section - Kong Admin GUI (0.33-x)
- [x] Section - Kong Developer Portal
- [x] Kong Deployment Guide (0.33-x)
- [ ] Kong Enterprise Changelog (**Remove version dropdown?**)
- [x] Upgrades and Migrations (0.32-x)
- [x] Setting Up Admin API RBAC (0.32-x)


coopr:
@vividh ok, first of all - AWESOME, this is so great!!

Second, the dropdown appearance and behavior for older docs version is interesting! Despite me specifying that users should be taken to the "home page" when an older docs version doesn't exist, now I'm second-guessing myself - and I want to check with some colleagues before I either accept your PR or request an adjustment. 

Additionally, could I ask that you make an attempt to add just a little documentation about this feature to a new section in CONTRIBUTING like https://github.com/Kong/docs.konghq.com/blob/master/CONTRIBUTING.md#table-of-contents-generator ? I'd suggest you clarify there how this feature works under at least these circumstances:

- A new docs page is added to a given version of Kong 
- An existing docs page is renamed from one version of Kong to the next
- A docs page in one version of Kong is dropped from the next version

app/_data/docs_version_timeline.yml
```diff
@@ -0,0 +1,18 @@
...
+#   slug: <article's latest slug>
+#   aliases: list of article's renamed slugs (if any)
+#   timeline:
+#     <release_version>: <slug_override>
+#
+# if override value is PAGE_CREATED, all versions below that will be set to redirect to index "/"
+
+-
+  slug: "getting-started/configuring-a-service"
+  aliases: ["getting-started/adding-your-api"]
```
coopr: Can you explain to me where these `aliases` come from? And why, in this case, is this the alias?

vividh:
> Second, the dropdown appearance and behavior for older docs version is interesting! Despite me specifying that users should be taken to the "home page" when an older docs version doesn't exist, now I'm second-guessing myself - and I want to check with some colleagues before I either accept your PR or request an adjustment.
> 

Sure! Please let me know if there is a better alternative to landing on the home page. Maybe a 404 page with the message **_This article doesn't exist_**, and a search bar for the user to engage with?

> Additionally, could I ask that you make an attempt to add just a little documentation about this feature to a new section in CONTRIBUTING like https://github.com/Kong/docs.konghq.com/blob/master/CONTRIBUTING.md#table-of-contents-generator ? I'd suggest you clarify there how this feature works under at least these circumstances:
> 
> * A new docs page is added to a given version of Kong
> * An existing docs page is renamed from one version of Kong to the next
> * A docs page in one version of Kong is dropped from the next version

Absolutely will do! The case where a doc is dropped in a version is not currently handled. I'll look into that as well.

app/_data/docs_version_timeline.yml
vividh: The `aliases` are basically names of an article in previous versions. Perhaps I should use a more descriptive name here?

Example:
The **Adding Your API** article (created in 0.2x) was renamed to **Configuring a Service** in 0.13x. The `aliases` property is required to make older versions of the page point to the newer renamed version. For example, when browsing **Adding Your API** in version 0.5x, it would now correctly point to **Configuring a Service** for versions above 0.12x.

coopr:
@steph-kalacis I'm looking forward to reviewing this with you soon and seeing your comments on ideal UX herein...

coopr:
Also FYI @vividh https://github.com/Kong/docs.konghq.com/pull/998 may affect your work in progress in this PR.

steph-kalacis:
In my opinion, the ideal UX is as follows:

If the user is trying to access a docs page that doesn't exist for the version they are on, instead of being sent home, they should be presented with a page that lets them know that page doesn't exist for that version. Similar to the 404 idea you had, @vividh. 

Example:
- I'm on https://docs.konghq.com/0.14.x/auth/ and want to go to version 0.3.x (which doesn't exist)
- I select the dropdown to choose version 0.3.x
- I get a page that says "Sorry, the documentation you're looking for does not exist in _(for?)_ this version. Please navigate to the `home page` to see all documentation for this version."

@coopr, how does that copy look?

coopr:
Sounds good to me @steph-kalacis - thanks for your UX advice!

Now, how does it sound to you @vividh ?

vividh:
> If the user is trying to access a docs page that doesn't exist for the version they are on, instead of being sent home, they should be presented with a page that lets them know that page doesn't exist for that version. Similar to the 404 idea you had, @vividh.

Agreed. I think this would be a good way to handle it. Additionally, I would like to add a small indicator in the version dropdown that informs the user that this article is not present in that version.

> Now, how does it sound to you @vividh?

@coopr This sounds good. I'll get started on the changes.

Follow up question regarding the new 404 page: Is there a design pipeline for Kong pages, wherein I should expect a design spec for the page, or could I come up with the page design myself and get your thoughts?

coopr:
I was reminded that we already have a design pattern and, apparently, code to handle docs pages that don't exist in prior and/or future versions - see https://docs.konghq.com/0.13.x/lua-reference/ 

lua_reference_-_v0_13_x___kong_-_open-source_api_management_and_microservice_management

coopr:
I'm trying to find out the historical reasons why the Lua Reference page uses a different version dropdown... will share what I learn.

coopr:
If the same `Not available for #.##` approach is straightforward to implement site-wide, I'd be fine with it - let's see if @steph-kalacis has an opinion too.

steph-kalacis:
Also fine with that! @coopr @vividh 

Sounds good to me.

coopr:
Wow, nice work, you are a star @vividh - I'll be reviewing this soon...

app/_data/docs_version_timeline.yml
thibaultcha: @vividh What happens/is there a procedure if a page is renamed more than once? Or a page is deleted from a particular version of the docs (e.g. 0.10), and added back in a subsequent version (0.11)?

CONTRIBUTING.md
```diff
@@ -374,6 +375,58 @@ will cause a broken ToC, and should be corrected to:
...
+If a doc page (e.g. [Authenticating the Developer Portal](https://docs.konghq.com/enterprise/0.31-x/developer-portal/authentication/)) is deleted in a version (e.g. `0.32-x`), then it must be configured as such:
+```yml
+-
+  slug: "developer-portal/authentication"
+  timeline:
+    0.32-x:  "PAGE_DELETED"
+```
+
+#### An existing doc page is renamed
+If a doc page (e.g. [Adding your API](https://docs.konghq.com/0.11.x/getting-started/adding-your-api/)) is renamed in a version (e.g. `0.13.x`), then it must be configured as such:
```
thibaultcha: The examples given in this document are meaningful and helpful since they highlight a document at a given version (e.g. 0.11.x), and what happens to this document in a later version (here, 0.13.x). However, once the Markdown rendered, those examples will be less meaningful, because the version number will now show up in the text, but only in the link's `href` attribute.

I suggest:
```suggestion
If a documentation page (e.g. [Adding your API in 0.11.x](https://docs.konghq.com/0.11.x/getting-started/adding-your-api/)) is renamed in a subsequent version (e.g. `0.13.x`), then its timeline must be specified as such:
```

app/_data/docs_version_timeline.yml
thibaultcha: This comment above in the diff seems to answer my question (or the part about multiple renames):

> Note: For a renamed article, specify the latest slug under `slug`, and other slugs under `aliases`.
> Specify the old renamed slug under the latest version that it existed in.

vividh: > @vividh What happens/is there a procedure if a page is renamed more than once? Or a page is deleted from a particular version of the docs (e.g. 0.10), and added back in a subsequent version (0.11)?

Yes, multiple renames, deletions and creations are handled as well. For example, following is a perfectly valid configuration:
```yml
-
  slug: "getting-started/configuring-a-service"
  aliases: ["getting-started/adding-your-api", "getting-started/service-configuration"]
  timeline:
    0.12.x: "getting-started/service-configuration"
    0.10.x: "PAGE_CREATED"
    0.9.x:  "PAGE_DELETED"
    0.8.x:  "getting-started/adding-your-api"
    0.3.x:  "PAGE_CREATED"
```
and the end result would be:
screen shot 2018-11-10 at 11 05 40 am

Here, versions `0.3.x - 0.8.x` point to `getting-started/adding-your-api`, versions `0.10.x - 0.12.x` point to `getting-started/service-configuration`, and versions above those point to the latest article slug i.e. `getting-started/configuring-a-service`.

thibaultcha: @vividh Great, thank you. That appeared more and more evident after the comment noted above :) Nice!

coopr:
More awesome improvements @vividh , I'm reviewing...

PR #920, opened 2018-10-04
Automatically build TOCs for pages

allejo:
### Summary

Greetings @coopr and Kong Doc maintainers!

Create an issue and you shall receive, especially during Hacktoberfest 🎃 I'm the guy lacking sanity who built [jekyll-toc](https://github.com/allejo/jekyll-toc), a GitHub Pages compatible way of building table of contents automagically ✨

Before continuing any further work, to save both our times I just wanted to present a quick proof of concept of how this would work.

`jekyll-toc` works by reading in HTML and extracting out H1-H6 headings, getting their IDs, and then building a TOC in markdown with that information and then `markdownify`ing it.

Since it seems that "Terminology" and "Configuration" are automatically rendered inside of the `extensions.html` layout, we'll need to combine that with `{{ content }}` and then feed that into `toc.html` so that it may generate all the headings correctly; this is being achieved by using a `{% capture %}`.

If this is something that interests you, I'd be more than happy to continue working on this but I would need *a lot* of help checking pages to make sure things are being rendered correctly. **And** checking that headings are in correct order. (e.g. ["Usage" is an H3 instead of H2 on the ACL page](https://docs.konghq.com//hub/kong-inc/acl/))

### Full changelog

* Added cool new automated ToC

### Issues resolved

Fix #894 

### Checklist:

- [ ] [Commit message & atomicity](https://github.com/Kong/docs.konghq.com/blob/master/CONTRIBUTING.md#commit-atomicity) checked
- [x] Documentation 
- [x] Spellchecked my updates
- [ ] Ready to be merged

PR #1015, opened 2018-11-01
Fix pdk doc headings

kikito:


app/0.14.x/pdk/index.md
```diff
@@ -20,7 +20,7 @@ The Plugin Development Kit (or "PDK") is set of Lua functions and variables
  `kong.request`, `kong.log`, etc...
 
 
-## Table of Contents
+### Table of Contents
```
coopr: Please remove the manual Table of Contents and allow the new auto-ToC to work its magic!

app/0.14.x/pdk/index.md
```diff
@@ -47,7 +47,7 @@ The Plugin Development Kit (or "PDK") is set of Lua functions and variables
 
 
 
-### <a name="kong_version"></a>kong.version
+#### <a name="kong_version"></a>kong.version
```
coopr: Please remove the `<a name="kong_version"></a>` code from headers - it is not needed

kikito:
@coopr 	I gave a look at the PR which adds the ToCs to all the pages (#920) and realized that it excludes `index.md` from the automatic ToC generation on purpose.

I can deactivate all the PDK ToCs at once, or leave them all active, but not deactivate "all ToCs but keep the one on `index.md`. I will need a bit more time to think how to handle this best. I will mark this as "do not merge" for the time being.

coopr:
Ah yeah, the reason auto-ToC excludes `index.md` was to avoid adding ToC to pages like https://docs.konghq.com/ and https://docs.konghq.com/enterprise/ - note that some other `index.md` pages _do_ get auto-ToCs, like https://docs.konghq.com/hub/kong-inc/canary/

thibaultcha:
In the case of the PDK, our documentation is auto-generated by LDoc, and we maintain our own template, which is different than the Jekyll template. It is most likely an exception that warrants the PDK to maintain its own ToCs.

kikito:
I decided to take the TOC out of the PDK and let the auto-doc handle that for two reasons:
* The TOCs we generate application-wise and the ones generated by the PDK were slightly different in format. Removing the TOC from the PDK and leaving that to the TOC plugin was less work than adapting the PDK to comply with the application-wide format.
* The LDoc templating is difficult to use (it is a sandboxed env where only the ldoc-approved functions are available - if I remember correctly even `pairs` was not available). Frankly, I think the less stuff we do with it, the better.

@coopr I fixed the problem with handling `index.md` on the PDK (it is now possible to "force-generate" the toc when a file is called `index.md` by setting `toc` to `true` at the top. The PDK pages now all contain `toc: true`, including the index).

I have removed the `do not merge` tag and consider this ready for review now.

thibaultcha:
@kikito Sounds good! 👌🏻

coopr:
Thanks @kikito - please add a brief mention of how to use the new `toc: true` functionality to https://github.com/Kong/docs.konghq.com/blob/master/CONTRIBUTING.md#table-of-contents-generator

coopr:
Also @kikito does https://github.com/Kong/docs.konghq.com#generating-the-plugin-development-kit-documentation need any updates? I don't see anything obvious that needs updating, but while you are updating the docs site docs, you might as well have a look :)

coopr:
> @coopr I fixed the problem with handling `index.md` on the PDK (it is now possible to "force-generate" the toc when a file is called `index.md` by setting `toc` to `true` at the top. The PDK pages now all contain `toc: true`, including the index).

@kikito can you clarify why _all_ the PDK pages contain `toc: true`? The non-index pages don't seem to need it (at least in my quick test I see the ToC is auto-generated whether or not `toc: true` is present) and to make life easier on future us, I'm reluctant to have pages include code that is unnecessary.

PR #1019, opened 2018-11-02
Removed button styling from <ul> items

### LINKED ISSUE #917: All pages should correctly render both unordered lists AND download buttons
Currently the pages under /install (all the pages linked to from the buttons on https://konghq.com/install/) have special handling for Markdown unordered lists - if you edit a page like https://github.com/Kong/docs.konghq.com/blob/master/app/install/kubernetes.md to add an unordered list without links:

```
## Table of Contents

- foo
- bar
- bang
```

It will be rendered strangely:

install_-_kong_on_kubernetes___kong_-_open-source_api_management_and_microservice_management

and if you make the unordered list contain links:

```
## Table of Contents

- [foo](http://www.konghq.com)
- [bar](http://www.konghq.com)
- [bang](http://www.konghq.com)
```

it will be rendered as download buttons:

install_-_kong_on_kubernetes___kong_-_open-source_api_management_and_microservice_management

Note that entering the same unordered list Markdown on pages outside of `/install` works as expected, and renders correctly as an unordered list.

To resolve this issue, we need a way to have a choice of showing download buttons, or unordered lists, on all pages in the site - note that this approach not only allows display of unordered lists on `/install` pages, but it also allows inclusion of download buttons on pages that are not in `/install`

### Suggested Resolution

There seem to be various options to resolve this https://stackoverflow.com/questions/40688633/how-can-i-add-a-button-in-a-md-file-with-jekyll - I think I prefer option 2, and I like https://stackoverflow.com/a/40753294

adamjkuhn:
added a button include to apply these styles, reverted <ul> to default bulleted list styles (install pages)

screen shot 2018-11-02 at 9 27 30 am

Button include markup is as follows:
{% include button.html button_name="Foo" button_class="download" %}
(classname 'download' will include download icon)

### Full changelog

* updated unordered list and unordered list links to standard disc formatting
* created new button tags w/ optional download styling
* added guideline unordered to /app/install/kubernetes.md

### Issues resolved

Fix #917

### DIFF
--- a/app/_assets/stylesheets/pages/install-method.less
+++ b/app/_assets/stylesheets/pages/install-method.less
@@ -64,59 +64,53 @@
     }
 
     ul {
-      list-style: none;
-      padding: 0;
+      list-style: disc;
+      padding: 0 0 0 20px;
       margin: 0;
 
       li {
         margin-right: 10px;
+        display: list-item;
 
-        display: inline-flex;
-        align-items: center;
+      }
+    }
 
-        a {
-          display: flex;
-          align-items: center;
+    .btn{
+      display: inline-block;
+      margin-right:10px;
+      position:relative;
 
-          &:extend(.button all);
-          border-color: #d4d7d9;
-          background-image: linear-gradient(to top, #f7f9fa, #fcfcfc);
-          color: @gray-dark;
-          font-size: 14px;
-          line-height: 39px;
-          height: 40px;
-
-          em {
-            margin-left: 5px;
-            font-style: normal;
-            color: #aab0b2;
-          }
-
-          &:active {
-            background: #f7f9fa;
-          }
-
-          &:before {
-            content: '';
-            display: inline-block;
-            width: 16px;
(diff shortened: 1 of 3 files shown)


coopr:
OK, progress! The unordered lists on http://localhost:3000/install/kubernetes/ now render as unordered lists. 

Next up - the styling of the unordered lists differs between the docs pages and /install pages - ideally it'd all be the same style:

install_-_kong_on_kubernetes___kong_-_open-source_api_management_and_microservice_management_and_proxy_reference_-_v0_14_x___kong_-_open-source_api_management_and_microservice_management

Also, while unordered lists look good on ^^, the former underordered lists on other pages, which need to be buttons, now appear as unordered lists - so we need the buttons back - can you update at least one page to show how those buttons should be coded, and I can do the rest of them?

install_-_ubuntu___kong_-_open-source_api_management_and_microservice_management_and_install_-_ubuntu___kong_-_open-source_api_management_and_microservice_management

PR #1042, opened 2018-11-14
docs (install) optional step to add kong.conf

polarweasel:
### Summary

The macOS (Homebrew) and source install instructions don't mention that you need to manually add a `kong.conf` file, nor where to find one. This fixes that. :)

(Also made a few very small Markdown and style changes, which should be very obvious in the diff.)

### Full changelog

* Added a step to `macos.md` and `source.md` to add `/etc/kong/kong.conf`.

* Fixed a bit of Markdown formatting (use `1.` to start each list item, add `bash` to a code snippet).

* A couple of small language edits.

### Checklist:

- [x] Documentation 
- [x] Spellchecked my updates
- [x] Ready to be merged

### DIFF
--- a/app/install/macos.md
+++ b/app/install/macos.md
@@ -19,39 +19,52 @@ breadcrumbs:
 
     Use the [Homebrew](https://brew.sh/) package manager to add Kong as a tap and install it:
 
-    ```
+    ```bash
     $ brew tap kong/kong
     $ brew install kong
     ```
 
-2. **Prepare your database**
+1. **Add `kong.conf`**
+
+    **Note**: This step is **required** if you are using Cassandra; it is **optional** for Postgres users.
+
+    By default, Kong is configured to communicate with a local Postgres instance.
+    If you are using Cassandra, or need to modify any settings, download the [`kong.conf.default`](https://raw.githubusercontent.com/Kong/kong/master/kong.conf.default) file and [adjust][configuration] it as necessary.
+    Then, as root, add it to `/etc`:
+
+    ```bash
+    $ sudo mkdir -p /etc/kong
+    $ sudo cp kong.conf.default /etc/kong/kong.conf
+    ```
+
+1. **Prepare your database**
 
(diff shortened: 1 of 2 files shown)


coopr:
Needs review and approval by @Kong/team-core

app/install/macos.md
```diff
@@ -19,39 +19,52 @@ breadcrumbs:
...
-    ```
+    ```bash
     $ brew tap kong/kong
     $ brew install kong
     ```
 
-2. **Prepare your database**
+1. **Add `kong.conf`**
+
+    **Note**: This step is **required** if you are using Cassandra; it is **optional** for Postgres users.
```
thibaultcha: No, this step is **never** required. One can create their own configuration file and store in anywhere. One can also rely on environment variables to configure Kong.

polarweasel: Thank you! Am I understanding correctly that:

- A change to _either_ the config file or some environment variables is required for Cassandra 
- You can point Kong to a config file wherever you want

My goal with this was to keep the Homebrew (and source) installs as simple as possible. Would you rather make this a bit more complicated with inline instructions laying out the options, or leave it as “do this for Cassandra” and add a note at the end explaining the other options? My preference is for the latter option.

thibaultcha:
The latter is too opinionated and not friendly (it also requires root access, which is not Homebrew-friendly). 

The proper fix for this is to have the Homebew formula install the config file somewhere and document it. Probably in the brew tap, I don’t remember from the top of my head. 

The documentation changes to make in the meantime, are to explain the above (hinting at future improvements to the formula, for which PRs would be welcome), and explain that one can create their own config file and use the `-c` CLI option. Or rely on environment variables. Most likely a link to the documentation around the CLI is enough.

app/install/macos.md
thibaultcha: Oh, I replied from a different UI, which didn’t add my comment to this specific thread.

polarweasel: Thanks! Will update PR. Agreed that the homebrew recipe change should be a separate issue, but happy to look into that one as well.

polarweasel: Further updates. Please review when you get a chance, @thibaultcha...

sagarkal:
I still had this issue: https://docs.konghq.com/install/macos/ it doesn't specify that kong.conf need to be downloaded

PR #1053, opened 2018-11-20
docs(ee/property-table) fix table formatting

h3xar0n:
### Summary

Formats the table for the `admin_gui_auth` reference. Instead of making a markdown table, the format is as a code block to be consistent with other property reference documentation.

### Full changelog

* Format table as a code block

### Checklist:
- [x] Documentation
- [x] Spellchecked my updates
- [x] Ready to be merged

### DIFF
--- a/app/enterprise/0.34-x/property-reference.md
+++ b/app/enterprise/0.34-x/property-reference.md
@@ -199,11 +199,12 @@ authentication plugin you have chosen.
 
 Supported Plugins:
 
-Value to Use        | Authentication Type
-----------------------+--------------------------
-basic-auth          | Basic Authentication
-ldap-auth-advanced  | LDAP Authentication
-
+```
+ Value to Use      | Authentication Type  
+-------------------|----------------------
+basic-auth         | Basic Authentication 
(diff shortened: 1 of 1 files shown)


coopr:
This doesn't make sense to me as a table, either table-formatted or code-formatted. 

I'd suggest simplifying and clarifying by just presenting the two options code-formatted, and linked to their respective plugin pages:

`[basic-auth](/hub/kong-inc/basic-auth/)` or `[ldap-auth-advanced](/hub/kong-inc/ldap-auth-advanced/)`

Ideally these URLs ^^ would be "version locked" to this version, but I versioned plugin docs either don't exist yet, or can't be referred to via URL, so... I guess that could get fixed later.

travispamaral:
I agree that the table is a little confusing (value to use / Authentication Type). Maybe reword or just show the format needed for conf file

kinman:
oops, didn't finish my thought. Thinking out loud on a PR, we might want to distinguish more clearly between authentication for Kong Manager and authentication for Kong-proxied services.

PR #1058, opened 2018-11-21
docs(ee/smtp) warning about not configuring smtp

h3xar0n:
### Summary

Makes users aware of a potential issue if they do not configure SMTP correctly. Currently, the application does not provide an error message; noting it in documentation is the best temporary solution.

### Full changelog

* Mention importance of SMTP configuration in Networking section

### Checklist:
- [x] Documentation
- [x] Spellchecked my updates
- [ ] Ready to be merged

### DIFF
--- a/app/enterprise/0.34-x/kong-manager/configuration/networking.md
+++ b/app/enterprise/0.34-x/kong-manager/configuration/networking.md
@@ -74,7 +74,7 @@ configured incorrectly, e.g. if they point to a non-existent
 email address, Kong Manager will not display an error message. 
 
 In addition, refer to the 
-[general SMTP configuration](enterprise/{{page.kong_version}}/property-reference/#general-smtp-configuration) 
+[general SMTP configuration](/enterprise/{{page.kong_version}}/property-reference/#general-smtp-configuration) 
 shared by Kong Manager and Dev Portal.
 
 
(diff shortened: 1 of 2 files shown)


app/enterprise/0.34-x/kong-manager/configuration/networking.md
```diff
@@ -68,6 +68,10 @@ Emails from Kong Manager require the following configuration:
 * [`admin_emails_reply_to`](/enterprise/{{page.kong_version}}/property-reference/#admin_emails_reply_to)
 * [`admin_invitation_expiry`](/enterprise/{{page.kong_version}}/property-reference/#admin_invitation_expiry)
 
+⚠️**Important:** If the SMTP settings are configured incorrectly,
```
coopr: These don't look like SMTP settings - for SMTP I'd expect things like `host` and `port`

Also... I wonder if this warning is even necessary - would anyone expect Kong to check if email addresses are "valid" and functional? If you enter non-functional email addresses, the emails from Kong Manager still get sent, right? It is just that the recipient of the email cannot reply to them and/or cannot understand who they come from. I don't see that that needs any warning, it seems sufficiently obvious.

coopr:
I Still don’t think that mentioning an error message is warranted in this situation, but this is a small change so LGTM 👍

PR #1060, opened 2018-11-27
docs(ee/vitals) fix path to spec

darrenjennings:


coopr:
Tried the Netlify preview - works good!

PR #1068, opened 2018-12-03
docs(prometheus) disable docs on routes for Prometheus plugin

hbagdi:
This was a documentation error as the plugin makes sense only at the
service level currently.

ad31ba0888 disabled it for API entity but during the review. @hbagdi
missed to disable it for Route entity as well.


hbagdi:
@coopr Could you merge this in for me? I don't have the necessary priveliges to do so. Thanks!

PR #1031, opened 2018-11-07
feat(versions) support version numbers without trailing x

coopr:
### Summary

Support version numbers without a trailing `x` - like the upcoming `Kong 1.0` release. 

### Issues resolved

Fix #947 

### Checklist:

- [x] [Commit message & atomicity](https://github.com/Kong/docs.konghq.com/blob/master/CONTRIBUTING.md#commit-atomicity) checked
- [ ] Documentation N/A
- [ ] Spellchecked my updates N/A
- [x] Ready to be merged

PR #1069, opened 2018-12-04
Docs 1.0.x remove apis

kikito:
### Summary

Removes deprecated API object from most of the docs.

### Full changelog
Divided into several commits:

* Added some elements (navigation) to make the 1.0.0 branch visible while making changes
* Remove references to APIs from the "main" docs
* Remove references to APIs from most plugins (excluded old EE versions and non-kong plugins)
* Remove code that generated APIs for plugins automatically (this commit was left separate in case it was deemed better not to remove it).


app/_data/kong_versions.yml
```diff
@@ -133,6 +133,16 @@
     cassandra: "3.x.x"
     postgres: "9.5+"
     openresty: "1.13.6.2"
+-
+  release: "1.0.x"
```
coopr: At @thibaultcha 's suggestion, we did work to make it so we can use `1.0` for the version (or the release, or maybe both, I'm not sure) - please have a look at https://github.com/Kong/docs.konghq.com/pull/1031 and modify this PR to not use the trailing .x

thibaultcha: Much preferable indeed!

app/_hub/_init/my-extension/index.md
```diff
@@ -136,8 +136,6 @@ kong_version_compatibility: # required
 
 params: # metadata about your plugin
   name: # name of the plugin in Kong (may differ from name: above)
-  api_id:
```
coopr: We can't remove this - it is needed for plugin versions that still support APIs.

app/_data/kong_versions.yml
kikito: Done, thanks!

app/1.0/configuration.md
```diff
@@ -311,10 +311,10 @@ and Kong will be running in that instance (as configured in `nginx-kong.conf`).
 
 ## Serving both a website and your APIs from Kong
 
-A common use case for API providers is to make Kong serve both a website
+A common use case for Service providers is to make Kong serve both a website
```
coopr: This looks like an inadvertent find-and-replace - I'd suggest that this be left `API`.

app/1.0/configuration.md
```diff
@@ -311,10 +311,10 @@ and Kong will be running in that instance (as configured in `nginx-kong.conf`).
...
 and the APIs themselves over the Proxy port &mdash; `80` or `443` in
-production. For example, `https://my-api.com` (Website) and
-`https://my-api.com/api/v1` (API).
+production. For example, `https://example.com` (Website) and
+`https://example.com/api/v1` (Service).
```
coopr: Same, this `Service` should probably be `API` for clarity

app/1.0/configuration.md
```diff
@@ -351,7 +351,7 @@ http {
     # here, we declare our custom location serving our website
     # (or API portal) which we can optimize for serving static assets
     location / {
-      root /var/www/my-api.com;
+      root /var/www/example.com;
```
coopr: I like the prior example better - it is more clear to me.

app/1.0/configuration.md
```diff
@@ -1158,9 +1158,9 @@ The origins configuration can be useful in complex networking configurations, an
...
+When an encrypted scheme like `tls` or `https` in the left origin is paired with an unencrypted scheme like `tcp` or `http` in the right origin, Kong will terminate TLS on incoming connections matching the left origin, and will then route traffic unencrypted to the specified right origin. This is useful when connections will be made to the Kong node over TLS, but the local service (for which Kong is proxying traffic) doesn't or can't terminate TLS. Similarly, if the left origin is `tcp` or `http` and the right origin is `tls` or `https`, Kong will accept unencrypted incoming traffic, and will then wrap that traffic in TLS as it is routed outbound. This capability is an important enabler of Kong Mesh.
```
coopr: What changed here?

app/1.0/configuration.md
```diff
@@ -1158,9 +1158,9 @@ The origins configuration can be useful in complex networking configurations, an
...
 
-Like all Kong configuration settings, the `origins` setting *can* be declared in the `Kong.conf` file - however **it is recommended that Kong administrators avoid doing so**. Instead, `origins` should be set on a per-node basis using [environment variables](https://docs.konghq.com/{{page.kong_version}}/configuration/#environment-variables). As such, `origins` is not present in [`kong.conf.default`](https://github.com/Kong/kong/blob/0.15.0/kong.conf.default).  
+Like all Kong configuration settings, the `origins` setting *can* be declared in the `Kong.conf` file - however **it is recommended that Kong administrators avoid doing so**. Instead, `origins` should be set on a per-node basis using [environment variables](https://docs.konghq.com/{{page.kong_version}}/configuration/#environment-variables). As such, `origins` is not present in [`kong.conf.default`](https://github.com/Kong/kong/blob/0.15.0/kong.conf.default).
```
coopr: What changed here?

app/1.0/configuration.md
kikito: Nothing changed. It looks like a github fluke.

app/1.0/configuration.md
kikito: Ok, undone

app/1.0/configuration.md
kikito: Undid this one too.

app/1.0/configuration.md
kikito: I'm sorry but I cannot comply with this one. www.my-api.com is a real business and we should not be linking to it in our docs. I have been burned by making a domain up from random words and it turned out to be ... not safe for work).

`example.net` is a reserved domain name that is guaranteed to be usable in documentation and manuals, so it doesn't have that problem.

kikito:
@coopr 	I have added an extra commit doing a copy of the 0.14 docs so that the changes I made for 1.0 are now visible in isolation on their commit:

https://github.com/Kong/docs.konghq.com/pull/1069/commits/01b696e8789f89910e8c55b775daaf49b03e14bb

Regarding Plugin versioning: I didn't know plugins were versioned now. I assume I can just rename the "api" ones to an older plugin version, and then create a new `index.md` page for each plugin? Only drawback I see with that approach is that you will not be able to easily see the differences between versions, as you can now. Maybe do a first pass on the plugin changes and if everything is ok I will redo them as new versions?

kikito:
Update regarding Plugin versions: I did my best to add new versions for the plugins. However I must have done something wrong while doing so because now the website doesn't render. It does not throw any error anywhere, it just says "could not render /".

app/_data/kong_versions.yml
```diff
@@ -133,6 +133,16 @@
...
+-
+  release: "1.0"
+  version: "1.0"
+  luarocks_version: "1.0.0-0"
+  dependencies:
+    luajit: "2.1.0-beta2"
+    luarocks: "2.4.3"
+    cassandra: "3.x.x"
+    postgres: "9.5+"
+    openresty: "1.13.6.2"
```
levfishbluefish: I'm not sure why yet but it appears this block is what is preventing the index from rendering.

levfishbluefish: Still debugging but as far as I can tell, the compiler is breaking because of the 1.0 release/version number

PR #1079, opened 2018-12-18
docs(plugins) add session priority

darrenjennings:


levfishbluefish:
@darrenjennings if this is still valid can you add it to the newer versions as well?

darrenjennings:
It is not currently bundled with Kong so was waiting until enterprise 0.35 is released. @bungle mentioned that maybe core would bundle it but it would be a luarock dependency, not traditional kong core plugin to use a rock so not sure that is something that we would want. @hishamhm et al, any thoughts?

levfishbluefish:
bump

lena-larionova:
Closing, as this was fixed by someone else at some point:

https://docs.konghq.com/1.4.x/plugin-development/custom-logic/#plugins-execution-order
https://docs.konghq.com/2.1.x/plugin-development/custom-logic/#plugins-execution-order

PR #1082, opened 2018-12-20
Docs / Make the 1.0.x dropdown appear

kikito:
* Changes the kong_versions.yml file so that it is "in sync" with the folder structure we have (1.0.x instead of 1.0)
* Copies 0.14.x files on the folder so we can see the website while we make changes

Things pending:

* The PDK needs to be regenerated
* The "install" section seems to be independent from the rest of the docs. It is unclear if changing the kong_versions.yml affects it or not
* 0.14.x plugins still have references to old APIs
* admin api is for 0.14.x. We might need to update changed objects (Upstreams, targets, plugins)
* We need service mesh / streams docs
* @thibaultcha mentioned that we might want to drop the .x on the 1.0.x urls

### DIFF
--- a/app/_data/kong_versions.yml
+++ b/app/_data/kong_versions.yml
@@ -134,6 +134,16 @@
     postgres: "9.5+"
     openresty: "1.13.6.2"
 -
+  release: "1.0"
+  version: "1.0"
+  luarocks_version: "1.0.0-0"
+  dependencies:
+    luajit: "2.1.0-beta2"
+    luarocks: "2.4.3"
+    cassandra: "3.x.x"
+    postgres: "9.5+"
+    openresty: "1.13.6.2"
+-
   release: "0.31-x"
   version: "0.31"
   edition: "enterprise"
--- a/app/_data/kong_versions.yml
+++ b/app/_data/kong_versions.yml
@@ -134,12 +134,12 @@
     postgres: "9.5+"
     openresty: "1.13.6.2"
 -
-  release: "1.0"
-  version: "1.0"
+  release: "1.0.x"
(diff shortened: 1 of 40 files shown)


app/_data/kong_versions.yml
```diff
@@ -133,6 +133,16 @@
     cassandra: "3.x.x"
     postgres: "9.5+"
     openresty: "1.13.6.2"
+-
+  release: "1.0.x"
+  version: "1.0.0"
+  luarocks_version: "1.0.0-0"
+  dependencies:
+    luajit: "2.1.0-beta2"
```
bungle: This should be `2.1.0-beta3` (I think it has been long like that, we just have forgotten it), you can correct it on `0.14.x` too.

thibaultcha:
> @thibaultcha mentioned that we might want to drop the .x on the 1.0.x urls

Any reason why we did not? 1.0 was supposed not to have said `x` in URLs indeed, because it makes no sense... It is not too late to address this without additional work.

levfishbluefish:
> > @thibaultcha mentioned that we might want to drop the .x on the 1.0.x urls
> 
> Any reason why we did not? 1.0 was supposed not to have said `x` in URLs indeed, because it makes no sense... It is not too late to address this without additional work.

I added it during debugging as I didn't realize we were trying to get rid of it! Feel free to remove.

thibaultcha:
Well, the issue now is that we cannot remove it without first implementing the proper redirects since links out there have been made with the `/1.0.x/` segment. Which is a bit more work that I was able to squeeze in yesterday, so I didn't.

PR #1096, opened 2019-01-02
Heroku deployment upgraded to Kong 1.0

mars:
### Summary

Heroku deployment upgraded to Kong 1.0.

### Full changelog

* Update "Kong Heroku app" Hub article for current release.

### DIFF
--- a/app/_hub/heroku/kong-app/index.md
+++ b/app/_hub/heroku/kong-app/index.md
@@ -2,22 +2,22 @@
 name: Kong Heroku app
 publisher: Heroku community
 
-categories: 
+categories:
   - deployment
 
 type: integration
 
-desc: Deploy Kong CE clusters to Heroku Common Runtime or Private Spaces
+desc: Deploy Kong clusters to Heroku Common Runtime or Private Spaces
 
 description: |
-  Easily deploy Kong Community Edition as a Heroku app. Supports both
+  Easily deploy Kong as a Heroku app. Supports both
   Heroku Common Runtime and Private Spaces. Includes a preset secure-by-default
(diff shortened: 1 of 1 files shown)


mars:
The Netlify deploy preview does not show what is expected:

* the [preview "Heroku Kong app" article](https://deploy-preview-1096--inspiring-goldwasser-3966ac.netlify.com/hub/heroku/kong-app/) does not contain the updates in this PR
* the [preview Hub listing](https://deploy-preview-1096--inspiring-goldwasser-3966ac.netlify.com/hub/) appears broken.

It's not apparent how I could have broken it 😬

thibaultcha:
@mars 

> the preview "Heroku Kong app" article does not contain the updates in this PR

It does. Note that only the compatibility section of the sidebar isn't showing up. That is because this PR isn't properly rebased on top of master (and thus the 1.0 documentation). Rebasing this PR will make the change take effect.

> the preview Hub listing appears broken.

This might be an issue with the Netlify environment. It builds fine locally.

A side note: we refer to Kong 1.0 as "1.0", and not "1.0.x" anymore (since the `x` in this case is the patch version, in which no breaking changes should occur, unlike 0.X.Y versions). I will make sure to update this while merging.

thibaultcha:
> we refer to Kong 1.0 as "1.0", and not "1.0.x" anymore (since the x in this case is the patch version, in which no breaking changes should occur, unlike 0.X.Y versions).

Uh, it looks like this was not taken care of during the 1.0 release... An item for later then.

thibaultcha:
@mars Thank you for both updating the deployment to Kong 1.0 _and_ for this PR!

PR #1097, opened 2019-01-03
docs(ee/config) complete property reference for enterprise TD-39

h3xar0n:
### Summary

Provides a complete property reference for Kong Enterprise; most sections outside of Kong Manager and Dev Portal were missing. Any shared properties with Core have links to the relevant doc.
 
### Full changelog

* Add either a description or a link to all missing properties in Enterprise
* Replace sub-section in Audit Log with a link to the main property reference
* Fix any noticeable formatting issues

### Issues resolved

Fix TD-39 (internal)

### Checklist:
- [x] Documentation
- [x] Spellchecked my updates
- [x] Ready to be merged 
  _Needed in order to merge_:
  - [x] Although we have an entry for `admin_invite_email`, it's nowhere in `kong.conf` and may need to be removed.
  - [x] Write all of the properties instead of linking to Kong reference. This will avoid confusion and improve the reader experience.
  - [x] Remove the ToC for now and create a manual ToC with general sections; this will all be replaced with a left-nav

### DIFF
--- a/app/enterprise/0.34-x/property-reference.md
+++ b/app/enterprise/0.34-x/property-reference.md
@@ -2,21 +2,55 @@
 title: Configuration Property Reference for Kong Enterprise
 ---
 
-## Kong Manager
+## General
 
 
-### admin_api_uri
+### prefix
 
-**Default:** `NONE` (auto generated)
+See the property description in Kong's configuration reference for 
+[prefix](https://docs.konghq.com/0.13.x/configuration/#prefix)
 
-**Description:**  
 
-Hierarchical part of a URI which is composed 
-optionally of a host, port, and path at which your 
-Admin interface API accepts HTTP or HTTPS traffic. 
-When this config is disabled, the gui will use the 
-window protocol + host and append the resolved 
-admin_gui_listen HTTP/HTTPS port.
+### log_level
+
+See the property description in Kong's configuration reference for 
+[log_level](https://docs.konghq.com/0.13.x/configuration/#log_level)
+
+
+### proxy_access_log
+
+See the property description in Kong's configuration reference for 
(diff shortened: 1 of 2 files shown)


app/enterprise/0.34-x/property-reference.md
```diff
@@ -2,27 +2,63 @@
 title: Configuration Property Reference for Kong Enterprise
 ---
 
-## Kong Manager
+## General
```
levfishbluefish: Could we include something here to explain why the first section is just links to another page? It does seem odd to just say "see Kong's configuration reference" as users might be like, "I thought I was!' IDK something like  - 'properties in this section are in both Kong and Kong Enterprise. Please see the Kong configuration file for further information"

levfishbluefish:
There are a LOT of duplicate properties, I am wondering if it would be easier to treat Kong and Kong Enterprise docs as totally separate docs (aka Enterprise customers should never have to view Kong and vice versa) and just include all the property info here as well so Enterprise users don't have to use two different docs

h3xar0n:
> There are a LOT of duplicate properties, I am wondering if it would be easier to treat Kong and Kong Enterprise docs as totally separate docs (aka Enterprise customers should never have to view Kong and vice versa) and just include all the property info here as well so Enterprise users don't have to use two different docs

That is definitely a better reader experience. To maintain it, we'll just need to be sure to diff the kong.conf every single version.

h3xar0n:
> Found an existing typo - line 822 - 'port_gui_use_subdomains' should be 'portal_gui_use_subdomains'

Since it may take longer to merge this one, I've made a separate PR here:
https://github.com/Kong/docs.konghq.com/pull/1099

PR #1100, opened 2019-01-04
Docs CLI migrations + upgrade path

kikito:
* Updates `kong migrations` CLI options
* Adds a *upgrading.md* section, with just the `1.0` guide, pointing to the full in-repo guide for older versions

### DIFF
--- a/app/1.0.x/upgrading.md
+++ b/app/1.0.x/upgrading.md
@@ -1,0 +1,289 @@
+---
+title: Upgrade guide
+---
+
+<div class="alert alert-warning">
+  <strong>Note:</strong> What follows is the upgrade guide for 1.0.x.
+  If you are trying to upgrade to an earlier version of Kong, please read
+  <a href="https://github.com/Kong/kong/blob/master/UPGRADE.md">UPGRADE.md file in the Kong repo</a>
+</div>
+
+This guide will inform you about breaking changes you should be aware of
+when upgrading, as well as take you through the correct sequence of steps
+in order to obtain a **no-downtime** migration in different upgrade
+scenarios.
+
+## Upgrade to `1.0.0`
+
+This is a major release of Kong, and includes a number of new features
+as well as breaking changes.
+
+This version introduces **a new schema format for plugins**, **changes in
+Admin API endpoints**, **database migrations**, **Nginx configuration
+changes**, and **removed configuration properties**.
+
+In this release, the **API entity is removed**, along with its related
+Admin API endpoints.
+
+This section will highlight breaking changes that you need to be aware of
+before upgrading and will describe the recommended upgrade path. We recommend
+that you consult the full [1.0.0
+Changelog](https://github.com/Kong/kong/blob/master/CHANGELOG.md) for a
+complete list of changes and new features.
+
+## 1. Breaking Changes
+
+### Dependencies
+
+- The required OpenResty version is 1.13.6.2, but for a full feature set,
+  including stream routing and service mesh abilities with mutual TLS, you need
+  Kong's [openresty-patches](https://github.com/kong/openresty-patches).
+- The minimum required OpenSSL version is 1.1.1. If you are building by hand,
+  make sure all dependencies, including LuaRocks modules, are compiled using
+  the same OpenSSL version. If you are installing Kong from one of our
+  distribution packages, you are not affected by this change.
+
+### Configuration
+
+- The `custom_plugins` directive is removed (deprecated since 0.14.0).
+  Use `plugins` instead, which you can use not only to enable
+  custom plugins, but also to disable bundled plugins.
+- The default value for `cassandra_lb_policy` changed from `RoundRobin`
+  to `RequestRoundRobin`.
+- Kong generates a new template file for stream routing,
+  `nginx-kong-stream.conf`, included in the `stream` block
+  of its top-level Nginx configuration file. If you use
+  a custom Nginx configuration and wish to use stream
+  routing, you can generate this file using `kong prepare`.
+- The Nginx configuration file has changed, which means that you need to update
+  it if you are using a custom template. The changes are detailed in the following diff:
+
+``` diff
+diff --git a/kong/templates/nginx_kong.lua b/kong/templates/nginx_kong.lua
+index d4e416bc..8f268ffd 100644
+--- a/kong/templates/nginx_kong.lua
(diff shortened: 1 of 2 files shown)


app/1.0.x/upgrading.md
```diff
@@ -0,0 +1,279 @@
...
+
+## Introduction
+
+<div class="alert alert-warning">
+  <strong>Note:</strong> What follows is the upgrade guide for 1.0.x.
+  If you are trying to upgrade to an earlier version of Kong, please read
+  <a href="https://github.com/Kong/kong/blob/master/UPGRADE.md">UPGRADE.md file in the Kong repo</a>
+</div>
+
+Unless indicated otherwise in one of the upgrade paths of this document, it is
```
hishamhm: > one of the upgrade paths of this document

This paragraph only makes sense in UPGRADE.md.

This section is not really an Introduction, it is really about upgrading patch releases (with no migrations). I would take everything from "Assuming that..." up to "...in-flight connections") and move it to a "Upgrading patch releases" at the bottom (with a new introductory paragraph explaining what patch releases are, i.e. "1.0.x to 1.0.y")

app/1.0.x/upgrading.md
```diff
@@ -0,0 +1,279 @@
+---
+title: Upgrading guide
```
hishamhm: The other page calls it "Upgrade guide"; this one says "Upgrading guide". I think "Upgrade guide" reads better.

app/1.0.x/cli.md
```diff
@@ -84,12 +84,37 @@ Usage: kong migrations COMMAND [OPTIONS]
...
+
+  --db-timeout     (default 60)      Timeout, in seconds, for all database
+                                     operations (including schema consensus for
+                                     Cassandra).
+
+  --lock-timeout   (default 60)      Timeout, in seconds, for nodes waiting on
+                                     the leader node to finish running
+                                     migrations.
+
+  -c,--conf        (optional string) Configuration file.
```
hishamhm: We really need to write a trivial script to generate this page out of the `--help` output of the various commands!

hishamhm: Aaaand I just did it :)
https://github.com/Kong/docs.konghq.com/pull/1102

Once that PR is merged, this is likely to cause a merge conflict. Better just remove the cli.md changes from this commit and keep the autogenerated one from there.

app/1.0.x/upgrading.md
kikito: I decided to just remove this part. We cal always add it later when we do a point release. And it is present in the `UPGRADING.md` file anyway.

app/1.0.x/cli.md
kikito: Nicely done! I have reviewed and approved it, and removed the CLI-related commit.

app/1.0.x/upgrading.md
hishamhm: > We cal always add it later when we do a point release

I bet we _will_ forget to do it when we do a point release ;) At the very least please add a sentence about 1.0.x point releases pointing to `UPGRADE.md`, in order to future-proof this page.

kikito: Ok. I ended up adding the whole thing in a section. It was faster to do than coming with a new section and then pointing to `UPGRADING.md`

app/1.0.x/upgrading.md
```diff
@@ -0,0 +1,285 @@
...
+   nodes pointing to the same datastore anymore. Only run
+   this command when you are confident that your migration
+   was successful. From now on, you can safely make Admin API
+   requests to your 1.0 nodes.
+
+### Upgrade Path from 1.0 Release Candidates
+
+The process is the same as for upgrading for 0.14 listed above, but on step 1 you should run `kong migrations up --force` instead.
+
+### Upgrade Path for patch releases
```
hishamhm: inconsistent capitalization

app/1.0.x/upgrading.md
```diff
@@ -0,0 +1,285 @@
...
+
+There are currently no point releases of kong 1.0.0 (there is no 1.0.1, 1.0.2, etc.) Once those are available,
+the update process will be the following:
```
hishamhm: Kong is uppercase. Also, this sentence is bound to be incorrect once 1.0.1 is out; better to write it in a future-proof way:

```
There are no migrations in upgrades between current or future patch releases of the same minor release of Kong (e.g. 1.0.0 to 1.0.1, 1.0.1 to 1.0.4, etc.). Therefore, the upgrade process is simpler.
```

app/1.0.x/upgrading.md
hishamhm: I think this sentence is still misplaced: "Unless indicated otherwise in one of the upgrade paths of this document", because no sections in this document indicate so. It is very much an artifact of `UPGRADE.md`.

hbagdi:
With new migrations, `kong migrations list` exits with different exit codes based on the state of the database, which is very helpful to figure out the right time to start/stop different versions of Kong during an upgrade rollout.

Can we document these in this PR?

hishamhm:
@hbagdi done!

app/1.0.x/upgrading.md
```diff
@@ -0,0 +1,307 @@
...
+are new migrations available. The status code of the process will
+also change accordingly:
+
+* `0` - migrations are up-to-date
+* `1` - failed inspecting the state of migrations (e.g. database is down)
+* `3` - database needs bootstrapping:
+  you should run `kong migrations bootstrap` to install on
+  a fresh datastore.
+* `4` - there are pending migrations: once your old cluster is
+  decomissioned you should run `kong migrations finish` (step 5 above).
```
hbagdi: Can clarify what is the state of the cluster when the exit code is 4?
As per my understanding, this exit code means that the newer version can now be safely started, the old version can safely run as well but there is another step (finish) that needs to be execute to finish the upgrade process.

We refer to the migrations in this state as pending I find the term a little ambiguous, hence this comment (I don't have a better term myself to suggest).

subnetmarco:
@kikito updates?

app/1.0.x/upgrading.md
hishamhm: > As per my understanding, this exit code means that the newer version can now be safely started, the old version can safely run as well but there is another step (finish) that needs to be execute to finish the upgrade process.

Yes, your understanding is correct.  Amended based on your description.

PR #1104, opened 2019-01-08
Docs/streams and sm odds and ends

kikito:
Includes some additional mentions of Stream routes into the 1.0 docs, as well as some typos.

### DIFF
--- a/app/1.0.x/health-checks-circuit-breakers.md
+++ b/app/1.0.x/health-checks-circuit-breakers.md
@@ -138,8 +138,8 @@ is set to zero, the checking is disabled at the corresponding scenario.
 When both are zero, active health checks are disabled altogether.
 
 <div class="alert alert-warning">
-<strong>Note:</strong> Active Health checks currently only support HTTP/HTTPS targets. They
-will self-deactivate when applied to TCP streams.
+<strong>Note:</strong> Active health checks currently only support HTTP/HTTPS targets. They
+do not apply to Upstreams assigned to Services with the protocol attribute set to "tcp" or "tls".
 </div>
 
 [Back to TOC](#table-of-contents)
--- a/app/1.0.x/health-checks-circuit-breakers.md
+++ b/app/1.0.x/health-checks-circuit-breakers.md
@@ -137,12 +137,17 @@ when a target is healthy or unhealthy. If the `interval` value for either
 is set to zero, the checking is disabled at the corresponding scenario.
 When both are zero, active health checks are disabled altogether.
 
+<div class="alert alert-warning">
+<strong>Note:</strong> Active Health checks currently only support HTTP/HTTPS targets. They
+will self-deactivate when applied to TCP streams.
+</div>
+
 [Back to TOC](#table-of-contents)
 
 ### Passive health checks (circuit breakers)
 
 Passive health checks, also known as circuit breakers, are
-checks performed based on the requests being proxied by Kong,
+checks performed based on the requests being proxied by Kong (HTTP/HTTPS/TCP),
 with no additional traffic being generated. When a target becomes
 unresponsive, the passive health checker will detect that and mark
 the target as unhealthy. The ring-balancer will start skipping this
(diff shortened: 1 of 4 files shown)


app/1.0.x/health-checks-circuit-breakers.md
```diff
@@ -137,12 +137,17 @@ when a target is healthy or unhealthy. If the `interval` value for either
 is set to zero, the checking is disabled at the corresponding scenario.
 When both are zero, active health checks are disabled altogether.
 
+<div class="alert alert-warning">
+<strong>Note:</strong> Active Health checks currently only support HTTP/HTTPS targets. They
+will self-deactivate when applied to TCP streams.
```
hishamhm: "TCP streams" here is ambiguous: HTTP requests after all also go through TCP.
This sentence "will self-deactivate" is very derivative of the actual implementation, and might be a bit misleading wrt the user-visible behavior.

Suggested:

> <strong>Note:</strong> Active health checks currently only support HTTP/HTTPS targets. They do not apply to Upstreams assigned to Services with the `protocol` attribute set to `"tcp"` or `"tls"`.

(capitalization to indicate Kong entities.)

app/1.0.x/proxy.md
```diff
@@ -18,8 +18,8 @@ properties:
   Kong's configuration capabilities: the **Admin API** (`8001` by default).
 
 <div class="alert alert-warning">
-<p><strong>Note:</strong> Starting with 0.13.0, the API entity was
-deprecated. This document will cover proxying with the new Routes and
+<p><strong>Note:</strong> Starting with 1.0.0, the API entity has been removed.
```
hishamhm: :+1:

app/1.0.x/proxy.md
```diff
@@ -925,14 +925,13 @@ the `cert.pem` certificate previously configured.
...
 [Back to TOC](#table-of-contents)
 
-### Restricting the client protocol (HTTP/HTTPS)
+### Restricting the client protocol (HTTP/HTTPS/TCP/TLS)
 
 Routes have a `protocols` property to restrict the client protocol they should
-listen for. This attribute accepts a set of values, which can be `http` or
-`https`.
+listen for. This attribute accepts a set of values, which can be `http`,
+`https`, `tcp` & `tls`.
```
hishamhm: `&` => `or`

Formal English writing avoids the use of `&` (I noticed you do this quite a bit: is the rule different is Spanish? :) )

app/1.0.x/proxy.md
```diff
@@ -946,22 +945,10 @@ the client is connecting with (plain-text or over TLS):
...
-    "methods": ["..."],
-    "protocols": ["http"],
-    "service": {
-        "id": "..."
-    }
-}
-```
-
-However, a Route with only `https` would _only_ accept traffic over TLS. It
+However, a Route with *only* `https` would _only_ accept traffic over https. It
```
hishamhm: https => HTTPS

app/1.0.x/proxy.md
```diff
@@ -994,6 +981,54 @@ Server: kong/x.y.z
 {"message":"Please use HTTPS protocol"}
 ```
 
+Since Kong 1.0 it's possible to create routes for raw TCP (not necessarily http)
```
hishamhm: http => HTTP

app/1.0.x/proxy.md
```diff
@@ -994,6 +981,54 @@ Server: kong/x.y.z
...
+connections by using `"tcp"` on the `protocols`:
```
hishamhm: on the `protocols` => in the `protocols` attribute:

app/1.0.x/proxy.md
```diff
@@ -994,6 +981,54 @@ Server: kong/x.y.z
...
+    "paths": ["..."],
+    "methods": ["..."],
+    "protocols": ["tcp"],
+    "service": {
+        "id": "..."
+    }
+}
+```
+
+Similarly, we can create routes which accept raw TLS traffic (not necessarily https) with
```
hishamhm: https => HTTPS

app/1.0.x/proxy.md
kikito: I honestly ignored that rule. I thought they were interchangeable. Thanks for teaching me something today!

PR #1138, opened 2019-01-31
style(*) shrink header spacing after small navbar

allejo:
**Before**

image

image

**After**

image

image

### Summary

The amount of spacing that currently exists when you click on headers pushes the heading a bit too far down because the old navbar was taller.

### Full changelog

* Pixel height in styles changed

### Issues resolved

None that I'm aware of

### Checklist:

- [ ] Documentation 
- [x] Spellchecked my updates
- [x] Ready to be merged

### DIFF
--- a/app/_assets/javascripts/app.js
+++ b/app/_assets/javascripts/app.js
@@ -3,6 +3,8 @@
 'use strict'
 
 $(function () {
+  var NAV_HEIGHT = 56
+
   var $window = $(window)
   var $docs = $('#documentation')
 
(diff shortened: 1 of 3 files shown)


app/_assets/javascripts/app.js
```diff
@@ -3,6 +3,8 @@
 'use strict'
 
 $(function () {
+  var NAV_HEIGHT = 56;
```
allejo: I'm not sure why there were so many different values for different scrollTop but I standardized it to 56px (the current height). If this is incorrect, hopefully, some light could be shed on why the other, seemingly, arbitrary values were picked?

PR #1177, opened 2019-03-14
feat(Sidebar) Collapse sidebar on small screens

travispamaral:
### Summary
Collapses sidebar on small devices instead of showing it above the main content. Sorry for giant screenshots!

_Closed_

_Open_

### Full changelog

* Adds class to sidebar
* Adds small JS to toggle open class

### Checklist:

- [ x ] [Commit message & atomicity](https://github.com/Kong/docs.konghq.com/blob/master/CONTRIBUTING.md#commit-atomicity) checked
- [ x ] Ready to be merged

### DIFF
--- a/app/_assets/javascripts/app.js
+++ b/app/_assets/javascripts/app.js
@@ -296,6 +296,15 @@ $(function () {
     }
 
     $navItems.on('click', setNavItemActive)
+
+    if ($(window).width() <= 800) {
+      $('.sidebar-toggle').click(function () {
+        $('.page-navigation').addClass('sidebar-open')
+      })
+      $('.page-navigation > .close-sidebar').click(function () {
+        $('.page-navigation').removeClass('sidebar-open')
+      })
+    }
   }
 
   // Analytics
--- a/app/_assets/stylesheets/header.less
+++ b/app/_assets/stylesheets/header.less
@@ -1034,7 +1034,7 @@ body {
       padding: 75px 0;
       color: #062f4d;
       overflow-y: auto;
-      z-index: 100;
+      z-index: 200;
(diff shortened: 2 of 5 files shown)


h3xar0n:
This is seriously awesome, @travispamaral . A much needed improvement over the current small nav. Points for all to consider:

Do you like how the *X* is underneath the hamburger? Purely subjective, but I think they would look better aligned.

Screen Shot 2019-03-14 at 4 32 25 PM


Testing on an iPhone X in Safari:

- The "Open Sidebar" text is too far right

IMG_1486
 
- The nav bar doesn't scroll but the rest of the page behind it does (this only happens occasionally so be sure to try several times from several positions in order to replicate).

IMG_1487

travispamaral:
@h3xar0n I can push it over to align with hamburger 👍 

Can you double check that when the menu is too far to the right / scroll prob - is there overflow somewhere below in the content?

h3xar0n:
I don't see any overflow below in the content in either case.

adamjkuhn:
this is such a great solution. Open Sidebar trigger doesn't seem to work in desktop browsers though (unless this is an issue with the Netlify preview?) and could use cursor:pointer - also not sure if you guys could use it or not but in addition to the device lab I built (and a handful of virtual machines) I have a Browserstack account that's quite helpful for this sort of thing.

travispamaral:
I wasn't able to reproduce the sidebar no opening. I did update the close icon as that was hard to click. I wasn't sure if cursor:pointer was necessary as this is really only focused on mobile. I suppose adding isn't the end of the world.

We actually have a browser stack account. I'll test once the preview updates

PR #1200, opened 2019-03-26
docs(1.0.x) document how to use the new migrations

kikito:
This change includes updates on both the Upgrade Guide and the Plugin
Development Guide for Kong 1.0.x

### DIFF
--- a/app/1.0.x/plugin-development/custom-entities.md
+++ b/app/1.0.x/plugin-development/custom-entities.md
@@ -6,9 +6,10 @@ chapter: 6
 
 ## Introduction
 
-Your plugin might need to store more than its configuration in the database. In
-that case, Kong provides you with an abstraction on top of its primary
-datastores which allows you to store custom entities.
+While not all plugins need it, your plugin might need to store more than
+its configuration in the database. In that case, Kong provides you with
+an abstraction on top of its primary datastores which allows you to store
+custom entities.
 
 As explained in the [previous chapter]({{page.book.previous}}), Kong interacts
 with the model layer through classes we refer to as "DAOs", and available on a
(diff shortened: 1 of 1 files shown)


app/1.0.x/plugin-development/custom-entities.md
```diff
@@ -32,88 +36,99 @@ returns them.
...
 migrations must be written.
 
-Each migration must bear a unique name, and `up` and `down` fields. Such fields
-can either be strings of SQL/CQL queries for simple migrations, or actual Lua
-code to execute for complex ones. The `up` field will be executed when Kong
-migrates **forward**. It must bring your database's schema to the latest state
-required by your plugin. The `down` field must execute the necessary actions to
-revert your schema to its previous state, before `up` was ran.
+If your plugin doesn't have it already, you should add a `<plugin_name>/migrations`
+folder to it. If there is no `init.lua` file inside. This is where all the migrations
```
hishamhm: `If there is no `init.lua` file inside. `
-> 
`If there is no `init.lua` file inside, you should create one.` (needs to explain contents?)

app/1.0.x/plugin-development/custom-entities.md
```diff
@@ -32,88 +36,99 @@ returns them.
...
 
-```lua
--- cassandra.lua
+## Adding a new migration to an existing plugin
+
+Sometimes it is necessary to introduce changes after a version of a plugin has already been
+released. A new functionality might be needed. A database table row might need changing.
+
+When this happens, it is strongly recommended that you create a new migrations file instead
+of modifying the existing ones.
```
hishamhm: not critical, but I'm concerned this indirectly suggests the wrong way as "a way" by making right way a "recommendation".

```
When this happens, it is strongly recommended that you create a new migrations file instead
of modifying the existing ones.
```
-> 
```
When this happens, you must create a new migrations file that performs this changes, so that users of your plugin can convert their existing databases to the new format. Migrations of past versions must never be modified.
```

app/1.0.x/plugin-development/custom-entities.md
```diff
@@ -32,88 +36,99 @@ returns them.
...
+  "000_base",
+  "001_add_foo",
 }
 ```
 
-```lua
--- postgres.lua
+## Migration File syntax
+
+While Kong's core migrations support both PostgreSQL and Cassadra, custom plugins
```
hishamhm: Cassadra -> Cassandra

app/1.0.x/upgrading.md
```diff
@@ -13,6 +13,41 @@ when upgrading, as well as take you through the correct sequence of steps
 in order to obtain a **no-downtime** migration in different upgrade
 scenarios.
 
+## Zero-downtime migrations
```
hishamhm: This section already exists below in the same document: `Migration Steps from 0.14`. Please remove this addition from this PR

PR #1210, opened 2019-03-27
feat(hub) add array examples to template

### LINKED ISSUE #1197: Update CORS plugin documentation with new Admin API notations
See: https://docs.konghq.com/hub/kong-inc/cors/

### Summary

In Kong 1.0, the old comma-separated array notation (and JSON dot-notation) were removed (after being deprecated for several versions) in favor of better standards for both MIME types.

The new array notation is only documented in the Admin API reference, see [Add Route](https://docs.konghq.com/1.0.x/admin-api/#add-route), in the `hosts` or `paths` properties for example.

However, the CORS plugin still uses the comma notation, which will not pass schema validation and produce an error, preventing a user from adding a plugin by simply following our examples.

### Additional Details

- The comma notation is still valid for Kong < 1.0 and Kong EE. This revigorates the hot topic of versioning our plugins' documentation
- Ideally, a tab shows the same request with various snippets (i.e. with httpsnippet?) and Content-Type, since the dot notation is accepted for `form-encoded`, but not for `application/json` which must specify nested JSON Objects :warning: 
- :warning: Other plugins may also be susceptible to the same flaws and lagging behind
- We should also consider documenting the new notations under the Admin API's [Supported Content Types](https://docs.konghq.com/1.0.x/admin-api/#supported-content-types) section.

rainest:
### Summary

Add support for array examples in Hub and update CORS and JWT plugins to use new syntax.

### Full changelog
* Add a new `array_value_in_examples` field to plugin template. This takes an array of values that are then rendered as separate array data parameters in the cURL example, e.g. `--data foo[]=one --data foo[]=two`.
* Update CORS plugin documentation to use `array_value_in_examples` for its `methods` configuration item.
* Update legacy (not templated) cURL example for JWT plugin to use new syntax.
* Update CONTRIBUTING.md with instructions for new field.

### Issues resolved

Fix #1197

### Checklist:

- [x] Documentation 
- [x] [Adheres to Kong style guide](https://github.com/Kong/docs.konghq.com/blob/master/STYLEGUIDE.md)
- [x] Spellchecked my updates
- [x] Tagged "Team Docs" as reviewers << failing to autocomplete for some reason; tagged individuals instead
- [ ] Ready to be merged 

Warrants discussion before merger, as that these changes break compatibility with pre-1.0.x releases of Kong (and releases of Kong Enterprise with pre-1.0.x cores). The old syntax (comma-separated strings) is still correct for older versions, but is not covered in examples. How should we handle breaking API changes for plugins?

On a more minor note, did not add a parallel feature for defaults, which are a bit tricky as we use them elsewhere as plain text. It's likely possible to add something that detects if defaults is an array and formats it appropriately when non-example representations are needed, but for now it's just necessary to provide an example array if your plugin needs array config.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -202,7 +202,7 @@ the existing plugins for examples, and see additional advice in
     the value to use. A required field with no `value_in_examples` entry
     will resort to the one in `default`.
     * `array_value_in_examples` - an array of values that should appear in
-    examples. This should always be provided if the field is is required,
+    examples. This should always be provided if the field is required,
     as the template does not translate default values into arrays.
     * `description` - description of the field.
     Use YAML's pipe notation if writing longer Markdown text.
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -201,6 +201,9 @@ the existing plugins for examples, and see additional advice in
     * `value_in_examples` - if the field is to appear in examples, this is
     the value to use. A required field with no `value_in_examples` entry
     will resort to the one in `default`.
+    * `array_value_in_examples` - an array of values that should appear in
+    examples. This should always be provided if the field is is required,
+    as the template does not translate default values into arrays.
     * `description` - description of the field.
     Use YAML's pipe notation if writing longer Markdown text.
 
--- a/app/_hub/kong-inc/cors/index.md
+++ b/app/_hub/kong-inc/cors/index.md
@@ -58,7 +58,9 @@ params:
     - name: methods
       required: false
       default: "`GET, HEAD, PUT, PATCH, POST`"
-      value_in_examples: GET, POST
+      array_value_in_examples:
+        - GET
+        - POST
       description:
         Value for the `Access-Control-Allow-Methods` header, expects a comma delimited string (e.g. `GET,POST`).
     - name: headers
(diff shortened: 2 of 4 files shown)


CONTRIBUTING.md
```diff
@@ -201,6 +201,9 @@ the existing plugins for examples, and see additional advice in
     * `value_in_examples` - if the field is to appear in examples, this is
     the value to use. A required field with no `value_in_examples` entry
     will resort to the one in `default`.
+    * `array_value_in_examples` - an array of values that should appear in
+    examples. This should always be provided if the field is is required,
```
h3xar0n:
```suggestion
    examples. This should always be provided if the field is required,
```

thibaultcha:
Let's not merge this without having a solution for conversioning the examples depending on 0.14 vs 1.0 notation. Merging this will not only break documentation for all <= 0.14 users, but also for all Enterprise and free trial versions currently available.

In the issue I opened, I suggested a tab to separate JSON and form-urlencoded examples, but we should prioritize a tab for `Kong <= 0.14` and `Kong >= 1.0 || Kong EE` instead. Suggesting a tab because while versioning whole plugin documentation should still be something we'd do (due to new features being added to plugins in-between Kong releases), it'd be significantly more effort than support a tab for the array notation at the moment.

Please feel free to reach out if you want to sort out any details. @bungle is also supposed to work on this documentation issue this week as part of his sprint (and btw, thank you @rainest for taking the initiative on this! :+1: ) and can help too.

bungle:
I didn't see that this was already there when I implemented:
https://github.com/Kong/docs.konghq.com/pull/1225

tlblessing:
Please see https://konghq.atlassian.net/browse/DOCS-1042. This PR is still on hold.

falondarville:
This has been addressed via: https://github.com/Kong/docs.konghq.com/pull/2238

PR #1220, opened 2019-04-08
docs(styleguide) update numbering rule and all rule formatting

h3xar0n:
**NOTE**: Please read the CONTRIBUTING.md guidelines before submitting your patch,
and ensure you followed them all:

https://github.com/Kong/docs.konghq.com/blob/master/CONTRIBUTING.md

### Summary

Changes the rule for numbering to reflect what is more consistent and easy to read. After examining many different contexts in our docs and exchanging feedback with a copyeditor, using digits instead of written-out numbers seemed like the better course.

There were also some lingering inconsistencies, both in rules and in examples, that are updated.

### Full changelog

* Alters **numbering** rules to use digits instead of written-out words.
* Fixes punctuation, formatting, and sentence structure for many rules.
* Enforces consistency with the latest rules, especially for entity names in bold text.

### Issues resolved

Fix #XXX

### Checklist:

- [x] Documentation 
- [x] [Adheres to Kong style guide](https://github.com/Kong/docs.konghq.com/blob/master/STYLEGUIDE.md)
- [x] Spellchecked my updates
- [x] Tagged "Team Docs" as reviewers
- [x] Ready to be merged

### DIFF
--- a/STYLEGUIDE.md
+++ b/STYLEGUIDE.md
@@ -227,7 +227,7 @@ Use proper case.
 Joined as a noun/modifier, separated as a phrasal verb; added prepositions are 
 also separated.
 
-* To see the dashboard, *log in* as an admin. 
+* To see the dashboard, *log in* as an **admin**. 
 * *Log in to* the app.
 * Reset your password at the *login* page.
 
@@ -276,20 +276,21 @@ Use proper case.
 Use lowercase, hyphen as a compound modifier. Note that an [Admin](#Admin) is a 
 Kong entity.
 
-* Invite a *super admin*.
-* An Admin account may invite others if it has *super-admin* permissions.
+* Invite a **_super admin_**.
+* An **Admin** account may invite others if it has **_super-admin_** 
+    **Permissions**.
 
 ### text in buttons, links
 Maintain the case, surround with double quotes
 
-* Click “Admins” in the sidebar.
+* Click **Admins** in the sidebar.
 
-* Use the “Add Role” button.
+* Use the **Add Role** button.
 
 ### titles
 Prepositions and articles are lower case, everything else is proper case.
 
-* Create RBAC Roles for an Admin in the New Workspace
+* Create RBAC **Roles** for an **Admin** in the new **Workspace**.
 
 ### utilize
 Has a specific meaning, “to use what is available”; it should not be used a 
--- a/STYLEGUIDE.md
+++ b/STYLEGUIDE.md
@@ -10,7 +10,7 @@ Use proper case for the Kong entity, lowercase for the RBAC Role.
 Use proper case.
 
 ### Admin GUI
-Use *Kong Manager* instead
+Use *Kong Manager* instead.
 
 ### American English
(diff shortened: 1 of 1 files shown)


levfishbluefish:
Hmm I have mixed feelings about the numbering, but because of the nature of our docs I agree that it could get too pedantic to dictate which contexts get digits vs words so I'll agree to just using digits.

STYLEGUIDE.md
```diff
@@ -1,30 +1,31 @@
...
+All upper case; if referring to an entity, use **Services** and **Routes** 
+instead.
 
 ### Basic Auth, Basic Authentication
-Use proper case, abbreviate if the context permits. Only use lowercase with a 
-hyphen in code/config samples
+Use proper case; abbreviate if the context permits. Only use lowercase (with a 
+hyphen) in code or configuration samples.
 
 * Bad: Options include LDAP Auth and basic auth.
```
levfishbluefish: Make this italics

levfishbluefish:
wondering if our use of italics and bolds in the examples follows our own style guide rules🤔

STYLEGUIDE.md
```diff
@@ -188,106 +200,112 @@ For multiple lines of code or an example command, use a
...
-[bold text](#bold-text)
+[bold text](#bold-text).
 
 ### Kong
-Use proper case
+Use proper case.
 
 ### Kong Admin
-Use Admin instead
+Use [**Admin** or **admin**](#admin-admin) instead.
```
levfishbluefish: This link was confusing, would it be better to say "Use **Admin** or **admin** (See #admin-admin)

h3xar0n: How's this?

Use **Admin** or **admin** (see [Admin, admin](#admin-admin)) instead.

```suggestion
Use **Admin** or **admin** (see [Admin, admin](#admin-admin)) instead.
```

STYLEGUIDE.md
h3xar0n: (Not sure what "this" refers to--just _configuration_?)

h3xar0n:
> wondering if our use of italics and bolds in the examples follows our own style guide rules🤔

The bold text is used correctly to refer to entities, specific names, etc. Any specific case in mind that isn't accounted for in the bulleted list?

I think our ad hoc use of italics is for _emphasis_, which is pretty normal. Do you think it would make sense to create a rule for italics?

h3xar0n:
> Hmm I have mixed feelings about the numbering, but because of the nature of our docs I agree that it could get too pedantic to dictate which contexts get digits vs words so I'll agree to just using digits.

I've seen other guides come to this concession. If we go the other route and insist on writing things out, we will definitely be correcting every PR. For technical material or situations with lots of numbering, digits will probably cause the least friction.

STYLEGUIDE.md
levfishbluefish: oh sorry, bad github linking - the "Bad" in this example is the only "Bad" not in italics

PR #1228, opened 2019-04-15
docs(style-guide) "new" is not a preposition

hishamhm:


hishamhm:
@h3xar0n I thought the use of bold there was to call attention to what was being exemplified. I just re-read the `bold text` section but (fortunately!) it doesn't say that all entities should be bold (otherwise there would be a _lot_ of bold text everywhere, defeating its purpose).

PR #1241, opened 2019-04-25
chore(upgrade) result of running yarn upgrade --latest

hutchic:
Ran `yarn upgrade --latest` to upgrade all out of date libraries and then backported a few until things worked again


levfishbluefish:
Started a slack thread but posting here for visibility:

- Are we able to update to yarn to 1.16.0?

- What in the repo isn't gulp 4.0 compatible?

Will approve this PR if the above questions are out of scope

hutchic:
> * Are we able to update to yarn to 1.16.0?

Where is the yarn version pinned I can't seem to locate it 🤔 

> * What in the repo isn't gulp 4.0 compatible?

The way our gulpfile is written needs to be migrated to 4.0

levfishbluefish:
@hutchic when running locally it came up as a warning

hutchic:
I'm inclined to ignore that yarn warning. I'm a little uncomfortable with our use of npm, yarn, gem, kitchensink all at the same time but also feel I'm not fluent enough in that particular ecosystem to make choices

PR #1269, opened 2019-05-23
docs(ldap-auth-advanced) `consumer_by` takes an array and the default…

dndx:
… value was

incorrect


h3xar0n:
LGTM, @dndx --Just need to see what's going on with the `.github/main.workflow` error.

PR #1272, opened 2019-05-24
Fix Dev Portal navigation 404s

levfishbluefish:
Updates url for managing developers and adding new registration field docs

### DIFF
--- a/app/_data/docs_nav_ee_0.35-x.yml
+++ b/app/_data/docs_nav_ee_0.35-x.yml
@@ -137,6 +137,7 @@
             - text: OpenID Connect
               url: /developer-portal/configuration/authentication/oidc
             - text: Adding Custom Registration Fields
+              url: /developer-portal/configuration/authentication/adding-registration-fields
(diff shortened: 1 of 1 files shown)


app/_data/docs_nav_ee_0.35-x.yml
```diff
@@ -137,6 +137,7 @@
             - text: OpenID Connect
               url: /developer-portal/configuration/authentication/oidc
             - text: Adding Custom Registration Fields
+              url: /developer-portal/configuration/authentication/adding-registration-fields
```
h3xar0n: the staging site is probably not updated, but `Adding Custom Registration Fields` is still taking me to the Enterprise landing page. Going to test locally.

levfishbluefish: Yes it hasn't updated to the second commit

h3xar0n: Works locally, going to approve.

PR #1288, opened 2019-05-30
docs(sg) add rules for active/passive voice

h3xar0n:


nijikokun:
Thoughts on using [multiple examples](https://plainlanguage.gov/guidelines/conversational/use-active-voice/)?

STYLEGUIDE.md
```diff
@@ -240,6 +247,8 @@ instead of "to" or "through".
 * *Bad*: This guide assumes steps one through three from the previous guide.
 * **_Good_: This guides assumes steps 1–3 from the previous guide.**
 
+### passive voice
+See [active voice](#active-voice)
```
h3xar0n:
```suggestion
See [active voice](#active-voice).
```

h3xar0n:
@nijikokun the more, the merrier. We should have one that does not include the subject since that example slips by more often.

STYLEGUIDE.md
```diff
@@ -1,5 +1,12 @@
 # Documentation Style Guide
 
+### active voice
+Use the active voice. See [personal pronouns](#personal-pronouns) for guidance 
+on when it is appropriate to use "you" and "your".
+
+* Bad: The files should be added to the directory by the owner.
+* **_Good_: The owner should add the files to the directory.**
+
```
h3xar0n:
```suggestion
* Bad: To access a network's LDAP services, your computer must first be authenticated
* **_Good_: To access a network's LDAP services, a server that supports the protocol must first authenticate your computer

```

h3xar0n: @nijikokun how's this?

levfishbluefish:
This level of granularity will make for difficult PRs - IMHO the Stye Guide should be reserved for Kong Enterprise specific terms, and other suggestions like this (and some others currently in the Style Guide) should be moved to the Technical Writing guide. 

The idea being that while we may not go through and change a PR's sentence structure, it is a tip for better writing

h3xar0n:
@levfishbluefish I feel it would be easier to link to a specific rule than saying, "Go read the writing tips in the technical writing guide." It's a use case for having a reference rather than an explanation. 

The reality is that most contributors do not have access to the Chicago Manual, which is the style we assume in our copyedits. As a result, we need to be explicit about the rules that matter, which exceed product terms. @nijikokun recently had a case where, because he couldn't refer to this rule, the suggestion was moot. 

Having general tips in the writing guide is great--and having a central source of truth (with discreet, linkeable rules) doesn't conflict with that goal. In fact, it supports it.

STYLEGUIDE.md
hishamhm: I like it! It's a good example, employing both active voice and a valid use of a personal pronoun (the examples are missing `.` at the end, though)

h3xar0n: fixed, thank you @hishamhm !

levfishbluefish:
Before adding this then I would like us to review the Style Guide as a whole and add any other grammar points like this all together

h3xar0n:
@levfishbluefish I think that's out of scope, maybe we can create a ticket. There are other usage and style rules already in the Style Guide.

PR #1316, opened 2019-06-04
docs(0.34-x/rbac-admin-api) Fix response when deleting a Role Entity …

hanweihsu:
…Permission

The response when deleting a RBAC Role Endpoint Permission via Admin API should be '_204_' instead of '_200_'.

### Summary

The response when deleting a RBAC Role Endpoint Permission should be '204' instead of '200'.

### Full changelog

* [app/enterprise/0.34-x/rbac/admin-api.md]: modified _Response_ when deleting a RBAC Role Endpoint Permission


h3xar0n:
Thanks, @hanweihsu .

PR #1322, opened 2019-06-06
docs(admin-api.md) new memory stats in /status

locao:
### Summary

Added memory entry in `/status` endpoint docs.

### Full changelog

* Created new dir `app/1.2.x/` and added `admin-api.md` file.
* Added memory entry in `/status` endpoint docs.

### DIFF
--- a/autodoc-admin-api/data.lua
+++ b/autodoc-admin-api/data.lua
@@ -151,7 +151,7 @@ return {
           description = [[
             Retrieve usage information about a node, with some basic information
             about the connections being processed by the underlying nginx process,
-            and the status of the database connection.
+            the status of the database connection, and node's memory usage.
 
             If you want to monitor the Kong process, since Kong is built on top
             of nginx, every existing nginx monitoring tool or agent can be used.
@@ -163,6 +163,28 @@ return {
 
             ```json
             {
+                "database": {
+                  "reachable": true
+                },
+                "memory": {
+                    "workers_lua_vms": [{
+                        "http_allocated_gc": "0.02 MiB",
+                        "pid": 18477
+                      }, {
+                        "http_allocated_gc": "0.02 MiB",
+                        "pid": 18478
+                    }],
+                    "lua_shared_dicts": {
+                        "kong": {
+                            "allocated_slabs": "0.04 MiB",
+                            "capacity": "5.00 MiB"
+                        },
+                        "kong_db_cache": {
+                            "allocated_slabs": "0.80 MiB",
+                            "capacity": "128.00 MiB"
+                        },
+                    }
+                },
                 "server": {
                     "total_requests": 3,
                     "connections_active": 1,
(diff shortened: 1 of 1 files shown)


levfishbluefish:
Please update this to PR into the `release/1.2.x` branch instead of master

app/1.2.x/admin-api.md
```diff
@@ -0,0 +1,2731 @@
...
+    "database": {
+        "reachable": true
+    }
+}
+```
+
+* `memory`: Metrics about the memory usage.
+    * `workers_lua_vms`: An array with all workers of the Kong node, where each
+      entry contains worker memory usage information (`http_allocated_gc`) and
+      its process identification number (`pid`).
```
thibaultcha: We should mention that the memory referred to here is the http submodule's Lua VM's Garbage Collector size as reported by `collectgarbage("count")` every 10s for each _active_ worker (it is only reported if a worker received a proxy call within the last 10s).

app/1.2.x/admin-api.md
```diff
@@ -0,0 +1,2731 @@
...
+    * `lua_shared_dicts`: An array of information about shared dictionaries,
+      where each node contains how much memory is dedicated for the specific
+      shared dictionary (`capacity`) and how much of said memory is in use
+      (`allocated_slabs`).
```
thibaultcha: Here, we should clarify that shared dictionaries have LRU eviction capabilities, meaning that a `lua_shared_dict` with `allocated_slabs == capacity` will still be able to function properly. That said, a note can be made that for _some_ shms (e.g. cache hit/miss shms), increasing their size if they are full can be beneficiary to the overall performance of a Kong node.

app/1.2.x/admin-api.md
```diff
@@ -0,0 +1,2731 @@
...
+			"pid": 18478
+		}],
+		"lua_shared_dicts": {
+			"kong": {
+				"allocated_slabs": "0.04 MiB",
+				"capacity": "5.00 MiB"
+			},
+			"kong_db_cache": {
+				"allocated_slabs": "0.80 MiB",
+				"capacity": "128.00 MiB"
```
thibaultcha: We need to mention the arguments accepted by this endpoint. Here is the description from the commit (https://github.com/Kong/kong/commit/389c8dd2aeabd90a23c0a7812a43fbb1b7fd9936):

* `unit`: can be one of `b/B`, `k/K`, `m/M`, `g/G`, which will return
    results in bytes, mebibytes, kibibytes, mebibytes, or gibibytes,
    respectively. When "bytes" are requested, the memory values
    in the response will have a number type instead of string.
   Defaults to `k`.
* `scale`: the number of digits to the right of the decimal points when
   values are given in human-readable memory strings (`unit` other than "bytes").
   Defaults to `2`.

app/1.2.x/admin-api.md
```diff
@@ -0,0 +1,2731 @@
...
+    "server": {
+        "total_requests": 3,
+        "connections_active": 1,
+        "connections_accepted": 1,
+        "connections_handled": 1,
+        "connections_reading": 0,
+        "connections_writing": 1,
+        "connections_waiting": 0
+    },
+    "database": {
```
thibaultcha: When hitting this endpoint today, this property comes _before_ the `memory` property (see example in the commit https://github.com/Kong/kong/commit/389c8dd2aeabd90a23c0a7812a43fbb1b7fd9936). This example should reflact that.

app/1.2.x/admin-api.md
thibaultcha: We should also mention that lua_shared_dicts are mapped memory zones shared by all Nginx workers.

app/1.2.x/admin-api.md
```diff
@@ -0,0 +1,2731 @@
...
+
+---
+
+### Retrieve Node Status
+
+Retrieve usage information about a node, with some basic information
+about the connections being processed by the underlying nginx process,
+the status of the database connection, and node's memory usage.
+
+If you want to monitor the Kong process, since Kong is built on top
```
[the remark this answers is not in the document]
thibaultcha: There are a number of mistakes I have also noted in other sections of this document, but there were written before this PR, so we should consider them outside the scope of this PR, for the time being

hishamhm:
@locao Besides changing the base in the Github UI, you need to rebase the branch locally as well and force-push it.

app/1.2.x/admin-api.md
```diff
@@ -613,13 +635,38 @@ HTTP 200 OK
...
+      capabilities, so a full dictionary, where `allocated_slabs == capacity`,
+      will work properly. However for some dictionaries, e.g. cache HIT/MISS
+      shared dictionaries, increasing their size can be beneficial for the
+      overall performance of a Kong node.
+    * The memory usage unit and precision can be changed using the querystring
+      arguments `unit` and `scale`:
+        * `unit`: one of `b/B`, `k/K`, `m/M`, `g/G`, which will return results
+          in bytes, kibibytes, mebibytes, or gibibytes, respectively. When
+          "bytes" are requested, the memory values in the response will have a
+          number type instead of string. Defaults to `k`.
```
thibaultcha: The default is `m` (my bad, for the comment I posted earlier was not taking subsequent changes I made to the endpoint in consideration).

thibaultcha: (As demonstrated by the above example response)

locao: Fixed!

hishamhm:
merged manually!

PR #1323, opened 2019-06-06
docs(plugins) document how to use plugins in DB-less mode

kikito:
* Adds a new way to add "tabs" using liquid templates (`{% tabs %} ... {% endtabs %}`). Since currently there's no CSS to render tabs, for now this utility renders both options one after the other.
* Uses the `{% tabs %}` utility to render "tabs" in all the generated code which precedes plugins.
* Includes manual fixes / rewordings of all Open Source plugin docs to include db-less instructions when possible.

Example of generated code:

Screenshot 2019-06-06 at 23 38 08


Gemfile
```diff
@@ -1,5 +1,5 @@
 source 'https://rubygems.org'
-gem 'jekyll', '3.3.1'
+gem 'jekyll', '~> 3.3.1'
```
levfishbluefish: We recently updated Jekyll to 3.8.5, you may want to confirm there are no issues with the new plugin with this update

kikito: Ok yes this is a problem. Jekyll internals have changed from 3.3.1 to 3.8.5.

Yes, this means I will have to update the plugin.

kikito:
I changed the plugin code, it should be good now

PR #1356, opened 2019-06-20
fix(js): Fix smooth scrolling

levfishbluefish:
### Summary

Fixes smooth scrolling when clicking on a Table of Contents link on documentation pages

### Full changelog

* Adds `anchor_class="scroll-to"` to toc call in doc.html (documentation layout) so that the smooth scrolling JS works (this was apparently broken)
* Adds a window width check to smooth-scrolling to reduce offset for smaller screens - this fixes the anchor-links cutting off the heading on screens smaller than 1000px wide

### DIFF
--- a/app/_assets/javascripts/app.js
+++ b/app/_assets/javascripts/app.js
@@ -15,9 +15,13 @@ $(function () {
 
   $('.scroll-to').on('click', function (e) {
     e.preventDefault()
-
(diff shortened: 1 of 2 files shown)


h3xar0n:
This fixes smooth scrolling for the auto-generated ToC, but not the hand-rolled ones we now make. Was that the intended scope of the fix?

PR #1357, opened 2019-06-20
fix(install): Fix generated TOC styling

### LINKED ISSUE #917: All pages should correctly render both unordered lists AND download buttons

levfishbluefish:
### Summary

Allow core installation docs to use the Table of Contents extension 

### Full changelog

* Adds call to toc.html in install page layout, moves it out of main content section to ensure it displays correctly
* Updates all installation pages to use the new TOC

### Issues resolved

Fix #917

### DIFF
--- a/app/_assets/stylesheets/pages/install-method.less
+++ b/app/_assets/stylesheets/pages/install-method.less
@@ -47,167 +47,167 @@
     }
   }
 
-  .content {
+  .page-content {
     padding: 50px 0;
+    .content {
 
-    h3 {
-      font-size: 22px;
-      margin-bottom: 1em;
-    }
-
-    pre {
-      margin: 1em 0;
-    }
-
-    hr {
-      margin: 1.5em 0px;
-    }
-
-    ul {
-      list-style: none;
-      padding: 0;
-      margin: 0;
-
-      li {
-        margin-right: 10px;
-
-        display: inline-flex;
-        align-items: center;
-
-        a {
-          display: flex;
+      h3 {
+        font-size: 22px;
+        margin-bottom: 1em;
+      }
+  
+      pre {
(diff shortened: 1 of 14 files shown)


h3xar0n:
One thing to note:

Kubernetes TOC changes from

Screen Shot 2019-06-20 at 3 21 22 PM

to

Screen Shot 2019-06-20 at 3 21 27 PM

That seems intentional and consistent, approving.

Issue #835, opened 2018-08-27
Table width exceeds content width


coopr:
- `docs.konghq.com` url(s): https://docs.konghq.com/enterprise/0.33-x/plugins/rate-limiting-advanced/ (this issue may exist on other pages)

### Summary

See attached screenshot
![rate_limiting_advanced_plugin_-_v0_33-x___kong_-_open-source_api_management_and_microservice_management](https://user-images.githubusercontent.com/94904/44675393-467cf780-a9e5-11e8-8f58-b59a18012941.jpg)

PR #1390, opened 2019-07-09
fix(styling): Fix table styling

levfishbluefish:
Changes table layout on plugin / extension pages so it no longer overflows main body

Fixes Issue #835 

### Desktop
- In `extensions.html`, moves `default` column in Plugin data tables to be nested under the property name, this helps prevent overflowing content
- Fixes overlapping styles causing some table text sizes to exceed 14px 
- Adds zebra-stripping for readability

Screen Shot 2019-07-08 at 9 46 58 PM

### Mobile:
- Collapses tables on mobile for readability
- Hides table headers on mobile
- JQuery adds the `data-label` attribute to table elements
- CSS shows `data-label` attr on mobile to make mobile tables more readable.
- Forces all text-alignment to `left` only on mobile -- allows authors to still use Markdown content alignment on Desktop

Before:
Screen Shot 2019-07-08 at 9 46 40 PM

After:
Screen Shot 2019-07-09 at 4 31 16 PM

### DIFF
--- a/app/_assets/stylesheets/page.less
+++ b/app/_assets/stylesheets/page.less
@@ -434,7 +434,7 @@
   {
     table, thead, tbody, th, td, tr {
       display: block;
-      text-align: left !important;
+      text-align: left;
     }
 
     table {
--- a/app/_assets/stylesheets/page.less
+++ b/app/_assets/stylesheets/page.less
@@ -440,7 +440,7 @@
     table {
         /* Hide table headers  */
       thead tr {
-       display: none;
+        display: none;
       }
 
       tr { border: 1px solid #ccc; }
--- a/app/_assets/stylesheets/page.less
+++ b/app/_assets/stylesheets/page.less
@@ -405,63 +405,62 @@
     font-style: italic;
   }
 
-    /* 
-  Generic Styling, for Desktops/Laptops 
+    /*
+  Generic Styling for Desktop
   */
-  table { 
-    width: 100%; 
-    border-collapse: collapse; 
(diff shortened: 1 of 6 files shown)


app/_assets/stylesheets/page.less
```diff
@@ -410,50 +405,63 @@
...
-      text-transform: lowercase;
-      .font-source-sans-sc();
-      font-size: 14px;
-      text-align: left !important;
-      background: @gray-light;
-      border-bottom: 1px solid @gray;
-      color: lighten(@gray-dark, 30%);
-      padding: 10px;
+  @media
+  only screen and (max-width: 1099px)
```
h3xar0n: Is there a reason to break the line here?

adamjkuhn:
perhaps set the column headers as data attributes for each <td> and including them as pseudo elements at specific breakpoints?
I would also mention that on collapse all text should maybe be aligned left- that said this is much much nicer; responsive tables are always a chore.

h3xar0n:
alignment example: 

Screen Shot 2019-07-09 at 10 57 55 AM

levfishbluefish:
It looks like smaller tables (not the main extension table I worked on) are inheriting text-align properties not associated with global table styles. Will have to track those down

app/_assets/stylesheets/page.less
levfishbluefish: ah good catch, I had it when I had more settings so it would stay at 80 chars, but removed the additional settings

PR #1398, opened 2019-07-11
docs(enterprise/changelog) clarify request-validator

gszr:
## Summary

Clarify request validator schema support.

### DIFF
--- a/app/enterprise/changelog.md
+++ b/app/enterprise/changelog.md
@@ -129,9 +129,10 @@ for details.
     parameter. Now you can provide status code ranges and single status codes
     together (e.g., `201-204,401`)
   - **Request Validator** (_NEW_):
-    - Validate the request body against a JSON schema; if the payload is not
+    - Validate the request body against a schema; if the payload is not
     valid JSON or if it doesn't conform to the schema, the plugin returns a 400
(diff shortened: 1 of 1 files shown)


app/enterprise/changelog.md
```diff
@@ -129,9 +129,10 @@ for details.
     parameter. Now you can provide status code ranges and single status codes
     together (e.g., `201-204,401`)
   - **Request Validator** (_NEW_):
-    - Validate the request body against a JSON schema; if the payload is not
+    - Validate the request body against a schema; if the payload is not
     valid JSON or if it doesn't conform to the schema, the plugin returns a 400
```
h3xar0n: shouldn't we remove `is not valid JSON` as well?

gszr: This PR only clarifies that the schema format used is not JSON schema (http://json-schema.org); but the request body still is validated, both syntactically (valid JSON) and semantically (conforms to the schema spec), so we shoudn't remove that.

h3xar0n: Understood, thank you for clarifying.

gszr:
Thank you @h3xar0n!

PR #1410, opened 2019-07-22
Install update

mannykhadilkar:


levfishbluefish:
Looks good! Thanks for the updates. While we are updating this guide, I'm going to do a bit of reformatting so it matches the other installation guides. Will commit to this PR

levfishbluefish:
Changes incorporated into #1501

PR #1414, opened 2019-07-24
docs(zipkin) add docs for include_credential config option

james-callahan:
Tied to https://github.com/Kong/kong-plugin-zipkin/pull/37


james-callahan:
Do we only merge this when a zipkin plugin release is done?

bungle:
@levfishbluefish it is now released, so this can be merged too.

PR #1437, opened 2019-08-08
docs(1.3.x) add upgrading.md instructions for 1.3.x

kikito:
Most changes from this commit are a straight replacement of the
text using https://github.com/Kong/kong/blob/next/UPGRADE.md .

There are two notable exceptions:

* Header levels have been changed
* The `Upgrade Path for Patch Releases` section does not exist in
  Kong's `UPGRADE.md` file.

Related: #1436

### DIFF
--- a/app/1.3.x/upgrading.md
+++ b/app/1.3.x/upgrading.md
@@ -3,7 +3,7 @@ title: Upgrade guide
 ---
 
 <div class="alert alert-warning">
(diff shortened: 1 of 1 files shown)


hishamhm:
@dndx looks like you force-pushed into `release/1.3.0`, which caused conflicts in this PR. Commit 7879e93 looks fine (it's a copy from UPGRADE.md which we already reviewed), I think it's the only one that should stay here, so could you fix this branch and merge it?

PR #1440, opened 2019-08-08
Add version independent references

levfishbluefish:
Creates a `references` folder in `enterprise`, but outside of specific versions. This will allow us to maintain certain references and guides separately from releases - allowing us to cut down on the amount of duplicated pages between versions.

As an example, this adds the Rate Limiting Library reference


h3xar0n:
How can we predict that the content will never be version-dependent? Not challenging the idea, because it seems really great, just want to know that particular point so that we have a standard for deciding which files meet the criteria to go into `references`.

levfishbluefish:
We can handle it the same way as plugins maybe? Plugins have their own versions, it would be very easy to add the version dropdown / sidebar. 

_Or_ we can note changes or version dependencies if they are updated

PR #1442, opened 2019-08-08
docs(proxy) add more description text for SNI matcher in Route entity

dndx:

### DIFF
--- a/app/1.3.x/proxy.md
+++ b/app/1.3.x/proxy.md
@@ -764,7 +764,8 @@ only to TLS, but also to other protocols carried over TLS - such as HTTPS and
 If multiple SNIs are specified in the Route, any of them can match with the incoming request's SNI.
 with the incoming request (OR relationship between the names).
 
-SNI is determined at TLS handshake time and can not be modified after TLS connection
+The SNI is indicated at TLS handshake time and cannot be modified after the TLS connection has been established. This means, for example, that multiple requests reusing the same keepalive connection
+will have the same SNI hostname while performing router match, regardless of the `Host` header.
 has been established. This means keepalive connections that send multiple requests
 will have the same SNI hostnames while performing router match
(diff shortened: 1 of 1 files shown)


app/1.3.x/proxy.md
```diff
@@ -761,7 +761,16 @@ allows routing via SNIs:
 Incoming requests with a matching hostname set in the TLS connection's SNI
 extension would be routed to this Route. As mentioned, SNI routing applies not
 only to TLS, but also to other protocols carried over TLS - such as HTTPS and
-gRPCs.
+gRPCs. If more than one hostnames are specified, then any one of them can match
```
gszr:
```suggestion
If multiple SNIs are specified in the Route, any of them can match with the incoming request's SNI.
```

gszr: Note sure what we meant with "(OR relationship between the names)."; it sounds a little obscure to me.

app/1.3.x/proxy.md
```diff
@@ -761,7 +761,16 @@ allows routing via SNIs:
...
+with the incoming request (OR relationship between the names).
+
+SNI is determined at TLS handshake time and can not be modified after TLS connection
```
gszr:
```suggestion
The SNI is indicated at TLS handshake time and can not be modified after the TLS connection has been established. This means, for example, that multiple requests reusing the same keepalive connection
will have the same SNI hostname while performing router match, regardless of the `Host` header.
```

app/1.3.x/proxy.md
gszr:
```suggestion
If multiple SNIs are specified in the Route, any of them can match with the incoming request's SNI.
```

gszr: (Not sure "(OR relationship between the names)." is necessary here)

PR #1449, opened 2019-08-12
docs(key-auth) remove the underscore in key name restriction from doc

dndx:
because it no longer applies

See: https://github.com/Kong/kong/pull/4896


dndx:
@levfishbluefish I didn't add any HTML tags in the change. Could you elaborate more on how to fix it?

bungle:
It seems to be supported here:
https://docs.konghq.com/hub/kong-inc/key-auth/#parameters (at least the `<br>` is outputted as is).

dndx:
Merged for now, we can fix any rendering issue later should they appear.

PR #1452, opened 2019-08-16
Release: 1.1.3

hishamhm:


hishamhm:
This is a leftover from the security releases, should be an easy merge

PR #1458, opened 2019-08-19
docs(ee-changelog) add notes for 0.36-1 and 0.35-4

h3xar0n:
Changelog for 0.36-1 and 0.35-4


UkiahSmith:
For this PR, this change looks good to me.

However if someone were to follow up and read the [CE changelog](https://github.com/Kong/kong/blob/master/CHANGELOG.md) they might be confused as 1.0.3 does not mention the CVEs, only 1.2.2 does. I understand that is out of scope of this PR.

PR #1460, opened 2019-08-21
docs(oidc) add Azure AD example

rainest:
Azure AD+OIDC documentation for use with a web application. As best I know, the combination of explicitly specifying the API version in the manifest and use of the `.default` scope *should* force Azure AD to return an access token we can handle. Please try in your own environments to confirm that I haven't inadvertently missed some configuration.

Please also comment on the general organization and content versus https://docs.konghq.com/enterprise/0.35-x/plugins/oidc-google/. I plan to use similar structure for Okta and updated versions of the Google/Auth0 docs.

### DIFF
--- a/app/enterprise/0.35-x/plugins/oidc-azuread.md
+++ b/app/enterprise/0.35-x/plugins/oidc-azuread.md
@@ -1,0 +1,87 @@
+---
+title: OpenID Connect with Azure AD
+---
+## Introduction
+
+This guide covers an example OpenID Connect plugin configuration to authenticate browser clients using an Azure AD identity provider.
+
+## Prerequisites
+
+Because OpenID Connect deals with user credentials, all transactions should take place over HTTPS. Although user passwords for third party identity providers are only submitted to those providers and not Kong, authentication tokens do grant access to a subset of user account data and protected APIs, and should be secured. As such, you should make Kong's proxy available via a fully-qualified domain name and [add a certificate][add-certificate] for it.
+
+## Kong Configuration
+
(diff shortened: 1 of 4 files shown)


felderi:
Man you guys are fast!  Really impressive.  I really like the new format.  Much clearer and to the point. The one addition I would make is some reference to the auth methods parameters to control which flow to enable.  Something to the extent of, by default, the plug-in enables all supported authentication flows (maybe a link to the supported flows?).  To enable only specific flows set auth methods to the desired flows (X, Y, Z).  For example, to enable only introspection, provide the command to do.  

The other comment I would make is that if this is for Enterprise we may want to include Kong Manager steps as well.  This can of course be done later.

app/enterprise/0.35-x/plugins/oidc-azuread.md
```diff
@@ -0,0 +1,87 @@
...
+
+Visiting a URL matched by that route in a browser will now redirect to Microsoft's authentication site and return you to the redirect URI after authenticating.
+
+### Access Restrictions
+
+The configuration above allows users to authenticate and access the Route even though no consumer was created for them: any user with a valid account in the directory will have access to the Route. The OIDC plugin allows this as the simplest authentication option, but you may wish to restrict access further. There are several options for this:
+
+#### Domain Restrictions
+
+Azure AD does provide identity tokens with the `hd` claim, and as such the OIDC plugin's `domains` configuration cannot restrict users based on their domain. Using a [single-tenant][azure-tenant] application will restrict access to users in your directory only. Multi-tenant apps allow users with Microsoft accounts from other directories and optionally any Microsoft account (e.g. live.com or Xbox accounts) to sign in.
```
deirdre-anderson: Azure AD does not instead of 'does' provide identity tokens...

deirdre-anderson:
Tested out the config locally from the steps above and it works perfectly

PR #1474, opened 2019-08-28
fix(passwords) use anchor links for page sections

rainest:
Fix Kong Manager password/token reset instructions to point to appropriate anchors within the page, rather than non-existent off-page links.


h3xar0n:
Thank you, @rainest . This page in 0.35 doc needs the same fix; I will commit it before approving.

PR #1480, opened 2019-09-06
remove purifycss

levfishbluefish:
Removes `purifycss` as it is an unmaintained project causing memory leak issues in the repo. 

Looking into alternative packages, but lets remove this for now.


h3xar0n:
Same-- @levfishbluefish , are we just keeping it there so that when we find an alternative, we know which classes to whitelist?

levfishbluefish:
Yes, I wanted to keep the whitelist in there for the moment while I look at alternative packages.

PR #1492, opened 2019-09-18
fix role endpoint description

levfishbluefish:
fixes role endpoint descp. in all versions, as well as persistent typos


h3xar0n:
The Permissions sections need to be reverted too, please see comment above.

PR #1522, opened 2019-10-07
Update "deleting an entity" docu

danielfuerst:
Deleting an entity required indeed a table. However it should be of this format.

### DIFF
--- a/app/1.3.x/plugin-development/custom-entities.md
+++ b/app/1.3.x/plugin-development/custom-entities.md
@@ -646,9 +646,9 @@ must do it manually, by checking with `select` before invoking `delete`.
 Example:
 
 ``` lua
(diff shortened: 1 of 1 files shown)


h3xar0n:
Does this only affect 1.3, or earlier? @Kong/team-core , please verify.

hishamhm:
> Does this only affect 1.3, or earlier? @Kong/team-core , please verify.

The proposed change is the docs is correct (it is also correct for every version since at least 1.0, if this mistake is present in earlier versions of the docs as well).

levfishbluefish:
Hi @danielfuerst looks like this error is also present in 1.1 and 1.0, would you mind updating those as well in this PR?

HCloward:
Closing this PR as the branch is old and I can't pull it down. I created a replacement PR #2655

PR #1570, opened 2019-11-07
fix(migrations) correct 1.3 links

rainest:
Use correct target for Enterprise 1.3 migration links. Links previously targeted guides for versions older than the prerequisite version.


h3xar0n:
Good catch, thank you, @rainest

PR #1577, opened 2019-11-11
docs(proxy) notes on route headers

jeremyjpj0916:
More details per: https://github.com/Kong/kong/issues/5213


hishamhm:
@h3xar0n this PR was prompted by this clarification I made: https://github.com/Kong/kong/issues/5213#issuecomment-552578137 — do you think we can reuse some more of that explanation or is the PR good as-is?

hishamhm:
@h3xar0n just to confirm, the change of this PR is correct as-is.

PR #1592, opened 2019-11-13
Change to parameters and curl

robbbittner:


robbbittner:
Updated doc to match changes in trial architecture, adding a route, updating curl and updating the response.

PR #1601, opened 2019-11-19
docs(*/k4k8s) add kubernetes docs

h3xar0n:


h3xar0n:
Please pardon the un-squashed commits; @levfishbluefish , we can discuss how to work around this in the public/private sync. 

Please refer to the actual files changed instead.

PR #1666, opened 2019-12-17
add images and bintray warning

levfishbluefish:
- Add images to install links
- Add warning about bintray 404 if you aren't logged in


levfishbluefish:
Tests are failing on the URLs of the icons I added but this is a false positive, pls ignore

h3xar0n:
When I copy the URLs from the link checker it looks like a genuine 404:
Screen Shot 2019-12-17 at 8 31 54 AM

h3xar0n:
And images don't render on the preview site: 
Screen Shot 2019-12-17 at 8 33 06 AM

h3xar0n:
So is it a false positive because the images are added in a separate PR? Right now it's a true positive...

levfishbluefish:
Turns out they got gitignored because the folder is named `install`, force added them

PR #1671, opened 2019-12-18
doc(plugins) add doc for acme plugin

fffonion:


fffonion:
@levfishbluefish Thanks! I added the missing assets.

app/_data/extensions/kong-inc/acme/versions.yml
```diff
@@ -0,0 +1 @@
+- release: 0.2.0
```
levfishbluefish:
```suggestion
- release: 0.2.0

```

levfishbluefish: (linter requires blank new line at the end of all files)

app/_hub/kong-inc/acme/index.md
```diff
@@ -0,0 +1,213 @@
...
+      default: "`shm`"
+      description: |
+        The backend storage type to use, choice of `"kong"`, `"shm"`, `"redis"`, `"consul"` or `"vault"`. In dbless mode, `"kong"` storage is unavailable.
+    - name: storage_config
+      required: false
+      default:
+      description: |
+        Storage configs for each backend storage. See below for its default value.
+    - name: tos_accepted
+      required: false
```
hbagdi: This should be always required?

hbagdi: We should ask user to always supply this as `true`.

app/_hub/kong-inc/acme/index.md
```diff
@@ -0,0 +1,213 @@
...
+                host = '127.0.0.1',
+                port = 8200,
+                -- secrets kv prefix path
+                kv_path = "acme",
+                -- Vault token
+                token = nil,
+                -- timeout in ms
+                timeout = 2000,
+            },
+        }
```
hbagdi: Please use JSON representation here as we do not use Lua code in the Admin API itself anywhere.

app/_hub/kong-inc/acme/index.md
```diff
@@ -0,0 +1,213 @@
...
+    To configure storage type other than `kong`, please refer to [lua-resty-acme](https://github.com/fffonion/lua-resty-acme#storage-adapters).
+
+---
+
+### Using the Plugin
+
+#### Configure Kong
+
+- Kong needs to listen 80 port or proxied by a load balancer that listens for 80 port.
+- `lua_ssl_trusted_certificate` needs to be set in `kong.conf` to ensure the plugin can properly
```
hbagdi: We should encourage the use of Nginx directive injection rather than this variable here to avoid confusing the user.

app/_hub/kong-inc/acme/index.md
```diff
@@ -0,0 +1,213 @@
...
+    - name: cert_type
+      required: false
+      default: "`rsa`"
+      description: |
+        The certificate type to create, choice of `"rsa"` for RSA certificate or `"ecc"` for EC certificate.
+    - name: domains
+      required: false
+      default: "`[]`"
+      description: |
+        The list of domains to create certificate for. Wildcard domain like `*.example.com` is also supported. Regex pattern is not supported.
```
hbagdi: > Wildcard domain like `*.example.com` is also supported.

I'd recommend that we should rephrase this.
This will be a source of confusion in future as some users will confuse this with wildcard certificates, which this plugin cannot support since those are based on DNS challenge.

app/_hub/kong-inc/acme/index.md
fffonion: We will check this value in schema if user is using let's encrypt and tos_accepted is not set to `true`. I feel it will be more user friendly when user omits this value to show `you must set this to true` instead of `setting this value is required`.

app/_hub/kong-inc/acme/index.md
fffonion: updated with a json text

fffonion: updated with a json text

app/_data/extensions/kong-inc/acme/versions.yml
h3xar0n: may need a newline if build fails

```suggestion
- release: 0.2.0

```

PR #1672, opened 2019-12-19
docs(hub) reformat route transformer as a hub page

h3xar0n:
@Tieske , when the Netlify preview is ready, please verify that the parameters are correctly rendered.


h3xar0n:
@levfishbluefish says we're clear to merge, this is just a CI fluke that will not affect master. Thanks @Tieske .

PR #1673, opened 2019-12-19
docs(1.4.x/admin) generate lists of possible values from `one_of`

hishamhm:
Our lists of accepted protocols in the service entity was outdated in the 1.4.x documentation.

Instead of fixing it once and risking getting this kind of thing out-of-date again, we now generate the list of accepted values in the documentation, the same way we do for presenting default values.

A [companion PR was sent to the Kong repo](https://github.com/kong/kong/pull/5358) with the updated script.


hishamhm:
Netlify seems to be broken at the moment, which explains the check failures (logs are empty).

@Kong/team-docs, a heads-up that you might want to check if any other places where Admin API is maintained by hand (Enterprise docs?) need to have those attributes fixed as well.

h3xar0n:
Noted internally, thanks, @hishamhm .

PR #1679, opened 2019-12-19
Release/1.3.0.1

h3xar0n:


app/_data/docs_nav_ee_1.3-x.yml
```diff
@@ -139,6 +141,8 @@
         url: /admin-api/vitals/vitals-prometheus-strategy
   - text: Audit Logging
     url: /admin-api/audit-log
+  - text: Keyring and Data Encryption
```
levfishbluefish: Looks like we are linking this in the nav twice? was this on purpose?

h3xar0n: @p0pr0ck5 , these seem to point to different paths.

h3xar0n: I see now. One is an overview, the other is API reference.

app/enterprise/changelog.md
```diff
@@ -3,6 +3,89 @@ title: Kong Enterprise Changelog
...
 ---
 
+## 1.3.0.1
+**Release Date:** 2019/12/19
+
+### Features
+
+* Kong Enterprise now officially supports RHEL 8
+* (Alpha feature) Kong Enterprise can now perform encryption-at-rest for sensitive fields within the data store (Postgres or Cassandra).
+* (Alpha feature) Kong Enterprise now has a License Reports module for customers to view current usage metrics. For more information, contact your Kong Account Executive.
```
levfishbluefish:
```suggestion
```

h3xar0n: This needs to be in both 1.3 and 1.3.0.1, correct, @GayleNeumann ?

PR #1691, opened 2019-12-31
docs(extension) add description for consumer.id

h3xar0n:
the conditional should check for the underscored value, whereas the description should have dot notation; these were reversed.

also improved the definition of "consumer" and style.


h3xar0n:
Screen Shot 2019-12-31 at 2 04 07 PM
Screen Shot 2019-12-31 at 2 04 23 PM

PR #1699, opened 2020-01-07
Fix broken link in install.md

deirdre-anderson:


h3xar0n:
Thanks, @deirdre-anderson !

PR #1721, opened 2020-01-17
updated installation to remove RBAC and httpie

mannykhadilkar:
in the interest of making the enterprise installation simpler and more streamlined i removed the rbac steps and removed httpie to stay with curl.

### DIFF
--- a/app/enterprise/1.3-x/deployment/installation/amazon-linux-2.md
+++ b/app/enterprise/1.3-x/deployment/installation/amazon-linux-2.md
@@ -216,21 +216,13 @@ Now that the database is set up, Kong must be given credentials to access the da
   > Note: If you used different values for the user and database name, uncomment the pg_user and pg_datastore variables and add the username and datastore values
 
 
-## Step 4. Customize Your Configuration
-Before starting Kong, you can further modify Kong’s configuration file to enable a host of different features:
+## Step 4. Enable Kong Manager and Developer portal
+Before starting Kong, you can further modify Kong’s configuration file to enable Kong Manager (a graphical user interface for administrators) and the Developer Portal. 
 
 
-### Enable RBAC
+### Seed Super Admin Password
 
-  Kong Enterprise allows applying RBAC to all requests. To enable RBAC, you must set RBAC to `on`, select an authentication plugin for Kong Manager, and configure the [Session plugin](/hub/kong-inc/session/).
-
-  ```
-  enforce_rbac = on
-
-  admin_gui_auth = basic-auth
-
-  admin_gui_session_conf = {"secret":"password", "cookie_secure": false, "cookie_samesite": "off", "cookie_name": "admin_session", "storage": "kong"}
-  ```
+Kong has a built in super admin account (kong_admin) that you will use when configuring RBAC. In order to provision the kong_admin account correctly during the initial installation, we need to seed the password. Please secure this password and save it for future configuration when we enable RBAC functionality in Kong. 
 
   Export the KONG_PASSWORD environment variable:
   ```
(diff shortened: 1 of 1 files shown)


app/enterprise/1.3-x/deployment/installation/amazon-linux-2.md
```diff
@@ -216,21 +216,13 @@ Now that the database is set up, Kong must be given credentials to access the da
...
-  Kong Enterprise allows applying RBAC to all requests. To enable RBAC, you must set RBAC to `on`, select an authentication plugin for Kong Manager, and configure the [Session plugin](/hub/kong-inc/session/).
-
-  ```
-  enforce_rbac = on
-
-  admin_gui_auth = basic-auth
-
-  admin_gui_session_conf = {"secret":"password", "cookie_secure": false, "cookie_samesite": "off", "cookie_name": "admin_session", "storage": "kong"}
-  ```
+Kong has a built in super admin account (kong_admin) that you will use when configuring RBAC. In order to provision the kong_admin account correctly during the initial installation, we need to seed the password. Please secure this password and save it for future configuration when we enable RBAC functionality in Kong.
```
h3xar0n: Interesting approach. The issue (as discussed with @GayleNeumann) is that many other guides assume RBAC is already enabled. If we make it optional for installation, then it will force people to restart Kong later. I think this has caused frustration in the past. 

On the other hand, I see the merits of treating it as optional, under the premise that this installation is with the most basic settings.

The fact is that Kong very much requires users to have an idea of what they want at a very early stage. That's only more the case with declarative config.

Given that this is an onboarding issue, I'd like to loop in @connorckong . I think for now, while the suggestion is completely warranted, this change may affect much more of the experience outside of this PR, so it may be out of scope here.Still very open to learning more about how we could work this in, @mannykhadilkar .

GayleNeumann: I'll schedule a meeting to discuss RBAC. My understanding is we need to seed the Super Admin initially, and configure RBAC later to avoid having to restart Kong. In any guide that assumes RBAC is enabled, we need to have a "prerequisites" section at the beginning of the topic to clarify the requirements.

h3xar0n: Configuring RBAC later does entail restarting Kong, as with anything in `kong.conf`. 

I think the issue we're trying to avoid is needing to run migrations later.

h3xar0n: Following today's discussion, @mannykhadilkar and @GayleNeumann , please feel free to adapt/revise:

```suggestion
Kong has a built-in Super Admin account (`kong_admin`) that you use when configuring RBAC. In order to seed the `kong_admin` account during the initial installation, we need to provide a password as an environment variable, `KONG_PASSWORD`. Note that this value serves as both the Super Admin's [API token](https://docs.konghq.com/enterprise/1.3-x/property-reference/#rbac_auth_header) and as their password for [logging into Kong Manager](https://docs.konghq.com/enterprise/1.3-x/kong-manager/administration/rbac/). Please secure this password and save it for future configuration when we [enable RBAC functionality](https://docs.konghq.com/enterprise/1.3-x/getting-started/start-kong/) in Kong Enterprise. 
```

h3xar0n:
Since HTTPie is a simpler alternative to cURL, and since cURL is provided nonetheless, could you elaborate the need to remove it, @mannykhadilkar ?

h3xar0n:
Based on the conflict, it looks like the newer topics cover this, so I will close for now, @mannykhadilkar . Thanks for leading the charge! :)

PR #1723, opened 2020-01-17
Added editing to Studio docs

GayleNeumann:


h3xar0n:
Addressed in `studio-1.2` comments. Thanks, @GayleNeumann 👍

PR #1745, opened 2020-01-28
Update docker.md - Install instructions

felderi:
I went through updated version of the Docker install steps using a similar template that Ahmed used for Centos.  The steps work well for me except that I still need to enable Dev Portal in Kong Manager.  Need to validate what's missing but otherwise, these steps work well.


h3xar0n:
Hi @felderi , did you move these changes to another PR or just not deem them important?

PR #1746, opened 2020-01-28
add itm query strings to Solutions urls - main nav

adamjkuhn:
* quick and simple addition of itm query strings to header navigation


h3xar0n:
Merge at will, @adamjkuhn !

PR #1747, opened 2020-01-30
Created rhel.md file with installation instructions for RHEL which are missing

felderi:


h3xar0n:
@felderi , I'm assuming we need to add navigation and an icon on the Enterprise installation page as well. If so, I'll make those changes after approving the other guides.

h3xar0n:
(will use this one)

Logo-RedHat-Hat-Color-CMYK (1)

PR #1750, opened 2020-01-30
docs(hub) update cors methods defaults

jeremyjpj0916:
As per: https://github.com/Kong/kong/blob/master/kong/plugins/cors/handler.lua#L218 , the defaults list has grown much larger.


jeremyjpj0916:
Note, this PR probably should be tweaked. Needs a newline in the list otherwise the table looks ugly in the demo.

h3xar0n:
Thank you, @jeremyjpj0916 . If worse comes to worst, it would still be better to have the missing content and look ugly than to not have it at all. I will see if I can tweak it later today, and otherwise, I'll approve by EOD.

jeremyjpj0916:
@h3xar0n cool deal, yeah I wasn't sure how to best force a line break in this doc app. Or if there is some standard Kong would want to follow there.

PR #1754, opened 2020-01-31
Update Ubuntu installation docs

peterskim12:
Update Ubuntu install docs to align with other install docs


peterskim12:
Ugh the numbered lists didn't get rendered as I expected. Need some more changes.

peterskim12:
Ok I've made a few changes after my initial PR submit and I feel good about where it's at now after those changes and confirming with Netlify.

PR #1770, opened 2020-02-06
Removing Host header known issue from CORS plugin

diequeiroz:
Since most browsers do send the Host header, the known issue in CORS plugin is outdated.

https://developer.mozilla.org/en-US/docs/Web/HTTP/CORS

image

### DIFF
--- a/app/_hub/kong-inc/cors/index.md
+++ b/app/_hub/kong-inc/cors/index.md
@@ -104,19 +104,3 @@ params:
 ## Known issues
 
 Below is a list of known issues or limitations for this plugin.
-
-### CORS Limitations
-
-If the client is a browser, there is a known issue with this plugin caused by a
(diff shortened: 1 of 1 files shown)


h3xar0n:
Please see this issue for context: https://github.com/Kong/kong/issues/3132#issuecomment-355083393

kikito:
@diequeiroz Could you clarify your reasoning here? The plugin can't handle "a custom `Host` header in a preflight `OPTIONS` request.", which is exactly what you are showing in your example (the `OPTIONS` request has a `Host` header).

What makes you think this isn't needed any more?

tlblessing:
@diequeiroz, @hishamhm, @kikito where does this PR stand? Should we close this PR?

tlblessing:
Closing this stale PR based on the convo from Enrique and lack of activity. If @kikito thinks it should be reopened, please let me/docs know.

PR #1785, opened 2020-02-12
docs(response_transformer_advanced) update remove.headers

murillopaula:
This will be available in the next 1.3.0.2 enterprise release.

### DIFF
--- a/app/_hub/kong-inc/response-transformer-advanced/index.md
+++ b/app/_hub/kong-inc/response-transformer-advanced/index.md
@@ -31,8 +31,8 @@ params:
   config:
     - name: remove.headers
       required: false
-      value_in_examples: "x-toremove, x-another-one:application/json, x-list-of-values:v1,v2,v3, Set-Cookie:/JSESSIONID=.*/, x-another-regex://status/$/, x-one-more-regex:/^/begin//"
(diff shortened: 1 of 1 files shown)


h3xar0n:
Thanks, @murillopaula . Since the doc version does not account for dot releases, do you think we should explicitly mention it's only available as of 1.3.0.2? 

(cc @GayleNeumann )

murillopaula:
Hey @h3xar0n that's what I was going to ask. If the documentation says this is available for Kong Enterprise "1.3-x" I think we need to make it explicit that it's only available for 1.3.0.2 onwards. Otherwise people would think it would be available for 1.3 and 1.3.0.1...

GayleNeumann:
> Hey @h3xar0n that's what I was going to ask. If the documentation says this is available for Kong Enterprise "1.3-x" I think we need to make it explicit that it's only available for 1.3.0.2 onwards. Otherwise people would think that it's available for 1.3 and 1.3.0.1...

Can we add a banner saying this is supported in 1.3.0.2 forward?

murillopaula:
@GayleNeumann @h3xar0n would also need to add a "this is available on 1.3.0.2 onwards" banner for this: https://github.com/Kong/docs.konghq.com/pull/1786

GayleNeumann:
> @GayleNeumann @h3xar0n would also need to add a "this is available on 1.3.0.2 onwards" banner for this: #1786

@murillopaula I'll work with @h3xar0n on getting a note saying this is available for 1.3.0.2 forward. It's a little more difficult being this param is in a table, but we'll get it done before 1.3.0.2 releases -- for this PR and also the other #1786.

PR #1787, opened 2020-02-13
docs(audit-log) clarify signature implementation pitfalls

p0pr0ck5:


p0pr0ck5:
Admin API Audit Log - v1 3-x   Kong - Open-Source API Management and Microservice Management

^ the `alert-info` callout style blends in with ``` blocks, so i went with `alert-info`. not sure if that's the best choice here- a callout box with a blue background seems like it would be the best fit IMO, not sure if we have a style for something like that

p0pr0ck5:
Admin API Audit Log - v1 3-x   Kong - Open-Source API Management and Microservice Management (1)

Although looking at this, it doesn't seem to severe- i expected the yellow to be sharper. i like this :)

PR #1804, opened 2020-02-21
Fix (Collector App) Fix Collector App docs to match 1.1.0

wongstein:
Two problems were fixed:
1. alerts/config description was incorrect
2. Setup instructions out of order and not in-depth enough.


h3xar0n:
Merge at will :shipit:

PR #1810, opened 2020-02-23
Invalid example, postgresql -> postgres

teunis90:
Example is invalid. Should be postgres, see for another example: https://github.com/Kong/kong/blob/2.0.1/kong/plugins/basic-auth/migrations/000_base_basic_auth.lua


hishamhm:
Yes, this PR is correct! Thanks @teunis90 for the fix and @h3xar0n for the ping!

teunis90:
Thanks for accepting.

PR #1817, opened 2020-02-26
redirect k8s sidecar injector to kuma

h3xar0n:
K8S Sidecar Injector is not supported, so we should remove the page and hub listing. Instead, we'll (re)direct users to Kuma.


hbagdi:
I'm not the owner of the plugin but this should go away and looks good.

h3xar0n:
I'm going to create a redirect instead of deleting it outright.

PR #1823, opened 2020-02-27
add grpc-web plugin to Kong Hub

javierguerragiraldez:


app/_hub/kong-inc/grpc-web/index.md
```diff
@@ -0,0 +1,158 @@
+---
+name: gRPC-Web
+publisher: Kong Inc.
+
+categories:
+  - transformations
+
+type: plugin
+
+desc: Allow browser clients to call gRCP services.
```
hishamhm:
```suggestion
desc: Allow browser clients to call gRPC services.
```

app/_hub/kong-inc/grpc-web/index.md
```diff
@@ -0,0 +1,158 @@
...
+
+## Purpose
+
+A service that presents a gRPC API can be used by clients written in many languages, but the network specifications are oriented primarily to connections within a datacenter.  In order to expose the API to the Internet, and to be called from brower-based JS apps, [gRPC-Web] was developed.
+
+This plugin translates requests and responses between gRPC-Web and "real" gRPC.  Supports both HTTP/1.1 and HTTP/2, over plaintext (HTTP) and TLS (HTTPS) connections.
+
+## Usage
+
+This plugin is intended to be used in a Kong route between a gRPC service and an HTTP endpoint.
```
hishamhm:
```suggestion
This plugin is intended to be used in a Kong route between an HTTP endpoint and a gRPC service.
```

Very minor/stylistic, but I think ordering by "source, then destination" makes it easier to read/understand.

app/_hub/kong-inc/grpc-web/index.md
```diff
@@ -0,0 +1,158 @@
...
+  routes:
+  - protocols:
+    - http
+    paths:
+    - /
+    plugins:
+    - name: grpc-web
+```
+
+Same thing via the administation API:
```
hishamhm:
```suggestion
Same thing via the Admin API:
```

(I think that's the nomenclature used by the docs style guide.)

app/_hub/kong-inc/grpc-web/index.md
```diff
@@ -0,0 +1,158 @@
...
+  - protocols:
+    - http
+    paths:
+    - /
+    plugins:
+    - name: grpc-web
+      proto: path/to/hello.proto
+```
+
+or via the administration API:
```
hishamhm:
```suggestion
or via the Admin API:
```

app/_hub/kong-inc/grpc-web/index.md
```diff
@@ -0,0 +1,158 @@
...
+  --data paths=/
+
+$ # add the plugin to the route
+$ curl -XPOST localhost:8001/routes/web-service/plugins \
+  --data name=grpc-web
+```
+
+In these examples we don't set any configuration for the plugin.  This minimal setup works for the default varieties of the [gRPC-Web protocol], which use ProtocolBuffer messages either directly in binary or with base64 encoding.  The related `Content-Type` headers are `application/grpc-web` or `application/grpc-web+proto` for binary and `application/grpc-web-text` or `application/grpc-web-text+proto`.
+
+If we wish to use JSON encoding, we have to provide the gRPC specification in a .proto file.  For example:
```
hishamhm:
```suggestion
If we wish to use JSON encoding, we have to provide the gRPC specification in a .proto file, which needs to be installed in the Kong node running the plugin.  For example:
```

^ is the clarification above right?

app/_hub/kong-inc/grpc-web/index.md
```diff
@@ -0,0 +1,158 @@
...
+  host: localhost
+  port: 9000
+  routes:
+  - protocols:
+    - http
+    paths:
+    - /
+    plugins:
+    - name: grpc-web
+      proto: path/to/hello.proto
```
hishamhm: A question that popped to my mind, which could probably be answered in the documentation: does the path need to be absolute or can in be relative? (if so, relative to what?)

app/_hub/kong-inc/grpc-web/index.md
```diff
@@ -0,0 +1,158 @@
...
+  --data proto=path/to/hello.proto
+```
+
+With this setup, we can support gRPC-Web/JSON clients using `Content-Type` headers like `application/grpc-web+json` or `application/grpc-web-text+json`.
+
+Additionaly, non-gRPC-Web clients can do simple POST requests with `Content-Type: application/json`.  In this case, the data payloads are simple JSON objects, without gPRC frames.  This allows clients to perform calls without using the gRPC-Web library, but streaming responses are delivered as a continous stream of JSON objects.  It's up to the client to split into separate objects if it has to support multiple response messages.
+
+## Dependencies
+
+The gRPC-Web plugin depends on [lua-protobuf], [lua-cjson] and [lua-pack]
```
hishamhm: Kong Hub docs don't list dependencies. These are handled by LuaRocks, so they're basically transparent to the user as they install the plugin. :grin:

javierguerragiraldez: done

app/_hub/kong-inc/grpc-web/index.md
javierguerragiraldez: good point.  I'll add some path handling code to make it similar to what other plugins assume

app/_hub/kong-inc/grpc-web/index.md
```diff
@@ -0,0 +1,154 @@
...
+      description: |
+        If present, describes the gRPC types and methods.
+        Required to support payload transcoding.  When absent, the
+        web client must use application/grpw-web+proto content.
+
+---
+
+## Purpose
+
+A service that presents a gRPC API can be used by clients written in many languages, but the network specifications are oriented primarily to connections within a datacenter.  In order to expose the API to the Internet, and to be called from brower-based JS apps, [gRPC-Web] was developed.
```
kikito: That last phrase is a bit Yoda.

```suggestion
A service that presents a gRPC API can be used by clients written in many languages, but the network specifications are oriented primarily to connections within a datacenter. [gRPC-Web] allows exposing such API to the Internet, and be consumed by brower-based JS apps. 
```

app/_hub/kong-inc/grpc-web/index.md
```diff
@@ -0,0 +1,154 @@
...
+
+This plugin translates requests and responses between gRPC-Web and "real" gRPC.  Supports both HTTP/1.1 and HTTP/2, over plaintext (HTTP) and TLS (HTTPS) connections.
```
kikito: Adding links to protocols.

```suggestion
This plugin translates requests and responses between [gRPC-Web](https://github.com/grpc/grpc-web) and ["real" gRPC](https://github.com/grpc/grpc).  Supports both HTTP/1.1 and HTTP/2, over plaintext (HTTP) and TLS (HTTPS) connections.
```

app/_hub/kong-inc/grpc-web/index.md
```diff
@@ -0,0 +1,154 @@
...
+
+## Usage
+
+This plugin is intended to be used in a Kong route between an HTTP endpoint and a gRPC service.
```
kikito:
```suggestion
This plugin is intended to be used in a Kong Route between an HTTP endpoint and a gRPC service.
```

app/_hub/kong-inc/grpc-web/index.md
```diff
@@ -0,0 +1,154 @@
...
+```bash
+$ # add the plugin to the route
+$ curl -XPOST localhost:8001/routes/web-service/plugins \
+  --data name=grpc-web \
+  --data proto=path/to/hello.proto
+```
+
+With this setup, we can support gRPC-Web/JSON clients using `Content-Type` headers like `application/grpc-web+json` or `application/grpc-web-text+json`.
+
+Additionaly, non-gRPC-Web clients can do simple POST requests with `Content-Type: application/json`.  In this case, the data payloads are simple JSON objects, without gPRC frames.  This allows clients to perform calls without using the gRPC-Web library, but streaming responses are delivered as a continous stream of JSON objects.  It's up to the client to split into separate objects if it has to support multiple response messages.
```
kikito: I don't understand what this paragraph means. When a user does a POST with `Content-Type: application/json`, what happens?

> the data payloads are simple JSON objects, without gPRC frames

What does "data payloads" mean here? Is it the consumer -> kong request, the kong -> upstream request, the upstream -> kong response, the kong -> user response, or all of them? My best understanding of this (with low trust) is "the JSON request is proxied verbatim to the upstream". Why? Is it just that the plugin isn't activated by this calls (so they are proxied transparently)?

> streaming responses are delivered as a continous stream of JSON objects

Does this mean that if I send a (non-html) TCP request the plugin is not activated and the TCP input will be proxied to the upstream without any transformation? Or I am not understanding something?

app/_hub/kong-inc/grpc-web/index.md
javierguerragiraldez: really?

no objections, but is this common use?

app/_hub/kong-inc/grpc-web/index.md
javierguerragiraldez: not sure where the confusion comes from.  Let me try to write this paragraph again.

app/_hub/kong-inc/grpc-web/index.md
hishamhm: ah, yeah, the docs style guide says that _entity_ names should be uppercase. It does end up looking odd at times, I agree, but generally it avoids confusion.

app/_hub/kong-inc/grpc-web/index.md
```diff
@@ -0,0 +1,158 @@
...
+      description: |
+        If present, describes the gRPC types and methods.
+        Required to support payload transcoding.  When absent, the
+        web client must use application/grpw-web+proto content.
+
+---
+
+## Purpose
+
+A service that presents a gRPC API can be used by clients written in many languages, but the network specifications are oriented primarily to connections within a datacenter. [gRPC-Web] allows exposing such API to the Internet, and be consumed by brower-based JS apps.
```
hishamhm:
```suggestion
A service that presents a gRPC API can be used by clients written in many languages, but the network specifications are oriented primarily to connections within a datacenter. [gRPC-Web] allows exposing such API to the Internet so that it can be consumed by browser-based JS apps.
```

I was going to suggest for the typo alone, but I ended up tweaking the sentence too :)

lena-larionova:
```suggestion
A service that presents a gRPC API can be used by clients written in many languages, but the network specifications are oriented primarily to connections within a datacenter. [gRPC-Web] lets you expose this API to the Internet so that it can be consumed by browser-based JS apps.
```
Slight further tweak to @hishamhm 's suggestion.

app/_hub/kong-inc/grpc-web/index.md
```diff
@@ -0,0 +1,158 @@
...
+$ curl -XPOST localhost:8001/routes/web-service/plugins \
+  --data name=grpc-web \
+  --data proto=path/to/hello.proto
+```
+
+With this setup, we can support gRPC-Web/JSON clients using `Content-Type` headers like `application/grpc-web+json` or `application/grpc-web-text+json`.
+
+Note that, even using JSON encoding, the [gRPC-Web protocol] specifies that both request and response data consist of a series of frames, in a similar way to the full [gRPC protocol].  The [gRPC-Web] library performs this framing as expected.
+
+As an extension, this plugin also allows "naked" JSON requests with POST method and `Content-Type: application/json` header.  These requests will be encoded to ProtocolBuffer, framed and forwarded to the gRPC service.  The responses are transformed in the same way back, allowing any HTTP client to use a gRPC service without special libraries.  This feature is limited to unary (non-streaming) requests.  Streaming responses are encoded into multiple JSON objects; it's up to the client to split into separate records if it has to support multiple response messages.
```
lena-larionova:
```suggestion
As an extension, this plugin also allows "naked" JSON requests with POST method and `Content-Type: application/json` header.  These requests are encoded to ProtocolBuffer, framed, and forwarded to the gRPC service.  Likewise, the responses are transformed on the way back, allowing any HTTP client to use a gRPC service without special libraries.  This feature is limited to unary (non-streaming) requests.  Streaming responses are encoded into multiple JSON objects; it's up to the client to split into separate records if it has to support multiple response messages.
```
Minor rephrasing for easier reading.

app/_hub/kong-inc/grpc-web/index.md
```diff
@@ -0,0 +1,158 @@
+---
+name: gRPC-Web
+publisher: Kong Inc.
+
+categories:
+  - transformations
+
+type: plugin
+
+desc: Allow browser clients to call gRPC services.
```
dndx: This can be more accurate. Maybe "Allow HTTP clients (such as browsers) to access gRPC services" is better? The plugin can definitely be used by non-browsers as well.

app/_hub/kong-inc/grpc-web/index.md
```diff
@@ -0,0 +1,158 @@
...
+  name: grpc-web
+  route_id: true
+  protocols: ["http", "https"]
+  dbless_compatible: true
+
+  config:
+    - name: proto
+      required: false
+      default:
+      value_in_examples: path/to/hello.proto
```
dndx: Storing filesystem path in the database is not a common practice in Kong (say, expect a few very specific case such as the File Log plugin). Instead we store those file's content directly inside the database (like certificates). This makes deployment and config management a lot easier.

(I know this is not the best place for feature related comments, but it seems that the other PR has been closed already because of the new repo).

app/_hub/kong-inc/grpc-web/index.md
```diff
@@ -0,0 +1,158 @@
...
+
+## Purpose
+
+A service that presents a gRPC API can be used by clients written in many languages, but the network specifications are oriented primarily to connections within a datacenter. [gRPC-Web] allows exposing such API to the Internet, and be consumed by brower-based JS apps.
+
+This plugin translates requests and responses between [gRPC-Web] and ["real" gRPC](https://github.com/grpc/grpc).  Supports both HTTP/1.1 and HTTP/2, over plaintext (HTTP) and TLS (HTTPS) connections.
+
+## Usage
+
+This plugin is intended to be used in a Kong Route between an HTTP endpoint and a gRPC service.
```
dndx:
```suggestion
This plugin should be enabled on a Kong `Route` that serves the `http(s)` protocol but proxies to a `Service` with the `grpc(s)` protocol.
```

app/_hub/kong-inc/grpc-web/index.md
```diff
@@ -0,0 +1,158 @@
...
+  --data paths=/
+
+$ # add the plugin to the route
+$ curl -XPOST localhost:8001/routes/web-service/plugins \
+  --data name=grpc-web
+```
+
+In these examples we don't set any configuration for the plugin.  This minimal setup works for the default varieties of the [gRPC-Web protocol], which use ProtocolBuffer messages either directly in binary or with base64 encoding.  The related `Content-Type` headers are `application/grpc-web` or `application/grpc-web+proto` for binary and `application/grpc-web-text` or `application/grpc-web-text+proto`.
+
+If we wish to use JSON encoding, we have to provide the gRPC specification in a .proto file, which needs to be installed in the Kong node running the plugin.  For example:
```
dndx:
```suggestion
If we wish to use JSON encoding, we have to provide the gRPC specification in a `.proto` file, which needs to be installed in the Kong node running the plugin.  For example:
```

app/_hub/kong-inc/grpc-web/index.md
javierguerragiraldez: done

app/_hub/kong-inc/grpc-web/index.md
javierguerragiraldez: Sure, gRPC are already accessible by HTTP(2) clients.  But gRPC-Web was designed specifically around the limitations of JS' Fetch API, and the gRPC-Web library is mainly for browsers.  (I guess it can be used on Node.js, but not without some extra work).

This plugin's extension to use `application/json` content is non-standard and actually eschews the gRPC-Web protocol, which is why it has some extra limitations around streaming I/O.

app/_hub/kong-inc/grpc-web/index.md
javierguerragiraldez: It can certainly be done, since the `protoc.lua` compiler takes a string as readily as a file; but the usual workflow on gRPC is to share the same .proto file between client and server, usually pulling from a common repository that defines the protocol for both (or a monorepo?  it's mostly google, after all).
Copying the content into the config (or database) would add extra work to keep client and server in sync across protocol evolution.

app/_hub/kong-inc/grpc-web/index.md
javierguerragiraldez: Not sure how other file type extensions are shown in the docs.  Probably most are referred to by the actual acronym.  "HTML files", not ".html files", images are JPEG files, not .jpeg.  What about .js files?  Few users concern with .patch or .rst, so I guess there's not much precedence.

app/_hub/kong-inc/grpc-web/index.md
kikito: I agree that with "routes" it doesn't seem very useful. It makes more sense with more commonly-used names, like Services and Plugins. Kong can be a service in Kubernetes, and have many Services configured pointing to the other services.

app/_hub/kong-inc/grpc-web/index.md
javierguerragiraldez: very few examples of paths in configs, but the code already does the (almost) obvious thing: the current directory is the Kong prefix (`/usr/local/kong/' by default).   added to doc

hishamhm:
@Kong/team-docs from Core Team's side, this is good to squash and merge

PR #1825, opened 2020-02-28
Add secrets for both Kong registries to K8S documentation

rainest:
Add instructions to create a secret for the `kong-enterprise-edition-docker` registry in addition to the existing `kong-enterprise-k8s` secret. We have a mix of customers using either image in practice, and it's not immediately obvious that they require separate secrets.

Several other changes in content to align the docs site example more closely with the example in https://github.com/Kong/charts/pull/65


rainest:
This change is contingent on https://github.com/Kong/kubernetes-ingress-controller/pull/556, which changes these values in the standard manifests. Both should be applied at the same time to avoid creating a discrepancy between the controller manifests and the docs site.

PR #1839, opened 2020-03-05
(docs/enterprise) Adding reading time estimate to enterprise pages

lena-larionova:
Testing out an estimated reading time property for the enterprise docs. Can also add to community docs if reviewers find value in it. 

Usage: estimated reading time will appear on all enterprise doc pages by default. If you don't want it to show up, set `skip_read_time: true`. For example, I set it on the enterprise docs landing page and on the "choose your own adventure" installation page.

### DIFF
--- a/app/_assets/stylesheets/page.less
+++ b/app/_assets/stylesheets/page.less
@@ -249,6 +249,12 @@
     border-right: none;
     margin-right: 0;
   }
+
+  .reading-time {
+      font-style: italic;
+      display: block;
+      color: rgb(133, 137, 140);
+  }
 }
 
 .content {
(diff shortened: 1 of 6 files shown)


GayleNeumann:
@rainest Thanks for the input. FYI, we are reorganizing the entire navigation tree in the docs for the next release. I agree we need to break apart the "Guides and References" section appropriately, with a section for only reference guides. I also believe this tag for reading time estimate can be applied or not applied to topics, so am sure we'll apply appropriately for task-based topics and not entire reference guides.

lena-larionova:
@rainest @GayleNeumann I can exempt by directory, no problem, but you're right - our directory structure causes more annoying issues. But like Gayle said, we can manually set the property on any pages we want to exclude, so at least there's a temp solution.

PR #1842, opened 2020-03-09
doc(vitals-prometheus) add grafana dashboard for vitals prometheus

fffonion:
strategy


app/enterprise/0.34-x/vitals-prometheus-strategy.md
```diff
@@ -161,6 +161,16 @@ mappings:
 # ...
 ```
 
+### Install grafana dashboard
```
lena-larionova: Can we add a short intro here (and use it to point out that this is optional): 

"If you use Grafana, you can import the Kong Vitals Prometheus dashboard to visualize the data."

PR #1871, opened 2020-03-20
clarify helm 2 vs helm 3 install commands

hbagdi:


hbagdi:
@lena-larionova Yes. Those are updated in the `next` branch. There will be a chart release tomorrow (Wednesday) which will update those values in the `master` branch of that Github repo.
Good catch on making sure that all references are up to date!

PR #1883, opened 2020-03-25
doc(request-transformer) note about X-Forwarded-* headers

locao:

### DIFF
--- a/app/_hub/kong-inc/request-transformer/index.md
+++ b/app/_hub/kong-inc/request-transformer/index.md
@@ -116,9 +116,10 @@ params:
       required: false
       description: List of paramname:value pairs. If the content-type is one the following [`application/json`, `application/x-www-form-urlencoded`], add a new parameter with the given value if the parameter is not present, otherwise if it is already present, the two values (old and new) will be aggregated in an array.
   extra: |
-    Note: if the value contains a `,` then the comma-separated format for lists cannot be used. The array notation must be used instead.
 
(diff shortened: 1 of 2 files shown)


app/_hub/kong-inc/request-transformer/index.md
```diff
@@ -118,6 +118,8 @@ params:
   extra: |
     Note: if the value contains a `,` then the comma-separated format for lists cannot be used. The array notation must be used instead.
 
+    Note: the `X-Forwarded-*` are non-standard header fields written by Nginx to inform the upstream about client details and can't be overwritten by this plugin. If overwriting these header fields is needed, please see [post-function plugin in Serverless Functions](https://docs.konghq.com/hub/kong-inc/serverless-functions/).
```
lena-larionova: For readability, I think it would be better to combine the notes into one section, eg:

```suggestion
**Notes**: 
* If the value contains a `,` then the comma-separated format for lists cannot be used. The array notation must be used instead.
* The `X-Forwarded-*` fields are non-standard header fields written by Nginx to inform the upstream about client details and can't be overwritten by this plugin. If you need to overwrite these header fields, see the [post-function plugin in Serverless Functions](https://docs.konghq.com/hub/kong-inc/serverless-functions/).
```

lena-larionova:
Does this note also apply to the advanced version of this plugin?

app/_hub/kong-inc/request-transformer/index.md
locao: Thank you!

locao:
> Does this note also apply to the advanced version of this plugin?

Yes, it does. Just updated there, thank you for pointing that.

PR #1889, opened 2020-03-27
docs(admin-api) Content types support multipart.

baokong:
Added multipart as supported content type, and moved
JSON format to beginning of section.

### DIFF
--- a/app/2.0.x/admin-api.md
+++ b/app/2.0.x/admin-api.md
@@ -559,17 +559,7 @@ of methods to secure the Admin API.
 
 ## Supported Content Types
 
-The Admin API accepts 2 content types on every endpoint:
-
-- **application/x-www-form-urlencoded**
-
-Simple enough for basic request bodies, you will probably use it most of the time.
-Note that when sending nested values, Kong expects nested objects to be referenced
-with dotted keys. Example:
-
-```
-config.limit=10&config.period=seconds
-```
+The Admin API accepts 3 content types on every endpoint:
 
 - **application/json**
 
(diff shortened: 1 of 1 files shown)


app/2.0.x/admin-api.md
```diff
@@ -585,6 +575,27 @@ a JSON representation of the data you want to send. Example:
...
+
+
+- **multipart/form-data**
+
+Similar to url-encoded, you will use dotted keys to reference nested objects.
+
+```
+    -F "config.limit=10"
+    -F "config.period=seconds" 
+```
```
murillopaula: Since `-F` is a `curl` thing, should we provide a full `curl` example?

I think that we can also mention that `multipart/form-data` is used more generally to send requests with files and binary data.

We can even show an example of configuring the `pre-function` serverless plugin:

```
curl -i -X POST http://localhost:8001/services/plugin-testing/plugins \
     -F "name=pre-function" \
     -F "config.functions=@custom-auth.lua"
```

https://docs.konghq.com/hub/kong-inc/serverless-functions/#with-a-database

murillopaula: Congrats on your first commit @baokong !

bungle: Yes, or if you want you can show it as:
```
--------------------------bfdd9fe2d3e813bb
Content-Disposition: form-data; name="config.period"

seconds
--------------------------bfdd9fe2d3e813bb
Content-Disposition: form-data; name="config.limit"

10
--------------------------bfdd9fe2d3e813bb--
```

baokong: Yes, I will push another commit. Sorry I took so long, did not have notifications turned on for PR review.

app/2.0.x/admin-api.md
```diff
@@ -585,6 +575,29 @@ a JSON representation of the data you want to send. Example:
...
+
+```
+config.limit=10&config.period=seconds
+```
+
+
+- **multipart/form-data**
+
+Similar to url-encoded, you will use dotted keys to reference nested objects.
+Here is an example of sending a Lua file to the pre-function Kong plugin.
```
lena-larionova, line 593:
```suggestion
Here is an example of sending a Lua file to the pre-function Kong plugin:
```

app/2.0.x/admin-api.md
```diff
@@ -585,6 +575,29 @@ a JSON representation of the data you want to send. Example:
...
+with dotted keys. Example:
+
+```
+config.limit=10&config.period=seconds
+```
+
+
+- **multipart/form-data**
+
+Similar to url-encoded, you will use dotted keys to reference nested objects.
```
lena-larionova, line 592:
```suggestion
Similar to url-encoded, this content type uses dotted keys to reference nested objects.
```

PR #1890, opened 2020-03-27
need and use of the go compiler image

javierguerragiraldez:

### DIFF
--- a/app/2.0.x/go.md
+++ b/app/2.0.x/go.md
@@ -37,7 +37,7 @@ The extension strategy chosen for Kong is common in other languages: a plugin is
 This is a relatively recent feature in Golang, and has some rough edges in tooling and deployability.  In particular, it's essential that the loading executable (`go-pluginserver` in our case) and the plugins have exactly the same linking behaviour.  This involves at least:
 
 * The same version of any common library, including:
-    * our `Kong/go-pdk`.
+    * `Kong/go-pdk`
     * all standard Go libraries (like `fmt`, `rpc`, `reflect`, etc)
     * OS libraries, like `libpthread`, `libc`, `ld-xxxx`, etc.
 * the exact same version of the Go compiler.
--- a/app/2.0.x/go.md
+++ b/app/2.0.x/go.md
@@ -6,6 +6,13 @@ title: Go language support
 
 Kong 2.0 introduces support for plugins written in Go.
 
+Until now, Lua was the only language supported for writing Kong plugins, and continues to be the main way to develop and extend Kong.  The addition of Go plugin support allows Kong users to tap into the Go ecosystem.  For example, there are databases like MS-SQLServer that don't have good client libraries for Lua, but are well supported in Go.  A Go plugin could directly access such a server without having to pass through Kong's Lua code.
+
+The architecture chosen for Go plugins is to have a separate process written in 100% Go, called `go-pluginserver`.  If so configured, Kong launches this process and opens a communications channel to pass events and function calls between them.  This means that Go plugins run in a real Go environment and are able to use Go's features like goroutines, I/O, IPC, etc.  Note also that this means any call to a PDK function has to be transferred to the Kong process and back.
+
+Go plugins are built with the `-buildmode=plugin` flag, which allows the pluginserver to dynamically load them.  In order to comply with Go's strict linking compatibility checks, they have to be compiled with the `kong/go-plugin-tool` Docker image, as described below.
+
+
 ## Prerequisites
 
 To use Go plugins:
@@ -36,18 +43,18 @@ The extension strategy chosen for Kong is common in other languages: a plugin is
 
 This is a relatively recent feature in Golang, and has some rough edges in tooling and deployability.  In particular, it's essential that the loading executable (`go-pluginserver` in our case) and the plugins have exactly the same linking behaviour.  This involves at least:
 
-* The same version of any common library, including:
+* The same version of any common libraries, including:
     * `Kong/go-pdk`
     * all standard Go libraries (like `fmt`, `rpc`, `reflect`, etc)
     * OS libraries, like `libpthread`, `libc`, `ld-xxxx`, etc.
-* the exact same version of the Go compiler.
-* the same Go environment variables like `$GOROOT` and `$GOPATH`
+* The exact same version of the Go compiler.
+* The same Go environment variables like `$GOROOT` and `$GOPATH`
 
-The first point is somewhat alleviated with `go.mod` dependency management, but that introduces more complex issues on the environment variables requirement.
+The common libraries version compatibility is partially handled by the `go.mod` dependency management, but that introduces more complex issues on the environment variables requirement.
 
-The last point, `$GOPATH` is a real problem, not only because one of the recommended patterns is `$HOME/go`, which includes the developer's username in their own system, but also because it's common that production builds (Dockerfiles, build scripts, CI/CD systems) use a very different pattern.
+For example, the evironment variable `$GOPATH` is a real problem, not only because one of the recommended patterns is `$HOME/go`, which includes the developer's username in their own system, but also because it's common that production builds (Dockerfiles, build scripts, CI/CD systems) use a very different pattern.
 
-To guarantee this consistency, the `kong/go-plugin-tool` is used as a wrapper to the Go compiler.  Kong distribution packages and images use it to compile the included `go-pluginserver`.
+To guarantee consistency, the `kong/go-plugin-tool` is used as a wrapper to the Go compiler.  Kong distribution packages and images use it to compile the included `go-pluginserver`.
 
 ### Development process
 
--- a/app/2.0.x/go.md
+++ b/app/2.0.x/go.md
@@ -4,11 +4,10 @@ title: Go language support
 
 ## Introduction
 
-Kong 2.0 introduces support for plugins written in Go.
-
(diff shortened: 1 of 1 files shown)


app/2.0.x/go.md
```diff
@@ -22,19 +22,39 @@ Refer to the the plugin via its filename (without the `.so` suffix).
...
+The extension strategy chosen for Kong is one very common in other languages: a plugin is a dynamically loaded module.  To allow this, the compiler no longer produces fully static programs, now the executable and plugins depend on the system libraries.
+
+This is a relatively recent feature in Golang, and has some rough edges in tooling and deployability.  In particular, it's essential that the loading executable (`go-pluginserver` in our case) and the plugins have exactly the same linking behaviour.  This involves at least:
+
+* The same version of any common library, including:
+    * our `Kong/go-pdk`.
+    * all standard Go libraries (like `fmt`, `rpc`, `reflect`, etc)
+    * OS libraries, like `libpthread`, `libc`, `ld-xxxx`, etc.
+* the exact same version of the Go compiler.
+* the same Go environment variables like `$GOROOT` and `$GOPATH`
```
lena-larionova, line 42:
```suggestion
* Go environment variables set to the same values across all of the above (e.g. `$GOROOT` and `$GOPATH`)
```
Same as what? I put "all of the above" in the suggestion, please change if that's wrong.

app/2.0.x/go.md
```diff
@@ -22,19 +22,39 @@ Refer to the the plugin via its filename (without the `.so` suffix).
...
+
+The extension strategy chosen for Kong is one very common in other languages: a plugin is a dynamically loaded module.  To allow this, the compiler no longer produces fully static programs, now the executable and plugins depend on the system libraries.
+
+This is a relatively recent feature in Golang, and has some rough edges in tooling and deployability.  In particular, it's essential that the loading executable (`go-pluginserver` in our case) and the plugins have exactly the same linking behaviour.  This involves at least:
+
+* The same version of any common library, including:
+    * our `Kong/go-pdk`.
+    * all standard Go libraries (like `fmt`, `rpc`, `reflect`, etc)
+    * OS libraries, like `libpthread`, `libc`, `ld-xxxx`, etc.
+* the exact same version of the Go compiler.
```
lena-larionova, line 41:
```suggestion
* A matching version of the Go compiler
```

app/2.0.x/go.md
```diff
@@ -22,19 +22,39 @@ Refer to the the plugin via its filename (without the `.so` suffix).
...
+* the same Go environment variables like `$GOROOT` and `$GOPATH`
+
+The first point is somewhat alleviated with `go.mod` dependency management, but that introduces more complex issues on the environment variables requirement.
```
lena-larionova, line 44: Is this referring to line 37? I would be more explicit here, eg:
```suggestion
Common library versions are partially covered with `go.mod` dependency management, but that introduces more complex issues for the environment variables requirement.
```
What's the solution/recommendation? Are we just telling users that this is an existing problem and there's nothing they can do? Or are you implying that they shouldn't use go.mod dependency management?

app/2.0.x/go.md
```diff
@@ -22,19 +22,39 @@ Refer to the the plugin via its filename (without the `.so` suffix).
...
+
+The last point, `$GOPATH` is a real problem, not only because one of the recommended patterns is `$HOME/go`, which includes the developer's username in their own system, but also because it's common that production builds (Dockerfiles, build scripts, CI/CD systems) use a very different pattern.
```
lena-larionova, line 46: Try to avoid referring to the order of bullet points in a list, since they could change in subsequent doc edits:
```suggestion
The biggest problem is the environment variable `$GOPATH`, not only because one of the recommended patterns is `$HOME/go`, which includes the developer's username in their own system, but also because production builds (e.g. Dockerfiles, build scripts, CI/CD systems) commonly use a different pattern.
```

app/2.0.x/go.md
```diff
@@ -22,19 +22,39 @@ Refer to the the plugin via its filename (without the `.so` suffix).
...
+
+To guarantee this consistency, the `kong/go-plugin-tool` is used as a wrapper to the Go compiler.  Kong distribution packages and images use it to compile the included `go-pluginserver`.
```
lena-larionova, line 48:
```suggestion
To guarantee consistency, the `kong/go-plugin-tool` is used as a wrapper to the Go compiler. Kong distribution packages and images use it to compile the included `go-pluginserver`.
```

lena-larionova: Does this guarantee consistency among all of the versions and env settings listed above?

app/2.0.x/go.md
```diff
@@ -22,19 +22,39 @@ Refer to the the plugin via its filename (without the `.so` suffix).
...
+
+### Development process
+
 To write a Kong plugin in Go, you need to:
 
 * Define a structure type to hold configuration.
 * Write a `New()` function to create instances of your structure.
 * Add methods on that structure to handle events.
-* Compile like: `go build -buildmode=plugin`.
+* Compile like: `docker run --rm -v $(pwd):/plugins kong/go-plugin-tool:<version> build <source>`.
```
lena-larionova, line 57:
```suggestion
* Compile using: `docker run --rm -v $(pwd):/plugins kong/go-plugin-tool:<version> build <source>`.
```

lena-larionova:
A further general comment about the Go PDK: we also need to add some info on what value this PDK provides that differs from the Lua PDK. Other than someone already being familiar with one language over the over, why would a developer choose to write their plugins in Go instead?

app/2.0.x/go.md
```diff
@@ -22,19 +22,39 @@ Refer to the the plugin via its filename (without the `.so` suffix).
...
+
+The extension strategy chosen for Kong is one very common in other languages: a plugin is a dynamically loaded module.  To allow this, the compiler no longer produces fully static programs, now the executable and plugins depend on the system libraries.
+
+This is a relatively recent feature in Golang, and has some rough edges in tooling and deployability.  In particular, it's essential that the loading executable (`go-pluginserver` in our case) and the plugins have exactly the same linking behaviour.  This involves at least:
+
+* The same version of any common library, including:
```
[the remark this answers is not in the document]
javierguerragiraldez: a dynamically linked program (both the executable and the plugin) includes references to other libraries (enumerated here), written by the compiler.

The executable and plugin are compiled separately (and by different persons on different machines), but to be able to load the plugin, those references must be exactly identical.

thus, the libraries present when compiling the plugin must be exactly the same version as those present when the executable was compiled.

unlike other languages, there's no "matching version" (6.2.4b matches 6.2.3g, or any 6.2.xx).  they must be exactly the same.  (6.4.2b must be 6.4.2b and not 6.4.2a)

app/2.0.x/go.md
javierguerragiraldez: the `go.mod` feature is relatively recent and helps a lot for source-level compatibility; but it doesn't solve plugin library compatibility.  we do include .mod files, but it doesn't do any difference when Kong tries to load a user-compiled plugin.   That's why we're providing the whole compiling environment in this docker image.

app/2.0.x/go.md
javierguerragiraldez: i'd prefer not to say "the biggest problem", because it makes it sound like if only this could be solved (and it does seem easy; variables can be tweaked, after all), the rest would be easy.

maybe just "A big part of the problem is the environment variable...." ?

app/2.0.x/go.md
javierguerragiraldez: it gives the user the exact same compiler, libraries, and environment used to compile the pluginserver included in the main Kong images.

in fact, we're no longer installing a go compiler to build it, we use this image

javierguerragiraldez:
> why would a developer choose to write their plugins in Go instead?

this would be a very different document.  not sure if there was something like that somewhere else.

surprisingly many users are already using it in production, even with all those limitations and much lower performance.  language chauvinism is a huge force; otherwise, Node.js would be a short joke...

app/2.0.x/go.md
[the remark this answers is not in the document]
lena-larionova: In that case, just a slight rephrasing to "The same versions of any common libraries" will work.

app/2.0.x/go.md
lena-larionova: New suggestion based on comment above:
```suggestion
* The exact same version of the Go compiler
```

app/2.0.x/go.md
```diff
@@ -6,6 +6,13 @@ title: Go language support
 
 Kong 2.0 introduces support for plugins written in Go.
 
+Until now, Lua was the only language supported for writing Kong plugins, and continues to be the main way to develop and extend Kong.  The addition of Go plugin support allows Kong users to tap into the Go ecosystem.  For example, there are databases like MS-SQLServer that don't have good client libraries for Lua, but are well supported in Go.  A Go plugin could directly access such a server without having to pass through Kong's Lua code.
```
lena-larionova, line 9:
```suggestion
Until Kong 2.0, Lua was the only language supported for writing Kong plugins, and continues to be the main way to develop and extend Kong. The addition of Go plugin support allows Kong users to tap into the Go ecosystem. For example, there are databases like MS SQL Server that don't have good client libraries for Lua, but are well supported in Go. A Go plugin could directly access such a server without having to pass through Kong's Lua code.
```

app/2.0.x/go.md
```diff
@@ -6,6 +6,13 @@ title: Go language support
 
 Kong 2.0 introduces support for plugins written in Go.
```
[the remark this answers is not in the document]
javierguerragiraldez: done

lena-larionova:
Looks good! And apparently I approved yesterday instead of just commenting? It was a long day.

PR #1947, opened 2020-04-15
Image expand functionality

wladimiiir:
Image expand functionality


wladimiiir:
> I like it, but I think it also needs the ability to close the modal by clicking outside of the image and/or by hitting the `Esc` key. Right now you can only close by clicking on X.
> 
> Edit: @wladimiiir

I have added both possibilities to close the modal.

PR #1952, opened 2020-04-15
feat(changelog) Update to new site layout

lena-larionova:


lena-larionova:
Closing because we need a slightly different CSS for the changelog before it can be updated.

PR #1953, opened 2020-04-16
added correlation-id to plugin priority list

GayleNeumann:
added correlation-id to plugin priority list as 100001


lena-larionova:
This change isn't needed, it's covered with other priority list updates in https://github.com/Kong/docs.konghq.com/pull/1955

PR #1957, opened 2020-04-16
Copy code snippet functionality

wladimiiir:
Copy code snippet functionality

### DIFF
--- a/app/_assets/javascripts/app.js
+++ b/app/_assets/javascripts/app.js
@@ -524,7 +524,21 @@ $(function () {
     }
   })
 
-  // Copy code snippet support
+  /**
+   * Copy code snippet support
+   *
+   * Example:
+   *
+   * ```bash
+   * $ curl -X GET http://kong:8001/basic-auths
+   * ```
+   * {: .copy-code-snippet data-copy-code="curl -X GET http://kong:8001/basic-auths" }
+   *
+   * where:
+   * .copy-code-snippet - adds class specifying that the code block is copyable
+   * data-copy-code="{custom-code}" - (optional) can be used to specify {custom-code} to be copied (only single-line text is supported)
+   *
+   */
   const copyInput = $('<textarea id="copy-code-input"></textarea>')
   $('body').append(copyInput)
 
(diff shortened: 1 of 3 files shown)


app/_assets/javascripts/app.js
```diff
@@ -523,4 +523,53 @@ $(function () {
...
+    action.click(function () {
+      if ($('#copy-code-success-info').length > 0) {
+        return
+      }
+
+      copyInput.text(snippet.data('copy-code') || snippet.find('code').text())
+      copyInput.select()
+      document.execCommand('copy')
+
+      const successInfo = $('<div id="copy-code-success-info">Code has been copied to clipboard.</div>')
```
lena-larionova, line 558:
```suggestion
      const successInfo = $('<div id="copy-code-success-info">Copied to clipboard!</div>')
```

lena-larionova:
Tested this out with a couple of codeblocks, some comments:
* Got some weird behavior in blocks that have a scrollbar, copy button should stick to the edge of the box instead of moving: 
Screen Shot 2020-04-16 at 10 59 41 AM
* I think this should be enabled on all codeblocks by default, not have to be manually added per codeblock.

wladimiiir:
> Tested this out with a couple of codeblocks, some comments:
> 
> * I think this should be enabled on all codeblocks by default, not have to be manually added per codeblock. If anything, we should treat this like the right-hand ToC: on by default, ability to disable if needed.
> * Can you get it to strip out the leading `$` in codeblocks on copy?
> * With thinner blocks, where the text sits right at the top (see screenshot), the icon hovers partially over the text. I know that's unavoidable if the icon stays inside the block - potential solutions/ideas might be to give it a background and raise opacity ([example](https://user-images.githubusercontent.com/1285584/73194694-812c1780-4124-11ea-989c-64dbb9d3b6ab.png)), or to move it outside the block ([example](https://i.stack.imgur.com/08Ed5.png))?
> * Got some weird behavior in blocks that have a scrollbar, copy button should stick to the edge of the box instead of moving:
> 
> Screen Shot 2020-04-16 at 10 59 41 AM

I have adjusted the functionality according your notes as well as updated comment doc section on how to use it.
One point though - the issue with button being moved with can only be experienced locally as there is some issue with highlighter plugin not triggering properly on first start when starting local server and not wrapping the code blocks into its specific elements.

PR #1994, opened 2020-04-27
docs/Fix GraphQL adv rate limiting plugin errors and inconsistencies

tlblessing:
https://konghq.atlassian.net/browse/DOCS-998

Direct link to generated PR output:

https://deploy-preview-1994--kongdocs.netlify.app/hub/kong-inc/graphql-rate-limiting-advanced/

### DIFF
--- a/app/_hub/kong-inc/graphql-rate-limiting-advanced/index.md
+++ b/app/_hub/kong-inc/graphql-rate-limiting-advanced/index.md
@@ -376,12 +376,26 @@ http -f :8001/services/example/routes hosts=example.com
 
 ### Enable the plugin on the service
 
+Example using a single window:
+
 ```
-$ curl -i -X POST http://kong:8001/services/{service}/plugins \
-  --data name=graphql-rate-limiting-advanced \
-  --data config.limit=100,10000 \
-  --data config.window_size=60,3600 \
-  --data config.sync_rate=10
+$ curl -i -X POST http://kong-ip:8001/services/{service}/plugins \
+     --data name=graphql-rate-limiting-advanced \
+     --data config.limit=100 \
+     --data config.window_size=60 \
+     --data config.sync_rate=10
+```
+
+Example using multiple windows:
+
+```
+$ curl -i -X POST --url http://kong-ip:8001/services/{service}/plugins \
+     --data 'name=graphql-rate-limiting-advanced' \
+     --data 'config.window_size[]=60' \
+     --data 'config.window_size[]=3600' \
+     --data 'config.limit[]=10' \
+     --data 'config.limit[]=100' \
+     --data 'config.sync_rate=10'
 ```
 
 ### Decorate gql schema for costs
--- a/app/_hub/kong-inc/graphql-rate-limiting-advanced/index.md
+++ b/app/_hub/kong-inc/graphql-rate-limiting-advanced/index.md
@@ -385,7 +385,7 @@ $ curl -X POST http://kong:8001/services/example/routes \
 Example using a single time window:
 
 ```
-$ curl -i -X POST http://kong-ip:8001/services/{service}/plugins \
+$ curl -i -X POST http://kong:8001/services/{service}/plugins \
      --data name=graphql-rate-limiting-advanced \
      --data config.limit=100 \
      --data config.window_size=60 \
@@ -395,7 +395,7 @@ $ curl -i -X POST http://kong-ip:8001/services/{service}/plugins \
 Example using multiple time windows:
 
 ```
-$ curl -i -X POST --url http://kong-ip:8001/services/{service}/plugins \
+$ curl -i -X POST --url http://kong:8001/services/{service}/plugins \
      --data 'name=graphql-rate-limiting-advanced' \
      --data 'config.window_size[]=60' \
      --data 'config.window_size[]=3600' \
@@ -428,22 +428,22 @@ Cost decoration is available on the following routes:
 For example:
 
 ```
-$ curl -X POST http://kong-ip:8001/services/{service}/graphql-rate-limiting-advanced/costs \
+$ curl -X POST http://kong:8001/services/{service}/graphql-rate-limiting-advanced/costs \
   --data type_path="Query.allPeople" \
   --data mul_arguments="first"
 
 
-$ curl -X POST http://kong-ip:8001/services/{service}/graphql-rate-limiting-advanced/costs \
+$ curl -X POST http://kong:8001/services/{service}/graphql-rate-limiting-advanced/costs \
   --data type_path="Person.vehicleConnection" \
   --data mul_arguments="first"
   --data add_constant=42
 
-$ curl -X POST http://kong-ip:8001/services/{service}/graphql-rate-limiting-advanced/costs \
+$ curl -X POST http://kong:8001/services/{service}/graphql-rate-limiting-advanced/costs \
   --data type_path="Vehicle.filmConnection" \
   --data mul_arguments="first"
 
 
-$ curl -X POST http://kong-ip:8001/services/{service}/graphql-rate-limiting-advanced/costs \
+$ curl -X POST http://kong:8001/services/{service}/graphql-rate-limiting-advanced/costs \
   --data type_path="Film.characterConnection" \
   --data mul_arguments="first"
 ```
@@ -465,7 +465,7 @@ $ curl -X POST http://kong-ip:8001/services/{service}/graphql-rate-limiting-adva
 ### Changing the default strategy
 
 ```
-$ curl -X POST http://kong-ip:8001/services/{service}/plugins \
+$ curl -X POST http://kong:8001/services/{service}/plugins \
   --data name=graphql-rate-limiting-advanced \
   --data config.limit=100 \
   --data config.limit=10000 \
@@ -476,7 +476,7 @@ $ curl -X POST http://kong-ip:8001/services/{service}/plugins \
 ```
 
 ```
-$ curl -i -X PATCH http://kong-ip:8001/plugins/{plugin_id} \
+$ curl -i -X PATCH http://kong:8001/plugins/{plugin_id} \
   --data config.cost_strategy=node_quantifier
 ```
 
@@ -490,7 +490,7 @@ a cost higher than our set `max_cost`. By default it's set to 0, which means
 no limit.
 
 ```
-$ curl -X POST http://kong-ip:8001/services/{service}/plugins \
+$ curl -X POST http://kong:8001/services/{service}/plugins \
   --data name=graphql-rate-limiting-advanced \
   --data config.limit=100 \
   --data config.limit=10000 \
@@ -503,7 +503,7 @@ $ curl -X POST http://kong-ip:8001/services/{service}/plugins \
 If `max_cost` needs to be updated, one can accomplish this by doing the following:
 
 ```
-$ curl -X PATCH http://kong-ip:8001/plugins/{plugin_id} \
+$ curl -X PATCH http://kong:8001/plugins/{plugin_id} \
   --data config.max_cost=10000
 ```
 
@@ -519,7 +519,7 @@ For example, a `score_factor` of `0.01` will divide the costs by 100, meaning
 every cost unit represents 100 nodes.
 
 ```
-$ curl -X POST http://kong-ip:8001/services/{service}/plugins \
+$ curl -X POST http://kong:8001/services/{service}/plugins \
   --data name=graphql-rate-limiting-advanced \
   --data config.limit=100 \
   --data config.limit=10000 \
@@ -533,6 +533,6 @@ $ curl -X POST http://kong-ip:8001/services/{service}/plugins \
 If `score_factor` needs to be updated, one can accomplish this by doing the following:
 
 ```
-$ curl -i -X PATCH http://kong-ip:8001/plugins/{plugin_id} \
+$ curl -i -X PATCH http://kong:8001/plugins/{plugin_id} \
   --data config.score_factor=1
 ```
--- a/app/_hub/kong-inc/graphql-rate-limiting-advanced/index.md
+++ b/app/_hub/kong-inc/graphql-rate-limiting-advanced/index.md
@@ -56,7 +56,7 @@ params:
       default:
       value_in_examples:
       description: |
-        One or more window sizes to apply limit to (defined in seconds).
+        One or more window sizes to apply a limit to (defined in seconds).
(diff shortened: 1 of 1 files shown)


app/_hub/kong-inc/graphql-rate-limiting-advanced/index.md
```diff
@@ -47,13 +47,13 @@ params:
...
       description: |
-        One of more request per window to apply
+        One of more requests per window to apply.
     - name: window_size
       required:
       default:
       value_in_examples:
       description: |
-        One more more window sizes to apply (defined in seconds)
+        One of more window sizes to apply (defined in seconds).
```
lena-larionova, line 56:
```suggestion
        One or more window sizes to apply (defined in seconds).
```

app/_hub/kong-inc/graphql-rate-limiting-advanced/index.md
```diff
@@ -47,13 +47,13 @@ params:
       default:
       value_in_examples:
       description: |
-        One of more request per window to apply
+        One of more requests per window to apply.
```
lena-larionova, line 50:
```suggestion
        One or more requests per window to apply.
```

tlblessing:
> I'm assuming the "of" is a typo and should be "or" - contextually, "of" doesn't make sense. If that is right after all, then the descriptions need some rephrasing.

That was my first inclination but then I wasn't sure. Perhaps it should be something like "One of multiple ... "? @marcoam do you have some insight for us?

tlblessing:
> Hey @tlblessing / @lena-larionova - Looks a LOT better! Thank you for working on this :)
> 
> While we've made a ton of progress, I will still ask for additional changes. Specifically, the curl commands aren't working as they are using arrays, which don't seem to work now (they used to work before). When making calls like this:
> 
> ```
>   --data config.limit=10,100 \
>   --data config.window_size=60,3600 \
> ```
> 
> It complains. Take a look at the following conversation:
> https://kongstrong.slack.com/archives/C0DSXDXV1/p1588149697286100
> 
> So we have a couple of options:
> 
> 1. Rather than using an array (which is a more complicated example) we simply work with one time window. What that means is that we put something like this:
> 
> ```
>   --data config.limit=10 \
>   --data config.window_size=60 \
> ```
> 
> But as I do feel that having multiple windows is important to highlight, that would mean that we need to add a new section to just address multiple time windows and give examples.
> 
> 1. We can use stick to the multiple windows and use a command that looks like this:
> 
> ```
> curl -i -X POST --url http://kong-ip:8001/services/{service}/plugins --data 'name=graphql-rate-limiting-advanced' --data 'config.window_size[]=60' --data 'config.window_size[]=3600' --data 'config.limit[]=10' --data 'config.limit[]=100' --data 'config.sync_rate=10'
> ```
> 
> or another call that a curl expert can come up with :)
> 
> 1. We can use httpie, which is a lot cleaner but requires our user to have httpie. As we don't always use that across our site, I would recommend against this option.
> 
> On an somewhat unrelated note, note that the non-GraphQL curl command is also broken: https://docs.konghq.com/hub/kong-inc/rate-limiting-advanced/#notes. So when fixing this, I would suggest that we fix the aforementioned page.
> 
> Thanks!

@marcoam thank you for your comments. We can show single and multiple window examples. When @GayleNeumann, @lena-larionova, and I reviewed the backlog tickets for plugins yesterday, we encountered this ticket https://konghq.atlassian.net/browse/DOCS-441 concerning array syntax. So it looks like this array syntax [] update will need to be made across the board where applicable. 

image
image

I already merged the advanced rate limiting PR that also added the sentinel pw config but will log another ticket for that. This ticket is increasing in scope but that is fine, we need to fix this. Were you going to provide cURL examples instead of the HTTPie as you mentioned in our convo? Thank you again for your assistance.

tlblessing:
@marcoam thanks for the updated array syntax. I tested it locally and it works:

`➜  ~ curl -i -X POST --url http://localhost:8001/services/example_service/plugins \
   --data 'name=graphql-rate-limiting-advanced' \
   --data 'config.window_size[]=60' \
   --data 'config.window_size[]=3600' \
   --data 'config.limit[]=10' \
   --data 'config.limit[]=100' \
   --data 'config.sync_rate=10'
HTTP/1.1 201 Created
Date: Wed, 29 Apr 2020 17:17:22 GMT
Content-Type: application/json; charset=utf-8
Connection: keep-alive
Access-Control-Allow-Origin: http://localhost:8002
X-Kong-Admin-Request-ID: UyEEDdB7ykJeq825U37HAqQdwktth1OQ
Vary: Origin
Access-Control-Allow-Credentials: true
Server: kong/1.5.0.1-enterprise-edition
Content-Length: 792
X-Kong-Admin-Latency: 157

{"created_at":1588180642,"config":{"redis":{"host":null,"sentinel_master":null,"timeout":2000,"cluster_addresses":null,"sentinel_role":null,"database":0,"port":null,"sentinel_addresses":null,"password":null,"sentinel_password":null},"score_factor":1,"window_type":"sliding","hide_client_headers":false,"dictionary_name":"kong_rate_limiting_counters","window_size":[60,3600],"sync_rate":10,"cost_strategy":"default","strategy":"cluster","namespace":"u88PdLjYgOkEj4j1FmydM6t05aaeWW4s","limit":[10,100],"identifier":"consumer","max_cost":0},"id":"03809793-1242-47e3-a5b8-7a455176cc13","service":{"id":"ef3c597b-8259-4437-84d2-38f33d160cdc"},"name":"graphql-rate-limiting-advanced","protocols":["grpc","grpcs","http","https"],"enabled":true,"run_on":null,"consumer":null,"route":null,"tags":null}%`

marcoam:
Hey @tlblessing , that's great! 

Ideally, we should have a standard throughout our docs to try and make sure that we use one notation. I talked to one of our engineers (https://kongstrong.slack.com/archives/C0DSXDXV1/p1588149697286100) and he even improved my call even more, which I think would be a good standard to keep. I'll go through this page and modify the calls to ones that work :)

marcoam:
Hi @tlblessing - I have gone ahead and updated the curl commands. I tested them, but please, do you mind testing them on your side? Want to make sure they work. A couple of random comments to mention of things I noticed: 
1) It's hard to understand, from a reader's perspective, if you are guiding me through a specific example/use case (as we do in the section "Usage") or if it's just general guidance (as we do in "Limit query cost by using config.max_cost"). The reason why this is important is because the curl command would change based on this. If we are talking about a specific use case, we can refer to the service, route or plugin we're talking about. In other words, calls can be like: 
`curl -X POST http://kong:8001/services/example/plugins`
and not
`curl -X POST http://kong:8001/services/{service}/plugins`

Does that make sense? So I will leave it up to you on what to do and how to organize it

2) While these curl commands aren't the most concise, I didn't want to introduce a new format when all of our other pages are using this type of curl calls. 

Hope this helps!

tlblessing:
> I am okay if this is changed back to {service} depends on what we're trying to do: follow the use case that we're building on top of or just mention how to change the default strategy...

image

I think it is understood to replace the values as stated upfront in the topic

tlblessing:
> Hi @tlblessing - I have gone ahead and updated the curl commands. Here is the commit:
> 
> #[4f8cdb6](https://github.com/Kong/docs.konghq.com/commit/4f8cdb62c6e740664b42aaac27565c7e392fced1)
> 
> I tested them, but please, do you mind testing them on your side? Want to make sure they work. A couple of random comments to mention of things I noticed:
> 
> 1. It's hard to understand, from a reader's perspective, if you are guiding me through a specific example/use case (as we do in the section "Usage") or if it's just general guidance (as we do in "Limit query cost by using config.max_cost"). The reason why this is important is because the curl command would change based on this. If we are talking about a specific use case, we can refer to the service, route or plugin we're talking about. In other words, calls can be like:
>    `curl -X POST http://kong:8001/services/example/plugins`
>    and not
>    `curl -X POST http://kong:8001/services/{service}/plugins`
> 
> Does that make sense? So I will leave it up to you on what to do and how to organize it
@marcoam, @lena-larionova just gave Gayle and I a walkthrough on plugin templates and how it is generated. Docs above the usage section are autogenerated. So we should try to stick with that convention {service}. In the Usage area, it is more freeform (not generated) but I still think an obvious placeholder is preferable. I also noticed the template has kong rather than kong-ip like your new examples, so I should probably reset that back. 
image

app/_hub/kong-inc/graphql-rate-limiting-advanced/index.md
```diff
@@ -4,9 +4,10 @@ name: GraphQL Rate Limiting Advanced
 publisher: Kong Inc.
 version: 1.3-x
 
-desc: Provide rate-limiting for GraphQL queries.
+desc: Provide rate limiting for GraphQL queries.
```
lena-larionova, line 7:
```suggestion
desc: Provide rate limiting for GraphQL queries
```
This is the short description for the cards on the main hub page, they don't have periods

app/_hub/kong-inc/graphql-rate-limiting-advanced/index.md
```diff
@@ -150,18 +161,20 @@ params:
       default: sliding
       value_in_examples:
       description: |
-        This sets the time window to either `sliding` or `fixed`
+        This sets the time window to either `sliding` or `fixed`.
   extra: |
     > Note:  Redis configuration values are ignored if the `cluster` strategy is used.
```
lena-larionova, line 166: Shouldn't this note also be part of the list below?

app/_hub/kong-inc/graphql-rate-limiting-advanced/index.md
```diff
@@ -150,18 +161,20 @@ params:
...
+    **Notes:**
 
-    **Note: The `dictionary_name` directive was added to prevent the usage of the `kong` shared dictionary, which could lead to `no memory` errors**
+     * PostgreSQL 9.5+ is required when using the `cluster` strategy with `postgres` as the backing Kong cluster data store. This requirement varies from the PostgreSQL 9.4+ requirement as described in the <a href="/install/source">Kong Community Edition documentation</a>.
+
+     * The `dictionary_name` directive was added to prevent the usage of the `kong` shared dictionary, which could lead to `no memory` errors.
 
 ---
 
 **GraphQL Rate Limiting Advanced** is an extension of
```
lena-larionova, line 176: This description is a bit cleaner/more grammatically correct than the description at the top of the topic. Maybe use this up above? I also don't know why we'd need the description again down here.

app/_hub/kong-inc/graphql-rate-limiting-advanced/index.md
```diff
@@ -398,10 +431,24 @@ Cost decoration is available on the following routes:
...
+  --data add_constant=42
+
+$ curl -X POST http://kong:8001/services/{service}/graphql-rate-limiting-advanced/costs \
+  --data type_path="Vehicle.filmConnection" \
+  --data mul_arguments="first"
+
+
+$ curl -X POST http://kong:8001/services/{service}/graphql-rate-limiting-advanced/costs \
+  --data type_path="Film.characterConnection" \
+  --data mul_arguments="first"
```
lena-larionova, line 451: In this example and the examples above, this topic uses three different syntax styles:

`--data "name=example"\`
`--data name=graphql-rate-limiting-advanced \`
`--data 'name=graphql-rate-limiting-advanced' \`

Any chance we can be more consistent?

app/_hub/kong-inc/graphql-rate-limiting-advanced/index.md
```diff
@@ -436,49 +485,57 @@ $ curl -i -X PATCH http://kong:8001/plugins/{plugin_id} \
...
-$ curl -i -X PATCH http://kong:8001/plugins/{plugin_id} \
-  --data config.max_cost=5000
+$ curl -X PATCH http://kong:8001/plugins/{plugin_id} \
+  --data config.max_cost=10000
 ```
 
 ### Using `config.score_factor` to modify costs
 
-GraphQL query cost depend on multiple factors depending on our resolvers
+GraphQL query cost depends on multiple factors depending on our resolvers
```
lena-larionova, line 515:
```suggestion
GraphQL query cost depends on multiple factors based on our resolvers
```
Suggest changing the second instance of "depend" in this sentence - "based on", "derived from", etc?

app/_hub/kong-inc/graphql-rate-limiting-advanced/index.md
```diff
@@ -47,13 +50,13 @@ params:
...
       description: |
-        One of more request per window to apply
+        One or more requests-per-window limits to apply.
     - name: window_size
       required:
       default:
       value_in_examples:
       description: |
-        One more more window sizes to apply (defined in seconds)
+        One or more window sizes to apply limit to (defined in seconds).
```
lena-larionova, line 59: I'd suggest either saying "apply a limit to" or "apply `limit` to"

tlblessing: > I'd suggest either saying "apply a limit to" or "apply `limit` to"

went with "apply a limit to"

app/_hub/kong-inc/graphql-rate-limiting-advanced/index.md
tlblessing: > In this example and the examples above, this topic uses three different syntax styles:
> 
> `--data "name=example"\`
> `--data name=graphql-rate-limiting-advanced \`
> `--data 'name=graphql-rate-limiting-advanced' \`
> 
> Any chance we can be more consistent?

@lena-larionova yes there is certainly a chance to be more consistent. Defining a style guide for preferred/recommended syntax should be done. Especially for array syntax as I saw mentioned in other channels that we need recommended and consistent syntax. I'm going to leave this for now because we don't have the bandwidth to test everything and I've already spent a week and a half on this PR.

app/_hub/kong-inc/graphql-rate-limiting-advanced/index.md
tlblessing: > This description is a bit cleaner/more grammatically correct than the description at the top of the topic. Maybe use this up above? I also don't know why we'd need the description again down here.

All I did was split out the note and remove the bold. I'd rather leave it be for now in the interest of time.

app/_hub/kong-inc/graphql-rate-limiting-advanced/index.md
tlblessing: > Shouldn't this note also be part of the list below?

It was my impression that it is a separate note directly under the table and so I did not more it.

PR #2015, opened 2020-05-07
Add overview of Kong for Kubernetes deployment options

rainest:
This document provides an overview of the two Enterprise image options for Kong for Kubernetes geared towards helping users decide which option fits their use case. It compares the features available in both, explains how database-backed Kong for Kubernetes deployments differ from non-controller-managed deployments, and provides basic information about migrating between the deployment types.

PR #2019, opened 2020-05-09
docs(install) Instructions for installing Kong Enterprise on Kubernetes

lena-larionova:
*  Adding descriptions of packages to intro based on https://github.com/Kong/docs.konghq.com/pull/2015
* Splitting out instructions for `kong-kubernetes-install` into separate section
* Rearranging topic to fit new instructions

Previews:
https://deploy-preview-2019--kongdocs.netlify.app/enterprise/1.5.x/kong-for-kubernetes/install/
https://deploy-preview-2019--kongdocs.netlify.app/enterprise/1.5.x/kong-for-kubernetes/install-on-kubernetes/

### DIFF
--- a/app/enterprise/1.5.x/kong-for-kubernetes/install-on-kubernetes.md
+++ b/app/enterprise/1.5.x/kong-for-kubernetes/install-on-kubernetes.md
@@ -236,8 +236,11 @@ The steps in this section show you how to install Kong Enterprise on Kubernetes 
 
 2. Copy the IP address from the output, then edit the `values.yaml` file to add the following line under `env` section:
 
+    > **Note:** Do not use IPs with RBAC. If you want to use RBAC, you need to set
+    up a DNS hostname first, instead of directly specifying an IP.
+
     ```
-    admin_api_uri: <your-ip>
+    admin_api_uri: <your-DNSorIP>
     ```
 
 3. Clean up:
@@ -271,8 +274,11 @@ The steps in this section show you how to install Kong Enterprise on Kubernetes 
 
 2. Copy the IP address from the output, then edit the `values.yaml` file to add the following line under `env` section:
 
+    > **Note:** Do not use IPs with RBAC. If you want to use RBAC, you need to set
+    up a DNS hostname first, instead of directly specifying an IP.
+
     ```
-    admin_api_uri: <your-ip>
+    admin_api_uri: <your-DNSorIP>
     ```
 
 3. Clean up:
--- a/app/enterprise/1.5.x/kong-for-kubernetes/install-on-kubernetes.md
+++ b/app/enterprise/1.5.x/kong-for-kubernetes/install-on-kubernetes.md
@@ -166,9 +166,7 @@ In the following steps, replace `<your-password>` with a secure password.
     |`image.tag` | The Docker image tag you want to pull down, e.g. `"1.5.0.2-alpine"`. |
     |`ingressController.enabled` | Set to `true` if you want to use the Kong Ingress Controller, or `false` if you don't want to install it. |
 
-3. (Optional) If not using the Kong Ingress Controller, remove the entire section of `Ingress Controller parameters`.
-
-4. In the `Kong Enterprise` section, enable Kong Manager (`manager`) and Kong Dev Portal (`portal`).
+3. In the `Kong Enterprise` section, enable Kong Manager (`manager`) and Kong Dev Portal (`portal`).
 
     For example:
     ```
@@ -193,7 +191,7 @@ In the following steps, replace `<your-password>` with a secure password.
         servicePort: 8446
     ```
 
-5. Fill in the rest of the parameters as they fit your implementation. Use the comments in the sample file to guide you, and see the documentation on [Kong Enterprise parameters](https://github.com/Kong/charts/blob/master/charts/kong/README.md#kong-enterprise-parameters) for more details.
+4. Fill in the rest of the parameters as they fit your implementation. Use the comments in the sample file to guide you, and see the documentation on [Kong Enterprise parameters](https://github.com/Kong/charts/blob/master/charts/kong/README.md#kong-enterprise-parameters) for more details.
 
 ## Step 8. Deploy Kong Enterprise on Kubernetes
 The steps in this section show you how to install Kong Enterprise on Kubernetes using Helm.
--- a/app/enterprise/1.5.x/kong-for-kubernetes/install-on-kubernetes.md
+++ b/app/enterprise/1.5.x/kong-for-kubernetes/install-on-kubernetes.md
@@ -1,0 +1,307 @@
+---
+title: Installing Kong Enterprise on Kubernetes
+---
+
+## Introduction
+
+Kong Enterprise on Kubernetes is recommended for deployments that require features not supported by Kong for Kubernetes Enterprise. It supports all Kong Enterprise plugins and features, but can't run in DB-less mode. See [Kong for Kubernetes deployment options](/enterprise/{{page.kong_version}}/kong-for-kubernetes/deployment-options) for more information.
+
+You can use kubectl or OpenShift oc to configure Kong Enterprise on Kubernetes, then deploy it using Helm.
+
+<img src="https://doc-assets.konghq.com/kubernetes/Kong-Enterprise-on-Kubernetes.png">
+
+## Prerequisites
+Before starting installation, be sure you have the following:
+
+- **Kubernetes cluster with load balancer**: Kong is compatible with all distributions of Kubernetes. You can use a [Minikube](https://kubernetes.io/docs/setup/minikube/), [GKE](https://cloud.google.com/kubernetes-engine/), or [OpenShift](https://www.openshift.com/products/container-platform) cluster.
+- **kubectl or oc access**: You should have `kubectl` or `oc` (if working with OpenShift) installed and configured to communicate to your Kubernetes cluster.
+- A valid Kong Enterprise License:
+  * If you have a license, continue to [Set Up Kong Enterprise License](#set-up-kong-enterprise-license) below. If you need your license file information, contact Kong Support.
+  * If you need a license, request a trial license through our [Request Demo](https://konghq.com/request-demo/) page.
+  * Or, try out Kong for Kubernetes Enterprise using a live tutorial at [https://kubecon.konglabs.io/](https://kubecon.konglabs.io/)
+- Kong Enterprise Docker registry access on Bintray.
+- Helm installed.
+
+## Step 1. Provision a namespace
+
+In order to create the secrets for license and Docker registry access,
+first provision the `kong` namespace:
+
+{% navtabs %}
+{% navtab kubectl %}
+```bash
+$ kubectl create namespace kong
+```
+{% endnavtab %}
+{% navtab OpenShift oc %}
+```bash
+$ oc new-project kong
+```
+{% endnavtab %}
+{% endnavtabs %}
+
+## Step 2. Set Up Kong Enterprise license
+Running Kong Enterprise on Kubernetes requires a valid license.
+
+As part of the sign-up process for Kong Enterprise, you should have received a license file. If you do not have one, contact your Kong sales representative. Save the license file temporarily to disk with filename `license` (no file extension) and execute the following:
+
+> Note:
+* There is no `.json` extension in the `--from-file` parameter.
+* `-n kong` specifies the namespace in which you are deploying Kong for Kubernetes Enterprise. If you are deploying in a different namespace, change this value.
+
+{% navtabs %}
+{% navtab kubectl %}
+```
+$ kubectl create secret generic kong-enterprise-license --from-file=./license -n kong
+```
+{% endnavtab %}
+{% navtab OpenShift oc %}
+```
+$ oc create secret generic kong-enterprise-license --from-file=./license -n kong
+```
+{% endnavtab %}
+{% endnavtabs %}
+
+## Step 3. Set up Helm
+
+1. Add the Kong charts repository:
+    ```
+    $ helm repo add kong https://charts.konghq.com
+    ```
+2. Update Helm:
+    ```
+    $ helm repo update
+    ```
+
+## Step 4. Configure Kong Enterprise Docker registry access
+Set up Docker credentials to allow Kubernetes nodes to pull down the Kong Enterprise Docker image, which is hosted in a private repository. You receive credentials for the Kong Enterprise Docker image when you sign up for Kong Enterprise.
+
+{% navtabs %}
+{% navtab kubectl %}
+Set up the credentials:
+```
+$ kubectl create secret -n kong docker-registry kong-enterprise-edition-docker \
+    --docker-server=kong-docker-kong-enterprise-edition-docker.bintray.io \
+    --docker-username=<your-bintray-username> \
+    --docker-password=<your-bintray-api-key>
+```
+{% endnavtab %}
+{% navtab OpenShift oc %}
+Set up the credentials:
+```
+$ oc create secret -n kong docker-registry kong-enterprise-edition-docker \
+    --docker-server=kong-docker-kong-enterprise-edition-docker.bintray.io \      
+    --docker-username=<your-bintray-username> \
+    --docker-password=<your-bintray-api-key>
+```
+{% endnavtab %}
+{% endnavtabs %}
+
+## Step 5. Seed the Super Admin password
+{% navtabs %}
+{% navtab kubectl %}
+(Optional) Create a password for the super admin:
+```
+kubectl create secret generic kong-enterprise-superuser-password  -n kong --from-literal=password=<your-password>
+```
+Though not required, this is recommended if you want to use RBAC, as it cannot be done after initial setup.
+{% endnavtab %}
+{% navtab OpenShift oc %}
+(Optional) Create a password for the super admin:
+```
+oc create secret generic kong-enterprise-superuser-password -n kong --from-literal=password=<your-password>
+```
+Though not required, this is recommended if you want to use RBAC, as it cannot be done after initial setup.
+{% endnavtab %}
+{% endnavtabs %}
+
+## Step 6. Prepare the sessions plugin for Manager and Dev Portal
+In the following steps, replace `<your-password>` with a secure password.
+
+{% navtabs %}
+{% navtab kubectl %}
+
+1. Create a sessions config file for Kong Manager:
+    ```
+    $ echo '{"cookie_name":"admin_session","cookie_samesite":"off","secret":"<your-password>","cookie_secure":false,"storage":"kong"}' > admin_gui_session_conf
+    ```
+2. Create a sessions config file for Kong Dev Portal:
+    ```
+    $ echo '{"cookie_name":"portal_session","cookie_samesite":"off","secret":"<your-password>","cookie_secure":false,"storage":"kong"}' > portal_session_conf
+    ```
+3. Create secret:
+    ```
+    kubectl create secret generic kong-session-config -n kong --from-file=admin_gui_session_conf=<manager-config-filename> --from-file=portal_session_conf=<portal-config-filename>
+    ```
+
+{% endnavtab %}
+{% navtab OpenShift oc %}
+
+1. Create a sessions config file for Kong Manager:
+    ```
+    $ echo '{"cookie_name":"admin_session","cookie_samesite":"off","secret":"<your-password>","cookie_secure":false,"storage":"kong"}' > admin_gui_session_conf
+    ```
+2. Create a sessions config file for Kong Dev Portal:
+    ```
+    $ echo '{"cookie_name":"portal_session","cookie_samesite":"off","secret":"<your-password>","cookie_secure":false,"storage":"kong"}' > portal_session_conf
+    ```
+3. Create secret:
+    ```
+    $ oc create secret generic kong-session-config -n kong --from-file=admin_gui_session_conf=<manager-config-filename> --from-file=portal_session_conf=<portal-config-filename>
+    ```
+{% endnavtab %}
+{% endnavtabs %}
+## Step 7. Prepare Kong's configuration file
+
+1. Create a `values.yaml` file based on the sample in the [Kong charts repository](https://github.com/Kong/charts/blob/master/charts/kong/values.yaml). This file sets all the necessary parameters for your Kong environment.
+
+2. Minimally, for setting up Kong Enterprise on Kubernetes, you will need to set the following parameters:
+
+    Parameter | Value  
+    ----------|-------
+    `env.database` | `"postgres"` or `"cassandra"`
+    `env.password.valueFrom.secretKeyRef.name` | Name of secret that holds the super admin password. In the example above, this is set to `kong-enterprise-superuser-password`.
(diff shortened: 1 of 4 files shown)


app/enterprise/1.5.x/kong-for-kubernetes/install.md
```diff
@@ -158,13 +152,223 @@ $ export PROXY_IP=$(kubectl get -o jsonpath="{.status.loadBalancer.ingress[0].ip
...
+    {"cookie_name":"portal_session","cookie_samesite":"off","secret":"<your-password>","cookie_secure":false,"storage":"kong"}
+    ```
+
+3. Create secret:
+    ```
+    $ oc create secret generic kong-session-config -n kong --from-file=admin_gui_session_conf --from-file=portal_session_conf
+    ```
+{% endnavtab %}
+{% endnavtabs %}
+### Prepare the values.yaml
```
lena-larionova, line 249: @hbagdi is there any way we can add a sample `values.yaml` specific to this package to the kong charts repo? That would make this section far easier to follow.

lena-larionova:
@felderi can you look over the openshift instructions?

app/enterprise/1.5.x/kong-for-kubernetes/install.md
```diff
@@ -1,33 +1,32 @@
 ---
-title: Installing Kong for Kubernetes Enterprise
+title: Installing Kong Enterprise with Kubernetes
```
hbagdi, line 2: Please split out the page into two separate pages, else this page is going to cause more problems. I myself got confused and had to scroll up and down a few times.

app/enterprise/1.5.x/kong-for-kubernetes/install.md
```diff
@@ -1,33 +1,32 @@
...
 ---
 
 ## Introduction
-This installation topic guides you through installing and deploying Kong for Kubernetes Enterprise (K4K8S-Enterprise), then directs you to the documentation for configuring and using the product.
+Kong for Kubernetes provides two options for its Kong Enterprise container image:
 
- You can install Kong for Kubernetes Enterprise using YAML with kubectl, or with OpenShift oc. Other deployment options, such as using Helm Chart and Kustomize, will be available at a later time.
+* Kong for Kubernetes Enterprise (`kong-enterprise-k8s`): provides most Kong Enterprise plugins and runs without a database, but does not include other Kong Enterprise features (Kong Manager, Dev Portal, Vitals, etc).
```
hbagdi, line 8: `kong-enterprise-k8s` is a behind the scenes name. Please remove it from this stanza.

app/enterprise/1.5.x/kong-for-kubernetes/install.md
```diff
@@ -158,13 +152,223 @@ $ export PROXY_IP=$(kubectl get -o jsonpath="{.status.loadBalancer.ingress[0].ip
...
+{% endnavtabs %}
+### Enable the sessions plugin for Manager and Dev Portal
+In the following steps, replace `<your-password>` with a secure password.
+
+{% navtabs %}
+{% navtab kubectl %}
+
+1. Create a sessions config file for Kong Manager:
+    ```
+    $ cat > admin_gui_session_conf
```
hbagdi, line 216: This should be:
` $ cat admin_gui_session_conf`

app/enterprise/1.5.x/kong-for-kubernetes/install.md
```diff
@@ -158,13 +152,223 @@ $ export PROXY_IP=$(kubectl get -o jsonpath="{.status.loadBalancer.ingress[0].ip
...
+    {"cookie_name":"admin_session","cookie_samesite":"off","secret":"<your-password>","cookie_secure":false,"storage":"kong"}
+    ```
+2. Create a sessions config file for Kong Dev Portal:
+    ```
+    $ cat > portal_session_conf
```
hbagdi, line 221: same here

app/enterprise/1.5.x/kong-for-kubernetes/install.md
```diff
@@ -158,13 +152,223 @@ $ export PROXY_IP=$(kubectl get -o jsonpath="{.status.loadBalancer.ingress[0].ip
...
+    {"cookie_name":"portal_session","cookie_samesite":"off","secret":"<your-password>","cookie_secure":false,"storage":"kong"}
+    ```
+
+3. Create secret:
+    ```
+    $ kubectl create secret generic kong-session-config -n kong --from-file=admin_gui_session_conf --from-file=portal_session_conf
```
hbagdi, line 227: We have seen that a lot of users run into issues when executing this command because they won't have the files name correctly. They will often have `.json` suffix in the filename. To get around this, I recommend instead using the following form:

```
kubectl create secret generic db-user-pass --from-file=username=./username.txt --from-file=password=./password.txt
```

Notice how `--from-file` is of the form `--from-file=<key-name>=<filename>`. This removes the problem I mentioned above.

app/enterprise/1.5.x/kong-for-kubernetes/install.md
hbagdi: ping @rainest

app/enterprise/1.5.x/kong-for-kubernetes/install.md
```diff
@@ -158,13 +152,223 @@ $ export PROXY_IP=$(kubectl get -o jsonpath="{.status.loadBalancer.ingress[0].ip
...
+
+2. Minimally, for setting up Kong Enterprise on Kubernetes, you will need to set the following parameters:
+
+    Parameter | Value  
+    ----------|-------
+    `env.database` | `"postgres"` or `"cassandra"`
+    `env.password.valueFrom.secretKeyRef.name` | Name of secret that holds the super admin password. In the example above, this is set to `kong-enterprise-superuser-password`.
+    `env.password.valueFrom.secretKeyRef.key` | The type of secret key used for authentication. If you followed the default settings in the example above, this is `password`.
+    `image.repository` | The Docker repository. In this case, `kong-docker-kong-enterprise-edition-docker.bintray.io/kong-enterprise-edition`
+    `image.tag` | The Docker image tag you want to pull down, e.g. `"1.5.0.2-alpine"`
```
hbagdi, line 261: Would `latest` be a better substitute here to avoid cases where this goes out of sync for newer versions in future?

app/enterprise/1.5.x/kong-for-kubernetes/install.md
lena-larionova: Wouldn't this just print the contents of the file? If it doesn't exist yet, it just errors out.

hbagdi: I see what you meant. I was executing `cat > admin_gui_session_conf` while we want the user to execute the multi-line command. This _could_ confuse other users too.

if we are doing this in other places in the site, let's keep it as.
If not, then does it make sense to ask users to create two files themselves, and show only the content for them?

app/enterprise/1.5.x/kong-for-kubernetes/install.md
lena-larionova: It does get exposed in the bintray link later in the topic though, so users are aware of it. The new deployment topic also refers to this name (pr #2015 )

If I split this topic out anyway, there will be no need to refer to this name in the intro, but something to be aware of for the other topic.

app/enterprise/1.5.x/kong-for-kubernetes/install.md
rainest: echo allows this to use a oneliner if we prefer:
```
echo '{"cookie_name":"admin_session","cookie_samesite":"off","secret":"<your-password>","cookie_secure":false,"storage":"kong"}' > admin_gui_session_conf
```

app/enterprise/1.5.x/kong-for-kubernetes/install.md
rainest: No--we shouldn't recommend `latest` since it might end up in actual configuration and result in unintentionally spawning a new version after a pod restart. Having a possibly outdated value in an example isn't a huge concern for me, though we should strive to keep actual defaults current.

app/enterprise/1.5.x/kong-for-kubernetes/install.md
rainest: I think the split is the best way to go. There are two concerns we want to address:
* Users have difficulty understanding the differences between the products and knowing which image to specify. My doc is focused on comparing them and listing the images prominently.
* Users want a step-by-step guide for installing their chosen product, which this provides. Splitting this into two documents duplicates a lot (and I don't think Jekyll provides a great way to write and reuse the shared sections), but is probably best to reduce decision points.

rainest: Further complicating the matter is the controller-managed versus not split: after reading I realized that the new documentation actually refers to the latter. I'd lean towards ignoring that for the most part and focusing on the DB-less (Kong for Kubernetes Enterprise) and DB-backed (Kong for Kubernetes with Kong Enterprise), with DB-backed, no controller clusters (Kong Enterprise on Kubernetes) in either a minor note or separate article.

@hbagdi aside from the article name, the only item of concern at installation should be whether or not the controller is enabled, correct? As far as I know there aren't any other settings that differ; all the Manager/Portal/RBAC/etc. configuration should be the same, though we should instruct users to create and configure an RBAC user for the controller.

app/enterprise/1.5.x/kong-for-kubernetes/install.md
```diff
@@ -158,13 +152,223 @@ $ export PROXY_IP=$(kubectl get -o jsonpath="{.status.loadBalancer.ingress[0].ip
...
+    `ingressController.enabled` | `false`
```
rainest, line 262: See comment at https://github.com/Kong/docs.konghq.com/pull/2019/files#r423333189

app/enterprise/1.5.x/kong-for-kubernetes/install.md
rainest: We plan to add these for a variety of scenarios: https://trello.com/c/t9sSfTi4/378-provide-example-valuesyamls-for-common-configurations

@lena-larionova do you have examples ready to go? The only things I'd want for a PR is that everything is placed under an `example-values` directory (doesn't exist yet) and that it either installs successfully or installs successfully if users create all referenced secrets.

rainest: Assuming probably no given the current format in this document, and we needed to write some anyway. https://github.com/Kong/charts/tree/feat/example-values with the caveat that this is still a draft and I haven't confirmed that these do in fact work.

lena-larionova: @rainest Harry gave me very bare bones example that I used for testing this out - I tweaked it to make it work with this install topic, I can send you that version.

app/enterprise/1.5.x/kong-for-kubernetes/install.md
lena-larionova: Addressed this by referring to db-backed vs db-less in the topic introductions and linking off to the the deployment topic for further info.

app/enterprise/1.5.x/kong-for-kubernetes/install.md
lena-larionova: Didn't remove it, but instead made the "remove ingress controller" section optional and expanded this parameter description to let people choose.

lena-larionova:
Ignore the broken links in the latest build, it's linking to the deployment topic that doesn't exist yet.

app/enterprise/1.5.x/kong-for-kubernetes/install-on-kubernetes.md
```diff
@@ -0,0 +1,306 @@
...
+    |Parameter      | Value                         |
+    |---------------|-------------------------------|
+    |`env.database` | `"postgres"` or `"cassandra"` |
+    |`env.password.valueFrom.secretKeyRef.name` | Name of secret that holds the super admin password. In the example above, this is set to `kong-enterprise-superuser-password`. |
+    |`env.password.valueFrom.secretKeyRef.key` | The type of secret key used for authentication. If you followed the default settings in the example above, this is `password`. |
+    |`image.repository` | The Docker repository. In this case, `kong-docker-kong-enterprise-edition-docker.bintray.io/ kong-enterprise-edition`. |
+    |`image.tag` | The Docker image tag you want to pull down, e.g. `"1.5.0.2-alpine"`. |
+    |`ingressController.enabled` | Set to `true` if you want to use the Kong Ingress Controller, or `false` if you don't want to install it. |
+
+3. (Optional) If not using the Kong Ingress Controller, remove the entire section of `Ingress Controller parameters`.
```
rainest, line 169: Probably remove this step entirely: the toggle is covered already in (2) and does need to remain in place: removing content from values.yaml will use the default from the base values.yaml, which will enable the controller.

The rest of the section doesn't do anything if `ingressController.enabled: false`, so it's safe to leave in place--removing it is mostly cosmetic, and instructing users to remove it runs the risk that they'll remove the toggle setting also.

app/enterprise/1.5.x/kong-for-kubernetes/install-on-kubernetes.md
```diff
@@ -0,0 +1,306 @@
...
+    ```
+    $ kubectl get svc -n kong my-kong-kong-admin --output=jsonpath='{.status.loadBalancer.ingress[0].ip}'
+    ```
+
+2. Copy the IP address from the output, then edit the `values.yaml` file to add the following line under `env` section:
+
+    ```
+    admin_api_uri: <your-ip>
+    ```
+
```
rainest, line 244: This will likely cause issues with RBAC: there's no way to configure the admin session configuration such that it can set a cookie on one IP and use it on another. We should add a note here that users should set up DNS and update these settings to use hostnames at their earliest convenience, and that they must do so before enabling RBAC.

We should provide documentation to both encourage users to use hostnames and explain how they relate to one another in the context of Kong, but that's fairly involved and I think we're best off with a brief note here for now to get the doc released quickly. This cuts across both Kubernetes and non-Kubernetes installations, though on Kubernetes the Admin and Manager IPs will differ since they're separate Services (this isn't the case if you're accessing them through an Ingress, but in that case you'd be using hostnames+paths rather than IPs). Jira ticket DOCS-1054 describes that proposed document in more detail.

app/enterprise/1.5.x/kong-for-kubernetes/install-on-kubernetes.md
```diff
@@ -0,0 +1,306 @@
...
+{% navtab OpenShift oc %}
+1. Run:
+    ```
+    $ oc get svc -n kong my-kong-kong-admin --output=jsonpath='{.status.loadBalancer.ingress[0].ip}'
+    ```
+
+2. Copy the IP address from the output, then edit the `values.yaml` file to add the following line under `env` section:
+
+    ```
+    admin_api_uri: <your-ip>
```
rainest, line 277: Ditto comment from line 239: we should add a note to encourage users to switch this to a hostname, and explain that they must do so before enabling RBAC.

app/enterprise/1.5.x/kong-for-kubernetes/install.md
```diff
@@ -44,10 +38,10 @@ $ oc new-project kong
 {% endnavtab %}
 {% endnavtabs %}
 
-## Set Up Kong Enterprise License
-Running Kong Enterprise requires a valid license.
+## Step 2. Set Up Kong Enterprise License
+Running Kong for Kubernetes Enterprise requires a valid license.
 
-As part of sign up for Kong Enterprise, you should have received a license file. If you do not have one, contact your Kong sales representative. Save the license file temporarily to disk with filename `license` (no file extension) and execute the following:
+As part of the sign-up process for Kong Enterprise, you should have received a license file. If you do not have one, contact your Kong sales represenative. Save the license file temporarily to disk with filename `license` (no file extension) and execute the following:
```
rainest, line 44:
```suggestion
As part of the sign-up process for Kong Enterprise, you should have received a license file. If you do not have one, contact your Kong sales representative. Save the license file temporarily to disk with filename `license` (no file extension) and execute the following:
```

app/enterprise/1.5.x/kong-for-kubernetes/install.md
rainest: https://github.com/Kong/charts/pull/134 has the same, now properly tested.

app/enterprise/1.5.x/kong-for-kubernetes/install-on-kubernetes.md
lena-larionova: > We should add a note here that users should set up DNS and update these settings to use hostnames at their earliest convenience, and that they must do so before enabling RBAC.

So we could go one of two ways here: 
- Keep the instructions as they are and add a note as you suggest, saying that this won't work in production and they need to set up a DNS, update this setting again, then optionally enable RBAC, or;
- Change the steps to tell users to set up a DNS before doing this, which becomes something like:

2. Copy the IP address from the output and set up a DNS pointing to that IP. 
- This is something you do outside of Kong, right? So we won't be telling them how to do it.
3. Set `admin_api_uri` to point to <DNS(:port?)>

rainest: Correct, though for the second option there are a several settings that interact with one another and have to be chosen to meet CORS and cookie-sending requirements. That won't easily fit into a brief paragraph here, so I think the best option is to briefly mention that you definitely need DNS hostnames for RBAC now and write something more detailed later. It's not perfect, but it's not worse than the status quo.

app/enterprise/1.5.x/kong-for-kubernetes/install-on-kubernetes.md
```diff
@@ -0,0 +1,310 @@
...
+      type: LoadBalancer
+      http:
+        enabled: true
+        servicePort: 8003
+      tls:
+        enabled: true
+        servicePort: 8446
+    ```
+
+4. Fill in the rest of the parameters as they fit your implementation. Use the comments in the sample file to guide you, and see the documentation on [Kong Enterprise parameters](https://github.com/Kong/charts/blob/master/charts/kong/README.md#kong-enterprise-parameters) for more details.
```
tlblessing, line 194:
```suggestion
4. Fill in the rest of the parameters as appropriate for your implementation. Use the comments in the sample file to guide you, and see the documentation on [Kong Enterprise parameters](https://github.com/Kong/charts/blob/master/charts/kong/README.md#kong-enterprise-parameters) for more details.
```

app/enterprise/1.5.x/kong-for-kubernetes/install-on-kubernetes.md
```diff
@@ -0,0 +1,310 @@
...
+    ```
+    $ helm install  my-kong kong/kong --namespace kong --values ~/values.yaml
+    ```
+    This may take some time.
+
+2. Check pod status:
+    ```
+    $ kubectl get pods -n kong
+    ```
+    Once migrations are complete and the `my-kong-kong-<ID>` pod is running, move on to the next section.
```
tlblessing, line 212: Prefer "After"  to "Once" to avoid ambiguity. This is a former employers style guide thing. Global comment.

app/enterprise/1.5.x/kong-for-kubernetes/install-on-kubernetes.md
```diff
@@ -0,0 +1,310 @@
...
+    ```
+    $ helm install  my-kong kong/kong --namespace kong --values ~/values.yaml
+    ```
+    This may take some time.
+
+2. Check pod status:
+    ```
+    $ oc get pods -n kong
+    ```
+    Once migrations are complete and the `my-kong-kong-<ID>` pod is running, move on to the next section.
```
tlblessing, line 225:
```suggestion
    Once migrations are complete and the `my-kong-kong-<ID>` pod is running, continue to the next section.
```

app/enterprise/1.5.x/kong-for-kubernetes/install.md
```diff
@@ -66,43 +59,32 @@ $ oc create secret generic kong-enterprise-license --from-file=./license -n kong
...
+    --docker-username=<your-bintray-username> \
     --docker-password=<your-bintray-api-key>
 ```
 {% endnavtab %}
 {% endnavtabs %}
 
-For future reference, make a note of the namespace in which you are deploying Kong.
-Once these credentials are created, you are ready to deploy Kong Enterprise Ingress Controller.
-
+## Step 4. Deploy Kong for Kubernetes Enteprise
```
tlblessing, line 85:
```suggestion
## Step 4. Deploy Kong for Kubernetes Enterprise
```

app/enterprise/1.5.x/kong-for-kubernetes/install-on-kubernetes.md
```diff
@@ -0,0 +1,310 @@
...
+- A valid Kong Enterprise License:
+  * If you have a license, continue to [Set Up Kong Enterprise License](#step-2-set-up-kong-enterprise-license) below. If you need your license file information, contact Kong Support.
+  * If you need a license, request a trial license through our [Request Demo](https://konghq.com/request-demo/) page.
+  * Or, try out Kong for Kubernetes Enterprise using a live tutorial at [https://kubecon.konglabs.io/](https://kubecon.konglabs.io/)
+- Kong Enterprise Docker registry access on Bintray.
+- Helm installed.
+
+## Step 1. Provision a namespace
+
+In order to create the secrets for license and Docker registry access,
```
tlblessing, line 27: "In order" is superfluous verbiage per many style guides.

app/enterprise/1.5.x/kong-for-kubernetes/install-on-kubernetes.md
```diff
@@ -0,0 +1,310 @@
...
+kubectl create secret generic kong-enterprise-superuser-password  -n kong --from-literal=password=<your-password>
+```
+Though not required, this is recommended if you want to use RBAC, as it cannot be done after initial setup.
+{% endnavtab %}
+{% navtab OpenShift oc %}
+(Optional) Create a password for the super admin:
+```
+oc create secret generic kong-enterprise-superuser-password -n kong --from-literal=password=<your-password>
+```
+Though not required, this is recommended if you want to use RBAC, as it cannot be done after initial setup.
```
tlblessing, line 114: This seems like it should have an Important note tag for RBAC users

app/enterprise/1.5.x/kong-for-kubernetes/install-on-kubernetes.md
```diff
@@ -0,0 +1,310 @@
...
+{% endnavtab %}
+{% endnavtabs %}
+
+## Step 6. Prepare the sessions plugin for Manager and Dev Portal
```
tlblessing, line 118:
```suggestion
## Step 6. Prepare the sessions plugin for Kong Manager and Dev Portal
```

app/enterprise/1.5.x/kong-for-kubernetes/install.md
```diff
@@ -1,33 +1,26 @@
 ---
-title: Installing Kong for Kubernetes Enterprise
+title: Installing Kong for Kubernetes Enteprise
```
tlblessing, line 2:
```suggestion
title: Installing Kong for Kubernetes Enterprise
```

PR #2027, opened 2020-05-12
fix(build workflow) Testing out [skip ci]

lena-larionova:


lena-larionova:
> Hope it works https://github.community/t5/GitHub-Actions/GitHub-Actions-does-not-respect-skip-ci/td-p/42834

This is the exact reason that we have to add this line to to the build. That discussion is where I got this solution from :)

PR #2028, opened 2020-05-12
docs(guidelines) Updating contribution guidelines

lena-larionova:


lena-larionova:
This method didn't end up working.

PR #2043, opened 2020-05-18
Adding ability to stay on the same topic when navigating between versions

wladimiiir:
Adding ability to stay on the same topic when navigating between versions

### DIFF
--- a/app/_layouts/docs-v2.html
+++ b/app/_layouts/docs-v2.html
@@ -56,8 +56,10 @@ id: documentation
             </button>
             <ul class="dropdown-menu dropdown-menu-right" role="menu" aria-labelledby="version-dropdown">
               {% for ver in page.kong_versions reversed %}
+              {% assign version_page_url = page.url | replace: page.kong_version, ver.release %}
+              {% capture version_page_exists %}{% page_exists {{version_page_url}} %}{% endcapture %}
               <li {% if page.kong_version==ver.release %} class="active" {% endif %}>
-                <a href="/{%if page.edition == 'enterprise' %}enterprise/{% endif %}{{ver.release}}"
+                <a href="{% if version_page_exists == 'true' %}{{version_page_url}}{% else %}/{%if page.edition == 'enterprise' %}enterprise/{% endif %}{{ver.release}}{% endif %}"
                     {% if ver.release==page.kong_version %} class="active" {% endif %}
                 >
                   {{ver.release}}
--- a/app/_layouts/docs.html
+++ b/app/_layouts/docs.html
@@ -54,9 +54,12 @@ id: documentation
         </button>
         <ul class="dropdown-menu dropdown-menu-right" role="menu" aria-labelledby="version-dropdown">
           {% for ver in page.kong_versions reversed %}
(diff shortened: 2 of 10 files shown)


lena-larionova:
Had a thought, not sure if this is possible: is there any way this feature can handle redirects? 
For example, this topic existed in 1.3-x:

https://docs.konghq.com/enterprise/1.3-x/getting-started

If you change the version in the URL to: 
https://docs.konghq.com/enterprise/1.5.x/getting-started

...then it redirects you to 
https://docs.konghq.com/getting-started-guide/latest/overview/

With the version switch from the dropdown, however, if you take the original topic in 1.3-x and switch to 1.5.x, you're redirected to the landing page of Enterprise. Any way to add a condition to redirect to a given page if a predefined redirect exists?

wladimiiir:
> Had a thought, not sure if this is possible: is there any way this feature can handle redirects?
> For example, this topic existed in 1.3-x:
> 
> https://docs.konghq.com/enterprise/1.3-x/getting-started
> 
> If you change the version in the URL to:
> https://docs.konghq.com/enterprise/1.5.x/getting-started
> 
> ...then it redirects you to
> https://docs.konghq.com/getting-started-guide/latest/overview/
> 
> With the version switch from the dropdown, however, if you take the original topic in 1.3-x and switch to 1.5.x, you're redirected to the landing page of Enterprise. Any way to add a condition to redirect to a given page if a predefined redirect exists?

It checks if page on the same path (../getting-started) exists within the different version, so what you can do is to create `/enterprise/1.5.x/getting-started.md` file with redirect content:
```
---
redirect_to: /getting-started-guide/latest/overview/
---
```

lena-larionova:
> It checks if page on the same path (../getting-started) exists within the different version, so what you can do is to create /enterprise/1.5.x/getting-started.md file with redirect content

So what you're saying is that we can use redirect_to in the front matter, but not redirect_from. I'm guessing there's no way it can know what to check for redirect_from.

wladimiiir:
> > It checks if page on the same path (../getting-started) exists within the different version, so what you can do is to create /enterprise/1.5.x/getting-started.md file with redirect content
> 
> So what you're saying is that we can use redirect_to in the front matter, but not redirect_from. I'm guessing there's no way it can know what to check for redirect_from.

I am saying that this is the way how to achieve your requirement, I just probably didn't explain it good enough :)
The dropdown item is either linked to `/enterprise/1.5.x/getting-started`, if there such document exists or `/enterprise` if the document file does not exists. So if you create getting-started.md file in /enterprise/1.5.x/ directory it will try to load that. And if you just put redirect_to in the file, it will redirect to your desired path. Hopefully, that's more understandable, if not, we can have a call and I can try to explain better :)

wladimiiir:
@lena

lena-larionova:
> And if you just put redirect_to in the file, it will redirect to your desired path. Hopefully, that's more understandable, if not, we can have a call and I can try to explain better :)

This makes sense :) I was hoping that there would be a way to avoid using redirect_to, as that requires us to keep old files around, but this is fine.

PR #2051, opened 2020-05-19
Right TOC sidebar changes

wladimiiir:
- close/reopen right-hand nav on desktop
- add lines to right-nav tree for nested headings

### DIFF
--- a/app/_assets/javascripts/app.js
+++ b/app/_assets/javascripts/app.js
@@ -321,11 +321,17 @@ $(function () {
     $('.docs-toc > .collapse-toc').click(function () {
       $('.docs-toc').addClass('collapsed')
       $('.page-content-container').addClass('toc-collapsed')
+      setCookie('toc-collapsed', 'true')
     })
     $('.docs-toc > .expand-toc').click(function () {
       $('.docs-toc').removeClass('collapsed')
       $('.page-content-container').removeClass('toc-collapsed')
+      setCookie('toc-collapsed', 'false')
     })
+    if (getCookie('toc-collapsed') === 'true') {
+      $('.docs-toc').addClass('collapsed')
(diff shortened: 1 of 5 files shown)


lena-larionova:
Looks like there's a slight misalignment for the last topic in a group:
Screen Shot 2020-05-19 at 11 26 12 AM

wladimiiir:
Fixed.

lena-larionova:
Sorry, another post-approval comment - found an issue. There's a problem with pages that have the right-hand nav disabled (eg https://deploy-preview-2051--kongdocs.netlify.app/enterprise/1.5.x/admin-api/) - now they have no right-hand nav but the page can't be expanded to fill the screen.

I don't think that having it fill the screen by default would make sense though, because it could be visually jarring to switch between unexpanded pages with right nav/pages with no right nav and expanded by default. Any ideas?

wladimiiir:
> I don't think that having it fill the screen by default would make sense though, because it could be visually jarring to switch between unexpanded pages with right nav/pages with no right nav and expanded by default. Any ideas?

I personally don't think having pages with no TOC expanded by default would be a visual issue. Even at current state, the navigating between the pages isn't as seamless as it could be due to the way how Jekyll generates the sites (and there is not much it can be done here to improve it probably), so this would not make it worse IMHO.
But what bothered me more is that when you collapsed the TOC and switched to different page, it would reset to default (expanded) state, so I have implemented a way to store the expanded state, which could be also a solution to your issue described above. I am not sure if this is behavior you are OK with, so feel free to revert the last commit if it's not the right way how to do it.
Unfortunately, I don't really have any other ideas how to solve the issue, as you cannot really put any collapse/expand button on these pages (would be strange from UX point of view IMHO). I personally would go for having them expanded by default.

PR #2057, opened 2020-05-20
Custom table column widths in markdown

wladimiiir:


wladimiiir:
@lena-larionova
Let me know, if such solution is acceptable.
Unfortunately, I could not find a way how to extend markdown parser, so it would be able to parse such specific tags to html element styles.

Another way how to do it (which is already available) is to add <style> to markdown file so it would look like this:
```
<style>
table th:nth-of-type(1) {
    width: 25%;
}
table th:nth-of-type(2) {
    width: 100px;
}
</style>

| Concept/Feature	| Description	| OSS or Enterprise	|
|-----------------|-------------|-------------------|
```

PR #2058, opened 2020-05-20
docs(proxy) detail that strip_path is true by default

David-Lor:
From this issue: https://github.com/Kong/kong/issues/5928

strip_path is True by default, but the documentation is confusing about it, as it leads to thinking it may be False by default.

### DIFF
--- a/app/2.0.x/proxy.md
+++ b/app/2.0.x/proxy.md
@@ -636,8 +636,8 @@ and interpreted as a space ` ` by Kong's Admin API.
 #### The `strip_path` property
 
 It may be desirable to specify a path prefix to match a Route, but not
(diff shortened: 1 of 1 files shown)


app/2.0.x/proxy.md
```diff
@@ -649,7 +649,7 @@ property by configuring a Route like so:
 }
 ```
 
-Enabling this flag instructs Kong that when matching this Route, and proceeding
+Enabling this flag (**which is enabled by default**) instructs Kong that when matching this Route, and proceeding
```
tlblessing, line 652:
```suggestion
Enabling this flag (which is enabled by default) instructs Kong that when matching this Route, and proceeding
```

PR #2060, opened 2020-05-20
docs(plugin-mtls_auth) Added known issue for revocation_check_mode

GayleNeumann:
- Added known issue for param "revocation_check_mode", as default IGNORE_CA_ERROR" does not work in version 1.5.0.0 and later. 
- Workaround is to set value to "SKIP"
- Did not want to change default value to Skip, so opted for known issue note. 
- See https://konghq.atlassian.net/browse/FTI-1509
- Need to remove this note when issue is fixed.

### DIFF
--- a/app/_hub/kong-inc/mtls-auth/index.md
+++ b/app/_hub/kong-inc/mtls-auth/index.md
@@ -53,7 +53,9 @@ params:
     - name: revocation_check_mode
       default: "`IGNORE_CA_ERROR`"
       description: |
-        ***Known Issue: The default value `IGNORE_CA_ERROR` has a known issue in versions 1.5.0.0 and later. To workaround, manually set the value to `SKIP`***.  Controls client certificate revocation check behavior. Valid values are `SKIP`, `IGNORE_CA_ERROR` or `STRICT`. If set to `SKIP`, no revocation check will be performed. If set to `IGNORE_CA_ERROR`, the plugin will respect the revocation status when either OCSP or CRL URL is set, and will not fail on network issues. If set to `STRICT`, the plugin will only treat the certificate as valid when it's able to verify the revocation status, and a missing OCSP or CRL URL in the certificate or a failure to connect to the server will result in a revoked status. If both OCSP and CRL URL are set, the plugin always checks OCSP first, and will only check CRL URL if it can't communicate with the OCSP server.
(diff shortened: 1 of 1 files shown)


app/_hub/kong-inc/mtls-auth/index.md
```diff
@@ -53,7 +53,7 @@ params:
...
+        ***Known Issue: The default value `IGNORE_CA_ERROR` has a known issue in versions 1.5.0.0 and later. To workaround, manually set the value to `SKIP`***.  Controls client certificate revocation check behavior. Valid values are `SKIP`, `IGNORE_CA_ERROR` or `STRICT`. If set to `SKIP`, no revocation check will be performed. If set to `IGNORE_CA_ERROR`, the plugin will respect the revocation status when either OCSP or CRL URL is set, and will not fail on network issues. If set to `STRICT`, the plugin will only treat the certificate as valid when it's able to verify the revocation status, and a missing OCSP or CRL URL in the certificate or a failure to connect to the server will result in a revoked status. If both OCSP and CRL URL are set, the plugin always checks OCSP first, and will only check CRL URL if it can't communicate with the OCSP server.
```
tlblessing, line 56:
```suggestion
        ***Known Issue: The default value `IGNORE_CA_ERROR` has a known issue in versions 1.5.0.0 and later. To workaround, manually set the value to `SKIP`***.  Controls client certificate revocation check behavior. Valid values are `SKIP`, `IGNORE_CA_ERROR`, or `STRICT`. If set to `SKIP`, no revocation check will be performed. If set to `IGNORE_CA_ERROR`, the plugin will respect the revocation status when either OCSP or CRL URL is set, and will not fail on network issues. If set to `STRICT`, the plugin will only treat the certificate as valid when it's able to verify the revocation status, and a missing OCSP or CRL URL in the certificate or a failure to connect to the server will result in a revoked status. If both OCSP and CRL URL are set, the plugin always checks OCSP first, and will only check CRL URL if it can't communicate with the OCSP server.
```

tlblessing: Due to the runon text on one line number, git can't diff this in a granular manner. This is what I suggested:

image

tlblessing:
@GayleNeumann and @lena-larionova there seems to be some build errors 

image

I have an mTLS pr in progress as well ...

lena-larionova:
> @GayleNeumann and @lena-larionova there seems to be some build errors

These build errors don't cause the link checker to fail. These are links to downloadable files and the checker doesn't know how to parse them, but it's not actually a problem as far as I know.

PR #2067, opened 2020-05-21
mTLS plugin ca_certs generated example field incorrect

tlblessing:
https://konghq.atlassian.net/browse/DOCS-1047

See Docs ticket for more information on testing the cert array syntax.

New generated examples:

image

It is singular cert but more can be added with another --data line. There isn't much control over plugin template-generated examples. Testing other array syntax formats was unsuccessful.

Former generated example showing incorrect id in string array:

image

Direct review link:
https://deploy-preview-2067--kongdocs.netlify.app/hub/kong-inc/mtls-auth/


app/_hub/kong-inc/mtls-auth/index.md
```diff
@@ -38,9 +38,9 @@ params:
         Whether to match the subject name of the client-supplied certificate against consumer's `username` and/or `custom_id` attribute. If set to `[]` (the empty array) then auto-matching is disabled.
     - name: ca_certificates
       required: true
-      value_in_examples: '`[ { "id": "fdac360e-7b19-4ade-a553-6dd22937c82f" }, { "id": "aabc360e-7b19-5aab-1231-6da229a7b82f"} ]`'
+      value_in_examples: fdac360e-7b19-4ade-a553-6dd22937c82f
       description: |
-        List of "CA Certificates" object to use as Certificate Authorities (CA) when validating client certificate. At least one is required but can specify as many as needed. The value of this array comprises of primary keys for the "Certificate Authority" object.
+        List of "CA Certificates" strings to use as Certificate Authorities (CA) when validating a client certificate. At least one is required but you can specify as many as needed. The value of this array is comprised of primary keys for the "Certificate Authority" object.
```
hbagdi, line 43: > The value of this array is comprised of primary keys for the "Certificate Authority" object.

1. It might be helpful to mention that `id` is the primary key as some users will not be familiar with what the primary key is. It could be as simple as "comprised of primary keys (`id`)".
2. There is no object in Kong that is known by "Certificate Authority". I think the doc is trying to refer to the "CACertificate" object and we should refer to as is. (This is a nit but is an important one to avoid misusing this term in other places because of copy-pasta techniques).

tlblessing: @hbagdi thank you for your review and feedback! Inc. in commit d714ac2.

PR #2073, opened 2020-05-22
Scroll to top button

wladimiiir:


lena-larionova:
@kikito with this change, we won't need the "Back to top" links on the auto-generated pages.

PR #2077, opened 2020-05-24
docs(nav) Add missing navigation items to Admin API and Guides

cwurm:
I noticed there are a few documentation pages that do not appear in the navigation of the Enterprise docs. Some can be reached through links on other pages, but some only through the search. Some of these pages contain important information, e.g. about conflict resolution between workspaces.

This PR adds them to the navigation as far back as `0.35` (when it seems a lot of the docs were reorganized).

It also replaces the old `API(s)` with the new `Service(s)` on one of those pages.

### DIFF
--- a/app/_data/docs_nav_ee_0.35-x.yml
+++ b/app/_data/docs_nav_ee_0.35-x.yml
@@ -73,8 +73,54 @@
       url: /secure-admin-api
     - text: Plugin Development Guide
       url: /plugin-development
+      items:
+        - text: Introduction
+          url: /plugin-development
+        - text: File structure
+          url: /plugin-development/file-structure
+        - text: Implementing custom logic
+          url: /plugin-development/custom-logic
+        - text: Plugin configuration
+          url: /plugin-development/plugin-configuration
+        - text: Accessing the datastore
+          url: /plugin-development/access-the-datastore
+        - text: Storing custom entities
+          url: /plugin-development/custom-entities
+        - text: Caching custom entities
+          url: /plugin-development/entities-cache
+        - text: Extending the Admin API
+          url: /plugin-development/admin-api
+        - text: Writing tests
+          url: /plugin-development/tests
+        - text: (un)Installing your plugin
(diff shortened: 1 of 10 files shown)


cwurm:
It seems because these pages were unlinked they have not been included in the automatic broken link checking for a while. I pushed a change to fix a number of broken links in 1.3 and 1.5 on the RBAC Examples page.

app/_data/docs_nav_ee_0.35-x.yml
```diff
@@ -103,8 +149,18 @@
...
+      - text: Workspace Examples
+        url: /admin-api/workspaces/examples
   - text: RBAC Object routes
     url: /admin-api/rbac/reference
+    items:
+      - text: RBAC Reference
+        url: /admin-api/rbac/reference
+      - text: RBAC Examples
+        url: /admin-api/rbac/examples
   - text: Admins
```
lena-larionova, line 164: I think these were left out because they need updating (the easy fix in the interest of time? I don't know the full history), but it does make sense to bring them in since people still use the topics. An FYI as well: by necessity, this is probably a temp location in the nav, as we want to automate the Admin API guide. But we can revisit that then...

cwurm: I'm not sure if they're outdated, possibly (I didn't read them end-to-end). But they're extremely important - e.g. the `Workspace Examples` page is the only place that explains conflict resolution between workspaces, and the `RBAC Examples` page is the only place that shows actual use of the `Kong-Admin-Token` to authenticate to the Admin API.

I stumbled on both of these "hidden pages" more or less by accident when I was searching for these topics.

lena-larionova: I used the RBAC examples page to figure out RBAC for myself and did run into some places where it didn't work as described, so it's definitely something that needs another look. However, hiding these pages doesn't solve anything, and at least exposing them first is a step forward - I want to work on incremental improvements here going forward.

PR #2080, opened 2020-05-26
docs(ToC) Adding release notes to ToC

lena-larionova:

### DIFF
--- a/app/_data/docs_nav_ee_1.5.x.yml
+++ b/app/_data/docs_nav_ee_1.5.x.yml
@@ -12,8 +12,10 @@
   items:
     - text: Release Notes
       url: /release-notes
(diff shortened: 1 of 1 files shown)


app/_data/docs_nav_ee_1.5.x.yml
```diff
@@ -10,8 +10,12 @@
   icon: /assets/images/icons/documentation/icn-references-color.svg
   url: /release-notes
   items:
+    - text: Release Notes
+      url: /release-notes
     - text: Changelog
       url: /enterprise/changelog
+    - text: Version Support Policy
+      url: /support-policy
       absolute_url: true
```
lena-larionova, line 19: Line 19 is in the wrong place, it should be for Changelog.

PR #2100, opened 2020-06-03
docs(portal-oidc) Notes on redirect_uri and developer accounts

cwurm:
Adds two important notes to the OIDC configuration for the Dev Portal on developer accounts and the IdP Redirect URI.

Especially without the hint on the `redirect_uri` pretty much the only way to know the correct configuration is to inspect the actual network requests and notice what's happening under the hood. I'm still not quite happy with it (still easy to overlook but essential for it to work) - but this is the best that I can come up with right now.

Some smaller fixes for spelling, replacing `Yahoo` with `Okta` as the much more common IdP, and removing the very Google-specific note on client id and secret with a pointer to the OIDC plugin docs that do a more thorough job.

Much of this is likely applicable to versions prior to 1.5, but it would take me a while to test this on those versions so hoping to get this in for 1.5 at least.

### DIFF
--- a/app/enterprise/1.5.x/developer-portal/configuration/authentication/oidc.md
+++ b/app/enterprise/1.5.x/developer-portal/configuration/authentication/oidc.md
@@ -1,12 +1,12 @@
 ---
-title: Enable OpenId Connect in the Dev Portal
+title: Enable OpenID Connect in the Dev Portal
 ---
 
 ### Introduction
 
 The [OpenID Connect Plugin](/hub/kong-inc/openid-connect/) (OIDC)
 allows the Kong Developer Portal to hook into existing authentication setups using third-party
-*Identity Providers* (IdP) such as Google, Yahoo, Microsoft Azure AD, etc.
+*Identity Providers* (IdP) such as Google, Okta, Microsoft Azure AD, etc.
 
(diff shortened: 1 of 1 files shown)


app/enterprise/1.5.x/developer-portal/configuration/authentication/oidc.md
```diff
@@ -49,20 +53,23 @@ Provider:
...
   - `<CLIENT_ID>` - Client ID provided by IdP
-        * For Example, Google credentials can be found here:
-        https://console.cloud.google.com/projectselector/apis/credentials
   - `<CLIENT_SECRET>` - Client secret provided by IdP
 
+See the [documentation of the OpenID Connect plugin](/hub/kong-inc/openid-connect/)
+for more information.
+
+**Important:** The `redirect_uri` needs to be configured as an allowed URI in the IdP.
+If not set explicitly in the configuration object it defaults to
```
tlblessing, line 67:
```suggestion
If not set explicitly in the configuration object, the URI default is
```

tlblessing:
Direct review link: https://deploy-preview-2100--kongdocs.netlify.app/enterprise/1.5.x/developer-portal/configuration/authentication/oidc/

PR #2119, opened 2020-06-18
docs(*) use current properties names.

locao:


hutchic:
could you update this snippet so it's for 2.1.0 and (enterprise 2.1.0)

image

PR #2125, opened 2020-06-23
docs(ee-install) Updating links to 1.5 and minor fixes

lena-larionova:
* Updating examples to current main version
* Adding hyperlinks
* Typo fix


lena-larionova:
> LGTM @lena-larionova but can we use a version variable for this?

Unfortunately, no - it won't work in code blocks. Variables are pretty much only useful in plaintext markdown.

PR #2147, opened 2020-07-01
chore(kubernetes) make redirect point to latest version of the document

bungle:


app/install/kubernetes.md
```diff
@@ -1,3 +1,3 @@
 ---
-redirect_to: /2.0.x/kong-for-kubernetes/install
+redirect_to: /2.1.x/kong-for-kubernetes/install
```
gszr, line 2: could we replace `2.1.x` with `latest`? I think `latest` will always redirect to the latest version, then we don't need to keep updating these

PR #2153, opened 2020-07-01
fix(kubernetes) install redirect page

bungle:
### Summary

I tried to use variable in #2147 but it just caused link-checker fail. Here back to static one.


lena-larionova:
Would redirecting to "latest" instead of the version number work instead? Ie, /latest/kong-for-kubernetes/install/

PR #2164, opened 2020-07-08
Updated download link to latest decK version

svenwal:
Very outdated link to 0.4 before - changed to 1.1.0
Please automate that it is always the latest link!


lena-larionova:
@svenwal this page is going away very soon (hopefully by the end of the week) and is getting replaced with a full decK doc set (migrated from Harry's repo). The version is correct on that new page, but good call on automating.

PR #2166, opened 2020-07-08
docs(plugins) Fixing filtering for enterprise plugins

lena-larionova:
Based on slack thread in `#docs`, mainly:
> Our hub lists ACME as being bundled with Kong Enterprise https://docs.konghq.com/hub/ - but it is not in the package.

Also happens with grpc-web and grpc-gateway plugins.

Filter tweak fixes the Enterprise plugin selector. Test by clicking "Bundled with Kong Enterprise".

Preview: https://deploy-preview-2166--kongdocs.netlify.app/hub/

### DIFF
--- a/app/_hub/index.html
+++ b/app/_hub/index.html
@@ -35,7 +35,7 @@ body_id: page-hub
               </a>
             </li>
             <li class="nav-item">
-              <a class="nav-link" data-filter="pub-konghq">
+              <a class="nav-link" data-filter="ee-compat">
                 Bundled with Kong Enterprise
               </a>
             </li>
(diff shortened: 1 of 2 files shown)


tlblessing:
@lena-larionova question for you: does bundled with Kong Enterprise mean that those plugins should all have the Enterprise label in the upper right? Are some of these plugins missing that label?  Or is the filter still not ideal?

image

lena-larionova:
@tlblessing 

> does bundled with Kong Enterprise mean that those plugins should all have the Enterprise label in the upper right? Are some of these plugins missing that label? Or is the filter still not ideal? Or is this a superset? 

No, Kong Enterprise has both Community and Enterprise plugins bundled with it. The plugins labelled as "Enterprise" in the corner of the tile are Enterprise-only plugins.

> Could we also have an option for Enterprise only plugins?

I get where you're coming from, but I'm not sure that this makes sense from a user perspective. Think of it this way: *
* Community users want to see Community plugins, and the Enterprise label is actually for them to know what they _don't_ have access to at a glance.
* Enterprise users don't care about whether a plugin is Enterprise-only or Community but bundled with Enterprise; they just want to know that they have access to the plugin.

PR #2183, opened 2020-07-21
correct spelling of permission

GayleNeumann:
-- spelled as permisson, corrected to permission


lena-larionova:
@GayleNeumann should we fix this through our supported versions? Or at least for 1.5.x, since that's the "latest stable" release?

GayleNeumann:
Yes, good catch @Lena! We should update this correct spelling in previous releases, and going back to 1.5.x is probably sufficient.

PR #2195, opened 2020-07-28
Make sure the image points to the 2.1

ikeike443:
👋 
when pulling the docker image, the version should be explicitly set. Because currently, the `latest` tag points to 1.5.x, if you don't specify the version.


ikeike443:
OR should we bump up the latest in Bintray repo instead of updating the doc?

tlblessing:
> OR should we bump up the latest in Bintray repo instead of updating the doc?

That sounds better than adding onto doc maintenance. @lena-larionova?

lena-larionova:
> > OR should we bump up the latest in Bintray repo instead of updating the doc?
> 
> That sounds better than adding onto doc maintenance. @lena-larionova?

Yep, that would be ideal... I'm curious if they're purposely not doing that because it's a beta, though.

ikeike443:
>  I'm curious if they're purposely not doing that because it's a beta, though.

This is also what comes to my mind.

PR #2200, opened 2020-07-31
Document response phase.

javierguerragiraldez:
New `:response()` plugin method.

### DIFF
--- a/app/2.1.x/plugin-development/custom-logic.md
+++ b/app/2.1.x/plugin-development/custom-logic.md
@@ -38,7 +38,7 @@ of the execution life-cycle of Kong:
 | `:certificate()`   | [ssl_certificate] | Executed during the SSL certificate serving phase of the SSL handshake.
 | `:rewrite()`       | [rewrite]         | Executed for every request upon its reception from a client as a rewrite phase handler. *NOTE* in this phase neither the `Service` nor the `Consumer` have been identified, hence this handler will only be executed if the plugin was configured as a global plugin!
 | `:access()`        | [access]          | Executed for every request from a client and before it is being proxied to the upstream service.
-| `:response()` | [access] | Replaces both `header_filter()` and `body_filter()`.  Executed after the whole response has been received from the upstream service, but before sending any part of it to the client.
+| `:response()` | [access] | Replaces both `header_filter()` and `body_filter()`.  Executed after the entire response has been received from the Upstream service, but before sending any part of the response to the client.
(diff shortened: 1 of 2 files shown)


app/2.1.x/plugin-development/custom-logic.md
```diff
@@ -38,10 +38,17 @@ of the execution life-cycle of Kong:
 | `:certificate()`   | [ssl_certificate] | Executed during the SSL certificate serving phase of the SSL handshake.
 | `:rewrite()`       | [rewrite]         | Executed for every request upon its reception from a client as a rewrite phase handler. *NOTE* in this phase neither the `Service` nor the `Consumer` have been identified, hence this handler will only be executed if the plugin was configured as a global plugin!
 | `:access()`        | [access]          | Executed for every request from a client and before it is being proxied to the upstream service.
+| `:response()` | [access] | Replaces both `header_filter()` and `body_filter()`.  Executed after the entire response has been received from the Upstream service, but before sending any part of the response to the client.
```
lena-larionova, line 41: Should this link to a response phase doc?

javierguerragiraldez: the phases linked by the other methods are OpenResty phases.  there's no OpenResty "response" phase, the `:response()` method is managed during the OpenResty "access" phase, using the buffered proxy mechanism (that handles the upstream during access)

javierguerragiraldez:
> Does this change mean that there is no longer a limitation for Go plugins (ie, previously no header_filter or body_filter phases)?

Go plugins still can't define `header_filter` or `body_filter` methods (because those OpenResty phases are non-yielding).  But now they can define a `response` method.

lena-larionova:
> Go plugins still can't define `header_filter` or `body_filter` methods (because those OpenResty phases are non-yielding). But now they can define a `response` method.

What's the upside/downside of this approach for Go plugins? In the Enterprise doc, currently the lack of `header_filter` and `body_filter` is listed as a limitation; by having `response` instead, does it provide equal configurability, or is there a downside (eg, I can see "lack of flexibility" with these methods being a "limitation").

javierguerragiraldez:
not sure if there's any missing information.  _filter phases work but can't yield, so Go plugins can't define them.  we added the new `:response()`, which can yield so are usable from Go.   as it relies on buffering, it's turned on automatically and as noted, can't handle HTTP/2 upstreams.

for Go plugins it's not much of a pros/cons thing, since it's the only way to handle responses

PR #2201, opened 2020-08-03
docs(autodoc) move autodocs scripts to the kong/kong repo

kikito:
While this repo will remain the final destination of the generated files, the generators themselves will serve us better in the Kong repo. This PR removes them from this repo. A sister PR in the Kong repo will add them there (with slight adaptations).


kikito:
@lena-larionova 

> After the import into kong/kong, will you regenerate the docs once again to pick up the correct warning path?

We ended up merging the docs into our `next` branch, which means that these changes won't make it to this repo until Kong 2.2.x. Regenerating for 2.1.x is possible, but it would be difficult and it would not point to the right place anyway (generators will not land in the 2.1.x branch) so let's leave it as it is now.

PR #2209, opened 2020-08-07
docs(deck) rename master to main and hbagdi to kong

hbagdi:


lena-larionova:
@hbagdi should we also add some info on how to upgrade to the new location? I thought I saw a thread about this on slack. 

Or - is there any issue with leaving the instructions as is? I just tested with a `brew upgrade deck` and noticed it still pulled from hbagdi/deck but the detailed info links to https://github.com/kong/deck. Will this be disabled in the future?

hbagdi:
We can document on how to upgrade but the way I've set it up requires no user action. Existing and new users will get the same artifact always.
The reasoning here is that the following repos are kept in sync:
github.com/kong/homebrew-deck/
github.com/hbagdi/homebrew-deck/

PR #2239, opened 2020-08-19
docs(ee-hybrid) Remove incorrect kubernetes info and add note about manager

lena-larionova:
* Remove note about nodes not being visible in Kong Manager
* Move Kubernetes section out of Limitations into Platform Compatibility

Preview: https://deploy-preview-2239--kongdocs.netlify.app/enterprise/2.1.x/deployment/hybrid-mode/

### DIFF
--- a/app/enterprise/2.1.x/deployment/hybrid-mode.md
+++ b/app/enterprise/2.1.x/deployment/hybrid-mode.md
@@ -72,12 +72,12 @@ Plane in Hybrid mode.
 ### Kubernetes support
 
 [Kong Enterprise on Kubernetes](/enterprise/{{page.kong_version}}/kong-for-kubernetes/install-on-kubernetes)
-supports Hybrid mode deployments. However, for the 2.1.x Beta release, the nodes
-will not appear in Kong Manager.
+fully supports Hybrid mode deployments.
 
-Kong for Kubernetes Enterprise (K4K8s, or the Kong Ingress Controller with
-{{site.base_gateway}}) does not use a database and is therefore not supported in
-Hybrid mode.
+[Kong for Kubernetes Enterprise]((/enterprise/{{page.kong_version}}/kong-for-kubernetes/install) 
+(K4K8s, or the Kong Ingress Controller with {{site.base_gateway}}) supports Hybrid 
+mode with one limitation: Kong Manager displays configuration in read-only format 
+for DB-less mode.
 
 * For a comparison of these images, see [Kubernetes Deployment Options](/enterprise/{{page.kong_version}}/kong-for-kubernetes/deployment-options).
 * For the full Kubernetes Hybrid mode documentation, see [Hybrid mode](https://github.com/Kong/charts/blob/main/charts/kong/README.md#hybrid-mode)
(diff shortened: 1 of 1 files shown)


app/enterprise/2.1.x/deployment/hybrid-mode.md
```diff
@@ -72,12 +72,12 @@ Plane in Hybrid mode.
...
-will not appear in Kong Manager.
+fully supports Hybrid mode deployments.
 
-Kong for Kubernetes Enterprise (K4K8s, or the Kong Ingress Controller with
-{{site.base_gateway}}) does not use a database and is therefore not supported in
-Hybrid mode.
+[Kong for Kubernetes Enterprise]((/enterprise/{{page.kong_version}}/kong-for-kubernetes/install) 
+(K4K8s, or the Kong Ingress Controller with {{site.base_gateway}}) supports Hybrid 
+mode with one limitation: Kong Manager displays configuration in read-only format 
+for DB-less mode.
```
rainest, line 80: This is I guess something we field back to Product, but essentially this distinction is going away; IMO we can just say it fully supports Hybrid Mode (the previous block) and drop this second block entirely.

For Hybrid Mode deployments, Kong Manager supports everything, since there's a DB-backed node somewhere it can read from and write config to. It's kind of in a weird space since we discourage mixing manual and controller-managed configuration, but there are some things that are never controller-managed (e.g. Dev Portal config) where Hybrid Mode has full Manager support, and we expect users to use Manager for them.

For _not_ Hybrid Mode deployments, Kong Manager supports can provide a read-only view of configuration on fully DB-less deployments, but that should be out of scope for the Hybrid Mode docs.

rainest, line 80: This is I guess something we field back to Product, but essentially this distinction is going away; IMO we can just say it fully supports Hybrid Mode (the previous block) and drop this second block entirely.

For Hybrid Mode deployments, Kong Manager supports everything, since there's a DB-backed node somewhere it can read from and write config to. It's kind of in a weird space since we discourage mixing manual and controller-managed configuration, but there are some things that are never controller-managed (e.g. Dev Portal config) where Hybrid Mode has full Manager support, and we expect users to use Manager for them.

For not Hybrid Mode deployments, Kong Manager supports can provide a read-only view of configuration on fully DB-less deployments, but that shouldn't be discussed in the Hybrid Mode docs (fully DB-less deployments aren't Hybrid Mode).

lena-larionova: > We can just say it fully supports Hybrid Mode (the previous block) and drop this second block entirely.

I'll have to reposition this, then. This section is under a "limitations" heading - and if there are no longer any limitations, then it doesn't need to be here at all.

We could fully leave out any mention of Kubernetes on this page, but I think some people wouldn't be very happy with that - so I think it need to sit either under ["Benefits"](https://docs.konghq.com/enterprise/2.1.x/deployment/hybrid-mode/#benefits), or in a new "Supported Platforms" section, which will really just say "everything that Kong supports, Hybrid mode does too".

PR #2272, opened 2020-08-27
docs(brain-immunity) added overview heading

GayleNeumann:
-- added introduction heading so that right page toc doesn't land on brain.

### DIFF
--- a/app/enterprise/2.1.x/brain-immunity/overview.md
+++ b/app/enterprise/2.1.x/brain-immunity/overview.md
@@ -2,7 +2,7 @@
 title: Kong Brain and Kong Immunity Overview
 ---
 
-### Introduction
+## Overview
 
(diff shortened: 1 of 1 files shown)


app/enterprise/2.1.x/brain-immunity/overview.md
```diff
@@ -2,6 +2,8 @@
 title: Kong Brain and Kong Immunity Overview
 ---
 
+### Introduction
```
lena-larionova, line 5: This should probably be an H2, but then so should the brain and immunity headers.

GayleNeumann: I just changed this to Overview, to match Vitals, and also the heading levels to H2s, and H3s for sub-sections. Hope it's correct. If not, will change back!

PR #2273, opened 2020-08-28
Adding request to the curl and http validation step

felderi:
Adding request to the curl and http validation step to improve how the output is displayed.

### DIFF
--- a/app/getting-started-guide/2.1.x/expose-services.md
+++ b/app/getting-started-guide/2.1.x/expose-services.md
@@ -124,12 +124,12 @@ Using the Admin API, issue the following:
 
 *Using cURL*:
 ```
-$ curl -i -X GET http://<admin-hostname>:8000/mock
+$ curl -i -X GET http://<admin-hostname>:8000/mock/request
 ```
 
 *Or using HTTPie*:
 ```
-$ http :8000/mock
+$ http :8000/mock/request
 ```
 
 {% endnavtab %}


lena-larionova:
@felderi what does this change add? I tested it everywhere, the output seems to be the same with or without `/request`. I guess if you already have plugins enabled on that route, it might get messy? That shouldn't be the case with the getting started guide, since they're ostensibly getting started on Kong and have nothing else enabled yet.

If you think this is necessary after all - can you also edit the browser request path in line 140?

felderi:
> @felderi what does this change add? I tested it everywhere, the output seems to be the same with or without `/request`. I guess if you already have plugins enabled on that route, it might get messy? That shouldn't be the case with the getting started guide, since they're ostensibly getting started on Kong and have nothing else enabled yet.
> 
> If you think this is necessary after all - can you also edit the browser request path in line 140?

Without the /request, the response is the content of the website mockbin.org which is messy.  When you add the /request that brings back a formatted json response.  We can go through it together if you like.

lena-larionova:
>Without the /request, the response is the content of the website mockbin.org which is messy. When you add the /request that brings back a formatted json response. We can go through it together if you like.

Interesting. I don't have that happen with the GET request for validation, I only see that formatting when the route is created. Environment differences?

Screen Shot 2020-08-28 at 1 13 31 PM

PR #2296, opened 2020-09-03
docs(plugins) Create Kubernetes example tabs

lena-larionova:
This PR introduces Kubernetes examples for (almost) all plugins.

### Testing and Remaining Work

Examples are based on docs for [Custom Resource Definitions](https://github.com/Kong/kubernetes-ingress-controller/blob/main/docs/references/custom-resources.md), the [Kong Plugin Resource](https://github.com/Kong/kubernetes-ingress-controller/blob/main/docs/guides/using-kongplugin-resource.md), and on existing examples for declarative config. 

Most plugins examples should now work, other than the small list below (same issues as in https://github.com/Kong/docs.konghq.com/pull/2238):
- Response Rate Limiting: unique syntax that isn't used in other plugins, this needs a non-templated example
- Kafka upstream and Kafka Log: template syntax still needs adjustment for mapping values
- Exit transformer: Needs both declarative YAML and k8s examples; cURL and YAML seem to require very different values here
- OAuth2: plugin not compatible with k8s
- OpenID Connect: Config section deactivated, as this plugin doc is complex and written in a different format from other plugins.

**TL;DR**, for OpenID Connect, Response Rate Limiting, and Exit Transformer, will need to write examples manually instead of relying on the template to generate them. Will revisit Kafka plugins in subsequent improvements.

Third-party plugins have been spot-tested and most work. Some still have missing values, and that will require further testing in a future edit.

### Previews
* [Sample Kong Enterprise plugin with K8s example](https://deploy-preview-2296--kongdocs.netlify.app/hub/kong-inc/application-registration/)
* [Sample Kong Community plugin with K8s examples for Service, ingress/Route, Consumer, and global](https://deploy-preview-2296--kongdocs.netlify.app/hub/kong-inc/proxy-cache/)
* [Sample plugin with schema value in example](https://deploy-preview-2296--kongdocs.netlify.app/hub/kong-inc/request-validator/)
* [Sample plugin with many nested arrays](https://deploy-preview-2296--kongdocs.netlify.app/hub/kong-inc/request-transformer/)

#### Why did you include the KongPlugin resource and the annotation instead of just showing the resource?

Because we show how to apply the plugin to each entity from the Admin API and in declarative format, so this way it's consistent. Yes, it does take up a lot of space on the page, but it also provides easy copy&paste blocks all in one place.

### DIFF
--- a/app/_layouts/extension.html
+++ b/app/_layouts/extension.html
@@ -120,7 +120,7 @@ $ curl -X POST http://<admin-hostname>:8001/services/<service>/plugins \
 {% unless page.params.k8s_compatible == false %}
 {% navtab Kubernetes %}
 
-First, create a [KongPlugin](https://github.com/Kong/kubernetes-ingress-controller/blob/main/docs/guides/using-kongplugin-resource.md)
+First, create a [KongPlugin](https://github.com/Kong/kubernetes-ingress-controller/blob/main/docs/references/custom-resources.md#kongplugin)
 resource:
 
 ```yaml
@@ -226,7 +226,7 @@ $ curl -X POST http://<admin-hostname>:8001/routes/<route>/plugins \
 {% unless page.params.k8s_compatible == false %}
 {% navtab Kubernetes %}
 
-First, create a [KongPlugin](https://github.com/Kong/kubernetes-ingress-controller/blob/main/docs/guides/using-kongplugin-resource.md)
+First, create a [KongPlugin](https://github.com/Kong/kubernetes-ingress-controller/blob/main/docs/references/custom-resources.md#kongplugin)
 resource:
 
 ```yaml
@@ -332,7 +332,7 @@ $ curl -X POST http://<admin-hostname>:8001/consumers/<consumer>/plugins \
 {% unless page.params.k8s_compatible == false %}
 {% navtab Kubernetes %}
 
-First, create a [KongPlugin](https://github.com/Kong/kubernetes-ingress-controller/blob/main/docs/guides/using-kongplugin-resource.md)
+First, create a [KongPlugin](https://github.com/Kong/kubernetes-ingress-controller/blob/main/docs/references/custom-resources.md#kongplugin)
 resource:
 
 ```yaml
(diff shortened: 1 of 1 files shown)


app/_layouts/extension.html
```diff
@@ -257,6 +421,32 @@
...
+{% if config_required_fields_k8s == 'config: ' %}config:
+  <optional_parameter>: <value>{% else %}{{ config_required_fields_k8s }}{% endif %}
+plugin: {{page.params.name}}
+```
+
+<div class="alert alert-ee blue">
+  <strong>Note:</strong> If you want the plugin to be available cluster-wide,
+  create the resource as a <code>KongClusterPlugin</code> instead of
+  <code>KongPlugin</code>.
+</div>
```
rainest, line 446: As of 0.10 (not yet released) we'll start requiring KongClusterPlugins for global plugins:

```suggestion
Create a [KongClusterPlugin](https://github.com/Kong/kubernetes-ingress-controller/blob/main/docs/references/custom-resources.md#kongclusterplugin)
resource and label it as global:

```yaml
apiVersion: configuration.konghq.com/v1
kind: KongClusterPlugin
metadata:
  name: <global-{{page.params.name}}>
  labels:
    global: \"true\"
{% if config_required_fields_k8s == 'config: ' %}config:
  <optional_parameter>: <value>{% else %}{{ config_required_fields_k8s }}{% endif %}
plugin: {{page.params.name}}
```
```

rainest (changes requested):
Global plugins need to be KongClusterPlugin, but otherwise looks good to go.

We may want to consider using the [KongPlugin reference](https://github.com/Kong/kubernetes-ingress-controller/blob/main/docs/references/custom-resources.md#kongplugin) rather than the guide: the examples here are already mini-guides (they show the link between the KongPlugin or Service/Route/etc.) and the reference goes into a bit more detail on the available fields.

app/_layouts/extension.html
rainest: Current target is the 16th. However, KongClusterPlugins are already available, so no harm in using them in advance--they aren't available in versions older than 0.8, but that's been out for a while now (since March).

rainest (approved):

lena-larionova:
@rainest FYI, I grabbed the info about global plugins here, I imagine you'll want to update it for 0.10: https://github.com/Kong/kubernetes-ingress-controller/blob/main/docs/guides/using-kongplugin-resource.md#configure-a-global-plugin

rainest:
Good to go on our end, just not visible from main yet: https://github.com/Kong/kubernetes-ingress-controller/blob/next/docs/guides/using-kongplugin-resource.md#configure-a-global-plugin is updated to use KongClusterPlugins.

PR #2304, opened 2020-09-10
docs(getting started guide) Example formatting and decK tab titles

lena-larionova:
* Change tab labels from "Using decK" to "Using decK (YAML)" for a bit more clarity (per Reza)
* add `sh` to cURL and HTTPie examples for syntax highlighting
* adjust cURL example syntax to be more consistent between topics (tested to make sure it works)
* fix: ` to ' in example


tlblessing (approved):
LGTM

PR #2308, opened 2020-09-14
header & footer navigation updates

adamjkuhn:
* updated footer navigation for style & content consistency with KongHQ
* minor content updates to header nav
* button style updates - header nav


lena-larionova (approved):
Looks like the logo and button are slightly oversized - they're bigger than their parallels on the main site. Not sure if that's an issue or not, but figured I'd point it out. Otherwise, looks good.

PR #2326, opened 2020-09-21
Add required param to examples

peterskim12:
lua_ssl_trusted_certificate needs to be set to the CA cert on the data plane. 
added /<path-to-file>/ to examples, especially for the kong.conf examples
changed examples of ca cert to use .pem extension to make clear those are typically PEM format certs.

when i was playing around with configs on the data node, it looked like cluster_ca_cert wasn't actually required; lua_ssl_trusted_certificate was. ideally this should be fixed so that cluster_ca_cert is where Kong looks for the CA cert because lua_ssl_trusted_certificate is buried in the MISCELLANEOUS section of kong.conf. but for now, i think the docs should just be fixed so it works and users don't get too confused. :)

### DIFF
--- a/app/enterprise/2.1.x/deployment/hybrid-mode-setup.md
+++ b/app/enterprise/2.1.x/deployment/hybrid-mode-setup.md
@@ -378,7 +378,7 @@ follow the instructions to:
     -e "KONG_CLUSTER_MTLS=pki" \
     -e "KONG_CLUSTER_SERVER_NAME=control-plane.kong.yourcorp.tld" \
     -e "KONG_CLUSTER_CERT=data-plane.crt" \
-    -e "KONG_CLUSTER_CERT_KEY=data-plane.crt" \
+    -e "KONG_CLUSTER_CERT_KEY=/<path-to-file>/data-plane.crt" \
     -e "KONG_CLUSTER_CA_CERT=ca-cert.pem" \
     -e "KONG_LUA_SSL_TRUSTED_CERTIFICATE=ca-cert.pem" \
     --mount type=bind,source="$(pwd)"/cluster,target=<path-to-keys-and-certs>,readonly \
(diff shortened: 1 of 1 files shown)


app/enterprise/2.1.x/deployment/hybrid-mode-setup.md
```diff
@@ -378,7 +378,9 @@ follow the instructions to:
     -e "KONG_CLUSTER_MTLS=pki" \
     -e "KONG_CLUSTER_SERVER_NAME=control-plane.kong.yourcorp.tld" \
     -e "KONG_CLUSTER_CERT=data-plane.crt" \
-    -e "KONG_CLUSTER_CERT_KEY=data-plane.crt"
+    -e "KONG_CLUSTER_CERT_KEY=data-plane.crt" \
```
lena-larionova, line 381:
```suggestion
    -e "KONG_CLUSTER_CERT_KEY=/<path-to-file>/data-plane.crt" \
```
Missing the <path-to-file> in these three (unless that's on purpose?)

app/enterprise/2.1.x/deployment/hybrid-mode-setup.md
```diff
@@ -378,7 +378,9 @@ follow the instructions to:
...
+    -e "KONG_CLUSTER_CA_CERT=ca-cert.pem" \
```
lena-larionova, line 382:
```suggestion
    -e "KONG_CLUSTER_CA_CERT=/<path-to-file>/ca-cert.pem" \
```

app/enterprise/2.1.x/deployment/hybrid-mode-setup.md
```diff
@@ -378,7 +378,9 @@ follow the instructions to:
...
+    -e "KONG_LUA_SSL_TRUSTED_CERTIFICATE=ca-cert.pem" \
```
lena-larionova, line 383:
```suggestion
    -e "KONG_LUA_SSL_TRUSTED_CERTIFICATE=/<path-to-file>/ca-cert.pem" \
```

lena-larionova (commented):
Thanks for doing the legwork on this!

app/enterprise/2.1.x/deployment/hybrid-mode-setup.md
peterskim12: no that's probably a good idea. i wasn't sure if the assumption with the --mount option was that the certs would live in the root of that mount path or in subdirs. thanks!

lena-larionova (approved):
LGTM

PR #2336, opened 2020-09-28
docs(styles) Fixing numbering issue with nested unordered lists 

lena-larionova:
Currently, the docs site has a problem with nested lists, in that unordered lists are implicitly counted as part of the step count. For example:

```
1. Example
2. List
  * Here's a point
  * And another one
5. List count doesn't continue at 3, but instead assumes the bullet points were part of it.
```

With this fix, it should all work as intended:
```
1. Example
2. List
  * Point
  * Point
3. Proper numbering continues
```

Also fixing font size to match across lists, whether the `<li>` has a `<p>` tag or not. Previous behaviour with two different font sizes:
https://docs.konghq.com/getting-started-guide/2.1.x/manage-teams/
Screen Shot 2020-09-28 at 12 40 13 PM

### DIFF
--- a/app/_assets/stylesheets/page.less
+++ b/app/_assets/stylesheets/page.less
@@ -621,7 +621,7 @@
 .content {
   padding: 50px 0;
   .font-source-sans();
-  font-size: 16px;
+  font-size: 17px;
   line-height: 1.5;
 
   .big {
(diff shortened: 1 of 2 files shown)


tlblessing (approved):
Thanks for fixing this! Had to do some awkward workarounds in past PRs.

PR #2353, opened 2020-10-03
Updated LICENSE from (C) to ©

Sumit1777:
Updated LICENSE from (C) to ©

### DIFF
--- a/LICENSE
+++ b/LICENSE
@@ -1,6 +1,6 @@
 The MIT License (MIT)
 
-Copyright © 2015-2018 Kong Inc. (https://konghq.com)
+Copyright (c) 2015-2020 Kong Inc. (https://konghq.com)
 
 Permission is hereby granted, free of charge, to any person obtaining a copy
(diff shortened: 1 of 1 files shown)


LICENSE
```diff
@@ -1,6 +1,6 @@
 The MIT License (MIT)
 
-Copyright (c) 2015-2018 Kong Inc. (https://konghq.com)
+Copyright © 2015-2018 Kong Inc. (https://konghq.com)
```
tlblessing, line 3: Should this be updated to year 2020? @hbagdi

hbagdi: Indeed. And `(c)` is good as is.

tlblessing: Thank you @hbagdi !

tlblessing (approved):
We are updating year only per Harry and not using the symbol.

tlblessing:
Thank you @Sumit1777 the PR and bringing this file to our attention.

PR #2358, opened 2020-10-06
feat(balancer) target objects can now be PATCHed and DELETEd

javierguerragiraldez:
2.2 now allows updating and removing targets via the admin-api.  Remove notes about previous caveats and document new method.

### DIFF
--- a/app/2.2.x/loadbalancing.md
+++ b/app/2.2.x/loadbalancing.md
@@ -166,13 +166,10 @@ upstreams is available in the `upstream` section of the
 
 ### Target
 
-Because the `upstream` maintains a history of changes, targets can only be
-added, not modified nor deleted. To change a target, just add a new entry for
-the target, and change the `weight` value. The last entry is the one that will
-be used. As such setting `weight=0` will disable a target, effectively
-deleting it from the balancer. Detailed information on adding and manipulating
-targets is available in the `target` section of the
-[Admin API reference][target-object-reference].
+A target is an ip address/hostname with a port that identifies an instance of
+a backend service. Each upstream can have many targets.
+Detailed information on adding and manipulating targets is available in the
+`target` section of the [Admin API reference][target-object-reference].
 
 The targets will be automatically cleaned when there are 10x more inactive
 entries than active ones. Cleaning will involve rebuilding the balancer, and


app/2.2.x/admin-api.md
```diff
@@ -3475,11 +3475,7 @@ HTTP 200 OK
 
 A target is an ip address/hostname with a port that identifies an instance of a backend
 service. Every upstream can have many targets, and the targets can be
-dynamically added. Changes are effectuated on the fly.
-
-Because the upstream maintains a history of target changes, the targets cannot
-be deleted or modified. To disable a target, post a new one with `weight=0`;
-alternatively, use the `DELETE` convenience method to accomplish the same.
+dynamically added, modified or deleted. Changes are effectuated on the fly.
```
tlblessing, line 3478:
```suggestion
dynamically added, modified, or deleted. Changes effectuate on the fly.
```

tlblessing (changes requested):
Suggested edits

lena-larionova:
@javierguerragiraldez can you make the changes to the Admin API (2.2.x/admin-api.md) in the source code instead? That 
 doc gets created using a script, and your changes will be overwritten here otherwise.

From the file:
> #  WARNING: this file was auto-generated by a script.
#  DO NOT edit this file directly. Instead, send a pull request to change
#  https://github.com/Kong/kong/blob/master/scripts/autodoc-admin-api
#  or its associated files instead.

javierguerragiraldez:
> can you make the changes to the Admin API (2.2.x/admin-api.md) in the source code instead?

:facepalm:   Kong/kong#6438

app/2.2.x/admin-api.md
tlblessing: @javierguerragiraldez The corresponding PR mentioned below used the language "takes effect" rather than "effectuate". Both are fine, and I guess this will get overwritten eventually if it isn't changed to match.

javierguerragiraldez: yes, i forgot that this is a generated file; so I removed the changes here.   as for the exact wording, i guess "take effect" is the most readable version.

tlblessing (approved):

PR #2364, opened 2020-10-08
docs(deck) Change version in install topic to variable

lena-larionova:
* update version to 1.2.1
* set variables so we only have to update one place (same as other install docs)

Preview: https://deploy-preview-2364--kongdocs.netlify.app/deck/installation/#linux


tlblessing (approved):
LGTM

PR #2375, opened 2020-10-13
Updating with correct docker registry and docker repo

felderi:
Updating with correct docker registry and docker repo

### DIFF
--- a/app/enterprise/2.1.x/kong-for-kubernetes/install.md
+++ b/app/enterprise/2.1.x/kong-for-kubernetes/install.md
@@ -75,16 +75,16 @@ Set up Docker credentials to allow Kubernetes nodes to pull down the Kong Enterp
 {% navtabs %}
 {% navtab kubectl %}
 ```
-$ kubectl create secret -n kong docker-registry kong-enterprise-edition-docker \
-    --docker-server=kong-docker-kong-enterprise-edition-docker.bintray.io \
-    --docker-username=<your-bintray-username> \
-    --docker-password=<your-bintray-api-key>
+$ kubectl create secret -n kong docker-registry kong-enterprise-k8s-docker \
+      --docker-server=kong-docker-kong-enterprise-k8s.bintray.io \
+      --docker-username=<your-bintray-username> \
+      --docker-password=<your-bintray-api-key>
 ```
 {% endnavtab %}
 {% navtab OpenShift oc %}
 ```
-$ oc create secret -n kong docker-registry kong-enterprise-edition-docker \
-    --docker-server=kong-docker-kong-enterprise-edition-docker.bintray.io \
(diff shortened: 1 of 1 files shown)


lena-larionova:
@felderi Looks like I missed something along the way - I thought we deprecated `kong-enterprise-k8s` in favor of `kong-enterprise-edition` across the board? 

For reference, here's our deployment options topic: https://docs.konghq.com/enterprise/2.1.x/kong-for-kubernetes/deployment-options/ 
and release notes: https://docs.konghq.com/enterprise/2.1.x/release-notes/#kong-for-kubernetes-enterprise-k4k8s-image-changes

Did that not actually end up happening?

felderi:
Probably worth double checking this but the instructions did not work for me @lena-larionova.  Changing to what I submitted in the PR worked for me.

felderi:
Li Yang ran into this as well:  https://kongstrong.slack.com/archives/C0DSXDXV1/p1602643762332900.  I also spoke to Damon that also ran into the same issue...

lena-larionova:
@rainest FYI - can you clarify what's going on here? I've checked the KIC repo, looks like `kong-enterprise-k8s` is still going strong with a 2.1.4.1 image. I guess there was a misunderstanding somewhere along the way?

lena-larionova (approved):

rainest:
I got all sorts of confused this morning, but I think I now understand more or less what happened:

- The kong-enterprise-k8s to kong-enterprise-edition switch was in a weird state when the deprecation happened. 2.1 was still in beta when we decided to make the switch, and as such we did not update documentation content (Helm and KIC READMEs and examples) that do not distinguish between beta and stable/prod-ready. The K8S team decided to leave them at stable during the beta period.
- Following the stable release of 2.1, we updated some examples, but not all. This change should have gone into the 1.0.0 KIC release, but we missed it.
- The docs site instructions _were_ updated during the beta because the docs site does make the beta/stable distinction.

https://github.com/Kong/kubernetes-ingress-controller/pull/909 corrects that, and should be the last holdout. Though we shouldn't encounter this particular problem going forward (the change to kong-enterprise-edition was a one-off change for 2.1), it does highlight a general class of issue we'll have with the new beta/stable Enterprise release pattern. We'll need to think of way to handle that for the Helm and KIC repo examples going forward, though I'm unsure what the best option is.

@felderi will go ahead and close this since we shouldn't merge it in: kong-enterprise-edition is correct going forward. If there any remaining issues, we should discuss those in https://github.com/Kong/kubernetes-ingress-controller/pull/909

Note that doc site instructions will still be misaligned until KIC#909 is merged, as instructions point to `main`. If you want to test in advance of that, replacing `main` with `fix/enterprise-version` should work temporarily.

PR #2191, opened 2020-07-24
doc(k4k8s) add kong-enterprise-edition DB-less doc

rainest:
Update K4K8S documentation to recommend the kong-enterprise-edition image for all deployments, DB-less or otherwise. See FT-1402 and DOCS-1166 for internal discussion of this change.

PR #2378, opened 2020-10-15
changing the name of the K8s secret

marcoam:
changing the name of the K8s secret creation command to match what's located here: https://bit.ly/k4k8s-enterprise. Note the following on the k4k8s-enterprise YAML file: 

      imagePullSecrets:
      - name: kong-enterprise-k8s-docker

I looked at the version 1.5.x of the doc and it had it right. So I dug a bit deeper and found this to be the cause: 
https://github.com/Kong/docs.konghq.com/pull/2191

which *used to* reference: https://bit.ly/k4k8s-enterprise-beta. In this "beta", the imagePullSecrets was set to: 

      imagePullSecrets:
      - name: kong-enterprise-edition-docker

so it worked. However, we're using the non-beta version which has the other name.

The question becomes which one do we want to use? Whichever we use, we should make sure that the instructions can be copied and pasted. As of now, it doesn't work due to the mismatch we have in the secret names.

### DIFF
--- a/app/enterprise/2.1.x/kong-for-kubernetes/install.md
+++ b/app/enterprise/2.1.x/kong-for-kubernetes/install.md
@@ -75,7 +75,7 @@ Set up Docker credentials to allow Kubernetes nodes to pull down the Kong Enterp
 {% navtabs %}
 {% navtab kubectl %}
 ```
-$ kubectl create secret -n kong docker-registry kong-enterprise-edition-docker \
+$ kubectl create secret -n kong docker-registry kong-enterprise-k8s-docker \
     --docker-server=kong-docker-kong-enterprise-edition-docker.bintray.io \
     --docker-username=<your-bintray-username> \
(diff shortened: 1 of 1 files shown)


lena-larionova:
The docs are actually correct here; Fel had attempted to make this exact change very recently, and we discovered yesterday that there's an error on the KIC side: https://github.com/Kong/docs.konghq.com/pull/2375#issuecomment-708593679

marcoam:
I'm okay with this as long as the documents can be copied and pasted :) Either we change the values.yaml file or the docs, but whichever it is, they should match :)

lena-larionova:
Yep, Travis is fixing the values.yaml. Hopefully that'll be merged soon.

lena-larionova:
Closing since the fix to the yaml file is merged: https://github.com/Kong/kubernetes-ingress-controller/pull/909

PR #2401, opened 2020-10-21
Forward proxy plugin missing https_verify field description

tlblessing:
https://konghq.atlassian.net/browse/DOCS-869

Direct review link:

https://deploy-preview-2401--kongdocs.netlify.app/hub/kong-inc/forward-proxy/#parameters

Question for reviewers (@gszr or @RobSerafini): I'm not sure which doc versions this field should be in? It looks like it was added 2 years ago per blame. Also, is port 80 appropriate for the existing port example?


lena-larionova (approved):
LGTM, can backport these changes as-is if needed.

PR #2404, opened 2020-10-22
plugin(prometheus) document status api endpoint

Tieske:
Using the status API is probably the preferred way of scraping metrics, and it was not documented, so added.

### DIFF
--- a/app/_hub/kong-inc/prometheus/index.md
+++ b/app/_hub/kong-inc/prometheus/index.md
@@ -41,17 +41,14 @@ params:
 
 ---
 
-Metrics are available on the Admin API at the `http://localhost:8001/metrics`
-endpoint. Note that the URL to the Admin API will be specific to your
+Metrics are available on both the Admin API and Status API at the
+`http://localhost:<port>/metrics`
+endpoint. Note that the URL to those APIs will be specific to your
 installation; see [Accessing the metrics](#accessing-the-metrics) below.
 
 This plugin records and exposes metrics at the node level. Your Prometheus
 server will need to discover all Kong nodes via a service discovery mechanism,
-and consume data from each node's configured `/metric` endpoint. Kong nodes
-that are set to proxy only (that is, their Admin API has been disabled by
-specifying `admin_listen = off`) will need to use a [custom Nginx
-configuration template](/latest/configuration/#custom-nginx-configuration)
-to expose the metrics data.
+and consume data from each node's configured `/metrics` endpoint.
 
 ### Grafana dashboard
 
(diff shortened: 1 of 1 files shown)


app/_hub/kong-inc/prometheus/index.md
```diff
@@ -185,14 +182,18 @@ allow access to the `/metrics` endpoint to Prometheus:
    Prometheus servers use to access the metric data. Access through any
    firewalls would also need to be configured.
 
-2. You can proxy the Admin API through Kong itself then use plugins to limit
+1. If the [Status API](https://docs.konghq.com/latest/configuration/#status_listen)
+   is enabled, then its `/metrics` endpoint can be used.
+   This is the preferred method for nodes configured as a Data plane.
```
tlblessing, line 187:
```suggestion
   This is the preferred method for nodes configured as a Data Plane.
```

tlblessing (changes requested):
Thank you @Tieske. Minor suggested edit for consistent capitalization.

app/_hub/kong-inc/prometheus/index.md
```diff
@@ -185,14 +182,18 @@ allow access to the `/metrics` endpoint to Prometheus:
    Prometheus servers use to access the metric data. Access through any
    firewalls would also need to be configured.
 
-2. You can proxy the Admin API through Kong itself then use plugins to limit
+1. If the [Status API](https://docs.konghq.com/latest/configuration/#status_listen)
+   is enabled, then its `/metrics` endpoint can be used.
+   This is the preferred method for nodes configured as a Data Plane.
```
hbagdi, line 187: This is the preferred method for **ALL** types of installations.

app/_hub/kong-inc/prometheus/index.md
```diff
@@ -185,14 +182,18 @@ allow access to the `/metrics` endpoint to Prometheus:
...
+
+1. You can proxy the Admin API through Kong itself then use plugins to limit
    access. For example, you can create a route `/metrics` endpoint and have
    Prometheus access this endpoint to slurp in the metrics while preventing
    others from accessing it. The specifics of how this is configured will depend
    on your specific setup. For details, see [Securing the Admin
    API](https://docs.konghq.com/latest/secure-admin-api/#kong-api-loopback).
```
hbagdi, line 194: This was a workaround that we had documented in absence of `status_listen`. This should be removed now.

app/_hub/kong-inc/prometheus/index.md
```diff
@@ -185,14 +182,18 @@ allow access to the `/metrics` endpoint to Prometheus:
...
 
-3. Finally, you could serve the content on a different port with a custom server
+1. Finally, you could serve the content on a different port with a custom server
```
hbagdi, line 196: Users pay high-tax for this method. I recommend that we should remove this as well.

app/_hub/kong-inc/prometheus/index.md
Tieske: @hbagdi just checking; did the first option ever work? using RBAC blocks access, and I couldn't get Prometheus to connect since it does not have an option for key-auth (basic auth + a token one only)?
Did I miss something? or did we document something that doesn't work?

hbagdi: You could use a request-transformer plugin to inject the token only for `/metrics` path.
The first option was primarily targeted towards OSS users.

Tieske: that was option 2. Option 1:

> Kong Enterprise users can protect the admin /metrics endpoint with an RBAC user that the Prometheus servers use to access the metric data. Access through any firewalls would also need to be configured.

I couldn't find any way for Prometheus to pass the access token? So this is just incorrect documentation then?

Tieske:
changed the options again, dropped some, left 2

hbagdi (approved):

Tieske:
@Kong/team-docs can we merge and deploy this one? it's important imo, since we're literally sending users down the wrong path currently

tlblessing (approved):

PR #2406, opened 2020-10-22
fix(styles) Bring cookie banner to top

lena-larionova:
* Fix issue of cookie banner appearing underneath other elements:

Screen Shot 2020-10-22 at 10 29 55 AM

* Remove scroll-to-top button from landing pages, positioning is off without the feedback widget there.

Preview:
* Scroll down any landing page, only the cookie banner should be there: https://deploy-preview-2406--kongdocs.netlify.app
* Without dismissing the cookie banner, also scroll down on any page one level down, eg https://deploy-preview-2406--kongdocs.netlify.app/2.1.x/admin-api/. Cookie banner should appear overtop other icons.


tlblessing (approved):
LGTM

PR #2417, opened 2020-10-26
Update bitly link

rainest:


lena-larionova (approved):
Thanks for making the new link!

One thing to note though: older doc versions point to the exact same link (https://bit.ly/k4k8s-enterprise). This means that they're pointing to the wrong version of the image no matter what - technically, people can keep using `kong-enterprise-k8s` in the older content, right? I wonder if we should just use the long link form in older content, at leas 1.3-x and 1.5.x.

rainest:
The older versions should point to the old image--for them, the separate image is indeed the only DB-less option. Hopefully shouldn't matter much in practice, as we shouldn't have new installs using those versions going forward.

PR #2420, opened 2020-10-27
regenerate configuration.md for 2.2

kikito:
There were some changes on the generated configuration.md file for 2.2.0

### DIFF
--- a/app/2.2.x/configuration.md
+++ b/app/2.2.x/configuration.md
@@ -620,6 +620,9 @@ Default: none
 
 ---
 
+
+### Hybrid Mode Data Plane section
+
 #### cluster_server_name
 
 The server name used in the SNI of the TLS connection from a DP node to a CP
@@ -644,6 +647,9 @@ Default: none
 
 ---
 
+
+### Hybrid Mode Control Plane section
+
 #### cluster_listen
 
 Comma-separated list of addresses and ports on which the cluster control plane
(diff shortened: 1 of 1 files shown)


app/2.2.x/configuration.md
```diff
@@ -1393,7 +1398,7 @@ name   | description  | default
 **pg_password** | Postgres user's password. | none
 **pg_database** | The database name to connect to. | `kong`
 **pg_schema** | The database schema to use. If unspecified, Kong will respect the `search_path` value of your PostgreSQL instance. | none
-**pg_ssl** | Toggles client-server TLS connections between Kong and PostgreSQL. | `off`
+**pg_ssl** | Toggles client-server TLS connections between Kong and PostgreSQL. Since PostgreSQL uses the same port for TLS and non-TLS this is only a hint. If the server does not support TLS the established connection will be a plain one. | `off`
```
tlblessing, line 1401:
```suggestion
**pg_ssl** | Toggles client-server TLS connections between Kong and PostgreSQL. Because PostgreSQL uses the same port for TLS and non-TLS, this is only a hint. If the server does not support TLS, the established connection will be a plain one. | `off`
```

tlblessing (approved):
@kikito  I get that this is a generated file, but docs should be involved upfront to edit the source of the generated files. Approved but might want to cherry-pick the suggested corrections where they belong. Thank you!

app/2.2.x/configuration.md
kikito: Thanks for the review. I've added these changes to the original config file and regenerated the docs. I will send other pull requests on the kong repo to make these changes permanent.

PR #2428, opened 2020-10-29
docs(plugins) Introduce datatype field for plugin template

lena-larionova:
### Issue
Plugin docs don't list config parameter datatypes, which is a standard piece of information in config reference documentation.

### Changes
* Adding "datatype" field to plugin template and doc layout, so that if a datatype is specified for a parameter, it'll appear below the config parameter name.
* Testing this out with the Basic Auth plugin

### Preview
[Basic Auth](https://deploy-preview-2428--kongdocs.netlify.app/hub/kong-inc/basic-auth/) plugin - see parameter table

### Notes 
* Based the `name` and `id` datatypes on our PDK doc: 
https://docs.konghq.com/2.2.x/plugin-development/plugin-configuration/#schemalua-specifications
* There's a lot of work to do to actually add the datatype field to the doc for each plugin; ideally, it should be automated.
* Will also update autogenerated example parser once this is applied to more plugins to get more accurate syntax
* Tested out adding a column for the datatype instead of including it in the same cell as the param name, and it looked pretty good. Unfortunately, because most of the plugins don't have this value set, we'd end up with a lot of empty cells until this was all automated or filled out manually.

Test:
Screen Shot 2020-10-29 at 2 27 36 PM

Reality for most plugins:
Screen Shot 2020-10-29 at 2 27 54 PM

### DIFF
--- a/app/_hub/_init/my-extension/index.md
+++ b/app/_hub/_init/my-extension/index.md
@@ -144,6 +144,7 @@ params: # Metadata about your plugin
         # options are 'yes', 'no', or 'semi'
         # 'semi' means dependent on other settings
       default: # any type - the default value (non-required settings only)
+      datatype: # specify the type of the value: e.g., string, array, boolean, etc.
       value_in_examples:
         # If the field is to appear in examples, this is the value to use.
         # A required field with no value_in_examples entry will resort to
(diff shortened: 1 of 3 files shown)


tlblessing (approved):
Thank you for accommodating my request so quickly! @lena-larionova you totally rock!

PR #2430, opened 2020-10-30
docs(rate-limiting) added missing conf options

locao:


tlblessing (changes requested):
Thank you @locao for the PR. I see the advanced rate limiting plugin has these fields but not the fallback information. Does that need to be ported to Enterprise?

locao:
> I see the advanced rate limiting plugin has these fields but not the fallback information. Does that need to be ported to Enterprise?

I just checked the code and yes, the advanced rate limiting also fallback to IP when the configured field can't be found. It should be ported.

tlblessing:
> > I see the advanced rate limiting plugin has these fields but not the fallback information. Does that need to be ported to Enterprise?
> 
> I just checked the code and yes, the advanced rate limiting also fallback to IP when the configured field can't be found. It should be ported.

OK I'm working on the advanced plugin in the private repo right now. I'll port it over. Thank you.

tlblessing (approved):

PR #2436, opened 2020-11-02
update sitemap priority levels

adamjkuhn:
* sitemap priority levels updated from 2.0/1.0 (for latest vs previous version pages) to 1.0/0.5 (as priority should be set at a maximum of 1.0)- this should help balance SEO results to prioritize most recent versions


lena-larionova (approved):
Thanks!

PR #2440, opened 2020-11-06
fix(docs-nav) Stop links to docs from opening in new window/tab

lena-larionova:
Docs ticket: https://konghq.atlassian.net/browse/DOCS-1133

There's no reason that docs should open in a new window or tab - might make sense on the main konghq.com site, but for the docs subdomain, it's logical to keep the user in the same tab as they navigate through the docs.


GayleNeumann (approved):
LGTM

PR #2441, opened 2020-11-06
fix(getting-started) Add missing argument for enabling RBAC with Docker

lena-larionova:
If you run the command as-is, you get the following error:

```
"docker exec" requires at least 2 arguments.
```

Fixing it, and formatting the codeblock for readability.

Preview:
https://deploy-preview-2441--kongdocs.netlify.app/getting-started-guide/2.1.x/manage-teams/#turn-on-rbac


GayleNeumann (approved):
LGTM

PR #2442, opened 2020-11-07
Docs(styles) Add support for tabbed codeblocks and make copy icon transparent

lena-larionova:
### Changes
- Add support for tabbed codeblocks
- make copy icon transparent so it doesn't obscure code in longer lines

### Usage
Same as current navtabs, except you can only include codeblocks in them. Use `navtabs` with the `codeblock` class, eg:
Screen Shot 2020-11-06 at 4 41 58 PM

Output:
Screen Shot 2020-11-06 at 4 34 46 PM

### Outstanding issues/questions
I'm not sure what to do with the copy-code icon when there's no header. Should I force a header on every codeblock? There's probably a more elegant solution than that...

Current behaviour:
Screen Shot 2020-11-06 at 4 43 54 PM

@adamjkuhn I think this PR definitely needs your expertise - I doubt my attempts are the most efficient or clean.

### Preview
Test topic: https://deploy-preview-2442--kongdocs.netlify.app/enterprise/2.1.x/deployment/installation/kong-for-kubernetes/

### DIFF
--- a/app/_assets/stylesheets/copy-code-snippet.less
+++ b/app/_assets/stylesheets/copy-code-snippet.less
@@ -7,14 +7,12 @@
     top: 4px;
     right: 4px;
     cursor: pointer;
-    color: rgba(black, 0.3);
+    color: rgba(black, 0.4);
     transition: color 0.3s;
-    background: white;
-    border-radius: 2px;
-    border: 1px solid #ebebeb;
+    z-index: 10;
 
     &:hover {
-      color: rgba(black, 0.5);
+      color: rgba(black, 0.8);
     }
   }
 }
--- a/app/_assets/stylesheets/copy-code-snippet.less
+++ b/app/_assets/stylesheets/copy-code-snippet.less
@@ -3,7 +3,7 @@
 
   .copy-action {
     position: absolute;
-    padding: 4px;
+    padding: 3px;
     top: 4px;
     right: 4px;
     cursor: pointer;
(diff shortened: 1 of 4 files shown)


adamjkuhn:
A couple of thoughts:

Conditionally add top padding to the editor (I'd recommend adding bottom padding as well for vertical balance)

block with top padding

otherwise perhaps adding a background & padding to the icon (or perhaps a backdrop-filter to blur content beneath the icon itself) is a viable option:
block with icon update

Let me know if either of these seem to work and I can implement whichever is best.

lena-larionova:
Thanks for taking a look @adamjkuhn !

> Conditionally add top padding to the editor (I'd recommend adding bottom padding as well for vertical balance)

I tried this one, but because the copy-code icon appears in every codeblock by default, it gives every codeblock a much larger footprint. However... Maybe I was overdoing it, so maybe you can give it a shot.

> otherwise perhaps adding a background & padding to the icon (or perhaps a backdrop-filter to blur content beneath the icon itself) is a viable option:

This is the current implementation, and that's causing problems at the moment with longer codeblocks, eg on the [Docker installation page](https://docs.konghq.com/enterprise/2.1.x/deployment/installation/docker/#pull-image). Maybe it's possible to conditionally extend the codeblock in those situations, so that it scrolls far enough past the code? Or is that a weird solution :)

tlblessing (approved):
Looks Great To Me @lena-larionova! Thank you

PR #2452, opened 2020-11-13
Serverless; update lambda and azure functions

Tieske:


Tieske:
direct preview link Lambda: https://deploy-preview-2452--kongdocs.netlify.app/hub/kong-inc/aws-lambda/
direct preview link Azure: https://deploy-preview-2452--kongdocs.netlify.app/hub/kong-inc/azure-functions/

@tlblessing seems that the tabs don't play nice with the numbering, how to fix that?

tlblessing:
> @tlblessing seems that the tabs don't play nice with the numbering, how to fix that?

Great question. We currently workaround that formatting issue. @lena-larionova would have more information about future improvements. You'll notice the prior versions of these plugins did not have the last testing example as a step. I removed the step and made it a heading. The navtabs weren't working because of the indents, which I moved back as they were before.

Does the Azure version need bumping? It is at 1.0.x with internal handler version 0.4.2. Thank you for the PR @Tieske.

tlblessing (approved):

PR #2477, opened 2020-11-24
docs(k8s) add ingress version documentation

rainest:
Add a concept document that covers the differences between v1beta1 and v1 versions of the Ingress resource.

Converted from https://github.com/Kong/kubernetes-ingress-controller/pull/950/

Open question regarding matching between `metadata.name` in IngressClass resources and `ingressClassName` in Ingress resources remains open, but I believe it's correct as written based on what info I can find. I can't test that in our GKE cluster, as GKE 1.18 still rejects v1 Ingresses and 1.19 isn't yet available. Not sure why this is, as [their release notes suggest otherwise](https://cloud.google.com/kubernetes-engine/docs/release-notes#known-20200908), but I can't find other documentation related to it. IMO better to go ahead and publish this with what we know, as we can always correct it if needed later.

### DIFF
--- a/app/_data/docs_nav_2.1.x.yml
+++ b/app/_data/docs_nav_2.1.x.yml
@@ -452,6 +452,8 @@
           url: /../../kubernetes-ingress-controller/latest/concepts/ingress-classes
         - text: Security
           url: /../../kubernetes-ingress-controller/latest/concepts/security
+        - text: Ingress Versions
+          url: /../../kubernetes-ingress-controller/latest/concepts/ingress-versions
     - text: Deployment
       items:
         - text: Overview
--- a/app/_data/docs_nav_2.1.x.yml
+++ b/app/_data/docs_nav_2.1.x.yml
@@ -452,7 +452,7 @@
           url: /../../kubernetes-ingress-controller/latest/concepts/ingress-classes
         - text: Security
           url: /../../kubernetes-ingress-controller/latest/concepts/security
-        - text: Ingress Versions
+        - text: Ingress Resource API Versions
           url: /../../kubernetes-ingress-controller/latest/concepts/ingress-versions
     - text: Deployment
       items:
--- a/app/kubernetes-ingress-controller/1.0.x/concepts/ingress-versions.md
+++ b/app/kubernetes-ingress-controller/1.0.x/concepts/ingress-versions.md
@@ -4,7 +4,7 @@ title: Ingress v1 and v1beta1 Differences
 
 ## Introduction
 
-Kubernetes 1.18 introduced a new `networking.k8s.io/v1` API for the [Ingress resource][kubernetes-ingress-doc].
+Kubernetes 1.19 introduced a new `networking.k8s.io/v1` API for the [Ingress resource][kubernetes-ingress-doc].
 It standardizes common practices and clarifies implementation requirements that
 were previously up to individual controller vendors. This document covers those
 changes as they relate to {{site.kic_product_name}} and provides sample
@@ -141,12 +141,13 @@ and `bar.example.com` with either v1 or v1beta1 Ingresses.
 Ingress v1 introduces support for backends other than Kubernetes Services through
 [resource backends][resource-backends].
 
-Kong has support for Routes without a Service in some cases, such as Routes
-that use the [AWS Lambda plugin][lambda-plugin]. However, it does not yet offer
-Kubernetes resources for them. As such, you should instead create a placeholder
-Kubernetes Service for them, using an [ExternalName Service][external-name]
-with an RFC 2606 invalid hostname (for example,`kong.invalid`). You can use these
-placeholder services with either v1 or v1beta1 Ingresses.
+Kong does not support any dedicated resource backend configurations, though it
+does have support for Routes without Services in some cases (for example, when
+using the [AWS Lambda plugin][lambda-plugin]). For these routes, you should
+create a placeholder Kubernetes Service for them, using an [ExternalName
+Service][external-name] with an RFC 2606 invalid hostname, e.g.
+`kong.invalid`. You can use these placeholder services with either v1 or
+v1beta1 Ingresses.
 
 [kubernetes-ingress-doc]: https://kubernetes.io/docs/concepts/services-networking/ingress/
 [ingress-class]: /kubernertes-ingress-controller/{{page.kong_version}}/concepts/ingress-classes
--- a/app/kubernetes-ingress-controller/1.0.x/concepts/ingress-versions.md
+++ b/app/kubernetes-ingress-controller/1.0.x/concepts/ingress-versions.md
@@ -1,5 +1,5 @@
 ---
-title: Ingress v1 and v1beta1 differences
+title: Ingress v1 and v1beta1 Differences
 ---
 
 ## Introduction
@@ -7,23 +7,23 @@ title: Ingress v1 and v1beta1 differences
 Kubernetes 1.18 introduced a new `networking.k8s.io/v1` API for the [Ingress resource][kubernetes-ingress-doc].
 It standardizes common practices and clarifies implementation requirements that
 were previously up to individual controller vendors. This document covers those
-changes as they relate to Kong Ingress Controller and provides example
+changes as they relate to {{site.kic_product_name}} and provides sample
 equivalent `networking.k8s.io/v1beta1` and `networking.k8s.io/v1` resources for comparison.
 
 ## Paths
 
 Both Ingress v1beta1 and v1 HTTP rules require a path, which represents a [URI
 path][uri-rfc-paths]. Although v1beta1 had specified that paths were [POSIX
-regular expressions][posix-regex] and enforced this, but in practice most
+regular expressions][posix-regex] and enforced this, in practice most
 controllers used other other implementations that did not match the
-specification. v1 seeks to reduce confusion by introducing several i[path
+specification. v1 seeks to reduce confusion by introducing several [path
 types][path-types] and lifting restrictions on regular expression grammars used
 by controllers.
 
-### `networking.k8s.io/v1beta1`
+### networking.k8s.io/v1beta1
 
 The controller passes paths directly to Kong and relies on its [path handling
-logic][kong-path]. The Kong proxy treats paths as a prefix unless they include
+logic][kong-paths]. The Kong proxy treats paths as a prefix unless they include
 characters [not allowed in RFC 3986 paths][uri-rfc-paths], in which case the
 proxy assumes they are a regular expression, and does not treat slashes as
 special characters. For example, the prefix `/foo` can match any of the
@@ -36,35 +36,35 @@ following:
 /foo/bar
 ```
 
-### `networking.k8s.io/v1`
+### networking.k8s.io/v1
 
 Although v1 Ingresses provide path types with more clearly-defined logic, the
 controller must still create Kong routes and work within the Kong proxy's
 routing logic. As such, the controller translates Ingress rule paths to create
-Kong routes that match the specification:
+Kong routes that match one of the following specifications: `Exact`, `Prefix`, or `ImplementationSpecific`.
 
 #### Exact
 
 If `pathType` is `Exact`, the controller creates a Kong route with a regular
-expression that matches the rule path only, e.g. an exact rule for `/foo` in an
+expression that matches the rule path only. For example, an exact rule for `/foo` in an
 Ingress translates to a Kong route with a `/foo$` regular expression path.
 
 #### Prefix
 
 If `pathType` is `Prefix`, the controller creates a Kong route with two path
-criteria, e.g. `/foo` will create a route with a `/foo$` regular expression and
+criteria. For example, `/foo` will create a route with a `/foo$` regular expression and
 `/foo/` plain path.
 
 #### ImplementationSpecific
 
-The controller leaves ImplementationSpecific path rules entirely up to the Kong
+The controller leaves `ImplementationSpecific` path rules entirely up to the Kong
 router. It creates a route with the exact same path string as the Ingress rule.
 
 ## Ingress class
 
 [Ingress class][ingress-class] indicates which resources an ingress controller
 should process. It provides a means to separate out configuration intended for
-other controllers or other instances of the Kong Ingress Controller.
+other controllers or other instances of the {{site.kic_product_name}}.
 
 In v1beta1, ingress class was handled informally using
 `kubernetes.io/ingress.class` [annotations][deprecated-annotation]. v1
@@ -74,7 +74,7 @@ their `ingressClassName` field.
 
 For example, consider this v1beta1 Ingress:
 
-```
+```yaml
 apiVersion: networking.k8s.io/v1beta1
 kind: Ingress
 metadata:
@@ -100,7 +100,7 @@ whose `metadata.name` field indicates the class name. The `ingressClassName`
 value of the Ingress object must match the value of the `name` field in the
 IngressClass metadata:
 
-```
+```yaml
 apiVersion: networking.k8s.io/v1
 kind: IngressClass
 metadata:
@@ -138,18 +138,18 @@ and `bar.example.com` with either v1 or v1beta1 Ingresses.
 
 ## Backend types
 
-Ingress v1 introduces support for backends other than Kubernetes Services, via
+Ingress v1 introduces support for backends other than Kubernetes Services through
 [resource backends][resource-backends].
 
-Kong has support for routes without a service in some cases, such as routes
+Kong has support for Routes without a Service in some cases, such as Routes
 that use the [AWS Lambda plugin][lambda-plugin]. However, it does not yet offer
 Kubernetes resources for them. As such, you should instead create a placeholder
 Kubernetes Service for them, using an [ExternalName Service][external-name]
-with an RFC 2606 invalid hostname, e.g. `kong.invalid`. You can use these
+with an RFC 2606 invalid hostname (for example,`kong.invalid`). You can use these
 placeholder services with either v1 or v1beta1 Ingresses.
 
 [kubernetes-ingress-doc]: https://kubernetes.io/docs/concepts/services-networking/ingress/
-[ingress-class]: ./ingress-classes.md
+[ingress-class]: /kubernertes-ingress-controller/{{page.kong_version}}/concepts/ingress-classes
 [uri-rfc-paths]: https://tools.ietf.org/html/rfc3986#section-3.3
 [posix-regex]: https://www.boost.org/doc/libs/1_38_0/libs/regex/doc/html/boost_regex/syntax/basic_extended.html
 [path-types]: https://kubernetes.io/docs/concepts/services-networking/ingress/#path-types
--- a/app/kubernetes-ingress-controller/1.0.x/concepts/ingress-versions.md
+++ b/app/kubernetes-ingress-controller/1.0.x/concepts/ingress-versions.md
@@ -1,0 +1,164 @@
+---
+title: Ingress v1 and v1beta1 differences
+---
+
+## Introduction
+
+Kubernetes 1.18 introduced a new v1 API for the [Ingress resource][kubernetes-ingress-doc].
+It standardizes common practices and clarifies implementation requirements that
+were previously up to individual controller vendors. This document covers those
+changes as they relate to Kong Ingress Controller and provides example
+equivalent v1beta1 and v1 resources for comparison.
+
+## Paths
+
+Both Ingress v1beta1 and v1 HTTP rules require a path, which represents a [URI
+path][uri-rfc-paths]. Although v1beta1 had specified that paths were [POSIX
+regular expressions][posix-regex] and enforced this, but in practice most
+controllers used other other implementations that did not match the
+specification. v1 seeks to reduce confusion by introducing several i[path
+types][path-types] and lifting restrictions on regular expression grammars used
+by controllers.
+
+### v1beta1
+
+The controller passes paths directly to Kong and relies on its [path handling
+logic][kong-path]. The Kong proxy treats paths as a prefix unless they include
+characters [not allowed in RFC 3986 paths][uri-rfc-paths], in which case the
+proxy assumes they are a regular expression, and does not treat slashes as
+special characters. For example, the prefix `/foo` can match any of the
+following:
+
+```
+/foo
+/foo/
+/foobar
+/foo/bar
+```
+
+### v1
+
+Although v1 Ingresses provide path types with more clearly-defined logic, the
+controller must still create Kong routes and work within the Kong proxy's
+routing logic. As such, the controller translates Ingress rule paths to create
+Kong routes that match the specification:
+
+#### Exact
+
+To handle exact matches, the controller creates a Kong route with a regular
+expression that matches the rule path only, e.g. an exact rule for `/foo` in an
+Ingress translates to a Kong route with a `/foo$` regular expression path.
+
+#### Prefix
+
+To handle prefix matches, the controller creates a Kong route with two path
+criteria, e.g. `/foo` will create a route with a `/foo$` regular expression and
+`/foo/` plain path.
+
+#### ImplementationSpecific
+
+The controller leaves ImplementationSpecific path rules entirely up to the Kong
+router. It creates a route with the exact same path string as the Ingress rule.
+
+## Ingress class
+
+[Ingress class][ingress-class] indicates which resources an ingress controller
+should process. It provides a means to separate out configuration intended for
+other controllers or other instances of the Kong Ingress Controller.
+
+In v1beta1, ingress class was handled informally using
+`kubernetes.io/ingress.class` [annotations][deprecated-annotation]. v1
+introduces a new [IngressClass resource][ingress-class-api] which provides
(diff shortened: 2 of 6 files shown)


app/_data/docs_nav_2.1.x.yml
```diff
@@ -452,6 +452,8 @@
           url: /../../kubernetes-ingress-controller/latest/concepts/ingress-classes
         - text: Security
           url: /../../kubernetes-ingress-controller/latest/concepts/security
+        - text: Ingress Versions
```
mflendrich, line 455: Not clear what "ingress versions" refers to. For example, is this versions of an ingress resource (if you apply consecutive patches to an ingress X in namespace Y) or API versions of the Ingress kind?

I recommend more specific alternatives:
- `networking.k8s.io` API versions (or `networking.k8s.io` versions)
- Ingress Resource API versions

mflendrich, line 455: Not clear what "ingress versions" refers to. For example, is this versions of an ingress resource (if you apply consecutive patches to an ingress X in namespace Y) or API versions of the Ingress kind?

I recommend more specific alternatives:
- `networking.k8s.io` API versions (or `networking.k8s.io` versions)
- Ingress Resource API versions

app/kubernetes-ingress-controller/1.0.x/concepts/ingress-versions.md
```diff
@@ -0,0 +1,164 @@
+---
+title: Ingress v1 and v1beta1 differences
+---
+
+## Introduction
+
+Kubernetes 1.18 introduced a new v1 API for the [Ingress resource][kubernetes-ingress-doc].
```
mflendrich, line 7: AFAIK `networking.k8s.io/v1` is available by default as of 1.19, not 1.18

mflendrich, line 7:
```suggestion
Kubernetes 1.18 introduced a new `networking.k8s.io/v1` API for the [Ingress resource][kubernetes-ingress-doc].
```

this is because Kubernetes API versions are always qualified like this; `v1` suggests it being part of core.

app/kubernetes-ingress-controller/1.0.x/concepts/ingress-versions.md
```diff
@@ -0,0 +1,164 @@
...
+
+Both Ingress v1beta1 and v1 HTTP rules require a path, which represents a [URI
+path][uri-rfc-paths]. Although v1beta1 had specified that paths were [POSIX
+regular expressions][posix-regex] and enforced this, but in practice most
+controllers used other other implementations that did not match the
+specification. v1 seeks to reduce confusion by introducing several i[path
+types][path-types] and lifting restrictions on regular expression grammars used
+by controllers.
+
+### v1beta1
```
mflendrich, line 23:
```suggestion
### `networking.k8s.io/v1beta1`
```

app/kubernetes-ingress-controller/1.0.x/concepts/ingress-versions.md
```diff
@@ -0,0 +1,164 @@
...
+following:
+
+```
+/foo
+/foo/
+/foobar
+/foo/bar
+```
+
+### v1
```
mflendrich, line 39:
```suggestion
### `networking.k8s.io/v1`
```

app/kubernetes-ingress-controller/1.0.x/concepts/ingress-versions.md
mflendrich: Also, `v1beta1` is ambiguous: it can be either `extensions/v1beta1` or ``extensions.k8s.io/v1beta1`. It's not unusual for apiversions to be moved between groups like that.

app/kubernetes-ingress-controller/1.0.x/concepts/ingress-versions.md
```diff
@@ -0,0 +1,164 @@
...
+path][uri-rfc-paths]. Although v1beta1 had specified that paths were [POSIX
+regular expressions][posix-regex] and enforced this, but in practice most
+controllers used other other implementations that did not match the
+specification. v1 seeks to reduce confusion by introducing several i[path
+types][path-types] and lifting restrictions on regular expression grammars used
+by controllers.
+
+### v1beta1
+
+The controller passes paths directly to Kong and relies on its [path handling
```
mflendrich, line 25:
```suggestion
For the `spec.rules[*].http.paths[*].path` field, the controller passes paths directly to Kong and relies on its [path handling
```

app/kubernetes-ingress-controller/1.0.x/concepts/ingress-versions.md
```diff
@@ -0,0 +1,164 @@
...
+/foobar
+/foo/bar
+```
+
+### v1
+
+Although v1 Ingresses provide path types with more clearly-defined logic, the
+controller must still create Kong routes and work within the Kong proxy's
+routing logic. As such, the controller translates Ingress rule paths to create
+Kong routes that match the specification:
```
mflendrich, line 44: Maybe we could frame this as a feature, instead of emphasizing KIC's heavy-lifting :stuck_out_tongue:

```suggestion
Ingress resources at this API version have explicitly defined matching semantics (as set in the `pathType` field). Under the hood of Kong Ingress Controller, a path (parameterized by `pathType`) gets translated to a regular expression in a Kong Route.
```

app/kubernetes-ingress-controller/1.0.x/concepts/ingress-versions.md
```diff
@@ -0,0 +1,164 @@
...
+
+#### Exact
+
+To handle exact matches, the controller creates a Kong route with a regular
```
mflendrich, line 48:
```suggestion
If `pathType` is `Exact`, the controller creates a Kong route with a regular
```

app/kubernetes-ingress-controller/1.0.x/concepts/ingress-versions.md
```diff
@@ -0,0 +1,164 @@
...
+expression that matches the rule path only, e.g. an exact rule for `/foo` in an
+Ingress translates to a Kong route with a `/foo$` regular expression path.
+
+#### Prefix
+
+To handle prefix matches, the controller creates a Kong route with two path
```
mflendrich, line 54:
```suggestion
If `pathType` is `Prefix`, the controller creates a Kong route with two path
```

app/kubernetes-ingress-controller/1.0.x/concepts/ingress-versions.md
```diff
@@ -0,0 +1,164 @@
...
+In v1beta1, ingress class was handled informally using
+`kubernetes.io/ingress.class` [annotations][deprecated-annotation]. v1
+introduces a new [IngressClass resource][ingress-class-api] which provides
+richer information about the controller. v1 Ingresses are bound to a class via
+their `ingressClassName` field.
+
+For example, consider this v1beta1 Ingress:
+
+```
+apiVersion: extensions/v1beta1
```
mflendrich, line 78:
```suggestion
apiVersion: networking.k8s.io/v1beta1
```

Let's explain the migration one step at a time - from `networking.k8s.io/v1beta1` to `networking.k8s.io/v1`. To avoid confusion, we should probably avoid mixing the two in one doc as well.

Particularly so because migration from `extensions/v1beta1` to `networking.k8s.io/v1beta1` is a no-op.

app/kubernetes-ingress-controller/1.0.x/concepts/ingress-versions.md
```diff
@@ -0,0 +1,164 @@
...
+value of the Ingress object must match the value of the `name` field in the
+IngressClass metadata:
+
+```
+apiVersion: networking.k8s.io/v1
+kind: IngressClass
+metadata:
+  name: kong
+spec:
+  controller: konghq.com/kong
```
mflendrich, line 109:
```suggestion
```

Unless k8s enforces the controller field being non-empty, let's leave it out, because KIC today disregards the IngressClass resource whatsoever. By specifying `controller` we might trick users into thinking that the field has a meaning today (which it doesn't have)

app/kubernetes-ingress-controller/1.0.x/concepts/ingress-versions.md
```diff
@@ -0,0 +1,164 @@
...
+title: Ingress v1 and v1beta1 differences
+---
+
+## Introduction
+
+Kubernetes 1.18 introduced a new v1 API for the [Ingress resource][kubernetes-ingress-doc].
+It standardizes common practices and clarifies implementation requirements that
+were previously up to individual controller vendors. This document covers those
+changes as they relate to Kong Ingress Controller and provides example
+equivalent v1beta1 and v1 resources for comparison.
```
mflendrich, line 11:
```suggestion
equivalent `networking.k8s.io/v1beta1` and `networking.k8s.io/v1` resources for comparison.
```

See below - let's be specific about exact API versions where it matters

app/kubernetes-ingress-controller/1.0.x/concepts/ingress-versions.md
rainest: There aren't any other paths in the Ingress spec that these could be confused with, so IMO this isn't adding much beyond making the sentence longer. Leaving this as-is.

app/kubernetes-ingress-controller/1.0.x/concepts/ingress-versions.md
rainest: It's an implementation detail, and I don't see a good reason we'd try to obfuscate what actually occurs. Users can check the admin API to see what happens, and the use of multiple path matches [can have practical implications](https://github.com/Kong/kubernetes-ingress-controller/issues/958) that they need to be aware of. Leaving this as-is.

app/kubernetes-ingress-controller/1.0.x/concepts/ingress-versions.md
rainest: 1.18 introduced it although it was a beta feature--it wasn't novel in 1.19, just graduated to GA.

The v1beta1 ambiguity is intentional here, since it exists in either extensions or networking depending on which K8S version you use, though its functionality is identical. That history isn't the focus here, and IMO the shorthand is fine for the differences this document seeks to cover.

mflendrich: With `networking.k8s.io/v1beta1` explicitly called out in the document, I'm on board :+1:

mflendrich: On 1.18 vs 1.19: with the current phrasing, I am expecting users coming to us with a complaint that "your docs said that I can use ingressv1 with 1.18, I tried it and I failed"

> I can't test that in our GKE cluster, as GKE 1.18 still rejects v1 Ingresses and 1.19 isn't yet available

this experience seems to indicate that confusion here (about the feature being opt-in) is a real thing.

mflendrich:
One discussion left, otherwise LGTM.

app/kubernetes-ingress-controller/1.0.x/concepts/ingress-versions.md
```diff
@@ -0,0 +1,163 @@
+---
+title: Ingress v1 and v1beta1 differences
```
lena-larionova, line 2:
```suggestion
title: Ingress v1 and v1beta1 Differences
```

app/kubernetes-ingress-controller/1.0.x/concepts/ingress-versions.md
```diff
@@ -0,0 +1,163 @@
...
+---
+
+## Introduction
+
+Kubernetes 1.18 introduced a new `networking.k8s.io/v1` API for the [Ingress resource][kubernetes-ingress-doc].
+It standardizes common practices and clarifies implementation requirements that
+were previously up to individual controller vendors. This document covers those
+changes as they relate to Kong Ingress Controller and provides example
```
lena-larionova, line 10:
```suggestion
changes as they relate to {{site.kic_product_name}} and provides sample
```
Might as well start using this variable now, in case the name changes again.

app/kubernetes-ingress-controller/1.0.x/concepts/ingress-versions.md
```diff
@@ -0,0 +1,163 @@
...
+equivalent `networking.k8s.io/v1beta1` and `networking.k8s.io/v1` resources for comparison.
+
+## Paths
+
+Both Ingress v1beta1 and v1 HTTP rules require a path, which represents a [URI
+path][uri-rfc-paths]. Although v1beta1 had specified that paths were [POSIX
+regular expressions][posix-regex] and enforced this, but in practice most
```
lena-larionova, line 17:
```suggestion
regular expressions][posix-regex] and enforced this, in practice most
```

app/kubernetes-ingress-controller/1.0.x/concepts/ingress-versions.md
```diff
@@ -0,0 +1,163 @@
...
+controllers used other other implementations that did not match the
+specification. v1 seeks to reduce confusion by introducing several i[path
```
lena-larionova, line 19:
```suggestion
specification. v1 seeks to reduce confusion by introducing several [path
```

app/kubernetes-ingress-controller/1.0.x/concepts/ingress-versions.md
```diff
@@ -0,0 +1,163 @@
...
+types][path-types] and lifting restrictions on regular expression grammars used
+by controllers.
+
+### `networking.k8s.io/v1beta1`
```
lena-larionova, line 23:
```suggestion
### networking.k8s.io/v1beta1
```
For the sidebar, mostly.

app/kubernetes-ingress-controller/1.0.x/concepts/ingress-versions.md
```diff
@@ -0,0 +1,163 @@
...
+following:
+
+```
+/foo
+/foo/
+/foobar
+/foo/bar
+```
+
+### `networking.k8s.io/v1`
```
lena-larionova, line 39:
```suggestion
### networking.k8s.io/v1
```

app/kubernetes-ingress-controller/1.0.x/concepts/ingress-versions.md
```diff
@@ -0,0 +1,163 @@
...
+
+Although v1 Ingresses provide path types with more clearly-defined logic, the
+controller must still create Kong routes and work within the Kong proxy's
+routing logic. As such, the controller translates Ingress rule paths to create
+Kong routes that match the specification:
```
lena-larionova, line 44:
```suggestion
Kong routes that match one of the following specifications: `Exact`, `Prefix`, or `ImplementationSpecific`.
```
Mostly because it's not great to leave a `:` right before a header.

app/kubernetes-ingress-controller/1.0.x/concepts/ingress-versions.md
```diff
@@ -0,0 +1,163 @@
...
+
+#### Exact
+
+If `pathType` is `Exact`, the controller creates a Kong route with a regular
+expression that matches the rule path only, e.g. an exact rule for `/foo` in an
```
lena-larionova, line 49:
```suggestion
expression that matches the rule path only. For example, an exact rule for `/foo` in an
```

app/kubernetes-ingress-controller/1.0.x/concepts/ingress-versions.md
```diff
@@ -0,0 +1,163 @@
...
+Ingress translates to a Kong route with a `/foo$` regular expression path.
+
+#### Prefix
+
+If `pathType` is `Prefix`, the controller creates a Kong route with two path
+criteria, e.g. `/foo` will create a route with a `/foo$` regular expression and
```
lena-larionova, line 55:
```suggestion
criteria. For example, `/foo` will create a route with a `/foo$` regular expression and
```

app/kubernetes-ingress-controller/1.0.x/concepts/ingress-versions.md
```diff
@@ -0,0 +1,163 @@
...
+`/foo/` plain path.
+
+#### ImplementationSpecific
+
+The controller leaves ImplementationSpecific path rules entirely up to the Kong
```
lena-larionova, line 60:
```suggestion
The controller leaves `ImplementationSpecific` path rules entirely up to the Kong
```

app/kubernetes-ingress-controller/1.0.x/concepts/ingress-versions.md
```diff
@@ -0,0 +1,163 @@
...
+router. It creates a route with the exact same path string as the Ingress rule.
+
+## Ingress class
+
+[Ingress class][ingress-class] indicates which resources an ingress controller
+should process. It provides a means to separate out configuration intended for
+other controllers or other instances of the Kong Ingress Controller.
```
lena-larionova, line 67:
```suggestion
other controllers or other instances of the {{site.kic_product_name}}.
```

app/kubernetes-ingress-controller/1.0.x/concepts/ingress-versions.md
```diff
@@ -0,0 +1,163 @@
...
+[existing support for them][kong-wildcard-hostnames].
+
+As such, while the v1beta1 specification did not officially support wildcard
+hostnames, you can use wildcard hostnames with either version. Setting a
+hostname like `*.example.com` will match requests for both `foo.example.com`
+and `bar.example.com` with either v1 or v1beta1 Ingresses.
+
+## Backend types
+
+Ingress v1 introduces support for backends other than Kubernetes Services, via
```
lena-larionova, line 141:
```suggestion
Ingress v1 introduces support for backends other than Kubernetes Services through
```

app/kubernetes-ingress-controller/1.0.x/concepts/ingress-versions.md
```diff
@@ -0,0 +1,163 @@
...
+[resource backends][resource-backends].
+
+Kong has support for routes without a service in some cases, such as routes
```
lena-larionova, line 144:
```suggestion
Kong has support for Routes without a Service in some cases, such as Routes
```

app/kubernetes-ingress-controller/1.0.x/concepts/ingress-versions.md
```diff
@@ -0,0 +1,163 @@
...
+that use the [AWS Lambda plugin][lambda-plugin]. However, it does not yet offer
```
lena-larionova, line 145: I don't know if it's a good idea to phrase something as "does not yet do this thing", since that sounds like you're promising a feature change in the future. Whether this feature is going to be available or not is not super relevant - timelines and roadmaps change, so I'd be wary of any pseudo-promises.

app/kubernetes-ingress-controller/1.0.x/concepts/ingress-versions.md
```diff
@@ -0,0 +1,163 @@
...
+Kubernetes resources for them. As such, you should instead create a placeholder
+Kubernetes Service for them, using an [ExternalName Service][external-name]
+with an RFC 2606 invalid hostname, e.g. `kong.invalid`. You can use these
```
lena-larionova, line 148:
```suggestion
with an RFC 2606 invalid hostname (for example,`kong.invalid`). You can use these
```

app/kubernetes-ingress-controller/1.0.x/concepts/ingress-versions.md
```diff
@@ -0,0 +1,163 @@
...
+regular expressions][posix-regex] and enforced this, but in practice most
+controllers used other other implementations that did not match the
+specification. v1 seeks to reduce confusion by introducing several i[path
+types][path-types] and lifting restrictions on regular expression grammars used
+by controllers.
+
+### `networking.k8s.io/v1beta1`
+
+The controller passes paths directly to Kong and relies on its [path handling
+logic][kong-path]. The Kong proxy treats paths as a prefix unless they include
```
lena-larionova, line 26:
```suggestion
logic][kong-paths]. The Kong proxy treats paths as a prefix unless they include
```

app/kubernetes-ingress-controller/1.0.x/concepts/ingress-versions.md
```diff
@@ -0,0 +1,163 @@
...
+
+Kong has support for routes without a service in some cases, such as routes
+that use the [AWS Lambda plugin][lambda-plugin]. However, it does not yet offer
+Kubernetes resources for them. As such, you should instead create a placeholder
+Kubernetes Service for them, using an [ExternalName Service][external-name]
+with an RFC 2606 invalid hostname, e.g. `kong.invalid`. You can use these
+placeholder services with either v1 or v1beta1 Ingresses.
+
+[kubernetes-ingress-doc]: https://kubernetes.io/docs/concepts/services-networking/ingress/
+[ingress-class]: ./ingress-classes.md
```
lena-larionova, line 152:
```suggestion
[ingress-class]: /kubernertes-ingress-controller/{{page.kong_version}}/concepts/ingress-classes
```

app/kubernetes-ingress-controller/1.0.x/concepts/ingress-versions.md
```diff
@@ -0,0 +1,163 @@
...
+
+In v1beta1, ingress class was handled informally using
+`kubernetes.io/ingress.class` [annotations][deprecated-annotation]. v1
+introduces a new [IngressClass resource][ingress-class-api] which provides
+richer information about the controller. v1 Ingresses are bound to a class via
+their `ingressClassName` field.
+
+For example, consider this v1beta1 Ingress:
+
+```
```
lena-larionova, line 77:
```suggestion
```yaml
```

app/kubernetes-ingress-controller/1.0.x/concepts/ingress-versions.md
```diff
@@ -0,0 +1,163 @@
...
+
+Its ingress class annotation is set to `kong`, and ingress controllers set to
+process `kong` class Ingresses will process it.
+
+In v1, the equivalent configuration declares a `kong` IngressClass resource
+whose `metadata.name` field indicates the class name. The `ingressClassName`
+value of the Ingress object must match the value of the `name` field in the
+IngressClass metadata:
+
+```
```
lena-larionova, line 103:
```suggestion
```yaml
```

lena-larionova (commented):
Mostly grammar, some variables, and link fixes.

app/kubernetes-ingress-controller/1.0.x/concepts/ingress-versions.md
rainest: Fair enough. Changed to 1.19 on the assumption that beta users are savvy enough to understand the distinction and know that they've enabled the gate, and that less savvy users will just use 1.19.

app/kubernetes-ingress-controller/1.0.x/concepts/ingress-versions.md
rainest: Yeah--we're in a weird position where we kinda supported something akin to resource backends through our existing features, and the question of how to write K8S configuration for those even though there's no real appropriate Service definition does come up in practice (recently https://kongstrong.slack.com/archives/CA42V003B/p1606771633206900).

We need to cover the workaround that enables those use cases (dummy Services) somewhere while acknowledging their relation to the new concept introduced in Ingress v1 that we don't support. Not sure where that will go in the future (since K8S' own Service APIs roadmap means many current Ingress concepts will change significantly); have reworded to say that we don't support any service backend stuff, but that you should use the dummy Services where appropriate.

app/kubernetes-ingress-controller/1.0.x/concepts/ingress-versions.md
```diff
@@ -0,0 +1,164 @@
...
+Kong does not support any dedicated resource backend configurations, though it
+does have support for Routes without Services in some cases (for example, when
+using the [AWS Lambda plugin][lambda-plugin]). For these routes, you should
+create a placeholder Kubernetes Service for them, using an [ExternalName
+Service][external-name] with an RFC 2606 invalid hostname, e.g.
+`kong.invalid`. You can use these placeholder services with either v1 or
+v1beta1 Ingresses.
+
+[kubernetes-ingress-doc]: https://kubernetes.io/docs/concepts/services-networking/ingress/
+[ingress-class]: /kubernertes-ingress-controller/{{page.kong_version}}/concepts/ingress-classes
```
lena-larionova, line 153:
```suggestion
[ingress-class]: /kubernetes-ingress-controller/{{page.kong_version}}/concepts/ingress-classes
```
Typo, my bad.

lena-larionova (approved):
LGTM, assuming the link typo is fixed.

rainest:
Small late edition in 1702261 to warn about the problem encountered in https://github.com/Kong/kubernetes-ingress-controller/issues/958

The core proxy behavior there is bizarre, but there shouldn't be cases where you'd truly need the other path types along with a regular expression path: not transforming those and leaving the expression up to users familiar with writing them should be the best course of action there, even if capture groups aren't involved.

PR #2482, opened 2020-12-01
Initial Draft on improving Kong with Kubernetes documentation

marcoam:
@felderi / @nedward - Here you will find a PR on the initial stab at helping with our Kong with Kubernetes documentation. Currently, this PR does the following: 
- On the "General" Install page, merges the two K8s options that the user had into one. 
- This Kubernetes option takes the user to a page where you will see the three options on how to deploy Kong in K8s: DB-less, DB-full with KIC and DB-full without KIC. Each of the options have its pros and cons.
- The links for all of these options send the user to detailed instructions on how to get their selected option up and running
- The K4K8s (aka, DB-less) option is complete

Now, there are a couple of things that still need to be done: 
- The DB-full with KIC option is still only using helm. There are no commands for kubectl
- The DB-full without KIC option is, as of now, simply a copy of the KIC option. Hence, it needs to be rewritten to address the complexities that come up when deploying without a KIC and potential solutions. Additionally, similar to above, we need to add the kubectl commands.

### DIFF
--- a/app/enterprise/2.2.x/deployment/install-kong-in-kubernetes.md
+++ b/app/enterprise/2.2.x/deployment/install-kong-in-kubernetes.md
@@ -1,0 +1,34 @@
+---
+title: Kong Enterprise in Kubernetes
+---
+
+# Install Options
+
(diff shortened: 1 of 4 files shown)


felderi:
Let's close the PR and iterate on the Branch.  Once we're in alignment we can create PR.

PR #2492, opened 2020-12-07
docs(kic) Replace product name with variables

lena-larionova:
Replaced any instances of "Kong Ingress Controller" with the variable `{{site.kic_product_name}}`. 

Since KIC is notorious for having issues with naming, this will make it easier to update everything in one go when it inevitably gets rebranded again.

Note: For page titles, did not use the variable but instead wrote out the name directly. Variables don't get parsed in browser tab labels.

https://deploy-preview-2492--kongdocs.netlify.app/kubernetes-ingress-controller/1.0.x/introduction/


tlblessing (approved):
One thing to consider is the use of long and short variables, which is what we did at Confluent. It is a bit tedious to read out the entire long product name every time. The first instance in a topic is long, subsequent are short.

lena-larionova:
> One thing to consider is the use of long and short variables, which is what we did at Confluent. It is a bit tedious to read out the entire long product name every time. The first instance in a topic is long, subsequent are short.

Good point - we did the same at SAP. For this PR, I only went through and changes instances of the full name to a variable, and any time it simply said "ingress controller", I left it alone. That could be a good candidate for the short name here. I'll note it as an improvement for the future.

PR #2494, opened 2020-12-07
docs(ee-license) Add note to use your own license for Docker installs

lena-larionova:
Based on docs ticket: https://konghq.atlassian.net/browse/DOCS-1435

Changed only Docker installation pages, as none of the other install topics include a license sample.


aaronhmiller:
Seeing a very clear NOTE in the license area. Exactly what I was looking for. This would have prevented a misunderstanding from a prospect and likely will fix many others issues as well.

aaronhmiller (approved):
Mainly looking for the callout about the license example being a SAMPLE. That is added clearly here in lines 78-81. Approved!

tlblessing (approved):

PR #2499, opened 2020-12-09
Fixed some leftover annotations from configuration.konghq.com to konghq.com/override

svenwal:
This is a migrated PR from the old Kong Ingress docs repo - see also https://github.com/Kong/kubernetes-ingress-controller/pull/954. Compared to this old text based changes already have been done, the image still has shown the old notation

**What this PR does / why we need it:**

Since 0.8 we are using konghq.com/override instead of configuration.konghq.com but some references in the docs (including a central overview graphic) still mentioned the old notation

**Special notes for your reviewer:**

Has been found by a prospect (in the way they tried it according to docs and have been unable to get method based routing to work. I have fixed their deployment and this PR should (hopefully) change all remaining references.


lena-larionova (changes requested):
@svenwal thanks for doing this! There are a couple of typos in the changed text though - and I actually just got the source files from Harry to update this image and another one. Here's a version you can upload instead: 
kong-ingress-association

lena-larionova (approved):

PR #2505, opened 2020-12-10
docs(kong-user) switching to a non-kong user

murillopaula:


app/2.2.x/kong-user.md
```diff
@@ -43,3 +43,9 @@ configuration property.
     ```sh
     kong start
     ```
+
+If running Kong as a non-root user other than the built-in `kong` user and group is desired, consider adding the user to the “kong” group with something like:
```
lena-larionova, line 47:
```suggestion
If running Kong as a custom non-root user is preferred, add the user to the `kong` group with a command similar to the following (the command may differ depending on your OS):
```
I might be wrong, but it seems like you wouldn't want to say "something other than the built-in `kong` user and group" if you're still telling them to use the `kong` group.

lena-larionova: After looking at this flow again - I suggest moving this up as an optional step 2, because they'd want to do this before starting Kong. Eg:

2. (Optional) If running Kong ...etc

So then the task becomes:
1. Switch to user
2. Optionally assign a group
3. Start Kong

app/2.2.x/kong-user.md
```diff
@@ -34,7 +34,7 @@ configuration property.
 
 ## Run Kong Gateway as a non-root user
 
-1. Switch to the `kong` user and group:
+1. Switch to the `kong` user:
```
lena-larionova, line 37:
```suggestion
1. Switch to the built-in `kong` user:
```

app/2.2.x/kong-user.md
```diff
@@ -34,7 +34,7 @@ configuration property.
...
     ```sh
     $ su kong
     ```
```
lena-larionova, line 40: I think there needs to be a line here to introduce the possibility of using a custom user, for example:

You can also switch to a custom non-root user. This will require an extra setup step.

lena-larionova (commented):
Comments all apply to both topics.

murillopaula:
@lena-larionova based on your comments, I pushed another commit with another approach. Let me know what you think. Thanks.

app/2.2.x/kong-user.md
```diff
@@ -32,9 +32,12 @@ configuration property.
 * [RedHat](/install/redhat)
 * [Ubuntu](/install/ubuntu)
 
-## Run Kong Gateway as a non-root user
+## Run Kong Gateway as the built-in kong user
+
+When {{site.ce_product_name}} is installed with a package management system such as `APT` and `YUM`, a default `kong` user and a default `kong` group are created. All the files installed by the package are owned by the `kong` user and group.
```
lena-larionova, line 37:
```suggestion
When {{site.ce_product_name}} is installed with a package management system such as `APT` or `YUM`, a default `kong` user and a default `kong` group are created. All the files installed by the package are owned by the `kong` user and group.
```

app/2.2.x/kong-user.md
```diff
@@ -43,3 +46,19 @@ configuration property.
     ```sh
     kong start
     ```
+
+## Run Kong Gateway as a custom non-root user
+
+It is also possible running Kong as a custom non-root user. Since all the files installed by the {{site.ce_product_name}} package are owned by the `kong` group, a user that belongs to that group should be permitted to perform the same operations as the `kong` user.
```
lena-larionova, line 52:
```suggestion
It is also possible to run Kong as a custom non-root user. Since all the files installed by the {{site.ce_product_name}} package are owned by the `kong` group, a user that belongs to that group should be permitted to perform the same operations as the `kong` user.
```

app/enterprise/2.2.x/deployment/kong-user.md
```diff
@@ -32,9 +32,12 @@ configuration property.
 * [RHEL](/enterprise/{{page.kong_version}}/deployment/installation/rhel)
 * [Ubuntu](/enterprise/{{page.kong_version}}/deployment/installation/ubuntu)
 
-## Run Kong Enterprise as a non-root user
+## Run {{site.ee_product_name}} as the built-in kong user
+
+When {{site.ee_product_name}} is installed with a package management system such as `APT` and `YUM`, a default `kong` user and a default `kong` group are created. All the files installed by the package are owned by the `kong` user and group.
```
lena-larionova, line 37:
```suggestion
When {{site.ee_product_name}} is installed with a package management system such as `APT` or `YUM`, a default `kong` user and a default `kong` group are created. All the files installed by the package are owned by the `kong` user and group.
```

app/enterprise/2.2.x/deployment/kong-user.md
```diff
@@ -43,3 +46,19 @@ configuration property.
     ```sh
     kong start
     ```
+
+## Run {{site.ee_product_name}} as a custom non-root user
+
+It is also possible running Kong as a custom non-root user. Since all the files installed by the {{site.ee_product_name}} package are owned by the `kong` group, a user that belongs to that group should be permitted to perform the same operations as the `kong` user.
```
lena-larionova, line 52:
```suggestion
It is also possible to run Kong as a custom non-root user. Since all the files installed by the {{site.ee_product_name}} package are owned by the `kong` group, a user that belongs to that group should be permitted to perform the same operations as the `kong` user.
```

lena-larionova (commented):
A couple of minor edits, but I really like the restructure.

murillopaula:
thx @lena-larionova when merging, please do a Squash and Merge

lena-larionova (approved):
LGTM!

PR #2517, opened 2020-12-14
Click on Add Route

theWilliamChia:
The button on the services overview page says "Add Route".


lena-larionova (approved):
LGTM, thanks for the fix.

PR #2524, opened 2020-12-22
Remove the description about Helm v2 due to it's deprecation.

ydFu:
Remove about Helm v2 due to is't deprecation.
After November 13, 2020,No further Helm v2 releases.

[ref. helm-v2-deprecation-timeline](https://helm.sh/blog/helm-v2-deprecation-timeline/)

Related documents:
* app/kubernetes-ingress-controller/1.1.x/deployment/aks.md
* app/kubernetes-ingress-controller/1.1.x/deployment/eks.md
* app/kubernetes-ingress-controller/1.1.x/deployment/gke.md
* app/kubernetes-ingress-controller/1.1.x/deployment/k4k8s-enterprise.md
* app/kubernetes-ingress-controller/1.1.x/deployment/k4k8s.md
* app/kubernetes-ingress-controller/1.1.x/deployment/minikube.md

### DIFF
--- a/app/kubernetes-ingress-controller/1.1.x/deployment/aks.md
+++ b/app/kubernetes-ingress-controller/1.1.x/deployment/aks.md
@@ -42,9 +42,6 @@ Please ensure that you've Tiller working and then execute:
 $ helm repo add kong https://charts.konghq.com
 $ helm repo update
 
-# Helm 2
-$ helm install kong/kong
-
 # Helm 3
 $ helm install kong/kong --generate-name --set ingressController.installCRDs=false
 ```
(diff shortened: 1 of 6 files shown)


lena-larionova:
Thanks for the update! Let me double-check something though:

@rainest @hbagdi Is it possible that there might be users still using Helm 2, or is this fine to remove? In the Helm changelog, they warned users to switch to Helm 3 in August, and Helm 2 is fully deprecated now. I assume that means this change is good, but wanted to pass it by you first.

hbagdi:
There certainly are users of Helm 2 but it is about time to start removing instructions for it now. Helm 2 is no longer supported and we should not have installation instructions which use software that is no longer maintained.

mflendrich (approved):
Given helm 2 end-of-life, I too think it's very clear that we should not recommend helm 2 as an installation option.

lena-larionova (approved):
In that case, full speed ahead on removing these. LGTM!

PR #2525, opened 2020-12-23
Changed url due to a maintainer change

danielkocot:


lena-larionova:
Thanks for the update! Does this mean the plugin branding also needs to change, or are you fine with leaving it as the "Reedelk transformer plugin" with the Reedelk logo as the icon?

danielkocot:
> Thanks for the update! Does this mean the plugin branding also needs to change, or are you fine with leaving it as the "Reedelk transformer plugin" with the Reedelk logo as the icon?

At the moment the plugin branding remains as it is.

lena-larionova (approved):
LGTM, thanks!

PR #2533, opened 2021-01-08
docs(ee-install) Change version variables to pull correct versions

lena-larionova:
Installation topics were pulling the latest real version of Enterprise. Instead, each version should be pulling its own most recent patch version number. Fixing 2.2.x, 2.1.x, and 1.5.x (versions that are still being actively maintained in patches).


tlblessing (approved):
Spot checked install topics for 1.5, 2.1, and 2.2 LGTM thank you @lena-larionova

PR #2534, opened 2021-01-08
docs(old versions) Adjust older versions banner

lena-larionova:
Based on docs ticket: https://konghq.atlassian.net/browse/DOCS-1518

**Issue:** Doc "browsing older version" banner links to the the landing page of the latest doc, while it should point to the same topic (if the topic exists).

Screen Shot 2021-01-07 at 2 44 05 PM
(eg see https://docs.konghq.com/1.5.x/proxy/)

**Changes:** 
* Instead of pointing to the doc landing page of the latest version, changing the link in the banner to check if a parallel page exists under `/latest/`. If the page exists, the link points to that page; if not, the link points to the doc landing page as before. 
* Rephrased the text to avoid calling out the product name, and shortened it to avoid banner wrapping onto a second line.

**Preview:**
Test by opening an older topic, then clicking on the link in the banner: 

https://deploy-preview-2534--kongdocs.netlify.app/enterprise/0.36-x/deployment/installation/docker/ 
^ This topic redirects to the latest version of the same topic.

https://deploy-preview-2534--kongdocs.netlify.app/enterprise/0.34-x/installation/docker/
^ This one redirects to docs.konghq.com/enterprise because the directory structure changed between 0.34-x and 0.36-x

### DIFF
--- a/app/_layouts/docs-v2.html
+++ b/app/_layouts/docs-v2.html
@@ -101,8 +101,7 @@ id: documentation
           {% endif %}
           {% if page.kong_version != page.kong_latest.release and page.no_version != true %}
           <div class="alert alert-warning">
-            <strong>Careful!</strong> You are browsing documentation for an outdated version.
-            See the
+            You are browsing documentation for an outdated version. See the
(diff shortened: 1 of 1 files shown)


tlblessing (approved):
Tested the links and they redirected as expected. 

Another thing to consider is the "Careful!". This is incomplete and incorrect grammar because it is an adjective. It could be "Be careful" or make it a Caution or Warning.

PR #2537, opened 2021-01-11
docs(plugins) Remove variables from front matter

lena-larionova:
Fixing my own mistake @tlblessing


tlblessing (approved):
oops I didn't look either. thanks.

PR #2540, opened 2021-01-14
Autodoc tool for updating plugin version information automatically

dndx:
This PR adds a new script that allows update of plugin version and Kong CE version information automatically.

To run it, you need to have Python 3, then setup the virtualenv:

```
python3 -m venv /path/to/docs.konghq.com/autodoc-kong-hub
source /path/to/docs.konghq.com/autodoc-kong-hub/bin/activate
pip install -r requirments.txt
```

Then you can run the tool like this:

```
python gen.py --latest-version 2.2.1 --kong-repo /path/to/kong/kong/ --kong-doc-repo /path/to/docs.konghq.com/
```

Running `python gen.py -h` will show the usage of the script.

The script is a bit hacky due to the fact that is is probably the only autodoc tool that updates instead of generating a file, so a lot of logic was involved in manipulating the file in place correctly.

There are also a list of plugins at the top of the script that had changed named over the years, the script handles these and considers them the same plugin.

There is a commit of me running the script against the `2.2.1` tag as a preview of what the script does, however we do not necessarily need to merge that commit if it is not needed. We can easily re run the script against 2.3.0 later.

We can easily make a similar script that handles the EE plugins, however let's make sure the behavior is reasonable before putting more work into it.

### DIFF
--- a/app/_hub/kong-inc/acl/index.md
+++ b/app/_hub/kong-inc/acl/index.md
@@ -1,7 +1,7 @@
 ---
 name: ACL
 publisher: Kong Inc.
-version: 1.0.0
+version: 3.0.1
 
 desc: Control which Consumers can access Services
 description: |
@@ -28,6 +28,7 @@ categories:
 kong_version_compatibility:
     community_edition:
       compatible:
+        - 2.3.x
         - 2.2.x
         - 2.1.x
         - 2.0.x
@@ -37,6 +38,7 @@ kong_version_compatibility:
         - 1.2.x
         - 1.1.x
         - 1.0.x
+        - 0.15.x
         - 0.14.x
         - 0.13.x
         - 0.12.x
@@ -46,6 +48,7 @@ kong_version_compatibility:
         - 0.8.x
         - 0.7.x
         - 0.6.x
+        - 0.5.x
         - 0.5.x
     enterprise_edition:
       compatible:
--- a/app/_hub/kong-inc/acme/index.md
+++ b/app/_hub/kong-inc/acme/index.md
@@ -1,7 +1,7 @@
 ---
 name: ACME
 publisher: Kong Inc.
-version: 0.2.13
+version: 0.2
 
 source_url: https://github.com/Kong/kong-plugin-acme
 
@@ -16,8 +16,10 @@ categories:
 kong_version_compatibility:
     community_edition:
       compatible:
+        - 2.3.x
         - 2.2.x
         - 2.1.x
+        - 2.0.x
         - 2.0.x
     enterprise_edition:
       compatible:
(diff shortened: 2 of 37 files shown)


tlblessing:
@dndx docs appreciates any efforts to automate plugin updates. However, we can see some issues with this approach. When there is a version bump, the index file is renamed to its last version, and a new index file is created as is documented on our wiki https://konghq.atlassian.net/wiki/spaces/KD/pages/1123254328/Add+and+Update+a+Plugin. This approach does not update the versions.yml file, which can cause 404 issues from the Version selection list in the plugin hub.

eskibars:
I'd like to revisit this now that we're past the initial Konnect launch.  I'm a big fan of doing something like this: there are far too many manual steps in our release process.  Let's try to work out the specific issues.

> This script approach does not update the versions.yml file, which can cause 404 issues from the Version selection list in the plugin hub. 

@dndx I've just learned from @lena-larionova how these things work and I think we can probably expand this PR to fix some of the underlying problems.  Today, versions.yml files exist on a per-plugin basis, e.g. https://github.com/Kong/docs.konghq.com/blob/release/2.2/app/_data/extensions/kong-inc/session/versions.yml .  It turns out these are read at https://github.com/Kong/docs.konghq.com/blob/master/app/_layouts/extension.html#L550-L562 .  If the script were to be modified to help with the versions problem, I think we'd be able to get over this issue.  @tlblessing do you agree?

> With the version in the index file, we have been trying to accommodate multiple versions by listing it as 2.3.x for example, instead of an exact patch number such as done in this PR for ACL version: 3.0.1
> The main versioned docs also have an x extension, such as 2.3.x, for the same maintenance reasons that can alleviate doc source bloat.

It would be good to retain the "across all patch versions in this minor" but I think we do need to have some flexibility around having patch-specific versions as well.  For example, even in the best cases, I can imagine some situation where we have to put out a new patch version of the gateway and a new patch version of some plugin which provides for some new optional parameter.  These should be very rare (though they aren't as rare as we'd like them to be today), but I doubt they'll ever 100% go away and we'll want some flexibility around this option

I don't know what the best technical solution for this is -- I can imagine us e.g. having a .x to indicate all versions or .1/.2/... if/when we have to.

> We need to determine if we are going to drop versioning from plugins and only use the most current version

We won't be dropping versioning from plugins, although there is an option on the table to tie the plugin version to the gateway versions, which might make things more understandable.  That still won't solve the versions problem (e.g. we'll probably still have very rare patch versions that add a new parameter) but may help with certain navigation issues if we do decide to incorporate that.

Bradamant3:
Closing in favor of
- solution planned elsewhere more recently
- tracking any ideas we still need to consider outside a PR

PR #2543, opened 2021-01-20
docs(2.3.x) release of 2.3.x docs

kikito:


app/_data/kong_versions.yml
```diff
@@ -387,6 +401,10 @@
   release: "2.2.x"
   version: "2.2"
   edition: "getting-started-guide"
+-
+  release: "2.3.x"
+  version: "2.3"
+  edition: "getting-started-guide"
```
lena-larionova, line 407: Looks like this is breaking the build, because the /getting-started-guide folder and nav weren't cloned to 2.3.x. Can you either remove this entry for now, or clone the folder and nav to 2.3.x?

kikito:
@lena-larionova thank you for the pointers, I added the missing files and now it is green.

On that note, I think it would be useful to have some sort of "list of steps" needed in order to make a new release. It begun being copying and renaming a single folder, but it has grown into its own little bundle of complexity.

lena-larionova:
> On that note, I think it would be useful to have some sort of "list of steps" needed in order to make a new release. It begun being copying and renaming a single folder, but it has grown into its own little bundle of complexity.

You're absolutely right. Working on a way to simplify it as well, but I'll make sure to update our release doc and share it with you guys - right now we keep one on the docs wiki, but it's for enterprise releases only. Nothing about scripts for the autodocs, or any of those pieces.

lena-larionova (approved):
In the interest of time and successful builds, approving this one to get it published whenever it needs to be. Will work out any wrinkles with the getting started guide and the enterprise docs later.

kikito:
@lena-larionova on #2057 you mentioned

> We normally create Enterprise branches in our private repo

It seems to me that this PR currently contains both CE and EE docs (copies of 2.2.0). Should I perhaps remove the EE ones and leave only CE-relevant docs here?

lena-larionova:
(@kikito accidentally edited your comment at first. Pre-coffee work is bad...)

> It seems to me that this PR currently contains both CE and EE docs (copies of 2.2.0). Should I perhaps remove the EE ones and leave only CE-relevant docs here?

I would normally say yes, but we're crunched for time on our side at the moment. In the preview, the 2.3.x EE docs don't appear in the dropdown and are only reachable through a direct URL - so I think we can let this one slide.

Elyahou:
It seems 2.3 introduces some breaking changes on the go plugins that are not reflected in documentation...

PR #2551, opened 2021-01-26
Removed all content related to /entities endpoint as it was removed from this version

HCloward:
I removed all of the content related to the `/entities` endpoint from version 2.1.x because according to [DOCS-1365](https://konghq.atlassian.net/browse/DOCS-1365), this endpoint was removed from this version of the product and it is misleading. The Jira issue only lists the reference page as needing updating, but I found that the examples page also had some use of the `/entities` endpoint and also needed updating. 

@aaronhmiller Can you tell me if the tutorial makes sense with the content I removed or if more updates need to be made? I wasn't sure if there was a way or what that way would be to add entities to workspaces that did not use the `/entities` endpoint . . . 

[Workspace Reference](https://deploy-preview-2551--kongdocs.netlify.app/enterprise/2.1.x/admin-api/workspaces/reference/)
[Workspace Examples](https://deploy-preview-2551--kongdocs.netlify.app/enterprise/2.1.x/admin-api/workspaces/examples/)


aaronhmiller:
I've reviewed the PR and it looks good. Should help cut down on misunderstandings about the `/entities` option not being available.

One suggestion. When I explain `workspaces` to prospects, it seems to help when I tell them it's the same concept as `namespaces` and they immediately grok the purpose. If you think it'll add clarity, you might want to add that here too.

lena-larionova:
@aaronhmiller is the /entities endpoint gone completely, or just from workspaces? For example, the RBAC topics refer to /entities a fair bit: 

https://deploy-preview-2551--kongdocs.netlify.app/enterprise/2.1.x/admin-api/rbac/reference/#add-a-role-entity-permission

https://docs.konghq.com/enterprise/2.2.x/kong-manager/administration/rbac/add-role/

And to double-check: it's gone from 2.1.x onward, correct? Not 2.1.x alone.

FYI @HCloward: a local search for the /entities endpoint will give you far better results than using the Algolia search bar on the site.

HCloward:
> @aaronhmiller is the /entities endpoint gone completely, or just from workspaces? For example, the RBAC topics refer to /entities a fair bit:
> 
> https://deploy-preview-2551--kongdocs.netlify.app/enterprise/2.1.x/admin-api/rbac/reference/#add-a-role-entity-permission
> 
> https://docs.konghq.com/enterprise/2.2.x/kong-manager/administration/rbac/add-role/
> 
> And to double-check: it's gone from 2.1.x onward, correct? Not 2.1.x alone.
> 
> FYI @HCloward: a local search for the /entities endpoint will give you far better results than using the Algolia search bar on the site.

@lena-larionova According to the comments on [FTI-1930](https://konghq.atlassian.net/browse/FTI-1930) which was referenced in the docs issue as the cause, only workspace entities endpoint was removed. However I probably should update the later versions as well. Will update.

lena-larionova (approved):
LGTM!

PR #2559, opened 2021-02-01
Update algolia config for KE

lena-larionova:
With Kong Enterprise getting removed from the tabs for Konnect, there's no top-level selector for Enterprise pages. 

Adding a selector and adjusting EE config file to match. Will be switching all content to this top-level hidden selector in the future.


GayleNeumann (approved):
LGTM

PR #2560, opened 2021-02-01
docs(algolia) Config updates and new full site config

lena-larionova:
### Changes
* Update `config-mesh` and `config-deck` to use the new selector class, removing the need for specifying URLs to exclude
* Remove `config-studio` as that doc set no longer exists
* Introduce `config-full-docs` to index the whole doc site; select only latest versions of every doc set, including plugins
 
Note for reviewers: There is one outstanding things I haven't fully worked out in the `full-docs` config file:
  *  I'd like to be able to select the latest version, but I'd not sure how, since this file is outside the `/app` directory. Environment variables maybe? 
 
### Previews
Not helpful, you won't see anything in them.


GayleNeumann (approved):
LGTM

PR #2562, opened 2021-02-02
docs(konnect) Removing khcp, replacing with konnect

lena-larionova:
Removing straggling instances of `khcp` in URLs that got missed in the launch.


GayleNeumann (approved):
LGTM

PR #2564, opened 2021-02-03
fix syntax error of kubernetes manifest in example

tydra-wang:
`apiVersion` missed in kubernetes manifest.


lena-larionova (approved):
LGTM, thanks!

PR #2566, opened 2021-02-04
fix(search) Adjust plugin hub value for algolia

lena-larionova:
Minor adjustments, as currently the "Plugin Hub" label isn't being picked up properly, and the plugins get categorized as "Kong" instead.


GayleNeumann (approved):
LGTM

PR #2571, opened 2021-02-08
About Kong.md

Rahul-160:
I have added a brief introduction about Kong.Inc as a new file.


lena-larionova:
Thanks for your contribution, but this repo is for the Kong docs site only. This information doesn't belong here.

PR #2574, opened 2021-02-09
Konnect Public Dev Portal

tlblessing:
https://konghq.atlassian.net/browse/DOCS-1567 per https://konghq.atlassian.net/browse/KHCP-646

I think the topics should be separate (from enable/disable dev and app reg) for SEO purposes. And to avoid confusion. I think the app and dev reg toggles should not be available if Public Portal is on. If someone enables public toggle, then that should disable enabled app and dev reg toggles. If nothing is enabled, selecting Public should make the disabled app/dev reg toggles unavailable. That would be typical expected behaviors _if_ I understand the features correctly.

This is a last minute ticket so more information appreciated.

Discussed with Kai and logged https://konghq.atlassian.net/browse/KHCP-1057.

Direct review link:

https://deploy-preview-2574--kongdocs.netlify.app/konnect/dev-portal/administrators/public-portal/


app/konnect/dev-portal/administrators/public-portal.md
```diff
@@ -0,0 +1,26 @@
+---
+title: Public Developer Portal
+no_version: true
+---
+
+Enable public access to your Dev Portal. When Public Portal is enabled, developers
+don't need to register for the portal access and any new application registrations aren't
+allowed. The Register options are not available in the portal.
```
lena-larionova, line 8:
```suggestion
Enable public access to your Dev Portal. With **Public Portal** enabled, anyone 
can access your Dev Portal without a login. Developers don't need to register 
for portal access and new application registrations aren't allowed. 

The Register options are not available in a public-facing portal.
```
Calling out that this disables portal login _in general_ - this was a common point of confusion from the bug bash right before launch, as an admin would try to access the portal and find themselves needing a specific dev portal account.

app/konnect/dev-portal/administrators/public-portal.md
```diff
@@ -0,0 +1,26 @@
+---
+title: Public Developer Portal
+no_version: true
+---
+
+Enable public access to your Dev Portal. When Public Portal is enabled, developers
+don't need to register for the portal access and any new application registrations aren't
```
lena-larionova, line 7: > new application registrations aren't allowed

Is there any security risk with existing application registrations? Or even if not, maybe this is something the admin should know: enabling public access means that app reg is disabled, but any existing registered applications will continue to work.

app/konnect/dev-portal/administrators/public-portal.md
```diff
@@ -0,0 +1,26 @@
...
+allowed. The Register options are not available in the portal.
+
+![Konnect Settings for a Public Portal](/assets/images/docs/konnect/konnect-portal-auto-approve.png)
+
+## Enable a Public Dev Portal
```
lena-larionova, line 12: This reads a bit awkwardly. The goal/action is to make your Dev Portal public, right? So maybe the title should be "Make a Dev Portal public" or something along those lines.

Same for the title below: "Disable public access to a Dev Portal"? As a user, I'd read the title below as disabling a public Dev Portal entirely, ie taking it down, which isn't the case.

app/konnect/dev-portal/administrators/public-portal.md
tlblessing: > > new application registrations aren't allowed
> 
> Is there any security risk with existing application registrations? Or even if not, maybe this is something the admin should know: enabling public access means that app reg is disabled, but any existing registered applications will continue to work.

I'm not sure if there are any security risks. @skibum would have to answer that question. Also it seems confusing to say app reg is disabled when the button can stay enabled!

app/konnect/dev-portal/administrators/public-portal.md
tlblessing: > This reads a bit awkwardly. The goal/action is to make your Dev Portal public, right? So maybe the title should be "Make a Dev Portal public" or something along those lines.
> 
> Same for the title below: Change "Disable a Public Dev Portal" to "Disable public access to a Dev Portal"? As a user, I'd read the title below as disabling a public Dev Portal entirely, ie taking it down, which isn't the case.

To be consistent with the remainder of the portal docs, we should stick with "Enable" rather than "Make".  So went with "Enable public access to a Dev Portal" and "Disable  Public Access to a Dev Portal" as you suggested. Thanks Lena!

app/konnect/dev-portal/administrators/public-portal.md
skibum: There is not a security risk for previously approved applications.

tlblessing: Thanks for your response @skibum.

lena-larionova (approved):
LGTM

kaiarrowood (approved):
Looks good to me too!

PR #2575, opened 2021-02-10
Docs/konnect advanced runtime

lena-larionova:
Docs ticket: https://konghq.atlassian.net/browse/DOCS-1572

Based on hybrid mode doc and eng PR: 
https://docs.konghq.com/enterprise/2.2.x/deployment/hybrid-mode-setup/
https://github.com/Kong/khcp/pull/994

### Changes
* Adding instructions for using the Advanced tab in Runtime Manager with Docker and with kong.conf
* Docker option (default advanced setup): including sample code for either new or existing instances. Need this double-checked, as I couldn't test it.

### Preview
https://deploy-preview-2575--kongdocs.netlify.app/konnect/runtime-manager/kong-gateway-runtime/


app/konnect/runtime-manager/kong-gateway-runtime.md
```diff
@@ -59,11 +58,186 @@ representative for access.
...
+      KONG_DATABASE=off \
+      KONG_ANONYMOUS_REPORTS=off \
+      KONG_CLUSTER_MTLS=pki \
+      KONG_CLUSTER_CONTROL_PLANE=<example.cp.konnect.foo>:443 \
+      KONG_CLUSTER_SERVER_NAME=<kong-cpoutlet-example.service> \
+      KONG_CLUSTER_TELEMETRY_ENDPOINT=<example.tp.konnect.foo>:443 \
+      KONG_CLUSTER_TELEMETRY_SERVER_NAME=<kong-telemetry-example.service> \
+      KONG_CLUSTER_CERT=/<path-to-file>/cluster.crt \
+      KONG_CLUSTER_CERT_KEY=/<path-to-file>/cluster.key \
+      KONG_LUA_SSL_TRUSTED_CERTIFICATE=/<path-to-file>/ca.crt \
```
lena-larionova, line 131: @artomal In the hybrid mode instructions, we also include `KONG_CLUSTER_CA_CERT`. According to the [Admin API doc](https://docs.konghq.com/enterprise/2.2.x/property-reference/#cluster_ca_cert), this is required when `cluster_mtls` is set to `pki`, but I didn't see it in the eng PR and previews. Is it no longer required?

artomal: I think if you include it in `cluster_cert` you don't need to set `cluster_ca_cert` correct me if I am wrong @fffonion

app/konnect/runtime-manager/kong-gateway-runtime.md
```diff
@@ -59,11 +58,186 @@ representative for access.
     <!-- and the tag in the **Node Status** column should say
     **Connected**.-->
 
+{% endnavtab %}
+{% navtab Advanced %}
+
+1. From the left navigation menu, open **Runtimes**.
```
HCloward, line 64:
```suggestion
1. From the left navigation menu, select **Runtimes**.
```

app/konnect/runtime-manager/kong-gateway-runtime.md
```diff
@@ -59,11 +58,186 @@ representative for access.
...
+
+6. On the **Configure New Runtime** page, click **Done** to go to the Runtime
+Manager overview.
+
+    The Runtime Manager will include a new entry for your instance.
+
+{% endnavtab %}
+{% endnavtabs %}
+
+{% endnavtab %}
```
HCloward, line 212: Should there be two sets of these ?

HCloward (approved):
LGTM

app/konnect/runtime-manager/kong-gateway-runtime.md
lena-larionova: yep - it's closing two different navtabs. Usually I put a comment in here to say something like `<!-- end of codeblock navtabs -->`, and I didn't this time, so I'll add one in :)

app/konnect/runtime-manager/kong-gateway-runtime.md
fffonion: If this is for bring-your-own-data-plane then `cluster_ca_cert` is still needed, `cluster_cert` is the client cert DP present to CP, and `cluster_ca_cert` is the CA for DP to verify CP.

Also we should use `KONG_LUA_SSL_TRUSTED_CERTIFICATE=system,/<path-to-file>/ca.crt` for including OS default trusted CA as well.

fffonion: If this is for bring-your-own-data-plane then `cluster_ca_cert` is not needed, `cluster_cert` is the client cert DP present to CP, and `cluster_ca_cert` is the CA for CP to verify DP.

Also we should use `KONG_LUA_SSL_TRUSTED_CERTIFICATE=system,/<path-to-file>/ca.crt` for including OS default trusted CA as well.

lena-larionova: Is `system` just that, ie a user would literally include `system` in the value? Not a variable?

artomal: @fffonion I think you added system by mistake no? I usually add just a path like in the example like this `KONG_LUA_SSL_TRUSTED_CERTIFICATE=/<path-to-file>/ca.crt` and it works just fine at least for BYODP case.

fffonion: yes `system` is just literally the value. it's the default value starting  2.2 (https://docs.konghq.com/gateway-oss/2.2.x/configuration/#lua_ssl_trusted_certificate). adding `system` does not influence hybrid mode, but some use case that requires lua land ssl verify will need the `system` to be set (acme plugin for example). so it's always a good pratice to append new path after `system`.

PR #2580, opened 2021-02-11
docs(ee/admin-api) Adding example of workspace segmentation

tyler-ball:
When I first started using the enterprise Admin API, the docs did not
make it clear that I was using the `default` workspace implicitly.
Adding a note to the Admin API docs to call this out and show users how
to explicitly select a workspace.

Signed-off-by: Tyler Ball <tyler.ball@konghq.com>

### DIFF
--- a/app/enterprise/2.3.x/admin-api/index.md
+++ b/app/enterprise/2.3.x/admin-api/index.md
@@ -737,6 +737,31 @@ HTTP 200 OK
 
 ---
 
+## Workspace Segmentation
+
+The following Admin API endpoints access the `default` workspace by default. To
+explicitely access a different workspace, prepend the endpoint with the
+workspace name in question:
+
+```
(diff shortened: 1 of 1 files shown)


app/enterprise/2.3.x/admin-api/index.md
```diff
@@ -735,6 +735,31 @@ HTTP 200 OK
       reflect the health of the database itself.
 
 
+---
+
+## Workspace Segmentation
```
tyler-ball, line 740: Does this seem like a good title for this subsection? I'm open to suggestions

tyler-ball:
If this seems like a good addition, I can also add this documentation to older Kong versions. Thoughts?

tyler-ball:
Closing this as recommended by another engineer because the workspaces page does contain this information

PR #2588, opened 2021-02-17
fix(jekyll) Adding a scope for the enterprise directory

lena-larionova:
Something is messing with our analytics data. It might be this, as this was the only metadata change to occur on April 6th last year, the day before analytics went messy.

Changes: setting a scope for the /enterprise path and removing an obsolete docs/ee path.

No preview, just check that nothing's broken.

### DIFF
--- a/jekyll-dev.yml
+++ b/jekyll-dev.yml
@@ -91,6 +91,11 @@ defaults:
       layout: 'docs-v2'
 
   - scope:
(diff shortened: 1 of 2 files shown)


tlblessing (approved):
I'm not sure what to check for this PR but I assume you'll be testing out results in analytics.

PR #2593, opened 2021-02-17
Proxy Cache plugin datatypes

tlblessing:
https://konghq.atlassian.net/browse/DOCS-1396

Adding for most current version only. Murillo and I already updated advanced version of this plugin so just copied that over.

Schema link:

https://github.com/Kong/kong-plugin-proxy-cache/blob/master/kong/plugins/proxy-cache/schema.lua

Direct review link:

https://deploy-preview-2593--kongdocs.netlify.app/hub/kong-inc/proxy-cache/

### DIFF
--- a/app/_hub/kong-inc/proxy-cache/index.md
+++ b/app/_hub/kong-inc/proxy-cache/index.md
@@ -43,7 +43,7 @@ params:
       required:
       default: 200, 301, 404
       value_in_examples: [ "200" ]
-      datatype: array of type integer
+      datatype: array integer elements
       description: |
         Upstream response status code considered cacheable.
     - name: request_method
--- a/app/_hub/kong-inc/proxy-cache/index.md
+++ b/app/_hub/kong-inc/proxy-cache/index.md
@@ -42,61 +42,71 @@ params:
     - name: response_code
       required:
       default: 200, 301, 404
-      value_in_examples:
+      value_in_examples: [ "200" ]
+      datatype: array of type integer
       description: |
         Upstream response status code considered cacheable.
     - name: request_method
       required:
       default: '`["GET","HEAD"]`'
-      value_in_examples:
+      value_in_examples: ["GET","HEAD"]
+      datatype: array of string elements
       description: |
         Downstream request methods considered cacheable.
     - name: content_type
       required:
       default: '`["text/plain", "application/json"]`'
-      value_in_examples:
+      value_in_examples: ["text/plain", "application/json"]
+      datatype: array of string elements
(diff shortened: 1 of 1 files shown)


app/_hub/kong-inc/proxy-cache/index.md
```diff
@@ -42,61 +42,71 @@ params:
     - name: response_code
       required:
       default: 200, 301, 404
-      value_in_examples:
+      value_in_examples: [ "200" ]
+      datatype: array of type integer
```
murillopaula, line 46: this way it's consistent with the already existing convention of `array of string elements`

```suggestion
      datatype: array integer elements
```

murillopaula (changes requested):

app/_hub/kong-inc/proxy-cache/index.md
```diff
@@ -42,61 +42,71 @@ params:
...
+      value_in_examples: ["GET","HEAD"]
+      datatype: array of string elements
       description: |
         Downstream request methods considered cacheable.
     - name: content_type
       required:
       default: '`["text/plain", "application/json"]`'
-      value_in_examples:
+      value_in_examples: ["text/plain", "application/json"]
+      datatype: array of string elements
```
lena-larionova, line 60: In other plugins (for example, key auth), you've used the construction "array of strings". I'm partial to that one - it's shorter, and still clear. Whichever option you go with though, ideally we should keep it consistent.

tlblessing: > In other plugins (for example, key auth), you've used the construction "array of strings". I'm partial to that one - it's shorter, and still clear. Whichever option you go with though, ideally we should keep it consistent.

@murillopaula has sated he prefers "elements" I think it depends on the plugin schema but agree we need to be consistent

murillopaula: `array of strings` is good, but we're already using the `array of string elements` convention... so my vote is to stay with the latter option for consistency and less work

murillopaula (approved):

app/_hub/kong-inc/proxy-cache/index.md
tlblessing: > `array of strings` is good, but we're already using the `array of string elements` convention... so my vote is to stay with the latter option for consistency and less work

Agree that it is more descriptive and preferable. It is already sprinkled all over beyond basic auth plugin.

lena-larionova (approved):
LGTM

PR #2595, opened 2021-02-17
Kafka Logging plugin datatypes

tlblessing:
https://konghq.atlassian.net/browse/DOCS-1396

Adding for most current version only. 

I had wanted to change the doc versions to match the handler versions but they are 0.1.0 and 0.1.1 respectively, so essentially 0.1.x. I did a diff on the doc source files and there are significant differences between the index and 1.3-x (api versions, protocols). 

Note: Kafka Upstream plugin had its install info removed as it is bundled with Enterprise. I think the same is true for this plugin. Will log another ticket after verifying. Logged follow-up ticket https://konghq.atlassian.net/browse/DOCS-1591.

Schema link:

https://github.com/Kong/kong-plugin-kafka-log/blob/master/kong/plugins/kafka-log/schema.lua

Direct review link:

https://deploy-preview-2595--kongdocs.netlify.app/hub/kong-inc/kafka-log/

### DIFF
--- a/app/_hub/kong-inc/kafka-log/1.3-x.md
+++ b/app/_hub/kong-inc/kafka-log/1.3-x.md
@@ -1,6 +1,7 @@
 ---
 name: Kafka Log
 publisher: Kong Inc.
+version: 1.3-x # 0.1.0 internal handler
 
 desc: Publish logs to a Kafka topic.
 description: |
--- a/app/_hub/kong-inc/kafka-log/index.md
+++ b/app/_hub/kong-inc/kafka-log/index.md
@@ -35,7 +35,7 @@ params:
       default:
       datatype: set of record elements
       description: |
-        List of bootstrap brokers in a `{host: host, port: port}` format.
+        Set of bootstrap brokers in a `{host: host, port: port}` list format. For examples, see the Quickstart section.
     - name: topic
       required: true
       value_in_examples: TOPIC
--- a/app/_hub/kong-inc/kafka-log/index.md
+++ b/app/_hub/kong-inc/kafka-log/index.md
@@ -130,7 +130,8 @@ $ cd /path/to/kong/plugins/kong-plugin-kafka-log
 $ luarocks make *.rockspec
 ```
 
-In both cases, you need to change your Kong [`plugins` configuration option](https://docs.konghq.com/2.3.x/configuration/#plugins)
+In both cases, you need to change your Kong `plugins`
+[configuration option](https://docs.konghq.com/latest/configuration/#plugins)
 to include this plugin:
 
 ```
--- a/app/_hub/kong-inc/kafka-log/index.md
+++ b/app/_hub/kong-inc/kafka-log/index.md
@@ -5,7 +5,10 @@ version: 1.5.x # 0.1.1 internal handler
 
 desc: Publish logs to a Kafka topic
 description: |
-   Publish request and response logs to a [Kafka](https://kafka.apache.org/) topic.
+   Publish request and response logs to an [Apache Kafka](https://kafka.apache.org/) topic.
+   For more information, see [Kafka topics](https://kafka.apache.org/documentation/#intro_concepts_and_terms).
+
+   Kong also provides a Kafka plugin for request transformations. See [Kafka Upstream](/hub/kong-inc/kafka-upstream/).
 
 type: plugin
 enterprise: true
@@ -21,7 +24,6 @@ kong_version_compatibility:
         - 2.2.x
         - 2.1.x
         - 1.5.x
-
 
 params:
 
--- a/app/_hub/kong-inc/kafka-log/index.md
+++ b/app/_hub/kong-inc/kafka-log/index.md
@@ -61,8 +61,9 @@ params:
       value_in_examples: PRODUCER_REQUEST_ACKS
       datatype: integer
       description: |
-        The number of acknowledgments the producer requires the leader to have received before considering a request complete.
-        Allowed values: 0 for no acknowledgments, 1 for only the leader, and -1 for the full ISR.
+         The number of acknowledgments the producer requires the leader to have received before
+         considering a request complete. Allowed values: 0 for no acknowledgments; 1 for only the leader;
+         and -1 for the full ISR (In-Sync Replica set).
     - name: producer_request_timeout
       required: false
       default: "`2000`"
--- a/app/_hub/kong-inc/kafka-log/index.md
+++ b/app/_hub/kong-inc/kafka-log/index.md
@@ -177,11 +177,11 @@ $ curl -X POST http://kong:8001/plugins \
 
 ## Log Format
 
-Similar to [HTTP Log Plugin](https://docs.konghq.com/hub/kong-inc/http-log#log-format).
+Similar to the [HTTP Log Plugin](https://docs.konghq.com/hub/kong-inc/http-log#log-format).
 
 ## Implementation details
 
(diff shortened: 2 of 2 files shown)


app/_hub/kong-inc/kafka-log/index.md
```diff
@@ -32,87 +33,103 @@ params:
       value_in_examples: BOOTSTRAP_SERVERS
       urlencode_in_examples: true
       default:
+      datatype:
```
tlblessing, line 36: @murillopaula I'm not sure about this one. 

Also, is the description correct? Should `{host: host, port: port}` be `{host:port, host: port}`?

app/_hub/kong-inc/kafka-log/index.md
```diff
@@ -32,87 +33,103 @@ params:
...
 Manually download and install:
+
 ```
 $ git clone https://github.com/kong/kong-plugin-kafka-log.git /path/to/kong/plugins/kong-plugin-kafka-log
 $ cd /path/to/kong/plugins/kong-plugin-kafka-log
 $ luarocks make *.rockspec
 ```
 
-In both cases, you need to change your Kong [`plugins` configuration option](https://docs.konghq.com/1.3.x/configuration/#plugins)
+In both cases, you need to change your Kong [`plugins` configuration option](https://docs.konghq.com/2.3.x/configuration/#plugins)
```
murillopaula, line 132: is there a way to not hardcode the version here? otherwise we'll always need to change it... maybe a template {KONG_LATEST} or something?

app/_hub/kong-inc/kafka-log/1.3-x.md
```diff
@@ -1,6 +1,7 @@
 ---
 name: Kafka Log
 publisher: Kong Inc.
+version: 1.3-x # 0.1.0 internal handler
```
murillopaula, line 4: is there a way to make the internal handler version show up in the hub, next to the plugin version menu? or is it better to just leave it as a comment?

app/_hub/kong-inc/kafka-log/index.md
```diff
@@ -32,87 +33,103 @@ params:
       value_in_examples: BOOTSTRAP_SERVERS
       urlencode_in_examples: true
       default:
+      datatype: set
```
murillopaula, line 36:
```suggestion
      datatype: set of record elements
```

app/_hub/kong-inc/kafka-log/index.md
```diff
@@ -32,87 +33,103 @@ params:
...
       description: |
         List of bootstrap brokers in a `{host: host, port: port}` format.
```
murillopaula, line 38:
```suggestion
        Set of bootstrap brokers in a `{host: host, port: port}` list format. Refer to the "Quickstart" section for examples.
```

we already have an `application/x-www-form-urlencoded` example in the "Quickstart" section. I'd suggest adding an `application/json` example there as well. Here is my example JSON config:

```json
{
  "name": "kafka-log",
  "config": {
    "topic": "kong-log",
    "bootstrap_servers": [
      {
        "host": "localhost",
        "port": 9092
      }
    ]
  }
}
```

murillopaula (changes requested):
@tlblessing 

The plugin is not showing up in the hub: https://deploy-preview-2595--kongdocs.netlify.app/hub/

tlblessing:
> @tlblessing
> 
> The plugin is not showing up in the hub: https://deploy-preview-2595--kongdocs.netlify.app/hub/

yes I know, that is why I didn't have a direct review link yet.  running a local build to troubleshoot. having local build issues after OS upgrade.

app/_hub/kong-inc/kafka-log/1.3-x.md
tlblessing: > is there a way to make the internal handler version show up in the hub, next to the plugin version menu? or is it better to just leave it as a comment?

That's a great question (I think this no because this is a yaml file but I can test)  but we do need to determine exactly what dictates a plugin version. Internal handler seems preferable to me and less confusing. Incrementing a plugin version based on KE/CE release number causes bloat in the doc source and is also confusing as it doesn't match the handler version number. cc: @eskibars

app/_hub/kong-inc/kafka-log/index.md
tlblessing: > is there a way to not hardcode the version here? otherwise we'll always need to change it... maybe a template {KONG_LATEST} or something?

Yes, this should be latest. will do now that release is out.

app/_hub/kong-inc/kafka-log/index.md
tlblessing: @murillopaula thanks for the extra example. We should avoid using double quotes and I'll make links between the sections. Also, I still think the description is wrong? Should be host:port, host:port  and not host:host, port:port?

app/_hub/kong-inc/kafka-log/1.3-x.md
tlblessing: It does work but I'm not sure if it adds to the confusion. Tested on other Kafka plugin.

image

murillopaula: can't I see it in the deploy preview... the version should be the internal handler only, not based on the KE/CE release number IMO

tlblessing: > can't I see it in the deploy preview... the version should be the internal handler only, not based on the KE/CE release number IMO

Agree versions should be internal handler only. Not sure how to handle all the legacy versions based on KE/CE. I'd prefer to just have the latest version only, but @eskibars will have to advise and approve.

I tested it locally on the other Kafka preview, so you wouldn't see it in this PR's preview. It just adds to the confusion IMO.

app/_hub/kong-inc/kafka-log/index.md
tlblessing: @murillopaula since the entire doc is currently `urlencode_in_examples: true`, and the main focus of this PR is to add datatypes, I'll log a ticket for us to work on an entire json example rather than just an inserted snippet.

app/_hub/kong-inc/kafka-log/index.md
```diff
@@ -156,27 +177,27 @@ $ curl -X POST http://kong:8001/plugins \
...
 
 ## Known issues and limitations
 
 Known limitations:
 
-1. There is no support for TLS
-2. There is no support for Authentication
-3. There is no support for message compression
+1. There is no support for TLS.
+2. There is no support for Authentication.
```
lena-larionova, line 191:
```suggestion
2. There is no support for authentication.
```

app/_hub/kong-inc/kafka-log/index.md
```diff
@@ -156,27 +177,27 @@ $ curl -X POST http://kong:8001/plugins \
...
 
 ## Quickstart
 
 The following guidelines assume that both `Kong` and `Kafka` have been installed on your local machine:
 
-1. Install `kong-plugin-kafka-log` as specified in the "Installation" section above.
+1. Install `kong-plugin-kafka-log` as specified in the above Installation section.
 
-2. Create `kong-log` topic in your `Kafka` cluster:
+2. Create a `kong-log` topic in your `Kafka` cluster:
```
lena-larionova, line 200:
```suggestion
2. Create a `kong-log` topic in your Kafka cluster:
```

lena-larionova (approved):
Minor edits, otherwise LGTM.

PR #2598, opened 2021-02-19
Update entity delete example

jcvincenti:
Remove curly brackets around id. The signature is the same as the select.


lena-larionova:
Thanks for the catch! Can you also make this change in `/enterprise/2.3.x/plugin-development/custom-entities/`? It's a duplicate of this topic.

jcvincenti:
> Thanks for the catch! Can you also make this change in [/enterprise/2.3.x/plugin-development/custom-entities/](https://github.com/Kong/docs.konghq.com/blob/master/app/enterprise/2.3.x/plugin-development/custom-entities.md)? It's a duplicate of this topic.

Change done !

lena-larionova (approved):
LGTM

PR #2604, opened 2021-02-19
Response Transformer datatypes

tlblessing:
https://konghq.atlassian.net/browse/DOCS-1396 epic to add datatypes to plugins beyond the few template-autogenerated parameters.

Ported the datatypes from the advanced version of the plugin that Murillo and I worked on.

Schema link:

https://github.com/Kong/kong-ee/blob/master/kong/plugins/response-transformer/schema.lua

Direct review link:

https://deploy-preview-2604--kongdocs.netlify.app/hub/kong-inc/response-transformer/

PR #2605, opened 2021-02-19
CORS plugin datatypes

tlblessing:
https://konghq.atlassian.net/browse/DOCS-1396 epic to add datatypes to plugins beyond the few template-autogenerated parameters.

Schema link:

https://github.com/Kong/kong-ee/blob/master/kong/plugins/cors/schema.lua

Direct review link:

https://deploy-preview-2605--kongdocs.netlify.app/hub/kong-inc/cors/

### DIFF
--- a/app/_hub/kong-inc/cors/index.md
+++ b/app/_hub/kong-inc/cors/index.md
@@ -99,7 +99,7 @@ params:
       description: |
         Indicates how long the results of the preflight request can be cached, in `seconds`.
     - name: preflight_continue
-      required: false
+      required: true
       default: "`false`"
       datatype: boolean
       description: A boolean value that instructs the plugin to proxy the `OPTIONS` preflight request to the Upstream service.
--- a/app/_hub/kong-inc/cors/index.md
+++ b/app/_hub/kong-inc/cors/index.md
@@ -62,14 +62,16 @@ params:
       value_in_examples: ["http://mockbin.com"]
       datatype: array of string elements
       description: |
-        List of allowed domains for the `Access-Control-Allow-Origin` header. If you want to allow all origins, add `*` as a single value to this configuration field. The accepted values can either be flat strings or PCRE regexes. **NOTE**: Prior to Kong 0.10.x, this parameter was `config.origin` (note the change in trailing `s`), and only accepted a single value, or the `*` special value.
+        List of allowed domains for the `Access-Control-Allow-Origin` header. If you want to allow all origins, add `*` as a single value to this configuration field. The accepted values can either be flat strings or PCRE regexes.
     - name: methods
       required: false
       default: "`GET, HEAD, PUT, PATCH, POST, DELETE, OPTIONS, TRACE, CONNECT`"
       value_in_examples: [ "GET", "POST" ]
       datatype: array of string elements
       description:
-        Value for the `Access-Control-Allow-Methods` header.
+        Value for the `Access-Control-Allow-Methods` header. Available
+        options include `GET`, `HEAD`, `PUT`, `PATCH`, `POST`, `DELETE`, `OPTIONS`, `TRACE`, `CONNECT`.
+        By default, all options are allowed.
     - name: headers
       required: false
       default: "Value of the `Access-Control-Request-Headers` request header"
--- a/app/_hub/kong-inc/cors/index.md
+++ b/app/_hub/kong-inc/cors/index.md
@@ -60,7 +60,7 @@ params:
       required: false
       default:
       value_in_examples: ["http://mockbin.com"]
-
+      datatype: array of string elements
       description: |
         List of allowed domains for the `Access-Control-Allow-Origin` header. If you want to allow all origins, add `*` as a single value to this configuration field. The accepted values can either be flat strings or PCRE regexes. **NOTE**: Prior to Kong 0.10.x, this parameter was `config.origin` (note the change in trailing `s`), and only accepted a single value, or the `*` special value.
     - name: methods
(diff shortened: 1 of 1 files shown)


lena-larionova (approved):

app/_hub/kong-inc/cors/index.md
```diff
@@ -60,41 +60,48 @@ params:
       required: false
       default:
       value_in_examples: ["http://mockbin.com"]
+
       description: |
         List of allowed domains for the `Access-Control-Allow-Origin` header. If you want to allow all origins, add `*` as a single value to this configuration field. The accepted values can either be flat strings or PCRE regexes. **NOTE**: Prior to Kong 0.10.x, this parameter was `config.origin` (note the change in trailing `s`), and only accepted a single value, or the `*` special value.
     - name: methods
       required: false
       default: "`GET, HEAD, PUT, PATCH, POST, DELETE, OPTIONS, TRACE, CONNECT`"
```
murillopaula, line 68: I have same question about the `one_of` asked here: https://github.com/Kong/docs.konghq.com/pull/2604

https://github.com/Kong/kong-ee/blob/master/kong/plugins/cors/schema.lua#L54

app/_hub/kong-inc/cors/index.md
```diff
@@ -60,41 +60,48 @@ params:
...
     - name: exposed_headers
       required: false
       default:
       value_in_examples: ["X-Auth-Token"]
+      datatype: array of string elements
       description: |
         Value for the `Access-Control-Expose-Headers` header. If not specified, no custom headers are exposed.
     - name: credentials
-      required: false
+      required: true
```
murillopaula, line 88: nice catch

app/_hub/kong-inc/cors/index.md
```diff
@@ -60,41 +60,48 @@ params:
...
         Flag to determine whether the `Access-Control-Allow-Credentials` header should be sent with `true` as the value.
     - name: max_age
       required: false
       default:
       value_in_examples: 3600
+      datatype: number
       description: |
         Indicates how long the results of the preflight request can be cached, in `seconds`.
     - name: preflight_continue
       required: false
```
murillopaula, line 102: this is required too https://github.com/Kong/kong-ee/blob/master/kong/plugins/cors/schema.lua#L58

app/_hub/kong-inc/cors/index.md
```diff
@@ -60,41 +60,48 @@ params:
...
       default:
       value_in_examples: ["http://mockbin.com"]
+
```
murillopaula, line 63: missing type for this config?

tlblessing: > missing type for this config? `origins`

oops good catch

app/_hub/kong-inc/cors/index.md
tlblessing: > this is required too https://github.com/Kong/kong-ee/blob/master/kong/plugins/cors/schema.lua#L58

good catch!

app/_hub/kong-inc/cors/index.md
tlblessing: > I have same question about the `one_of` asked here: #2604
> 
> https://github.com/Kong/kong-ee/blob/master/kong/plugins/cors/schema.lua#L54

@murillopaula I'm not sure about this one (no pun intended). The default lists all options, and the example lists 2 options, so what is the intent of `one_of ` for this plugin parameter?

murillopaula: The intent of the `one_of` is to specify what are the allowed values for the field. In this case, it's only one of: `GET, HEAD, PUT, PATCH, POST, DELETE, OPTIONS, TRACE, CONNECT`.

My question was how we could explicitly communicate the above information to the user. What do you think about doing the same approach as the response-ratelimiting plugin? See my comment at https://github.com/Kong/docs.konghq.com/pull/2617/#discussion_r580996833

The `config.origin` legacy info I think we should remove as well since we're in 2.3.x already...

tlblessing: > The intent of the `one_of` is to specify what are the allowed values for the field. In this case, the allowed values can only be: `GET, HEAD, PUT, PATCH, POST, DELETE, OPTIONS, TRACE or CONNECT`. Anything outside this list if not allowed.

@murillopaula the * char is still allowed when configuring the CORS plugin (I had tested it in Konnect), but it is not reflected in the schema. I think we need to mention it.

> My question was how we could explicitly communicate the above information to the user. What do you think about doing the same approach as the response-ratelimiting plugin? See my comment at #2617

Yes agree.

> The `config.origin` legacy info I think we should remove as well since we're in 2.3.x already...

Will do.

app/_hub/kong-inc/cors/index.md
```diff
@@ -62,14 +62,16 @@ params:
...
+        List of allowed domains for the `Access-Control-Allow-Origin` header. If you want to allow all origins, add `*` as a single value to this configuration field. The accepted values can either be flat strings or PCRE regexes.
     - name: methods
       required: false
       default: "`GET, HEAD, PUT, PATCH, POST, DELETE, OPTIONS, TRACE, CONNECT`"
       value_in_examples: [ "GET", "POST" ]
       datatype: array of string elements
       description:
-        Value for the `Access-Control-Allow-Methods` header.
+        Value for the `Access-Control-Allow-Methods` header. Available
+        options include `GET`, `HEAD`, `PUT`, `PATCH`, `POST`, `DELETE`, `OPTIONS`, `TRACE`, `CONNECT`.
```
tlblessing, line 73: When I Available options: , I got a yaml error. not sure why some of these older plugins have odd template issues.

murillopaula (approved):

PR #2611, opened 2021-02-22
Fix Immunity docker URLs

lena-larionova:
Docs ticket: https://konghq.atlassian.net/browse/DOCS-1602

Fixing URL for Immunity Docker image in 2.2.x and 2.3.x. Checked with a quick `docker pull` to make sure this one works.
This image didn't exist before 2.2.x, it was Brain&Immunity before that.


HCloward (approved):
This is the docker image I was able to pull recently.

PR #2617, opened 2021-02-22
Response Rate Limiting plugin add datatypes

tlblessing:
https://konghq.atlassian.net/browse/DOCS-1396 epic to add datatypes to plugins beyond the few template-autogenerated parameters.

Schema link:

https://github.com/Kong/kong-ee/blob/master/kong/plugins/response-ratelimiting/schema.lua

@murillopaula not clear on how to phrase datatype for limits. Map of key:values? Went with string, please let me know if that is correct.

image

Direct review link:

https://deploy-preview-2617--kongdocs.netlify.app/hub/kong-inc/response-ratelimiting/

### DIFF
--- a/app/_hub/kong-inc/response-ratelimiting/index.md
+++ b/app/_hub/kong-inc/response-ratelimiting/index.md
@@ -71,7 +71,7 @@ params:
     - name: limits.{limit_name}
       required: true
       value_in_examples: <SMS>
-      datatype:
+      datatype: string
       description: This is a list of custom objects that you can set, with arbitrary names set in the `{limit_name`} placeholder, like `config.limits.sms.minute=20` if your object is called "SMS".
     - name: limits.{limit_name}.second
       required: semi
--- a/app/_hub/kong-inc/response-ratelimiting/index.md
+++ b/app/_hub/kong-inc/response-ratelimiting/index.md
@@ -12,7 +12,7 @@ description: |
   rate-limiting object can limit the inbound requests per seconds, minutes, hours,
   days, months, or years.
 
-  If the underlying Service/Route (or deprecated API entity) has no authentication
+  If the underlying Service/Route has no authentication
   layer, the **Client IP** address will be used; otherwise, the Consumer will be
   used if an authentication plugin has been configured.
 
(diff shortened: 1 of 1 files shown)


app/_hub/kong-inc/response-ratelimiting/index.md
```diff
@@ -71,67 +71,85 @@ params:
...
-      required: false
+      required: true
       default: "`false`"
+      datatype: boolean
       description: A boolean value that determines if the requests should be blocked as soon as one limit is being exceeded. This will block requests that are supposed to consume other limits too.
     - name: limit_by
       required: false
       default: "`consumer`"
+      datatype: string
       description: "The entity that will be used when aggregating the limits: `consumer`, `credential`, `ip`. If the `consumer` or the `credential` cannot be determined, the system will always fallback to `ip`."
```
murillopaula, line 115: Here in this plugin the `one_of` is represented in the description of the config field. Is it okay? Should we be adopting this convention in the plugins that are missing it such as https://github.com/Kong/docs.konghq.com/pull/2604?

murillopaula:
@tlblessing I think it's ok to leave it as it is now. As I see it, the field is just a placeholder for explaining the general behavior since it's not possible to create a plugin instance with just `config.limits.{limit_name}`. You need to specify the values such as `seconds`, `minutes`, etc. And we're already specifying the types of each of these value in the next fields.

Also, please remove the `api_id` config from the plugin since it's a legacy config field.

app/_hub/kong-inc/response-ratelimiting/index.md
tlblessing: > Here in this plugin the `one_of` is represented in the description of the config field. Is it okay? Should we be adopting this convention in the plugins that are missing it such as #2604?

I think that is our best option really. We state in many places what the available options are, so this is an established convention.

tlblessing:
> Also, please remove the `api_id` config from the plugin since it's a legacy config field.

Removed in commit 9de562c. (set to false)

app/_hub/kong-inc/response-ratelimiting/index.md
tlblessing: resolving convo per https://github.com/Kong/docs.konghq.com/pull/2617#issuecomment-784187024

murillopaula (approved):

Bradamant3 (approved):
LGTM

PR #2628, opened 2021-02-24
add info on new config properties introduced in 2.2.1.0

RobSerafini:
Copied the doc section from 2.3.x version docs.

The additional configuration was added in 2.2.1.0 to match 2.3.0.0. The default value for `untrusted_lua` is different from 2.3.0.0.

FT-1688

### DIFF
--- a/app/enterprise/2.2.x/property-reference.md
+++ b/app/enterprise/2.2.x/property-reference.md
@@ -3178,6 +3178,55 @@ Defines the token value used to communicate with the v2 KV Vault HTTP(S) API.
 
 ---
 
+#### untrusted_lua
+
+Accepted values are:
+
+- `off`: disallow any loading of Lua functions from admin supplied sources
+  (such as via the Admin API).
+
+Note using the `off` option will render plugins such as Serverless Functions
+unusable.
+
+- `sandbox`: allow loading of Lua functions from admin supplied sources, but
+  use a sandbox when executing them. The sandboxed function will have restricted
+  access to the global environment and only have access to standard Lua
+  functions that will generally not cause harm to the Kong node.
+
+In this mode, the `require` function inside the sandbox only allows loading
+external Lua modules that are explicitly listed in
+`untrusted_lua_sandbox_requires` below.
+
+LuaJIT bytecode loading is disabled.
+
+Warning: LuaJIT is not designed as a secure runtime for running malicious code,
+therefore, you should properly protect your Admin API endpoint even with
+sandboxing enabled. The sandbox only provides protection against trivial
+attackers or unintentional modification of the Kong global environment.
+
+- `on`: allow loading of Lua functions from admin supplied sources and do not
+  use a sandbox when executing them. Functions will have unrestricted access to
+  global environment and able to load any Lua modules. This is similar to the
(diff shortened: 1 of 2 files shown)


app/enterprise/2.2.x/property-reference.md
```diff
@@ -3178,6 +3178,55 @@ Defines the token value used to communicate with the v2 KV Vault HTTP(S) API.
...
+
+untrusted_lua_sandbox_requires = Comma-separated list of modules allowed to be
+loaded with `require` inside the sandboxed environment. Ignored if
+`untrusted_lua` is not `sandbox`.
+
+Note: certain modules, when allowed, may cause sandbox escaping trivial.
+
+untrusted_lua_sandbox_environment = Comma-separated list of global Lua
+variables that should be made available inside the sandboxed environment.
+Ignored if `untrusted_lua` is not `sandbox`.
```
lena-larionova, line 3222: These two properties don't seem to be generating properly from the code. They're getting listed as part of the `untrusted_lua` property. I just tested, and it looks like the spacing is off for in the source file? Should be this (ie, no blank line without a hash on the same level as the property name):

```
#untrusted_lua_sandbox_requires = # Comma-separated list of modules allowed to be loaded
                                  # with `require` inside the sandboxed environment. Ignored
                                  # if `untrusted_lua` is not `sandbox`.
                                  #
                                  # Note: certain modules, when allowed, may cause sandbox
                                  # escaping trivial.

#untrusted_lua_sandbox_environment = # Comma-separated list of global Lua variables
                                     # that should be made available inside the sandboxed
                                     # environment. Ignored if `untrusted_lua` is not `sandbox`.
                                     #
                                     # Note: certain variables, when made available,
                                     # may cause sandbox escaping trivial.
```

lena-larionova: I can take over and regenerate this+log PR in kong-ee, if that would be easiest.

app/enterprise/2.2.x/property-reference.md
```diff
@@ -3178,6 +3178,55 @@ Defines the token value used to communicate with the v2 KV Vault HTTP(S) API.
...
+
+Warning: LuaJIT is not designed as a secure runtime for running malicious code,
+therefore, you should properly protect your Admin API endpoint even with
+sandboxing enabled. The sandbox only provides protection against trivial
+attackers or unintentional modification of the Kong global environment.
+
+- `on`: allow loading of Lua functions from admin supplied sources and do not
+  use a sandbox when executing them. Functions will have unrestricted access to
+  global environment and able to load any Lua modules. This is similar to the
+  behavior in Kong prior to 2.3.0.
```
lena-larionova, line 3210: From this line, I'm assuming this doc was generated from the 2.3 file version?
```suggestion
  behavior in Kong prior to 2.2.0.
```

app/enterprise/2.2.x/property-reference.md
lena-larionova: Made the changes here and opened a PR against the source in kong/kong.

lena-larionova (approved):

lena-larionova:
Note for future reviewers: checks are all good, just had to cancel a duplicate.

PR #2639, opened 2021-02-25
[DOCS-1564] Add canonical links for all version URLs

adamjkuhn:
### Summary
* conditionally add <link rel=“canonical”> in head of URL paths with versions directing to latest versioned pages for improved search indexing

Sample output:
Screenshot of preview version canonical versus URL

Screenshot of preview version canonical with subdirectory

Screenshot of self-referencing canonical

* I've chosen to leave in self-referencing canonicals as this is acceptable practice and some argue it improves indexing structures.

* In order to fully ensure proper indexing after each version update it may be necessary to manually force a recrawl via Google Webmaster Tools

### DIFF
--- a/app/_includes/head.html
+++ b/app/_includes/head.html
@@ -163,7 +163,7 @@ site.title }}{% endcapture %} {% else %} {% capture page_title %}{{ page.name }}
   {% capture latest_page_exists %}{% page_exists {{latest_page_url}} %}{% endcapture %}
 
   {% if page.kong_version == page.kong_latest.release and page.no_version != true %}<link rel="canonical" href="{{ site.links.web }}{{ page.url }}" />
-    {% elsif latest_page_exists == 'true' %}<link rel="canonical" href="{% capture latest_path %}{{ site.links.web }}{% assign latest_page_url = page.url | replace: page.kong_version, page.kong_latest.release %}{{ latest_page_url }}{% endcapture %}{{ latest_path }}" />
+    {% elsif latest_page_exists == 'true' %}<link rel="canonical" href="{% capture latest_path %}{{ site.links.web }}{{ latest_page_url }}{% endcapture %}{{ latest_path }}" />
   {% endif %}
 
 </head>
--- a/app/_includes/head.html
+++ b/app/_includes/head.html
@@ -158,6 +158,13 @@ site.title }}{% endcapture %} {% else %} {% capture page_title %}{{ page.name }}
     href="/assets/hub.css?v={{ site.time | date: '%s' }}"
   />
   {% endif %}
+
+  {% if page.no_version != true %}
+  {% if page.kong_version == site.data.kong_latest.release %}<link rel="canonical" href="{{ site.links.web }}{{ page.url }}" />{% else %}<link rel="canonical" href="{{ site.links.web }}{% assign latest_page_url = page.url | replace: page.kong_version, page.kong_latest.release %}{% capture latest_page_exists %}{% page_exists {{latest_page_url}} %}FOUR{% endcapture %}{{ latest_page_url }}" />{%endif%}
+  {% endif %}
+
+
+
 </head>
 
 <!-- Kong Cookie Policy -->
--- a/app/_includes/head.html
+++ b/app/_includes/head.html
@@ -159,11 +159,12 @@ site.title }}{% endcapture %} {% else %} {% capture page_title %}{{ page.name }}
   />
   {% endif %}
 
-  {% if page.no_version != true %}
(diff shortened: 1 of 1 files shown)


app/_includes/head.html
```diff
@@ -158,6 +158,13 @@
     href="/assets/hub.css?v={{ site.time | date: '%s' }}"
   />
   {% endif %}
+
+  {% if page.no_version != true %}
+  {% if page.kong_version == site.data.kong_latest.release %}<link rel="canonical" href="{{ site.links.web }}{{ page.url }}" />{% else %}<link rel="canonical" href="{{ site.links.web }}{% assign latest_page_url = page.url | replace: page.kong_version, page.kong_latest.release %}{% capture latest_page_exists %}{% page_exists {{latest_page_url}} %}FOUR{% endcapture %}{{ latest_page_url }}" />{%endif%}
```
lena-larionova, line 163: Something looks off with the last portion:
```
{% capture latest_page_exists %}{% page_exists {{latest_page_url}} %}FOUR{% endcapture %}{{ latest_page_url }}
```
This is capturing `latest_page_exists` but then the capture isn't used anywhere. As well, `latest_page_url` isn't defined here.

I tested out a couple of older docs that don't have a parallel latest page, and they're indiscriminately pointing to a 2.3.x doc that doesn't exist. For example:

https://deploy-preview-2639--kongdocs.netlify.app/enterprise/0.36-x/getting-started/add-admin/ 
- If you look at the link in the yellow banner, it take you to `/enterprise`. That's because `/enterprise/2.3.x/getting-started/add-admin/` doesn't exist. When you check the header, this non-existent page is set as the canonical url:

Screen Shot 2021-02-25 at 9 14 33 AM

adamjkuhn: Apologies that this took me a bit but I've leverged `page.exists` in order to avoid applying canonical links to pages that do not resolve to a latest version. 

Unfortunately this means that pages such as [https://deploy-preview-2639--kongdocs.netlify.app/enterprise/0.36-x/getting-started/add-admin/](https://deploy-preview-2639--kongdocs.netlify.app/enterprise/0.36-x/getting-started/add-admin/) will not have a canonical link defined in the page head, so as of now only pages that are either self-referential canonicals or that have canonicals that exist as the latest version will be defined. This should still provide a net benefit regarding SEO and indexing.

app/_includes/head.html
```diff
@@ -158,6 +158,14 @@
     href="/assets/hub.css?v={{ site.time | date: '%s' }}"
   />
   {% endif %}
+
+  {% assign latest_page_url = page.url | replace: page.kong_version, page.kong_latest.release %}
+  {% capture latest_page_exists %}{% page_exists {{latest_page_url}} %}{% endcapture %}
+
+  {% if page.kong_version == site.data.kong_latest.release and page.no_version != true %}<link rel="canonical" href="{{ site.links.web }}{{ page.url }}" />
```
lena-larionova, line 165:
```suggestion
  {% if page.kong_version == page.kong_latest.release and page.no_version != true %}<link rel="canonical" href="{{ site.links.web }}{{ page.url }}" />
```
You want this piece of metadata here instead, the one you're using will only pull the latest version for CE gateway - a remnant from the way the site was originally setup with no other expected products/versions.

app/_includes/head.html
```diff
@@ -158,6 +158,14 @@
...
+    {% elsif latest_page_exists == 'true' %}<link rel="canonical" href="{% capture latest_path %}{{ site.links.web }}{% assign latest_page_url = page.url | replace: page.kong_version, page.kong_latest.release %}{{ latest_page_url }}{% endcapture %}{{ latest_path }}" />
```
lena-larionova, line 166:
```suggestion
    {% elsif latest_page_exists == 'true' %}<link rel="canonical" href="{% capture latest_path %}{{ site.links.web }}{{ latest_page_url }}{% endcapture %}{{ latest_path }}" />
```
Don't need the `assign` again since it's already assigned above.

lena-larionova (commented):
Couple of tweaks, and then I think we're good.

app/_includes/head.html
lena-larionova: Hmm. We'll need to find another way to deal with those, then, because they're a big part of what's polluting our search results. Maybe this is a case for `noindex`, then? I can work on that in the future.

app/_includes/head.html
adamjkuhn: Good to know-thanks

app/_includes/head.html
adamjkuhn: 🤦‍♂️

app/_includes/head.html
adamjkuhn: I'd imagine there's a better way of handling this than running a loop through pages to compare version numbers and ultimately terminating at the greatest available version but I'd have to dig a bit more into the nuances of Liquid, since this ends up in a soul-crushingly slow build time.
I'd hate to go nuclear with `noindex` - hopefully as a half-measure this will at least help clean up some of the results.

lena-larionova:
Ignore the "failed" checks, I cancelled them because this PR is still pulling in an older version where the check ran on push.

app/_includes/head.html
lena-larionova: Yeah, and the loop is used a couple of times in other situations... that can't be good for the build time. If you ever have the time to take a look, the doc team will be eternally grateful 🙏

lena-larionova (approved):
LGTM! Thanks for all your help @adamjkuhn

PR #2642, opened 2021-02-26
Fix Markdown markup in kong.response.md

mloskot:
### Review

@reviewer

### Summary

Fix broken rendering of `kong.response.error` section

### Reason

image


tlblessing (approved):
Thank you for the PR @mloskot. Looks like this is an issue in 2.1.x and 2.2.x as well. I'll get those versions fixed.

PR #2662, opened 2021-03-05
fix(analytics) Remove duplicate event tracker

lena-larionova:
### Summary
Removing an event that triggers on every single page view in docs, giving us a really weird bounce rate.

### Reason
To remove a duplicate google analytics event:
Screen Shot 2021-03-05 at 1 14 19 PM

### Testing
TBA


GayleNeumann (approved):
LGTM

PR #2680, opened 2021-03-10
Correct invalid Deployment apiVersion

rainest:
### Summary
Pull fallback-svc example out of bit.ly/fallback-svc and correct the apiVersion for its Deployment.

### Reason
The original link was a specific gist version with an error, and can't be modified without creating a new link. The example here is short enough that we can pull it out.

Originally reported by https://twitter.com/SixteenOne/status/1368510931211259921

### DIFF
--- a/app/kubernetes-ingress-controller/1.0.x/guides/configuring-fallback-service.md
+++ b/app/kubernetes-ingress-controller/1.0.x/guides/configuring-fallback-service.md
@@ -39,8 +39,8 @@ This is expected as Kong does not yet know how to proxy the request.
 
 ## Setup a Sample Service
 
-For the purpose of this guide, we will setup an [httpbin](https://httpbin.org)
-service in the cluster and proxy it.
+For the purpose of this guide, we will setup a simple HTTP service in the
+cluster and proxy it.
 
 ```bash
 $ echo '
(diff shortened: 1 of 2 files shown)


app/kubernetes-ingress-controller/1.0.x/guides/configuring-fallback-service.md
```diff
@@ -43,9 +43,48 @@ For the purpose of this guide, we will setup an [httpbin](https://httpbin.org)
...
+  ports:
+  - port: 80
+    targetPort: 5678
+    protocol: TCP
+    name: http
+  selector:
+    app: fallback-svc
+' | kubectl apply -f -
+deployment.apps/fallback-svc created
+service/fallback-svc created
```
lena-larionova, line 87: Can you split out the result from the codeblock so that it's easily copyable? E.g:

Result:
```bash
deployment.apps/fallback-svc created
service/fallback-svc created
```

app/kubernetes-ingress-controller/1.0.x/guides/configuring-fallback-service.md
```diff
@@ -43,9 +43,53 @@ For the purpose of this guide, we will setup an [httpbin](https://httpbin.org)
 service in the cluster and proxy it.
```
HCloward, line 43: I'm confused. The text says this is an example for httpbin, but the example appears to be for fallback-svc - was this intentional?

rainest: Probably copy-pasta from some other example that did use HTTPbin, whereas this uses `image: hashicorp/http-echo`.

b5e95d5 uses generic language to describe the application instead. "fallback" here more describes what we're using it for (a service that handles requests that don't match a rule for something else), and the actual application used isn't particularly important.

lena-larionova (approved):

PR #2691, opened 2021-03-16
docs(changelog) kong user change arrived in EE 2.2.1.0 not EE 2.3.0.0…

murillopaula:
I think this was inherited from the OSS changelog: https://github.com/Kong/kong/blob/master/CHANGELOG.md#configuration

But in EE, it arrived first in EE 2.2.1.0.

Sometimes the things inherited from the OSS changelog don’t necessarily map 1:1 to the EE changelog.

In my case for example, I had this changed merged into CE master that arrived first in CE 2.3.0 but because of the CE->EE merges, it went to EE 2.2.1.0 first, not EE 2.3.0.0 first.

### DIFF
--- a/app/enterprise/changelog.md
+++ b/app/enterprise/changelog.md
@@ -269,10 +269,6 @@ being shown in the logs.
   `lua-resty-healthcheck`.
 - Certificates for database connections now are loaded in the right order
   avoiding failures to connect to Postgres databases.
-  
-#### Configuration
-- Kong now runs as a `kong` user if it exists; if user does not exist
-  in the system, the `nobody` user is used, as before.
 
 #### CLI
 - Fixed issue where `kong reload -c <config>` would fail.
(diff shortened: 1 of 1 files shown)


lena-larionova:
Hmm. I _think_ the approach is that we want to keep both entries, depending on when the change was introduced in each minor version. So if this feature went into 2.2.1.0 and 2.3.2.0, the entry goes into both. For example, open up the [Enterprise changelog](https://docs.konghq.com/enterprise/changelog/) and search for "OpenSSL version bumped to 1.1.1i. " - it shows up three times, the paragraph being exactly the same each time.

I'm not saying that this is the _best_ way of doing this, but it is how it's done now...

murillopaula:
@lena-larionova thanks. Just FYI, this specific `kong user` change now appears two times:

1) EE 2.2.1.0
2) EE 2.3.0.0 beta

It appears in 1) because it landed in EE for the time in EE 2.2.1.0 and in 2) because it's inheriting from the OSS changelog

plz review

lena-larionova (approved):

PR #2693, opened 2021-03-17
[DOCS-1675] Remove read time from docs layout

lena-larionova:
### Summary
Removing "Estimated read time" from docs. Also removing all instances of `skip_read_time` from page front matter so that it doesn't pollute docs for the future.

### Reason
Feedback that the information is unnecessary and blog-like. See ticket: https://konghq.atlassian.net/browse/DOCS-1675

### Testing
TBA


HCloward (approved):
LGTM

PR #2695, opened 2021-03-17
[DOCS-1368] Route transformer plugin version menu

tlblessing:
https://konghq.atlassian.net/browse/DOCS-1368

Some plugins follow internal handler version (app reg plugin and more). Since this plugin was never properly versioned, and I don't know historical changes or have much bandwidth to reconstruct history (there are only 4 config params anyway), I'm going with that (0.2.x) rather than bloating the docs with EE version of source files that are the same.

Direct review link:

https://deploy-preview-2695--kongdocs.netlify.app/hub/kong-inc/route-transformer-advanced/

image


murillopaula (approved):

Bradamant3 (approved):
LGTM

PR #2696, opened 2021-03-18
[DOCS-1674] fixed link issue redirecting to wrong version

HCloward:
### Review
Anyone on the docs team can take a look at this. It is just a link fix.

### Summary
Someone noticed that the link on the Classic Install page was redirecting to the wrong version. I noticed that part of the link was missing and it seems to have fixed it locally for me anyway. I looked and it was affecting all versions following 1.5 - seems like there was some rework.

### Reason
DOCS-1674

### Testing
Will include link to sample build when available.


Bradamant3 (approved):
LGTM&TYSM!

PR #2698, opened 2021-03-18
[DOCS-1650] Konnect runtime manager: Helm config instructions

lena-larionova:
### Review

@nikirago, @rainest 

### Summary
Adding Helm option when setting up a Konnect runtime through Runtime Manager's advanced config.

Questions for reviewers:
* The [sample values.yaml](https://github.com/Kong/charts/blob/main/charts/kong/example-values/minimal-kong-hybrid-data.yaml) file I based this on has `postgres: enabled` for the data plane, which can't be right - and in fact, I couldn't get it to launch until I set it to `postgres: disabled`. Was I missing something, or does the sample file need correcting? Or is it just different for `pki` mode?
* How do you add/specify the CA cert? In the doc for kubernetes tls secrets, I couldn't find any info on adding anything beyond one cert and key: https://kubernetes.io/docs/concepts/configuration/secret/#tls-secrets
* Any other steps or instructions missing?

### Reason
Helm/YAML config is a common use case:
https://konghq.atlassian.net/browse/DOCS-1650

### Testing
https://deploy-preview-2698--kongdocs.netlify.app/konnect/runtime-manager/kong-gateway-runtime/ - open the Advanced tab, then Helm.

### DIFF
--- a/app/konnect/runtime-manager/kong-gateway-runtime.md
+++ b/app/konnect/runtime-manager/kong-gateway-runtime.md
@@ -230,7 +230,8 @@ move onto the next step.
         $ helm repo update
         ```
 
-2. Create a `tls` secret using the files you saved earlier:
+2. Create a `tls` secret using the `cluster.cert` and `cluster.key` files
+you saved earlier:
 
     ```bash
     $ kubectl create secret tls kong-cluster-cert \
@@ -238,9 +239,16 @@ move onto the next step.
       --key=/tmp/cluster.key
     ```
 
-3. In the **Configuration Parameters** section, copy the codeblock.
+3. Create a generic secret for the CA cert:
 
-4. Open your instance's `values.yml` file, or use the
+    ```bash
+    $ kubectl create secret generic kong-cluster-ca \
+      --from-file=ca.crt=/tmp/ca.crt
+    ```
+
+4. In the **Configuration Parameters** section, copy the codeblock.
+
+5. Open your instance's `values.yml` file, or use the
 [data plane template](https://github.com/Kong/charts/blob/main/charts/kong/example-values/minimal-kong-hybrid-data.yaml).
 Remove the `KONG_` prefix from the parameters in the sample codeblock and add
 the following parameters to the file.
@@ -251,6 +259,7 @@ the following parameters to the file.
     ```yaml
     secretVolumes:
     - kong-cluster-cert
+    - kong-cluster-ca
     admin:
       enabled: false
     env:
@@ -262,15 +271,16 @@ the following parameters to the file.
       cluster_server_name: <kong-cpoutlet-example.service>
       cluster_telemetry_endpoint: <example.tp.konnect.foo>:443
       cluster_telemetry_server_name: <kong-telemetry-example.service>
+      cluster_ca_cert: /etc/secrets/kong-cluster-ca/ca.crt
       cluster_cert: /etc/secrets/kong-cluster-cert/tls.crt
       cluster_cert_key: /etc/secrets/kong-cluster-cert/tls.key
-      lua_ssl_trusted_certificate: system,/<path-to-file>/ca.crt
+      lua_ssl_trusted_certificate: system,/etc/secrets/kong-cluster-ca/ca.crt
     ```
 
     See [Parameters](#parameters) for descriptions and the matching fields
     in {{site.konnect_short_name}}.
 
-5. Apply the `values.yml`.
+6. Apply the `values.yml`.
 
     Existing instance:
     ```bash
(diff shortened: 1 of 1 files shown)


app/konnect/runtime-manager/kong-gateway-runtime.md
```diff
@@ -206,6 +207,88 @@ Manager overview.
...
+        ```bash
+        $ helm repo update
+        ```
+
+2. Create a `tls` secret using the files you saved earlier:
+
+    ```bash
+    $ kubectl create secret tls kong-cluster-cert \
+      --cert=/tmp/cluster.crt \
+      --key=/tmp/cluster.key
```
rainest, line 238: The dedicated `tls` Secret type only supports cert/key pairs, so CA certificates where you only have the certificate need to go into a generic certificate instead:

```suggestion
    $ kubectl create secret tls kong-cluster-cert \
      --cert=/tmp/cluster.crt \
      --key=/tmp/cluster.key

    $ kubectl create secret generic kong-cluster-ca \
      --from-file=ca.crt=/tmp/ca.crt
```
Technically speaking they could go into a ConfigMap instead, but our mount mechanism doesn't support those yet.

app/konnect/runtime-manager/kong-gateway-runtime.md
```diff
@@ -206,6 +207,88 @@ Manager overview.
...
+    env:
+      role: data_plane
+      database: "off"
+      anonymous_reports: off
+      cluster_mtls: pki
+      cluster_control_plane: <example.cp.konnect.foo>:443
+      cluster_server_name: <kong-cpoutlet-example.service>
+      cluster_telemetry_endpoint: <example.tp.konnect.foo>:443
+      cluster_telemetry_server_name: <kong-telemetry-example.service>
+      cluster_cert: /etc/secrets/kong-cluster-cert/tls.crt
```
rainest, line 265:
```suggestion
      cluster_ca_cert: /etc/secrets/kong-cluster-ca/ca.crt
      cluster_cert: /etc/secrets/kong-cluster-cert/tls.crt
```

app/konnect/runtime-manager/kong-gateway-runtime.md
```diff
@@ -206,6 +207,88 @@ Manager overview.
...
+[data plane template](https://github.com/Kong/charts/blob/main/charts/kong/example-values/minimal-kong-hybrid-data.yaml).
+Remove the `KONG_` prefix from the parameters in the sample codeblock and add
+the following parameters to the file.
+
+    Make sure to replace the values in `cluster_cert`, `cluster_cert_key`,
+    and `cluster_ca_cert` with references to the secret you created earlier:
+
+    ```yaml
+    secretVolumes:
+    - kong-cluster-cert
```
rainest, line 253:
```suggestion
    - kong-cluster-cert
    - kong-cluster-ca
```

rainest (changes requested):
Yeah, that example had bad copy pasta from elsewhere. Spawning Postgres shouldn't actually prevent startup, though it's pointless. https://github.com/Kong/charts/pull/321 will remove it from upstream.

lena-larionova:
@rainest Does it make sense now to set `lua_ssl_trusted_certificate: system,/etc/secrets/kong-cluster-ca/ca.crt`? I.e, does it still make sense to use `system` here? (Here's the [reference doc](https://docs.konghq.com/enterprise/2.3.x/property-reference/#lua_ssl_trusted_certificate))

rainest (approved):
AFAIK `lua_ssl_trusted_certificate` is used for other trust checks, but not for this at all. https://docs.konghq.com/gateway-oss/2.3.x/hybrid-mode/#pki-mode doesn't indicate that it needs to be set, so this doc doesn't need to modify it either.

lena-larionova:
Not sure if the PKI mode doc is entirely correct - that, or it's slightly different for Konnect. The auto-generated chunk of code you get in the Konnect UI includes this parameter (but no `cluster_ca_cert`, which is necessary here):

Screen Shot 2021-03-22 at 10 46 28 AM

rainest:
The generated Konnect config seems wrong but I'm unsure who's the best person to contact about it. PKI mode [enforces a hard requirement on the dedicated setting](https://github.com/Kong/kong/blob/2.3.3/kong/conf_loader/init.lua#L1024-L1026), so I wouldn't expect a DP node to start in PKI mode without it.

lena-larionova:
> The generated Konnect config seems wrong but I'm unsure who's the best person to contact about it. PKI mode [enforces a hard requirement on the dedicated setting](https://github.com/Kong/kong/blob/2.3.3/kong/conf_loader/init.lua#L1024-L1026), so I wouldn't expect a DP node to start in PKI mode without it.

@artomal can you take a look?

lena-larionova:
Never mind - I vaguely remembered a similar discussion on an earlier ticket: https://github.com/Kong/docs.konghq.com/pull/2575#discussion_r575564875

The specific info to note:
> If this is for bring-your-own-data-plane then cluster_ca_cert is not needed, cluster_cert is the client cert DP present to CP, and cluster_ca_cert is the CA for CP to verify DP.

Since Konnect config is managed by Kong, generally this setting isn't needed on the DP side?

HCloward (approved):

PR #2697, opened 2021-03-18
update names of mesh components

Bradamant3:
### Review

@subnetmarco lmk if older versions also need to be changed. This PR addresses only v1.2. These are the only instances grep found ...

BUT note also that the Helm instructions include a user-provided namespace, and we currently suggest `kong-mesh-system` ... potential new confusion?

### Summary
Update names of mesh components

### Reason
Inaccurate names currently published

### Testing
Will post link to netlify preview after it builds

PR #2703, opened 2021-03-22
Fixed the namespace for Kong mesh

ikeike443:
### Review

@Kong/team-docs 

### Summary

I've noticed the namespace Kong mesh actually create through the process has been changed from `kuma-system` to `kong-mesh-system`.


Bradamant3:
@ikeike443 thanks much for this, but #2697 addresses the issue (sorry, needed to double-check something last week so we didn't get it in in timely fashion) in helm charts and openshift also. 🙏

PR #2706, opened 2021-03-22
[DOCS-1660] Vault plugin add datatypes

tlblessing:
https://konghq.atlassian.net/browse/DOCS-1660 part of epic https://konghq.atlassian.net/browse/DOCS-1396.

Schema:

https://github.com/Kong/kong-plugin-vault-auth/blob/master/kong/plugins/vault-auth/schema.lua

Direct review link:

https://deploy-preview-2706--kongdocs.netlify.app/hub/kong-inc/vault-auth/

@murillopaula i'm not sure about the vault.id:

>      { vault = { type = "foreign", reference = "vaults", required = true } },

what is foreign? I just think of a foreign key in relational db.

### DIFF
--- a/app/_hub/kong-inc/vault-auth/index.md
+++ b/app/_hub/kong-inc/vault-auth/index.md
@@ -33,37 +33,44 @@ params:
     - name: access_token_name
       required: true
       default: "`access_token`"
+      datatype: array of string elements
       description: |
         Describes an array of comma-separated parameter names where the plugin will look for an access token. The client must send the access token in one of those key names, and the plugin will try to read the credential from a header or the querystring parameter with the same name.<br>*note*: the key names may only contain [a-z], [A-Z], [0-9], [_] and [-].
     - name: vault.id
       required: true
       default:
       value_in_examples: "<UUID>"
+      datatype: foreign
       description: |
         A reference to an existing `vault` object within the database. `vault` entities define the connection and authentication parameters used to connect to a Vault HTTP(S) API.
     - name: secret_token_name
       required: true
       default: "`secret_token`"
+      datatype: array of string elements
       description: |
         Describes an array of comma-separated parameter names where the plugin will look for a secret token. The client must send the secret in one of those key names, and the plugin will try to read the credential from a header or the querystring parameter with the same name.<br>*note*: the key names may only contain [a-z], [A-Z], [0-9], [_] and [-].
     - name: key_in_body
       required: false
       default: "`false`"
+      datatype:
       description: |
         If enabled, the plugin will read the request body (if said request has one and its MIME type is supported) and try to find the key in it. Supported MIME types are `application/www-form-urlencoded`, `application/json`, and `multipart/form-data`.
     - name: hide_credentials
       required: false
       default: "`false`"
+      datatype: boolean
       description: |
         An optional boolean value telling the plugin to show or hide the credential from the upstream service. If `true`, the plugin will strip the credential from the request (i.e. the header or querystring containing the key) before proxying it.
     - name: anonymous
       required: false
       default:
+      datatype: string
       description: |
-        An optional string (consumer uuid) value to use as an "anonymous" consumer if authentication fails. If empty (default), the request will fail with an authentication failure `4xx`. Please note that this value must refer to the Consumer `id` attribute which is internal to Kong, and **not** its `custom_id`.
+        An optional string (consumer uuid) value to use as an "anonymous" consumer if authentication fails. If empty (default), the request will fail with an authentication failure `4xx`. Note that this value must refer to the Consumer `id` attribute that is internal to Kong, and **not** its `custom_id`.
     - name: run_on_preflight
       required: false
       default: "`true`"
+      datatype: boolean
       description: |
         A boolean value that indicates whether the plugin should run (and try to authenticate) on `OPTIONS` preflight requests, if set to `false` then `OPTIONS` requests will always be allowed.
 
--- a/app/_hub/kong-inc/vault-auth/index.md
+++ b/app/_hub/kong-inc/vault-auth/index.md
@@ -49,7 +49,7 @@ params:
       datatype: array of string elements
       description: |
         Describes an array of comma-separated parameter names where the plugin will look for a secret token. The client must send the secret in one of those key names, and the plugin will try to read the credential from a header or the querystring parameter with the same name.<br>*note*: the key names may only contain [a-z], [A-Z], [0-9], [_] and [-].
-    - name: key_in_body
+    - name: tokens_in_body
       required: false
       default: "`false`"
       datatype:
--- a/app/_hub/kong-inc/vault-auth/index.md
+++ b/app/_hub/kong-inc/vault-auth/index.md
@@ -39,7 +39,7 @@ params:
       required: true
       default:
       value_in_examples: "<UUID>"
-      datatype: foreign
+      datatype: foreign ID
       description: |
         A reference to an existing `vault` object within the database. `vault` entities define the connection and authentication parameters used to connect to a Vault HTTP(S) API.
     - name: secret_token_name
@@ -49,13 +49,13 @@ params:
       description: |
         Describes an array of comma-separated parameter names where the plugin will look for a secret token. The client must send the secret in one of those key names, and the plugin will try to read the credential from a header or the querystring parameter with the same name. The key names can only contain [a-z], [A-Z], [0-9], [_], and [-].
     - name: tokens_in_body
-      required: false
+      required: true
       default: "`false`"
-      datatype:
+      datatype: boolean
       description: |
         If enabled, the plugin will read the request body (if said request has one and its MIME type is supported) and try to find the key in it. Supported MIME types are `application/www-form-urlencoded`, `application/json`, and `multipart/form-data`.
     - name: hide_credentials
-      required: false
+      required: true
       default: "`false`"
       datatype: boolean
       description: |
@@ -63,15 +63,15 @@ params:
     - name: anonymous
       required: false
       default:
-      datatype: string
+      datatype: string UUID
       description: |
         An optional string (consumer uuid) value to use as an "anonymous" consumer if authentication fails. If empty (default), the request will fail with an authentication failure `4xx`. Note that this value must refer to the Consumer `id` attribute that is internal to Kong, and **not** its `custom_id`.
     - name: run_on_preflight
-      required: false
+      required: true
       default: "`true`"
       datatype: boolean
       description: |
-        A boolean value that indicates whether the plugin should run (and try to authenticate) on `OPTIONS` preflight requests, if set to `false` then `OPTIONS` requests will always be allowed.
+        A boolean value that indicates whether the plugin should run (and try to authenticate) on `OPTIONS` preflight requests. If set to `false`, then `OPTIONS` requests will always be allowed.
 
 ---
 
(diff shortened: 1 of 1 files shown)


app/_hub/kong-inc/vault-auth/index.md
```diff
@@ -33,37 +33,44 @@ params:
...
       required: true
       default: "`access_token`"
+      datatype: array of string elements
       description: |
         Describes an array of comma-separated parameter names where the plugin will look for an access token. The client must send the access token in one of those key names, and the plugin will try to read the credential from a header or the querystring parameter with the same name.<br>*note*: the key names may only contain [a-z], [A-Z], [0-9], [_] and [-].
     - name: vault.id
       required: true
       default:
       value_in_examples: "<UUID>"
+      datatype: foreign
```
murillopaula, line 43:
```suggestion
      datatype: foreign UUID
```

vs

```suggestion
      datatype: foreign ID
```

what do you prefer? or just `foreign` is fine?

what is the current convention used in the site across plugins? If there is one existing convention, I'd stick to that. If not, then we create a convention now.

`foreign` is an ID referencing an existing element in the database which is identified by its ID

app/_hub/kong-inc/vault-auth/index.md
```diff
@@ -33,37 +33,44 @@ params:
...
       description: |
         A reference to an existing `vault` object within the database. `vault` entities define the connection and authentication parameters used to connect to a Vault HTTP(S) API.
     - name: secret_token_name
       required: true
       default: "`secret_token`"
+      datatype: array of string elements
       description: |
         Describes an array of comma-separated parameter names where the plugin will look for a secret token. The client must send the secret in one of those key names, and the plugin will try to read the credential from a header or the querystring parameter with the same name.<br>*note*: the key names may only contain [a-z], [A-Z], [0-9], [_] and [-].
     - name: key_in_body
```
murillopaula, line 52: I don't see this field in the plugin's schema 🤔 

There is `tokens_in_body` though. Perhaps a typo from the past?

app/_hub/kong-inc/vault-auth/index.md
```diff
@@ -33,37 +33,44 @@ params:
...
       required: false
       default: "`false`"
+      datatype:
       description: |
         If enabled, the plugin will read the request body (if said request has one and its MIME type is supported) and try to find the key in it. Supported MIME types are `application/www-form-urlencoded`, `application/json`, and `multipart/form-data`.
     - name: hide_credentials
       required: false
```
murillopaula, line 59: `required` should be `true` here right ? given that it has a default field

app/_hub/kong-inc/vault-auth/index.md
```diff
@@ -33,37 +33,44 @@ params:
...
       default: "`false`"
+      datatype: boolean
       description: |
         An optional boolean value telling the plugin to show or hide the credential from the upstream service. If `true`, the plugin will strip the credential from the request (i.e. the header or querystring containing the key) before proxying it.
     - name: anonymous
       required: false
       default:
+      datatype: string
```
murillopaula, line 67: it's not a simple string though, it must follow the `UUID` format.

app/_hub/kong-inc/vault-auth/index.md
```diff
@@ -33,37 +33,44 @@ params:
...
       description: |
-        An optional string (consumer uuid) value to use as an "anonymous" consumer if authentication fails. If empty (default), the request will fail with an authentication failure `4xx`. Please note that this value must refer to the Consumer `id` attribute which is internal to Kong, and **not** its `custom_id`.
+        An optional string (consumer uuid) value to use as an "anonymous" consumer if authentication fails. If empty (default), the request will fail with an authentication failure `4xx`. Note that this value must refer to the Consumer `id` attribute that is internal to Kong, and **not** its `custom_id`.
     - name: run_on_preflight
       required: false
```
murillopaula, line 71: `required` should be `true` here right ? given that it has a default field

murillopaula (changes requested):

app/_hub/kong-inc/vault-auth/index.md
tlblessing: > what do you prefer? or just foreign is fine?

@murillopaula I'm not aware yet of any other plugin that has this so we are creating a convention now. I think foreign ID is better.

app/_hub/kong-inc/vault-auth/index.md
tlblessing: > I don't see this field in the plugin's schema 🤔
> 
> There is `tokens_in_body` though. Perhaps a typo from the past?

Great catch!

tlblessing: I should've noticed that myself. Just testing you! Just kidding! Done in commit 6372df2.

app/_hub/kong-inc/vault-auth/index.md
murillopaula: okay, let's go with foreign ID then

app/_hub/kong-inc/vault-auth/index.md
```diff
@@ -33,39 +32,46 @@ params:
...
-      required: false
+      required: true
       default: "`false`"
+      datatype: boolean
       description: |
         An optional boolean value telling the plugin to show or hide the credential from the upstream service. If `true`, the plugin will strip the credential from the request (i.e. the header or querystring containing the key) before proxying it.
     - name: anonymous
       required: false
       default:
+      datatype: string UUID
```
murillopaula, line 66: since we're going with foreign ID, what do you think about string ID here?

app/_hub/kong-inc/vault-auth/index.md
tlblessing: okay we can always change it later

app/_hub/kong-inc/vault-auth/index.md
tlblessing: > since we're going with foreign ID, what do you think about string ID here?

@murillopaula so the description below mentions UUID. Are we making this more complicated than it needs to be? As long as we mention ID or UUID in the description and examples, is that sufficient? I'm waffling now, but I'm willing to follow your guidance as to what a developer expects to see as the datatype.

murillopaula (approved):

app/_hub/kong-inc/vault-auth/index.md
```diff
@@ -33,39 +32,46 @@ params:
     - name: access_token_name
       required: true
       default: "`access_token`"
+      datatype: array of string elements
       description: |
-        Describes an array of comma-separated parameter names where the plugin will look for an access token. The client must send the access token in one of those key names, and the plugin will try to read the credential from a header or the querystring parameter with the same name.<br>*note*: the key names may only contain [a-z], [A-Z], [0-9], [_] and [-].
+        Describes an array of comma-separated parameter names where the plugin will look for an access token. The client must send the access token in one of those key names, and the plugin will try to read the credential from a header or the querystring parameter with the same name. The key names can only contain [a-z], [A-Z], [0-9], [_], and [-].
```
lena-larionova, line 37:
```suggestion
        Describes an array of comma-separated parameter names where the plugin looks for an access token. The client must send the access token in one of those key names, and the plugin will try to read the credential from a header or the querystring parameter with the same name. The key names can only contain [a-z], [A-Z], [0-9], [_], and [-].
```

app/_hub/kong-inc/vault-auth/index.md
```diff
@@ -33,39 +32,46 @@ params:
...
     - name: vault.id
       required: true
       default:
       value_in_examples: "<UUID>"
+      datatype: foreign ID
```
lena-larionova, line 42: What exactly is a "foreign ID" datatype? We should really try to standardize here. 

Looking at your conversation earlier, I see:

> I'm not aware yet of any other plugin that has this so we are creating a convention now. I think foreign ID is better. There is no mention of UUID in the description. But the existing example shows UUID. I'll go with whatever you think is best.

There is no UUID in the description, but the value in examples one line above is a UUID. The ID in question is a UUID, so can we specify that?

app/_hub/kong-inc/vault-auth/index.md
```diff
@@ -33,39 +32,46 @@ params:
...
         A reference to an existing `vault` object within the database. `vault` entities define the connection and authentication parameters used to connect to a Vault HTTP(S) API.
     - name: secret_token_name
       required: true
       default: "`secret_token`"
+      datatype: array of string elements
       description: |
-        Describes an array of comma-separated parameter names where the plugin will look for a secret token. The client must send the secret in one of those key names, and the plugin will try to read the credential from a header or the querystring parameter with the same name.<br>*note*: the key names may only contain [a-z], [A-Z], [0-9], [_] and [-].
-    - name: key_in_body
-      required: false
+        Describes an array of comma-separated parameter names where the plugin will look for a secret token. The client must send the secret in one of those key names, and the plugin will try to read the credential from a header or the querystring parameter with the same name. The key names can only contain [a-z], [A-Z], [0-9], [_], and [-].
```
lena-larionova, line 50:
```suggestion
        Describes an array of comma-separated parameter names where the plugin looks for a secret token. The client must send the secret in one of those key names, and the plugin will try to read the credential from a header or the querystring parameter with the same name. The key names can only contain [a-z], [A-Z], [0-9], [_], and [-].
```

app/_hub/kong-inc/vault-auth/index.md
lena-larionova, line 66: I've seen UUID and i've seen string as a datatype in the past, generally not both. A UUID is a string, but it's a very specific one - a subset of the string datatype. Honestly, changing this to simply "UUID" will be just as clear, if not clearer, than the current label.

app/_hub/kong-inc/vault-auth/index.md
```diff
@@ -33,39 +32,46 @@ params:
...
+      datatype: boolean
       description: |
         An optional boolean value telling the plugin to show or hide the credential from the upstream service. If `true`, the plugin will strip the credential from the request (i.e. the header or querystring containing the key) before proxying it.
     - name: anonymous
       required: false
       default:
+      datatype: string UUID
       description: |
-        An optional string (consumer uuid) value to use as an "anonymous" consumer if authentication fails. If empty (default), the request will fail with an authentication failure `4xx`. Please note that this value must refer to the Consumer `id` attribute which is internal to Kong, and **not** its `custom_id`.
+        An optional string (consumer uuid) value to use as an "anonymous" consumer if authentication fails. If empty (default), the request will fail with an authentication failure `4xx`. Note that this value must refer to the Consumer `id` attribute that is internal to Kong, and **not** its `custom_id`.
```
lena-larionova, line 68:
```suggestion
        An optional string (consumer uuid) value to use as an "anonymous" consumer if authentication fails. If empty (default), the request will fail with an authentication failure `4xx`. 
        
        **Note:** This value must refer to the Consumer `id` attribute that is internal to Kong Gateway, and **not** its `custom_id`.
```

app/_hub/kong-inc/vault-auth/index.md
tlblessing: > I've seen UUID and i've seen string as a datatype in the past, generally not both. A UUID is a string, but it's a very specific one - a subset of the string datatype. Honestly, changing this to simply "UUID" will be just as clear, if not clearer, than the current label.

@lena-larionova and @murillopaula the schema entry for anonymous is:

 { anonymous = { type = "string", uuid = true, legacy = true }, },

The description mentions that this is a consumer uuid so I think this should just be string.

lena-larionova (approved):
LGTM

PR #2714, opened 2021-03-23
(no ticket) Add links to github from main doc set

lena-larionova:
### Summary
Adding links to edit this page and open an issue to the right sidebar, above "On this page". 
Changing the existing links on the plugins layout to open in a new tab and directly to editing status.

### Reason
Link to github was removed and overlooked during site redesign. Issue was recently brought up on Slack.

### Testing
Netlify links TBA.


HCloward (approved):
Looks great to me. I particularly like that it opens in a new tab and it opens in edit mode.

PR #2720, opened 2021-03-26
doc(kic) update Kong version compatibility

rainest:
### Summary
Add 2.2 and 2.3, and their equivalent Enterprise versions, to the KIC-Kong version compatibility tables. These inherit from 2.1; we know of no new incompatibilities since.

### Reason
These were missing before.


lena-larionova (approved):
LGTM

Future improvements: I need to make these files unversioned... it's weird that you keep having to edit multiple pages for this info. Also, I wonder if we can do something about the horrid markdown tables that'll only keep getting worse...

rainest:
Probably two options there that are better than writing the tables by hand:

- Something that takes some sort of structured data (a CSV of compatibilities or whatever) and renders the tables in the generated HTML
- Do away with the tables entirely. They're perhaps easier to scan visually than plain text, but that's already breaking down on smaller screens, since you often need to scroll in both directions to find your specific cell even now, and that will only get worse over time. I don't see any reason our supported version ranges would ever not be contiguous, so something like "Controller 1.0 through 1.2 supports Kong 2.1 through 2.3 and Kong Enterprise 2.1 through 2.3" can communicate the same info.

PR #2722, opened 2021-03-26
(no ticket) Limit existing GH actions to public repo only

lena-larionova:
### Summary
Setting an if condition on the build/test and algolia index workflows so that they will only run in the public repo.

### Reason
Working on introducing repo sync, which means I'll need to re-enable github actions on the docs private repo. Don't want every single action running in both repos.

### Testing
N/A


GayleNeumann (approved):
LGTM

PR #2724, opened 2021-03-29
[DOCS-1361] Line numbers for codeblocks

lena-larionova:
### Summary
Adding line numbers to codeblocks by default. 

With this feature, all codeblocks in the docs will have numbered lines. This will help authors call out specific content in the docs, and when referencing code examples externally (for support, for example). 

One-line codeblocks will also be numbered. This is mainly for visual consistency and alignment; it would also be a decent chunk of work to get one-liners to pick up a different style by default, so ideally we won't be doing that. 

#### Using codeblocks without line numbers
You can disable line numbers for a codeblock by using liquid tags instead of fenced codeblocks.

Default (fenced code blocks):
\```bash
some code here
\```

To remove line numbers, use the `{% highlight %}` tag with an optional language class:

{% highlight bash %}
some code here
{% endhighlight %}

### Reason
https://konghq.atlassian.net/browse/DOCS-1361

### Testing
Anywhere that codeblocks exist in the docs. Eg https://deploy-preview-2724--kongdocs.netlify.app/hub/kong-inc/application-registration/

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -32,6 +32,7 @@ Consult the Table of Contents below, and jump to the desired section.
     - [Links](#links)
     - [Info Blocks](#info-blocks)
     - [Table of Contents generator](#table-of-contents-generator)
+    - [Codeblocks](#codeblocks)
     - [Using navtabs within topics](#using-navtabs-within-topics)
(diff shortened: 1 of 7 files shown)


GayleNeumann (approved):
LGTM

PR #2733, opened 2021-03-31
(no ticket) Fix key_name syntax for key auth plugins

lena-larionova:
### Review
@ashosborne 

### Summary
Add `value_in_examples` to display the desired array syntax for the `key_name` value in the key auth plugins.
Also fixing the extra blank space that appears before declarative config example arrays.

### Reason
The `key_name` value is not displaying correctly in declarative examples. It should be an array.

Issue reported on Slack.

### Testing
See the declarative tab in
https://deploy-preview-2733--kongdocs.netlify.app/hub/kong-inc/key-auth-enc/ and
https://deploy-preview-2733--kongdocs.netlify.app/hub/kong-inc/key-auth/

### DIFF
--- a/app/_hub/kong-inc/key-auth-enc/1.0.x.md
+++ b/app/_hub/kong-inc/key-auth-enc/1.0.x.md
@@ -33,6 +33,7 @@ params:
     - name: key_names
       required: false
       default: "`apikey`"
+      value_in_examples: ["apikey"]
       description: |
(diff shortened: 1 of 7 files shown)


ashosborne (approved):
Copied + pasted the decK config in the preview and it works fine. Also checked key-auth encrypted which is also correct. Thanks Lena.

Bradamant3 (approved):
LGTM

PR #2734, opened 2021-04-01
update wording using a database for Kong

ctrongminh:
### Review

@reviewer @Kong/team-docs

### Summary
IMHO, in the section "Using a database for Kong for Kubernetes", there would be 2 cases: the first case is that user does not want to run Kong with the DB, and the other case is "decide to use a DB".
Please decline this PR, if my understanding is incorrect


lena-larionova (approved):

lena-larionova:
Thanks for catching this - the phrasing was definitely wrong before.

PR #2735, opened 2021-04-01
added ee 2.1.4.5 changelog entry

HCloward:
### Review
@tyler-ball 

### Summary
The Fast Track team released a patch release yesterday to fix an issue users were having when migrating from 1.5 to 2.1.

### Reason
Need to add 2.1.4.5 patch release note to inform users of the fix.

### Testing
[2.1.4.5 changelog](https://deploy-preview-2735--kongdocs.netlify.app/enterprise/changelog/#2145)

### DIFF
--- a/app/enterprise/changelog.md
+++ b/app/enterprise/changelog.md
@@ -756,7 +756,6 @@ open-source **Kong Gateway 2.2.0.0**:
 ### Fixes
 
 #### Enterprise
-
 Fixed an issue when upgrading Kong Enterprise Gateway from v1.5.x to v2.1.x. Before the fix, when admin consumers were shared across multiple workspaces, it was possible for the migration to fail. This happened because plugin entities that depend on consumer entities must live in the same workspace as the consumer entity. This fix migrates these plugin entities to the same workspace as the consumer entities. Customers affected by this issue will need to upgrade to at least v2.1.4.5 before attempting to migrate from v1.5.x to v2.1.x.
 
 ## 2.1.4.4
--- a/app/enterprise/changelog.md
+++ b/app/enterprise/changelog.md
@@ -750,6 +750,15 @@ open-source **Kong Gateway 2.2.0.0**:
 - The `shorthands` attribute in schema definitions is deprecated in favor of
   the new `shorthand_fields` top-level attribute.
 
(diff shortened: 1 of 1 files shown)


app/enterprise/changelog.md
```diff
@@ -750,6 +750,14 @@ open-source **Kong Gateway 2.2.0.0**:
 - The `shorthands` attribute in schema definitions is deprecated in favor of
   the new `shorthand_fields` top-level attribute.
 
+## 2.1.4.5
+**Release Date** 2021/03/31
+
+### Fixes
+
+#### Enterprise
+Fixed an issue when upgrading Kong Enterprise Gateway from v1.5.x to v2.1.x. Before the fix, when admin consumers were shared across multiple workspaces, it was possible for the migration to fail. This happened because plugin entities that depend on consumer entities must live in the same workspace as the consumer entity. This fix migrates these plugin entities to the same workspace as the consumer entities. Customers affected by this issue will need to upgrade to at least v2.1.4.5 before attempting to migrate from v1.5.x to v2.1.x.
```
lena-larionova, line 759:
```suggestion
Fixed an issue when upgrading Kong Gateway (Enterprise) from v1.5.x to v2.1.x. Before the fix, when admin consumers were shared across multiple workspaces, it was possible for the migration to fail. This happened because plugin entities that depend on consumer entities must live in the same workspace as the consumer entity. This fix migrates these plugin entities to the same workspace as the consumer entities. Customers affected by this issue will need to upgrade to at least v2.1.4.5 before attempting to migrate from v1.5.x to v2.1.x.
```

lena-larionova, line 759: > will need to upgrade to at least v2.1.4.5 before attempting to migrate from v1.5.x to v2.1.x.

This is confusing phrasing; "upgrade to 2.1.4.5 before upgrading to 2.1.x" doesn't really make sense. Do they need to upgrade 1.5.x to 2.1.4.5 and then to whatever other version? That is, we're telling them to skip all 2.1.x versions before 2.1.4.5.

HCloward: ok, I updated that line - let me know what you think

lena-larionova (approved):

PR #2743, opened 2021-04-06
Mark Okta as an enterprise integration

bforbis:
The Okta integration guide requires a Kong Enterprise license, as it makes use of the Kong OIDC Enterprise plugin

### Summary
Document Okta integration guide as "enterprise" 

### Reason
The Okta integration guide is not marked as enterprise, however the instructions in the guide requires you to have an enterprise account

### DIFF
--- a/app/_hub/okta/okta/index.md
+++ b/app/_hub/okta/okta/index.md
@@ -7,6 +7,8 @@ categories:
 
 type: integration
 
+enterprise: true
+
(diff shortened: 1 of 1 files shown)


lena-larionova:
Thanks for the contribution, good catch!

bforbis:
@lena-larionova it looks like making this change did not mark this as an enterprise integration? Perhaps because the page is of type "integration" and not of type "plugin"? Is it possible to mark an integration as enterprise?

PR #2745, opened 2021-04-06
Update rate-limiting.md

aaronArinder:
### Summary
Typo: `he` => `The`

### Reason
The hairs on the back of my neck stand up when I see the sheer chaos of a typo, just like every other internet denizen with too much time to kill.

### Testing
See if your hairs stand up when reading the fixed sentence.


lena-larionova (approved):
You're the first to point this out in a topic that has been around and mostly unchanged for years... nice catch 😁.

PR #2749, opened 2021-04-07
fix(kic) correct IngressClass example

### LINKED ISSUE #2585: IngressClass definition is missing value for spec.controller
- `docs.konghq.com` url(s): https://docs.konghq.com/kubernetes-ingress-controller/1.1.x/concepts/ingress-versions/

### Summary

Creating IngressClass using documentation instructions fails due to missing `spec.controller` field. I searched a bit, but I could not yet find the intended value for this parameter. Documentation fix would be good for this.

(The helm chart did not seem to create this object automatically)

### Steps To Reproduce

1. Create IngressClass YAML per documentation
```
$ cat ingressclass.yaml 
---
apiVersion: networking.k8s.io/v1
kind: IngressClass
metadata:
  name: kong
spec:
```
2. Applying the YAML fails
```
$ kubectl apply -f ingressclass.yaml
The IngressClass "kong" is invalid: spec.controller: Required value
```

### Additional Details

Kubernetes server & client version: 1.20.2

rainest:
### Summary
Add a controller field to the IngressClass example. Although we don't yet make use of this value in our implementation, Kubernetes requires it.

### Reason
Fix #2585

### DIFF
--- a/app/kubernetes-ingress-controller/1.0.x/concepts/ingress-versions.md
+++ b/app/kubernetes-ingress-controller/1.0.x/concepts/ingress-versions.md
@@ -114,6 +114,7 @@ kind: IngressClass
 metadata:
   name: kong
 spec:
+  controller: ingress-controllers.konghq.com/kong
 ---
 apiVersion: networking.k8s.io/v1
 kind: Ingress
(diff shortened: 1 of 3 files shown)


lena-larionova (approved):
LGTM

PR #2742, opened 2021-04-05
Minor updates to the Konnect Helm instructions

rossmcdonald:
### Review

@reviewer

### Summary

I ran through these instructions and found a few minor errors that may trip up users.

### Reason
To improve the Konnect user onboarding/setup experience.

### Testing
Run the example commands.

PR #2750, opened 2021-04-07
[DOCU-1388] Konnect runtimes rework

lena-larionova:
### Review
@rossmcdonald @skibum @jackkav 

### Summary
Reworking the docs for the Konnect runtime manager.
* Splitting one long config page into three: docker, kubernetes, and kong.conf (other). 
* Validated Docker install, found a bunch of issues, got it working.
* Pulled in various feedback on Helm install:
  * Removed link to sample `values.yaml`, as it wasn't appropriate for Konnect. Directing users to write a config file from scratch; maybe we can add a sample Konnect-specific data plane config file in the future?
  * Cleanup for consistency:
    * `yaml` vs `yml` - went with `yaml`, since that's what we normally use. Doesn't actually matter as long as its consistent, which it wasn't.
    * added missing `-n kong` to secrets
    * changed secret paths to `<path-to-file>` instead of `/tmp/` so that users know it needs to be set
    * sample values.yaml now pulls down the gateway enterprise image instead of OSS; version is a variable, so we don't need to manually update it.
    * Remove all options to update an existing system. Doc now assumes that you're starting from scratch, for simplicity.

**To-do:** figure out how to find the proxy URL on kubernetes post-install, and how to set a custom one. @rossmcdonald 

### Reason
Helm setup was confusing and also not entirely accurate.
Along the way, realized that the Docker setup was also not fully accurate, so updated that as well.

Doc ticket: https://konghq.atlassian.net/browse/DOCU-1388

Covering issues logged in https://github.com/Kong/docs.konghq.com/issues/2728 and https://github.com/Kong/docs.konghq.com/pull/2742. 

### Testing
[Overview page with tiles](https://deploy-preview-2750--kongdocs.netlify.app/konnect/runtime-manager/#types-of-runtimes)
[Set up a gateway runtime on Docker](https://deploy-preview-2750--kongdocs.netlify.app/konnect/runtime-manager/gateway-runtime-docker/)
[Set up a gateway runtime on Kubernetes](https://deploy-preview-2750--kongdocs.netlify.app/konnect/runtime-manager/gateway-runtime-kubernetes)
[Set up a gateway runtime with kong.conf](https://deploy-preview-2750--kongdocs.netlify.app/konnect/runtime-manager/gateway-runtime-conf)


jackkav:
Hey nice work so far! I also noticed this copy is a little misleading `Important: Do not navigate away from this page while saving the certificate and key files. They are unique and won't display again.`

It's my understanding that certs can be regenerated if lost, 'won't display again' seems too strong. Not blocking feedback.

lena-larionova:
> Hey nice work so far! I also noticed this copy is a little misleading `Important: Do not navigate away from this page while saving the certificate and key files. They are unique and won't display again.`
> 
> It's my understanding that certs can be regenerated if lost, 'won't display again' seems too strong. Not blocking feedback.

Hmm. Yeah, the exact same certs won't display again, but you can just regenerate them. I can adjust the phrasing here to not be quite as strong/scary, and maybe instead focus on the fact that they need to stay here to grab the config parameters anyway.

app/_includes/md/konnect/runtime-certs.md
```diff
@@ -0,0 +1,30 @@
+<!-- Shared between Konnect gateway runtime config topics: Docker, Kubernetes, and kong.conf -->
+1. In Konnect, from the left navigation menu, select **Runtimes**.
+
+    For the first runtime, the page opens to a **Configure New Runtime** form.
+
+    Once configured, this page lists all runtimes associated with the
+    {{site.konnect_short_name}} SaaS account.
+
+2. (Optional) If this is not the first runtime configuration, click
+**Configure New Runtime**.
```
skibum, line 10: In order to generate certs the user must also click into the Advanced tab

app/konnect/runtime-manager/gateway-runtime-kubernetes.md
```diff
@@ -0,0 +1,152 @@
...
+    The Runtime Manager will include a new entry for your instance.
+
+<div class="alert alert-ee warning">
+<b>Important:</b> Certificates expire after 30 days and must be renewed. See
+<a href="/konnect/runtime-manager/renew-certificates">Renew Certificates</a>.
+</div>
+
+## Access services using the proxy
+
+Run:
```
skibum, line 143: I'm wondering if we should add an brief into to this section like "In order to proxy traffic through the runtime you need the external IP address. To obtain that address:"

app/_data/docs_nav_konnect.yml
```diff
@@ -60,7 +60,15 @@
     - text: Overview
       url: /runtime-manager
     - text: Manage a Kong Gateway Runtime
-      url: /runtime-manager/kong-gateway-runtime
+      items:
+        - text: Docker
+          url: /runtime-manager/gateway-runtime-docker
+        - text: Kubernetes
+          url: /runtime-manager/gateway-runtime-kubernetes
+        - text: Universal (kong.conf)
```
skibum, line 68: Do we use this Universal label elsewhere in the docs? Should this just be 'Linux'?

skibum (changes requested):

app/konnect/runtime-manager/gateway-runtime-kubernetes.md
lena-larionova: You're right - I was leaving this section sparse until I could verify what the actual process is. But no reason not to write it already.

app/_data/docs_nav_konnect.yml
lena-larionova: Only in the Mesh docs. Linux works!

lena-larionova:
Reviewers: could someone to validate out how to set up the proxy? I'm looking at https://github.com/Kong/charts/blob/39b83558c8e8efc40d48db808647b313a2a3f57f/charts/kong/example-values/full-k4k8s-with-kong-enterprise.yaml#L47 - looks like you need to add another section to the config file that I'm missing?

app/konnect/runtime-manager/gateway-runtime-kubernetes.md
```diff
@@ -0,0 +1,168 @@
...
+Kubernetes secrets.
+
+1. Create a `tls` secret using the `cluster.cert` and `cluster.key` files
+you saved earlier:
+
+    ```bash
+    $ kubectl create secret tls kong-cluster-cert -n kong \
+      --cert=/<path-to-file>/cluster.crt \
+      --key=/<path-to-file>/cluster.key
+    ```
```
jackkav, line 69: these files are mapped to tls.key and tls.cert inside the cluster which could cause some confusion. Might help to make these match. As if things aren't working it could cause undue concern.

app/konnect/runtime-manager/gateway-runtime-kubernetes.md
```diff
@@ -0,0 +1,168 @@
...
+
+1. Back in {{site.konnect_short_name}}, copy the
+codeblock from the **Configuration Parameters** section.
+
+2. Create a `values.yaml` file. Remove the `KONG_` prefix from the parameters
+in the sample codeblock and add the following parameters to the file.
+
+    Make sure to replace the values in `cluster_cert`, `cluster_cert_key`,
+    and `cluster_ca_cert` with references to the secrets you created earlier:
+
```
jackkav, line 88: The `Write configuration` section could use some tweaks:
- this code block could be more specific there are multiple this could refer to
- remove the kong prefix is not clear enough as its also changes from upper to lower case, either way expecting a person following the docs to transform key casing introduces too much uncertainty. @skibum ideally, this should be in the advanced konnect tab ready to be copy pasted.
- replace the values isn't clear about what should be replaced

app/konnect/runtime-manager/gateway-runtime-kubernetes.md
```diff
@@ -0,0 +1,168 @@
...
+6. Apply the `values.yaml`:
+
+    ```bash
+    $ helm install my-kong kong/kong -n kong \
+      --values ./values.yaml
+    ```
+
+7. Clean up:
+    ```bash
+    $ kubectl delete jobs -n kong --all
```
jackkav, line 133: this could use some explanation about why this might be used

jackkav:
Since changing the doc from kong OSS 2.3 to kong-ee 2.3.2.0 I'm also seeing the dataplanes list is being updated with a new instance, but the containers keep restarting complaining about the missing license json

lena-larionova:
> Since changing the doc from kong OSS 2.3 to kong-ee 2.3.2.0 I'm also seeing the dataplanes list is being updated with a new instance, but the containers keep restarting complaining about the missing license json, I tried adding the license key but no joy. Couldn't get it working.

Maybe this portion is missing: https://github.com/Kong/charts/blob/39b83558c8e8efc40d48db808647b313a2a3f57f/charts/kong/values.yaml#L592

The only thing here is that for hybrid mode, the control plane is supposed to populate the license, right? 
I'm not actually sure how that's supposed to work for enterprise images. Here's what I know:
* For enterprise free mode, you can run without a license by using a dummy license secret: https://docs.konghq.com/enterprise/2.3.x/deployment/installation/kong-on-kubernetes/#license
* For enterprise with a license, you populate the secret with a real license
* For hybrid mode - we have no doc here, and I can't find anything about it in the Kong/charts repo. For Konnect, afaik the control plane is supposed to populate the data plane with a license automatically, and that seems to work with Docker. For Konnect, it's basically hybrid mode + enterprise gateway, and that's a use case the examples don't cover at all...

skibum:
@lena-larionova yes, in Konnect the cloud control plane provides the license to all the connected GW runtimes (data planes). @jackkav if you are seeing licensing errors then it may be an issue with your control plane not having a license. We can investigate.

app/konnect/runtime-manager/gateway-runtime-kubernetes.md
lena-larionova: Good question - pulled it from an [existing installation topic](https://docs.konghq.com/enterprise/2.3.x/deployment/installation/kong-on-kubernetes/#step-8-finalize-configuration-and-verify-installation). I'm guessing the answer is probably "good practice"?

jackkav: My understanding is the doc is targeting a user who already uses k8s. I'm no expert just trying to be helpful, but it seems to me k8s good practice is out of scope for this doc. Also as far as I can tell there aren't any jobs to run for a data plane, a typical job would be a db migration which happens at the control plane.

jackkav:
@Kong/team-k8s

app/konnect/runtime-manager/gateway-runtime-kubernetes.md
rossmcdonald: > remove the kong prefix is not clear enough as its also changes from upper to lower case, either way expecting a person following the docs to transform key casing introduces too much uncertainty.

I see this as the biggest barrier to this deployment option. There is no level of documentation that is going to make this less confusing or error-prone, and it's a fairly simple change from a UI perspective (just remove the prefix and lower-case the variable names that are already shown). Even better would be a download option for a zip with a pre-filled `values.yaml` and all of the certs/keys ready to go.

skibum: We are planning to update the UI to fix the casing and the prefix issue.

skibum (approved):

app/konnect/runtime-manager/gateway-runtime-kubernetes.md
lena-larionova: Addressed bullet points 1 and 3 in doc - you caught the fact that the order of operations was just... not logical, so I reworked it. Point 2 I _sort of_ covered, but really it'll be fixed in the UI, and that will be way better.

GayleNeumann (approved):
Looks great!

app/konnect/runtime-manager/gateway-runtime-kubernetes.md
```diff
@@ -0,0 +1,168 @@
...
+    - kong-cluster-ca
+
+    admin:
+      enabled: false
+
+    env:
+      role: data_plane
+      database: "off"
+      anonymous_reports: off
+      vitals_ttl_days: 732
```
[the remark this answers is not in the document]
jackkav: This is supposed to be 2 years in days, 366 * 2 = 732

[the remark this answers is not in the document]
jackkav: @lena-larionova

app/konnect/runtime-manager/gateway-runtime-conf.md
```diff
@@ -0,0 +1,100 @@
...
+to the file, remove the `KONG_` prefix, and turn them into lowercase.
+
+    The result should look something like this, replacing placeholder values
+    with your own from {{site.konnect_short_name}}:
+
+    ```sh
+    role = data_plane
+    database = off
+    anonymous_reports = off
+    vitals_ttl_days = 723
```
jackkav, line 56: 732

app/konnect/runtime-manager/gateway-runtime-docker.md
```diff
@@ -0,0 +1,155 @@
...
+
+    See [Parameters](/konnect/runtime-manager/runtime-parameter-reference) for
+    descriptions and the matching fields in {{site.konnect_short_name}}.
+
+    ```sh
+    $ docker run -d --name kong-gateway-dp1 \
+      -e "KONG_ROLE=data_plane" \
+      -e "KONG_DATABASE=off" \
+      -e "KONG_ANONYMOUS_REPORTS=off" \
+      -e "KONG_VITALS_TTL_DAYS=723" \
```
jackkav, line 113: 732

app/konnect/runtime-manager/gateway-runtime-kubernetes.md
```diff
@@ -0,0 +1,197 @@
...
+    - kong-cluster-ca
+
+    admin:
+      enabled: false
+
+    env:
+      role: data_plane
+      database: "off"
+      anonymous_reports: off
+      vitals_ttl_days: 723
```
jackkav, line 100: 732

jackkav (changes requested):

app/konnect/runtime-manager/gateway-runtime-kubernetes.md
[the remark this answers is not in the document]
rossmcdonald: @jackkav The generated config says `723`:

image

We should fix it in Konnect as well if that's the case.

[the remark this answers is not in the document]
skibum: Thanks for catching that @jackkav . I'll flag to be fixed in the UI.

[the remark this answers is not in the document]
jackkav: I fixed it in a PR merged a few days ago.

jackkav (approved):

PR #2772, opened 2021-04-12
fix curl command url 

david972:
This proposition is to fix a mistake in url in curl command in Set up the Key Authentication Plugin

### Review

@reviewer

### Summary
In Set up the Key Authentication Plugin part
This PR fix the URI example for curl command

### Reason
because Set up the Key Authentication Plugin the URI routes is not existing in this case

### Testing
In Set up the Key Authentication Plugin
By following the example in CURL mode you can reproduce the issue.
And the fix correct this

### DIFF
--- a/app/getting-started-guide/2.3.x/secure-services.md
+++ b/app/getting-started-guide/2.3.x/secure-services.md
@@ -61,7 +61,7 @@ created:
 {% navtabs codeblock %}
 {% navtab cURL %}
 ```sh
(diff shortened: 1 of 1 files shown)


lena-larionova:
Hi @david972, thanks for your submission. Unfortunately, this seems like an implementation error - the routes endpoint does indeed support plugins: https://docs.konghq.com/gateway-oss/2.3.x/admin-api/#add-plugin

PR #2790, opened 2021-04-14
fix version variable for mesh

Bradamant3:
### Review
@reviewer

### Summary
Call Mesh version consistently in Mesh install docs

### Reason
Stray odd call to version variable was creating inconsistent version numbers in Mesh install docs, leading to user confusion

### Testing
will post Netlify preview when it's ready


lena-larionova (approved):
LGTM - something to revisit for the next Mesh release though, as we should probably have a versioned install script.

PR #2794, opened 2021-04-15
Added note about using the global catalog port for Active Directory

felderi:
### Review

@reviewer

### DIFF
--- a/app/enterprise/2.3.x/kong-manager/service-directory-mapping.md
+++ b/app/enterprise/2.3.x/kong-manager/service-directory-mapping.md
@@ -62,6 +62,8 @@ admin_gui_session_conf = { "secret":"set-your-string-here" }
 
 Configure LDAP Authentication for Kong Manager with the following properties. Note the attribute variables are defined below:
 
+For Active Directory environments, instead of port 389, user the Global Catalog port which is 3268 by default.
+
 ```
 admin_gui_auth_conf = {
  "anonymous":"", \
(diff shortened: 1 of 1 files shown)


app/enterprise/2.3.x/kong-manager/service-directory-mapping.md
```diff
@@ -62,6 +62,8 @@ admin_gui_session_conf = { "secret":"set-your-string-here" }
 
 Configure LDAP Authentication for Kong Manager with the following properties. Note the attribute variables are defined below:
 
+For Active Directory environments, instead of port 389, user the Global Catalog port which is 3268 by default.
```
lena-larionova, line 65: This might make more sense as part of the breakdown of properties below the codeblock, right after `ldap_password`. Something like this:

* `ldap_port`: the LDAP port, which defaults to `389`. 
  * For Active Directory environments, use the Global Catalog port, which is `3268` by default.

tlblessing: "defaults" is not a word but is frequently in docs written by developers. I'll check out the branch and edit.

tlblessing:
@murillopaula can you please look over this suggested change for LDAP? Should we add this info to the plugin hub doc?

tlblessing (approved):

murillopaula (approved):

murillopaula:
@tlblessing no, we already mention about the global directory catalog in the plugin hub.

PR #2797, opened 2021-04-16
(no ticket) Change gateway version in konnect helm config

lena-larionova:
### Summary
Change variable to hardcoded version, as the latest version doesn't work with Konnect.

### Reason
There's a breaking change with the OIDC plugin that breaks compatibility with 2.3.3.x

### Testing
https://deploy-preview-2797--kongdocs.netlify.app/konnect/runtime-manager/gateway-runtime-kubernetes/#write-and-apply-configuration - see codeblock


GayleNeumann (approved):
LGTM

PR #2803, opened 2021-04-20
fix(hybrid) correct example for compatibility check

fffonion:
### Review

@reviewer


tlblessing (approved):

tlblessing:
Thanks for the PR @fffonion. Looks like the Enterprise side is already corrected, so going to go ahead and merge.

PR #2824, opened 2021-04-29
[DOCU-1432] Bintray migration: k8s docs

lena-larionova:
### Review
@mflendrich @rainest 

### Summary
Updating docs to point to new image location and removing requirement for a Docker registry secret. 

### Reason
Bintray migration.

### Testing
Netlify links TBA.


rainest:
The short-lived kong-enterprise-k8s image is the gift that keeps on giving. That wasn't actually migrated yet, and presumably won't be getting a public home (possibly, but unlikely--I don't know what all was needed to produce public builds of the standard Enterprise images, but kong-enterprise-k8s had its own build process and it's probably not worth porting the public build stuff to it).

2.1+ are fine as such, but we'll need to figure out the new home for kong-enterprise-k8s and preserve the credential instructions for it on the older pages.

lena-larionova:
@rainest Added a note to everywhere I could find that mentioned that image with instructions to contact support if they need it. I think that covers it from the docs side? Originally, I was thinking I might need to provide instructions on how to access it, but it seems like support will be covering all of that.

rainest (approved):
Yep, sounds good to me.

GayleNeumann (approved):
LGTM. Nice work!

PR #2830, opened 2021-05-07
Fix documented priority of the ip-restriction plugin

ndemengel:
### Summary
According to https://github.com/Kong/kong/blob/master/kong/plugins/ip-restriction/handler.lua#L10,
the value is 990 and not 3000 (because of this PR: Kong/kong#3526)

Therefore the documentation is misleading :-)


Bradamant3 (approved):
LGTM and thanks!

PR #2835, opened 2021-05-10
doc(kic) remove stderrthreshold flag

rainest:
### Summary
Remove the --stderrthreshold flag from KIC documentation. This is a leftover from an old logging library we used prior to 0.10.0. It hasn't actually been available since.

### Reason
Found when reviewing log flag updates for KIC 2.0: https://github.com/Kong/kubernetes-ingress-controller/pull/1294#pullrequestreview-655935956


rainest:
Aight, nevermind then. https://github.com/Kong/kubernetes-ingress-controller/issues/1297 instead.

PR #2842, opened 2021-05-13
docs(kong-mesh) describe how to configure vault step by step

jakubdyszkiewicz:
* Added a description of how to configure Vault with Kong Mesh
* Changed inline to inlineString, which is probably more convenient for the token, you don't have to do base64
* Changed the structure of Authentication to have `## Using Vault CA` -> `### Configure Vault`, `### Configure Mesh`. mTLS to Vault is optional.

TODO:
can't get the tabs to be properly put into steps!


Bradamant3:
@jakubdyszkiewicz I copied your changes over to my release PR (#2845) where I had to fix the tabs issue differently (turns out we don't support tabs nested inside lists, it was not you). So closing this one unmerged bc taken care of there. Thank you!

PR #2857, opened 2021-05-18
Add Konnect promo banner

adamjkuhn:
### Review
@lena-larionova 

### Summary
* add promo banner
* add necessary scripts to dismiss/compress promo banner including updated localStorage var
* style promo banner & open/closed/scrolled states
* add conditional margin to page content to accomodate additional header size

### Reason
Marketing update


lena-larionova (approved):
LGTM - the broken check is due to an outage on Github Actions, and from what I can see, nothing here should cause issues with it. I ran a quick test locally as well to double-check.

PR #2872, opened 2021-05-19
[DOCU-1503] Clean up straggling 2.3 content in Gateway EE 2.4

lena-larionova:
### Review
@HCloward 

### Summary
* Update install version in EE 2.4 topics
* Update Docker note for accessing old images
* Fix statsd include
* Add 2.4 to immunity version support table
* Some minor adjustments to version-specific mentions and URLs

### Reason
https://konghq.atlassian.net/browse/DOCU-1503

### Testing
TBA


HCloward (approved):
I added an update for Immunity I needed to do for the Bintray sunsetting. Also, I think this will fix things for 2.4 anyway, but in general I think we should consider an approach that will not require an update with every release. Perhaps we just use some kind if sample text instead of a specific version number? What do you think?

lena-larionova:
> Also, I think this will fix things for 2.4 anyway, but in general I think we should consider an approach that will not require an update with every release. Perhaps we just use some kind if sample text instead of a specific version number? What do you think?

I'm inclined to agree, but we've had pushback on this before, from support. They like being able to simply copy & paste, and I'm guessing customers do as well. We should be able to figure out something with variables though... I'm just not sure how to pull the correct "real" version for every release right now.

HCloward:
Ran checks locally bc issues with GitHub Issues

PR #2879, opened 2021-05-21
Rename X-Kong-Debug to Kong-Debug

B-Galati:
### Summary
It looks like there is a typo in the doc about this header.
I found the solution on this topic https://discuss.konghq.com/t/kong-1-4-0-using-x-kong-debug-1-doesnt-seem-to-work/4949.

I know the PR is not ok. That is a conversation start. I'll be happy to finish it as required.


gszr (commented):
I can confirm the change is correct.

B-Galati:
@lena-larionova I don't need to update all the other doc version ?

lena-larionova:
> I don't need to update all the other doc version ?

@B-Galati No need, though I won't stop you if you decide to update :) We're currently sticking with updating just the latest version of the open-source gateway docs, as we're working on a better way to handle versions in the future and don't want to create extra work.

PR #2884, opened 2021-05-26
remove alerts and redundant content, light copyedits

Bradamant3:
### Review

@lena-larionova 

### Summary

Clean up Mesh install content for usability:

- removed all alerts
- also removed redundant content about needing a license
- changes only for v1.2

### Reason

UX

### Testing

Will post netlify links here when deploy preview ready

### DIFF
--- a/app/_includes/md/mesh/install-universal-run.md
+++ b/app/_includes/md/mesh/install-universal-run.md
@@ -8,13 +8,6 @@ Navigate to the `bin` folder:
 $ cd kong-mesh-{{page.kong_latest.version}}/bin
 ```
 
-<div class="alert alert-ee blue">
-<strong>Note:</strong> Before running the {{site.mesh_product_name}}
-control plane process in the next step &mdash; which is served by the <code>kuma-cp</code>
-executable &mdash; you need to have a valid {{site.mesh_product_name}} license
-in place.
-</div>
-
 Then, run the control plane with:
 
 ```sh
@@ -36,8 +29,7 @@ in `/usr/local/bin/` by executing:
 $ ln -s ./kumactl /usr/local/bin/kumactl
 ```
 
-<div class="alert alert-ee blue">
 <strong>Note:</strong> By default, this will run {{site.mesh_product_name}} with a memory
 <a href="https://kuma.io/docs/latest/documentation/backends/">backend</a>, but you can use a persistent storage like PostgreSQL by updating the
 <code>conf/kuma-cp.conf</code> file.
-</div>
+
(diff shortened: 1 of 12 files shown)


app/_includes/md/mesh/install-universal-run.md
```diff
@@ -36,8 +29,6 @@ in `/usr/local/bin/` by executing:
 $ ln -s ./kumactl /usr/local/bin/kumactl
 ```
 
-<div class="alert alert-ee blue">
 <strong>Note:</strong> By default, this will run {{site.mesh_product_name}} with a memory
```
lena-larionova, line 32: With the <div> tag removed, these are all still labelled as notes but are not using Note styling. Is that something we want to do? Might be better to be more discerning for whether something is a note or not, and always use the blockquote style if it is one. Plus, if you don't want too many panels on a page, then it makes you start thinking about why you're adding so many notes, and whether they're actually better as part of the content flow. 

Tl;dr, I suggest removing the "Note" prefix in these sections that you've de-blued. The final solution might be to rework these topics altogether, but I realize that you're not trying to do that here.

(Applies to all changes in this PR)

lena-larionova (commented):
(I thought I had left a review yesterday, but apparently I forgot to actually submit it 🤦‍♀️ )

Looks good, I'm just concerned about all the leftover "note" labels.

app/mesh/1.2.x/installation/openshift.md
```diff
@@ -231,46 +222,44 @@ $ kumactl config control-planes add --name=XYZ --address=http://{address-to-kong
...
+{{site.mesh_product_name}} explicitly specifies a UID
+for the `kuma-dp` sidecar to avoid capturing traffic from
+`kuma-dp` itself. You must grant a `nonroot` [Security Context Constraint]
+(https://docs.openshift.com/container-platform/latest/authentication/managing-security-context-constraints.html)
+to the application namespace:
 
-<pre>
-<code>$ oc adm policy add-scc-to-group nonroot system:serviceaccounts:&lt;app-namespace&gt;</code></pre>
+```sh
+$ oc adm policy add-scc-to-group nonroot system:serviceaccounts:&lt;app-namespace&gt;
```
lena-larionova, line 235:
```suggestion
$ oc adm policy add-scc-to-group nonroot system:serviceaccounts:<app-namespace>
```
The HTML entity names for the `<` and `>` characters don't render in fenced codeblocks.

lena-larionova (approved):
LGTM - just one tiny rendering fix, now that this content isn't using HTML tags.

PR #2886, opened 2021-05-26
updates required fields for request-validator plugin

dimitropoulos:
it appears that these are not actually required because they have defaults

### Review

@reviewer

### DIFF
--- a/app/_hub/kong-inc/request-validator/index.md
+++ b/app/_hub/kong-inc/request-validator/index.md
@@ -42,7 +42,6 @@ params:
         must be specified.
 
     - name: allowed_content_types
-      required: true
       default: "application/json"
       value_in_examples:
       datatype: Set of string elements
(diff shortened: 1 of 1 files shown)


tlblessing:
@dimitropoulos regardless of defaults, the schema indicates allowed_content_types is a required field 
image
cc: @murillopaula .

dimitropoulos:
I'm sorry, I don't understand.  A default refers to the behavior when it's not set, but required means that it cannot _not_ be set.  This seems to be a contradiction to me.  Am I missing something?

In other words, if it's required then it does't make sense to have a default because that default value would never be used (because the user must always provide their own value because it's required).

tlblessing:
> I'm sorry, I don't understand. A default refers to the behavior when it's not set, but required means that it cannot _not_ be set. This seems to be a contradiction to me. Am I missing something?
> 
> In other words, if it's required then it does't make sense to have a default because that default value would never be used (because the user must always provide their own value because it's required).

Perhaps a bug ticket should be logged to update the schema.

lena-larionova:
Closing this and taking to a separate project/conversation, as there is a bigger problem here.

PR #2891, opened 2021-05-27
fix/(robots.txt) Add Disallow statements

hhoover:
### Summary
This fix removes old versions of Kong documentation from search
results.

Follows Google Robot Spec https://developers.google.com/search/docs/advanced/robots/robots_txt

### Reason
Issue 2890

### Testing
https://support.google.com/webmasters/answer/6062598?hl=en

### DIFF
--- a/app/robots.txt
+++ b/app/robots.txt
@@ -2,4 +2,25 @@
 layout: null
 ---
 User-agent: *
+Disallow: /enterprise/0*
+Disallow: /enterprise/1*
+Disallow: /enterprise/2.1.x/
+Disallow: /enterprise/2.2.x/
+Disallow: /enterprise/2.3.x/
+Disallow: /mesh/1.0*
+Disallow: /mesh/1.1*
+Disallow: /gateway-oss/0*
+Disallow: /gateway-oss/1*
+Disallow: /gateway-oss/2.0*
+Disallow: /gateway-oss/2.1*
+Disallow: /gateway-oss/2.2*
+Disallow: /gateway-oss/2.3*
+Disallow: /kubernetes-ingress-controller/1.0*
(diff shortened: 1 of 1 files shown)


app/robots.txt
```diff
@@ -2,4 +2,20 @@
...
+Disallow: /mesh/1.1*
+Disallow: /gateway-oss/0*
+Disallow: /gateway-oss/1*
+Disallow: /gateway-oss/2.0*
+Disallow: /gateway-oss/2.1*
+Disallow: /gateway-oss/2.2*
+Disallow: /gateway-oss/2.3*
+Disallow: /kubernetes-ingress-controller/1.0*
+Disallow: /kubernetes-ingress-controller/1.1*
+
```
lena-larionova, line 20: I'd also add the getting started guide here:

```suggestion
Disallow: /getting-started-guide/CE*
Disallow: /getting-started-guide/2.0*
Disallow: /getting-started-guide/2.1*
Disallow: /getting-started-guide/2.2*
Disallow: /getting-started-guide/2.3*
```

lena-larionova (commented):
Thanks for putting this together! Do you know, by any chance, if adding a version to this file the same day as releasing a new version would hurt site SEO? Or would it be fine due to already indexed and cached entries that would cover it for a short time, before the new version gets crawled?

app/robots.txt
hhoover: if we're talking about Google specifically, someone can control the crawl using Google's tools.

https://support.google.com/webmasters/answer/48620?hl=en

I believe Google crawls every few days. So there potentially could be an overlap on new versions but it wouldn't last long.

hhoover:
updated with requested changes

lena-larionova (approved):

PR #2893, opened 2021-05-27
Move link checker to a weekly action

lena-larionova:
### Summary
* Removing link checker from the build/test Github action that runs on every PR.
* Adding a weekly scheduled Github action to run the link check at the start of the week. This action can also be kicked off manually, in case of releases.
* Adding a local link checker option to the Makefile. To use it, first `make run` then `make check-links-local`.

### Reason
While the link checker is an absolutely necessary tool, it takes upwards of 15 minutes to run and keeps getting longer as the repo size grows.

### Testing
The build/test check in this PR should not include a link check, and should only be about 5-8 min long. The weekly action has to be tested after a merge.

Notice that checking links is no longer a step in the build/test process, and took only 6 minutes: https://github.com/Kong/docs.konghq.com/pull/2893/checks?check_run_id=2687650847


tlblessing (approved):
awesome

PR #2895, opened 2021-05-27
Add KIC 1.3.0

rainest:
### Summary
Documentation for KIC 1.3.0

### Reason
https://github.com/Kong/kubernetes-ingress-controller/releases/tag/1.3.0

### Testing
Version matrices need render check.


rainest:
There's either something I've forgotten or that's changed since I added 1.2. Apparently I broke all the links, not sure how :(

lena-larionova:
> There's either something I've forgotten or that's changed since I added 1.2. Apparently I broke all the links, not sure how :(

Looks like the index file for the version is missing in 1.3.x, not sure how that happened...

rainest:
Matrices looked fine but were a bit out of date with Kong releases. Now fixed.

app/kubernetes-ingress-controller/1.3.x/references/annotations.md
```diff
@@ -0,0 +1,528 @@
...
+```
+
+Will result in two routes:
+
+```
+{"hosts":["foo.example", "example.com", "example.net"], "paths":["/foo"]}
+{"hosts":["bar.example", "example.com", "example.net"], "paths":["/bar"]}
+
+Take care not to accidentally create unintentionally overlapping routes by not
+including the same path in multiple rules.
```
lena-larionova, line 415:
```suggestion

{:.important}
> To avoid creating overlapping routes, don't reuse the same path in multiple rules.
```
I hope I understood what you were going for correctly, the number of negatives was a bit confusing. Got rid of both "accidentally" and "unintentional", as they're adding complexity - intentionality is up to the user to decide, and we should just warn them about the possibility and how to avoid it.

lena-larionova (changes requested):
Some minor review comments on the new annotation.

lena-larionova (approved):

PR #2897, opened 2021-05-28
Add gemfile to dockerfile and adjust develop command

lena-larionova:
### Summary
Attempt to fix issue with broken docker build.

### Reason
Docker-compose development is broken. 

### Testing
Pull down branch and run `make develop`. Also, if build/test action passes fine, then there shouldn't be an issue.


lena-larionova:
Permissions error, this solution won't work.

PR #2909, opened 2021-06-01
[DOCU-47] Improving the plugin and OSS install layouts

lena-larionova:
### Review
@tlblessing 

### Summary
* Widened the plugin doc layout
* Adjusted the header to match the rest of the site better
* Moved plugin version selector into the side nav
* Mobile responsiveness stuff
* Slight adjustments to the plugin params table styling
* Removed unnecessary layout files that are no longer in use
* Adjusted OSS install docs (eg docs.konghq.com/install/centos) to work with new header changes

### Reason
Main reason is that the plugin docs need to be wider. They contain many reference tables, and those tables are always squished into a narrow column and can be hard to read.

Addresses part of https://konghq.atlassian.net/browse/DOCU-47.

### Testing
Take a look at any plugin under https://deploy-preview-2909--kongdocs.netlify.app/hub/
Test that different window sizes (mobile etc) don't break anything.

Without a version dropdown:
https://deploy-preview-2909--kongdocs.netlify.app/hub/optum/kong-service-virtualization/

With a version dropdown:
https://deploy-preview-2909--kongdocs.netlify.app/hub/kong-inc/key-auth/

OSS installation (check that header isn't broken):
https://deploy-preview-2909--kongdocs.netlify.app/install/centos/


HCloward (approved):
I like the larger text area and the updates to the header. I think the plugin version drop down should go, because of all of the confusion it causes. In the Jira, there is mention of moving items over to the left to more closely mimic the rest of the site. Was this decided against or is this a first step towards that end? Either way, this is a definite improvement. Thanks Lena!

lena-larionova:
> I think the plugin version drop down should go, because of all of the confusion it causes.

Sorry if the preview made it look like there were options - I was just showing previews of two possible setups, where the plugins have versions and where they don't. I think ultimately, it would be nice to get rid of the versioned plugin docs, but that's a big decision that I don't want to make in this PR.

>  In the Jira, there is mention of moving items over to the left to more closely mimic the rest of the site. Was this decided against or is this a first step towards that end? Either way, this is a definite improvement. Thanks Lena!

Nice catch. I was thinking that leaving it on the right currently makes more sense, as it's a page-level nav, and the rest of our docs have page-level navs on the right. I'd like to do a bigger overhaul later where we revisit this, though (for the plugins specifically).

PR #2915, opened 2021-06-03
[DOCU-1496]Updates vitalsSpec.yaml files for Kong versions 2.2, 2.3, and 2.4

HCloward:
### Reason
[DOCU-1496](https://konghq.atlassian.net/browse/DOCU-1496)

### Testing
Will include sample build link when available.


Bradamant3 (approved):
Approving from the docs side in case that's all you need. Why don't we generate this? 😿

PR #2921, opened 2021-06-04
Site width and responsiveness

lena-larionova:
### Summary
Adjusting site widths so that the page ToC collapses earlier and makes more room for the content. 

Setting an absolute width for the main container so that percentage widths have something to relate to and stop awkward jumpy transitions.

### Reason
Lots of issues with jumpy site window resizing, and awkward section widths at smaller sizes. 

Bad:
Screen Shot 2021-06-04 at 3 32 53 PM

### Testing
Old/bad behavior: open [any docs page](https://docs.konghq.com/enterprise/) in full screen, then slowly resize the width of the window. Notice the sudden jump from whitespace on either side to no whitespace. Also notice how awkwardly narrow the middle column becomes before the page ToC collapses.

New/better behavior: test the same page in the netlify preview: https://deploy-preview-2921--kongdocs.netlify.app/enterprise/

### DIFF
--- a/app/_assets/stylesheets/page.less
+++ b/app/_assets/stylesheets/page.less
@@ -8,7 +8,7 @@
         width: 99%;
       }
 
-      @media screen and (max-width: 1100px) {
+      @media screen and (max-width: 1300px) {
         padding: 8px;
       }
     }
@@ -233,7 +233,7 @@
           min-height: 100px;
         }
 
-        @media screen and (max-width: 1080px) {
+        @media screen and (max-width: 1300px) {
           .title-version-search-container {
             padding-left: 316px;
             padding-right: 32px;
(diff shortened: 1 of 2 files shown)


HCloward (approved):
Very nicely done. I like that the side icon for the right nav appears sooner and the transitions are definitely more seamless. I noticed there is a point where the top margin above the page-content-container-v2 is a bit larger than at any other point - somewhere between 1300 px width and 1080. Maybe has something to do with the right nav toggle?

lena-larionova:
> I noticed there is a point where the top margin above the page-content-container-v2 is a bit larger than at any other point - somewhere between 1300 px width and 1080. Maybe has something to do with the right nav toggle?

Nice catch. Fixed in https://github.com/Kong/docs.konghq.com/pull/2921/commits/dc4aa23f3ff6f1c37255126676472379b6a8a536.

PR #2923, opened 2021-06-07
Update AWS AMI links in installation docs

rossmcdonald:
### Review

@reviewer

### Summary
The links currently being used are pointing to a removed marketplace entry.

### Reason
The current links are broken.

### Testing
Validate that the links work and point to active marketplace entry.


lena-larionova (approved):
LGTM, thanks for the fix!

Issue #2919, opened 2021-06-04
Bad link: Configuration Reference


Yiyiyimu:
- `docs.konghq.com` url(s): https://docs.konghq.com/gateway-oss/2.4.x/getting-started/configuring-a-grpc-service/

### Summary

```
In Kong 1.3, gRPC support assumes gRPC over HTTP/2 framing. As such, make sure you have at least one HTTP/2 proxy listener (check out the [Configuration Reference][configuration-rerefence] for how to). In this guide, we will assume Kong is listening for HTTP/2 proxy requests on port 9080.
```

PR #2924, opened 2021-06-07
Fix typo causing broken link

lena-larionova:
### Summary
Fixing a typo.

### Reason
Issue #2919.


HCloward (approved):

lena-larionova:
Turns out it was not only a typo, but the source link was wrong entirely. Fixing that too.

PR #2926, opened 2021-06-07
Removing clearbit from docs

lena-larionova:
### Summary
Removing clearbit.

### Reason
Doing this through Google Tag Manager instead.


GayleNeumann (approved):
LGTM

PR #2930, opened 2021-06-09
Algolia search index improvements

lena-larionova:
### Summary
* Adding changelogs to index
* Adding an entry for Konnect Platform
* Excluding summary/next steps sections from polluting search results
* Delete old search selector in favor of better config

Q for reviewers: should we also add the contributing guide to the search? 

### Reason
Missing content from search indices and old search selector adding slightly to longer site build times.

### Testing
Check that changelogs and Konnect Platform appear in search (it should already work, I ran the indexer locally).
https://deploy-preview-2930--kongdocs.netlify.app/

### DIFF
--- a/algolia/config-full-docs.json
+++ b/algolia/config-full-docs.json
@@ -14,6 +14,10 @@
       "url": "https://docs.konghq.com/enterprise/2.4.x/"
     },
     {
+      "url": "https://docs.konghq.com/enterprise/changelog/",
+      "selectors_key": "changelog"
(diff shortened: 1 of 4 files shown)


HCloward (approved):
Konnect docs and changelog entries do now appear in the search. When I search for release notes though the changelog is not the first entry - we may want to reconsider our naming of the changelog - depending on if we think users will be searching for changelog or release notes. Otherwise, LGTM!

PR #2944, opened 2021-06-15
Fix table of contents for kic 1.3

lena-larionova:
### Summary
Fixing table of contents.

### Reason
Old ToC got pulled in when the KIC docs were versioned for 1.3.

Currently, that means the ToC is all one level and not expanded: https://docs.konghq.com/kubernetes-ingress-controller/

### Testing
Check that the nav for https://deploy-preview-2944--kongdocs.netlify.app/kubernetes-ingress-controller/ matches version 1.2.x.


Bradamant3 (approved):
Oh my TYSM. I evidently did _not_ check when 1.3 shipped :-(

PR #2945, opened 2021-06-16
Fixing issues with decK links and versions

lena-larionova:
### Summary
* Redirecting unversioned decK links to /latest/
* Setting algolia config to only index the latest deck version
* Stopping pre-1.7 decK docs from being indexed by search engines

### Reason
Links were all updated in the docs but unversioned links exist out in the world. Issue was reported on Slack: https://kongstrong.slack.com/archives/CDSTDSG9J/p1623879076092400

### Testing
Test any of the redirects in the file to make sure they work. For example: https://deploy-preview-2945--kongdocs.netlify.app/deck/installation/

### DIFF
--- a/algolia/config-full-docs.json
+++ b/algolia/config-full-docs.json
@@ -5,7 +5,7 @@
       "url": "https://docs.konghq.com/konnect/"
     },
     {
-      "url": "https://docs.konghq.com/deck/"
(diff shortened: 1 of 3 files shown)


falondarville (approved):
Verified that version number appears in URLs.

PR #2957, opened 2021-06-22
Update external-plugins.md

anton-mesnyankin:
full pathes are required here (https://github.com/Kong/kong/issues/7481)

### Review

@reviewer

### Summary

This fix runtime issue.

### Reason

https://github.com/Kong/kong/issues/7481

### Testing

https://github.com/Kong/kong/issues/7481

### DIFF
--- a/app/gateway-oss/2.4.x/external-plugins.md
+++ b/app/gateway-oss/2.4.x/external-plugins.md
@@ -40,12 +40,12 @@ an hypothetical Python plugin server called `pypluginserver.py`):
 pluginserver_names = go,python
 
 pluginserver_go_socket = /usr/local/kong/go_pluginserver.sock
-pluginserver_go_start_cmd = go-pluginserver -kong-prefix /usr/local/kong/ -plugins-directory /usr/local/kong/go-plugins
-pluginserver_go_query_cmd = go-pluginserver -dump-all-plugins -plugins-directory /usr/local/kong/go-plugins
+pluginserver_go_start_cmd = /usr/local/bin/go-pluginserver -kong-prefix /usr/local/kong/ -plugins-directory /usr/local/kong/go-plugins
+pluginserver_go_query_cmd = /usr/local/bin/go-pluginserver -dump-all-plugins -plugins-directory /usr/local/kong/go-plugins
 
 pluginserver_python_socket = /usr/local/kong/python_pluginserver.sock
-pluginserver_python_start_cmd = kong-python-pluginserver
-pluginserver_python_query_cmd = kong-python-pluginserver --dump-all-plugins
+pluginserver_python_start_cmd = /usr/local/bin/kong-python-pluginserver
+pluginserver_python_query_cmd = /usr/local/bin/kong-python-pluginserver --dump-all-plugins
 ```
 
 To enable those plugins, add the each plugin name to the `plugins` config. Assume we have those hello plugins
--- a/app/gateway-oss/2.4.x/external-plugins.md
+++ b/app/gateway-oss/2.4.x/external-plugins.md
@@ -55,6 +55,11 @@ in each language:
 plugins = bundled, go-hello, js-hello, py-hello
 ```
 
+{:.note}
+> **Note:** The `pluginserver_XXX_start_cmd` and `pluginserver_XXX_query_cmd` commands use
+ a limited default `PATH` variable. In most cases, you have to specify the full executable
+ path instead.
+
 ### Legacy configuration
 
 {{site.base_gateway}} versions 2.0.x to 2.2.x supported only Go external plugins and a single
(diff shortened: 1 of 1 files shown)


lena-larionova:
@fffonion can you take a look at this submission?

app/gateway-oss/2.4.x/external-plugins.md
```diff
@@ -40,12 +40,12 @@ an hypothetical Python plugin server called `pypluginserver.py`):
 pluginserver_names = go,python
 
 pluginserver_go_socket = /usr/local/kong/go_pluginserver.sock
-pluginserver_go_start_cmd = go-pluginserver -kong-prefix /usr/local/kong/ -plugins-directory /usr/local/kong/go-plugins
-pluginserver_go_query_cmd = go-pluginserver -dump-all-plugins -plugins-directory /usr/local/kong/go-plugins
+pluginserver_go_start_cmd = /usr/local/bin/go-pluginserver -kong-prefix /usr/local/kong/ -plugins-directory /usr/local/kong/go-plugins
```
fffonion, line 43: I think the python plugin server indeed installed in /usr/local/bin if it's not installed with `-u`.
But i'm not sure about go, i think normally it's in GOPATH/bin. @anton-mesnyankin could you
describe how do you install the go-pluginserver?

anton-mesnyankin:
@fffonion , I'm using docker (multistage image build). my docker file looks like:
---
....
RUN git clone https://github.com/Kong/go-pluginserver.git /go/go-pluginserver && \
    cd /go/go-pluginserver; \
    git checkout v0.6.0; \
    go build

...

FROM kong:2.4.1-alpine

COPY --from=builder /go/go-pluginserver/go-pluginserver /usr/local/bin/
---
so yes, go-pluginserver is available as /usr/local/bin/go-pluginserver , but I received errors 
"
[notice] 34#0: *14 [kong] process.lua:248 Starting go, context: ngx.timer
[notice] 34#0: signal 17 (SIGCHLD) received from 37
[notice] 34#0: *14 [kong] process.lua:258 external pluginserver 'go' terminated: exit 127, context: ngx.timer
" 
without proposed fix.

fffonion:
@anton-mesnyankin Thank you for confirming that for me. In that case we shouldn't assume go-pluginserver
is always available in `/usr/local/bin` in others' env. In fact, you can change your dockerfile to 
```
COPY --from=builder /go/go-pluginserver/go-pluginserver /usr/bin/
```
and then don't need to change start_cmd.

Since in our case openresty spawns a shell to execute those plugin server commands, it should has default
PATH env variable of `/usr/bin:/bin:/usr/sbin:/sbin` (need double check, but i'm sure it doesn't include
`/usr/local/bin`). So instead of documenting go-pluginserver the way in this PR, we should probably
instead say something like "by default Kong only looks up in those locations,
if you install the plugin servers elsewhere, you need to specify the full path".

anton-mesnyankin:
Hi @fffonion , 
I agree that /usr/bin/go-pluginserver will work well and that can be more universal solution than hardcoding it to /usr/local/bin.
In general: I think that the documentation can be impoved for that, because it has a lot of references to '/usr/**local**/bin/go-pluginserver' which is confusing and goes to error (because some people (who read the official documentation) will think that go-pluginserver binary should be placed at /usr/local/bin.....

fffonion:
@anton-mesnyankin That's a good point. I replied you at the issue from Kong side as well. I think it's a valid
fix needs to go into Kong. In the meanwhile let's avoid hard code the go-pluginserver path here, but instead
mentioning binaries in `/usr/local/bin` needs to be in full path.

app/gateway-oss/2.4.x/external-plugins.md
lena-larionova: Looks like Javier confirmed this with https://github.com/Kong/docs.konghq.com/pull/2979.

lena-larionova:
Just to put this info into full view, and not get it lost in a comment: looks like Javier confirmed this with #2979 as well.

@anton-mesnyankin can you also add the following note to this doc at line 58, to mirror his changes?

```
{:.note}
> **Note:** The `start` and `query` commands use a limited default PATH variable. In most cases, you have to specify the full executable path instead.
```

anton-mesnyankin:
@lena-larionova, done

app/gateway-oss/2.4.x/external-plugins.md
```diff
@@ -55,6 +55,8 @@ in each language:
 plugins = bundled, go-hello, js-hello, py-hello
 ```
 
+**Note:** the start and query commands use a very limited default PATH variable.  In most cases you'll have to specify the full executable path.
```
lena-larionova, line 58:
```suggestion
{:.note}
> **Note:** The `pluginserver_XXX_start_cmd` and `pluginserver_XXX_query_cmd` commands use
 a limited default `PATH` variable. In most cases, you have to specify the full executable
 path instead.
```
Some rephrasing that also applied to the linked PR.

anton-mesnyankin:
recommended changes are added

lena-larionova (approved):
LGTM!

PR #2965, opened 2021-06-24
Performance: Remove unused CSS

lena-larionova:
### Review
@falondarville 

### Summary
* Removing homepage.less
* Removing most of the contents in header.less

### Reason
* `homepage.less` is not in use, this styling was removed from docs in March 2021 with a site redesign. Homepages (eg docs.konghq.com/enterprise or docs.konhq.com/mesh) are no longer any different from other docs pages
* `header.less` was mostly replaced by `header-v2.less` in Feb 2021. The only thing remaining here that's relevant is the promo banner.

### Testing
Link TBA
Will need to check that nothing breaks in the header or on homepages, and that tests pass without issues.


falondarville (approved):
Looks good to me

PR #2968, opened 2021-06-25
[DOCU-1544] Doc feature: Editable placeholders

lena-larionova:
### Review
@HCloward @falondarville 

### Summary
* Adding support for editable placeholders in the documentation, inspired by [Google's implementation](https://developers.google.com/style/placeholders?hl=en).

* Removing line numbers from codeblocks: we've tested them out and found they weren't adding much value and were instead making codeblocks look busy.

* Enabling language guessing for syntax highlighting: instead of leaving codeblocks unformatted, if there is a codeblock without a language explicitly set on it, the syntax highlighter will attempt to guess the language. If it can't figure it out, it defaults to plaintext. 

#### Using editable codeblocks

Since you can't insert HTML into fenced codeblocks, you must set `pre` and `code` tags instead. To include a placeholder, use a `<div contenteditable="true">` tag. For example:

```
<pre><code>http://<div contenteditable="true">{HOSTNAME}</div>:8002 </code></pre>
```
Whatever content you enter will be converted to uppercase visually. 

Note that you have to include the pre and code tags inline with a line of code, otherwise they'll generate an extra empty line. For multiline codeblocks, do this:

```
<pre><code> curl -X PATCH http://<div contenteditable="true">{HOSTNAME}</div>:8001/workspaces/default \
      --data "config.portal=true" </code></pre>
```

Unfortunately, Rouge (our syntax highlighter) can only highlight fenced codeblocks, that is, codeblocks enclosed with three backticks. I've tested all the possible combinations of classes and tags to get stuff picked up by the highlighter, but I'm out of ideas. Reviewers, I welcome suggestions.

There is also a tiny bit of javascript that strips formatting from pasted content, and that registers the enter key as a submission/escape instead of a newline. @falondarville can you look at that and let me know if it needs any tweaking?

### Reason
Many confused users not noticing that they have to edit a variable in a codeblock, copying code as is, and running it with the placeholders.

### Testing
https://deploy-preview-2968--kongdocs.netlify.app/enterprise/2.4.x/deployment/installation/docker/

Reviewers: please test the functionality and look for anything that stands out as odd/awkward/needs work.
* Enter a value and press enter to save, then click the copy code icon to make sure it copies the new text
* Color choice
* Any other ideas for better UX
* Comment on using a mix of highlighted and non-highlighted codeblocks in a topic - does it look strange? Should we stick with all non-highlighted or all highlighted?

### DIFF
--- a/app/_assets/javascripts/app.js
+++ b/app/_assets/javascripts/app.js
@@ -663,7 +663,7 @@ $(function () {
       copyInput.text(
         snippet.data("copy-code") ||
           snippet
-            .find(".rouge-code")
+            .find("code")
             .text()
             .replace(/^\s*\$\s*/gi, "")
       );
@@ -877,4 +877,16 @@ setInterval(function () {
 $(".closebanner").click(function () {
   $(".navbar-v2").addClass("closed");
   localStorage.setItem("closebanner-konnect", "closebanner");
+});
+
+
+$('[contenteditable]').on('paste', function(e) {
+    //strips elements added to the editable tag when pasting
+    var $self = $(this);
+    setTimeout(function() {$self.html($self.text());}, 0);
+}).on('keypress', function(e) {
+    if (event.keyCode === 13) {
+     // unfocus when hitting enter key
+     $('[contenteditable]').blur();
+   }
 });
--- a/app/_assets/javascripts/app.js
+++ b/app/_assets/javascripts/app.js
@@ -886,7 +886,7 @@ $('[contenteditable]').on('paste', function(e) {
     setTimeout(function() {$self.html($self.text());}, 0);
 }).on('keypress', function(e) {
     if (event.keyCode === 13) {
-     // unfocus when hitting enter key
-     $('[contenteditable]').blur();
-   }
+      // unfocus when hitting enter key
+      $('[contenteditable]').blur();
+    }
 });
--- a/app/_assets/stylesheets/base.less
+++ b/app/_assets/stylesheets/base.less
@@ -159,7 +159,6 @@ code {
     line-height: 16px;
     border-radius: 2px;
     -webkit-user-modify: read-write-plaintext-only;
-    text-transform: uppercase;
 
     &::after {
       content: "\f040";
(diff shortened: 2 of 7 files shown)


falondarville (approved):
1. Enter a value and press enter to save, then click the copy code icon to make sure it copies the new text: Done, looks great. 
2. Color choice: ADA OK for user input: https://color.a11y.com/ContrastPair/?bgcolor=ffffff&fgcolor=19a654
3. Any other ideas for better UX: future refactor to rm text upon click https://stackoverflow.com/a/20300212/8986774
4.Comment on using a mix of highlighted and non-highlighted codeblocks in a topic - does it look strange? Should we stick with all non-highlighted or all highlighted? - non-highlighted more than OK for bash commands

lena-larionova:
> Any other ideas for better UX: future refactor to rm text upon click https://stackoverflow.com/a/20300212/8986774

👀 let me test

lena-larionova:
> > Any other ideas for better UX: future refactor to rm text upon click https://stackoverflow.com/a/20300212/8986774
> 
> 👀 let me test

Results of test:

* Using the method suggested in that conversation, tried the pure CSS route to clean out the field. It worked, but the text field shrinks to the size of the icon on click, which looks a bit odd. I can set a min-width on the field, but for some reason, the icon wants to float at the edge of the text and not the field, so it aligns left. Need some more work here.
* It might be better to select all text on click. Then it's very clear that the content is editable, and the box size doesn't change on click. Clearing on click makes the text box jump around in width, and this method would solve that. However, I need to do more testing/exploring to figure out how to make that work; the "simple" solution of setting `user-select` all hasn't worked for me yet, requires some more trial and error.

lena-larionova:
> I do think we should be consistent with color highlighting - otherwise I think it would be confusing - why is this highlighted and this other block not

This one I can do for this iteration, at least on this one topic. So going forward, just have to make sure that we set fenced codeblocks to `plaintext` if there's editable placeholders being used on the page.

I'll write all this up in the contributing guidelines.

PR #2974, opened 2021-06-28
Add PDF support

mheap:
### Review
@lena-larionova 

### Summary
Add the ability to generate PDFs of any documentation set. This will add a `workflow_dispatch` job that can be triggered from `main` once merged

image

On job completion an artefact will be uploaded containing the generated PDFs and can be downloaded as required

image

### Reason

DOCU-10 - request for PDF versions of our documentation. There is no table of contents and are no internal links as it significantly increases the complexity of the work.

### Testing

First, build the docs by running `./node_modules/.bin/gulp build` in the root directory.

Once that completes, we can generate a PDF from the files by running the following:

```
cd pdf-generation
npm ci
KONG_DOC_VERSIONS="ce_2.4.x" node run.js
```

This will produce a file named `gateway-oss-2.4.x.pdf`


mheap:
This PR now adds a separate job for generating plugin documentation.

I changed tactic a little which has made the process simpler. Instead of using a separate Jekyll config file (`jekyll_pdf.yml`) I now generate the full docs pages with header before extracting everything in `.page-content` and setting the whole page to be just that content (https://github.com/Kong/docs.konghq.com/pull/2974/files#diff-7093148ef2941db81d51d859207c87387fdc248262b75ee51cbb220fc6fdee57R74-R78)

This has the effect of stripping the header, footer + sidebars automatically, along with any cookie banners etc.

I also worked out how to move the tabs inline so that each option (Admin API, k8s etc) gets it's own header (https://github.com/Kong/docs.konghq.com/pull/2974/files#diff-7093148ef2941db81d51d859207c87387fdc248262b75ee51cbb220fc6fdee57R80-R91)

image

lena-larionova (approved):
LGTM! Thanks for putting all of this work in - we're going to have some very happy people that have been asking for PDF docs for some time :) 

For tracking purposes: reviewed these changes directly with Michael.

PR #2977, opened 2021-06-28
explain regex dangers and limits applied

javierguerragiraldez:


Bradamant3 (approved):
LGTM, although I'm adding it to a list of pages we might want to consider trimming. Feels overexplained in context, but isn't wrong. Would also be good to say something about _why_ it matters. Also -- do we provide a useful error for cases where matching fails?

kikito (approved):

PR #2978, opened 2021-06-28
Konnect update 06/24

lena-larionova:
### Summary
Adding update on global plugins.

### Reason
Deploy went out last week. Bigger docs update in progress, but release notes should at least be updated to say that this is available. 

### Testing
Link TBA


GayleNeumann (approved):
LGTM

PR #2980, opened 2021-06-28
doc(kic) add section on using tcdpump sidecars

rainest:
### Summary
Adds a KIC troubleshooting section covering basic usage of a tcpdump sidecar.

### Reason
This is a useful debugging tool and instructions for it should live somewhere more accessible than the internal Slack conversation I keep linking people to.

### DIFF
--- a/app/kubernetes-ingress-controller/1.3.x/troubleshooting.md
+++ b/app/kubernetes-ingress-controller/1.3.x/troubleshooting.md
@@ -219,3 +219,56 @@ approaches to isolate issues:
   and responses (passing `--verbose 2` to decK will show all requests) and
    add debug Kong Lua code when controller requests result in an
   unhandled error (500 response).
+
+## Inspecting network traffic with a tcpdump sidecar
+
+Inspecting network traffic allows you to review traffic between the ingress
+controller and Kong admin API and/or between the Kong proxy and upstream
+applications. You can use this in situations where logged information does not
+provide you sufficient data on the contents of requests and you wish to see
+exactly what was sent over the network.
+
+Although you cannot install and use tcpdump within the controller
+or Kong containers, you can add a tcpdump sidecar to your Pod's containers. The
+sidecar will be able to sniff traffic from other containers in the Pod. You can
+edit your Deployment (to add the sidecar to all managed Pods) or a single Pod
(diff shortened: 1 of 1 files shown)


app/kubernetes-ingress-controller/1.3.x/troubleshooting.md
```diff
@@ -219,3 +219,56 @@ approaches to isolate issues:
...
+controller and Kong admin API and/or between the Kong proxy and upstream
+applications. You can use this in situations where logged information does not
+provide you sufficient data on the contents of requests and you wish to see
+exactly what was sent over the network.
+
+Although you cannot install and use tcpdump within the controller
+or Kong containers, you can add a tcpdump sidecar to your Pod's containers. The
+sidecar will be able to sniff traffic from other containers in the Pod. You can
+edit your Deployment (to add the sidecar to all managed Pods) or a single Pod
+and add the following under the `containers` section of the Pod spec:
```
Bradamant3, line 235: A few thoughts here:

- Should we add an example for the Deployment case?
- Wonder whether the Pod example would be more helpful if we provided a (truncated) complete Pod spec, with a bit more context (the `containers` section at least, and preferably starting with the Pod itself so as to make indentation very precise bc copy/paste can mess it up pretty badly)

Bradamant3 (commented):
A couple of questions about the example. The whole page could use a rework just to make it easier to find the right section, but that's a docs team task. Will create Jira to track.

app/kubernetes-ingress-controller/1.3.x/troubleshooting.md
rainest: The Pod/Deployment examples are basically equivalent. The Deployment contains a Pod template, and the change you'd make to it is the same change you'd make when editing a Pod directly.

Context shouldn't be necessary as the example block is a complete entry under `containers`. While you can mess up indentation, the correct indentation should be apparent when editing it (there will be other container blocks there already) and the impact of doing it incorrectly is minimal (if it is wrong, Kubernetes will simply reject the edit and continue running the existing version of the Pod(s)).

Bradamant3 (approved):
Approving as is to get the information into the docs. See also https://konghq.atlassian.net/browse/DOCU-1586

PR #2993, opened 2021-06-30
[DOCU-1587] Update Jekyll to 4.2

lena-larionova:
### Review
@falondarville @HCloward @Bradamant3 

### Summary
Upgrading jekyll from 3.9.0 to 4.2.0.

Removing `markdown.rb`: this plugin defines rules for redcarpet, a Rouge syntax highlighter and codeblock parser. Jekyll dropped support for it in 4.0.0 in favor of native Rouge syntax highlighting. See [changelog](https://jekyllrb.com/docs/history/#v4-0-0)

Jekyll now generates a `.jekyll-cache` at build time. Like the `dist` directory, it should never be committed to github, so I've added it to the gitignore file (@Bradamant3 , might be worth noting that for the Kuma website)

Adjusting our alias generating plugin, as it was making jekyll think that there are duplicate files pointing to the same URL. 
There are two warnings remaining (instead of hundreds):

```
Conflict: The following destination is shared by multiple files.
                    The written file may end up with unexpected contents.
                   /docs.konghq.com/dist/enterprise/1.5.x/introduction/index.html
                     - enterprise/1.5.x/introduction/index.md
                     - enterprise/1.5.x/introduction.md
                    
Conflict: The following destination is shared by multiple files.
                    The written file may end up with unexpected contents.
                  /docs.konghq.com/dist/enterprise/2.1.x/introduction/index.html
                     - enterprise/2.1.x/introduction/index.md
                     - enterprise/2.1.x/introduction.md
```

If anyone can figure out why they're happening,  that would be helpful. Otherwise, it's not actually a problem - this has been an issue for a very long time, we just never ran `jekyll doctor` before, and now it's baked into the build.

### Reason
To take advantage of build performance enhancements in 4.x versions.

### Testing
Team, I'm going to need your help poking around the site to look for anything wrong, broken, or somehow odd.
https://deploy-preview-2993--kongdocs.netlify.app

* Codeblock syntax highlighting should still work without redcarpet
* Test `latest` URLs
* Test versioned URLs
* Test that index pages load fine

Might be worth glancing over the Jekyll changelog for 4.0.0-4.2.0 to see if there's anything I might've missed that needs adjusting: https://jekyllrb.com/docs/history/


falondarville (approved):
Clicked through site and tested. LGTM

PR #2997, opened 2021-07-01
Remove old tabs plugin

lena-larionova:
### Summary
Removing tabs.rb, a tab plugin that doesn't work. We currently use navtabs.rb instead.

Replacing all instances of {% tabs %} in the docs with {% navtabs %}. There was a small enough number that this is easy enough to do, and doesn't leave random unused liquid strings all over the place.

### Reason

We have both tabs.rb and navtabs.rb as plugins that generate doc tabs. tabs.rb doesn’t work, and is only used on a handful of pages.

See entry here: https://konghq.atlassian.net/wiki/spaces/KD/pages/2463629315/Site+Performance+Improvements

### Testing
Any plugin page with tabs in the Usage section. For example:
https://deploy-preview-2997--kongdocs.netlify.app/hub/kong-inc/basic-auth/2.1-x.html#create-a-consumer

Mainly, need to make sure that the build runs smoothly and nothing breaks.


falondarville (approved):
LGTM

PR #2999, opened 2021-07-02
docs(plugins/syslog) docs for facility option

locao:


app/_hub/kong-inc/syslog/index.md
```diff
@@ -90,6 +91,15 @@ params:
         A list of key-value pairs, where the key is the name of a log field and
         the value is a chunk of Lua code, whose return value sets or replaces
         the log field value.
+    - name: facility
+      required: false
+      default: "`user`"
+      datatype: string
+      description: |
+        The facility is used by the operating system to decide how to handle each log message. This
```
HCloward, line 99: Should we include a description of what each of these means? Also, what does "how to handle each log message" mean? What exactly is done here? @dndx suggested this is a way for the Syslog plugin to group messages from different sources - if so I would mention that here.

HCloward (commented):
I am not sure the description is very clear. What is the value of using this parameter?

app/_hub/kong-inc/syslog/index.md
dndx: Yes that's what it does. See: https://en.wikipedia.org/wiki/Syslog#Facility

HCloward: ok, that's fine with me

HCloward (approved):

PR #3003, opened 2021-07-02
Reduce build time using liquid-c and jekyll-include-cache

mheap:
### Review

@lena-larionova 

### Summary
Add a liquid parser written in C to reduce the build time. This is a quick win to speed up the build.

It also adds the `jekyll-include-cache` gem which speeds up `include` usage. We use `include` heavily when generating navigation.

Anecdotal evidence shows a ~70% improvement on my local machine and a ~60% increase on netlify.

## Locally

Without Cache, without liquid-c

* 103.905 seconds
* 72.69 seconds
* 70.078 seconds
* 70.817 seconds

Without Cache, with liquid-c

* 89.737 seconds
* 73.613 seconds
* 69.113 seconds
* 69.413 seconds

With Cache, with liquid-c

* 54.5 seconds
* 26.62 seconds
* 27.44 seconds
* 27.026 seconds

## On Netlify

Before change:

* 142.009 seconds
* 153.586 seconds
* 138.788 seconds

After change (liquid-c):

* 115.321 seconds

After change (liquid-c and include_cached)

* 59.0772

### Reason
Less time waiting 🥳

### Testing
Spot check various pages. Please pay attention to things that change per page such as navigation, breadcrumbs, table of contents etc


Gemfile
```diff
@@ -3,3 +3,4 @@ gem 'jekyll', '4.2.0'
 gem 'kramdown-parser-gfm'
 gem 'jekyll-redirect-from'
 gem 'rouge', '3.26.0'
+gem 'liquid-c'
```
lena-larionova, line 6:
```suggestion
gem 'liquid-c'

```
Adding a newline to appease the linter.

app/_layouts/docs-v2.html
```diff
@@ -4,38 +4,42 @@
...
+{% assign latest_page_url = page.url | replace: page.kong_version,
+page.kong_latest.release %}
+{% capture latest_page_exists %}{% page_exists {{latest_page_url}} %}
+{% endcapture %}
 
 <div class="page v2 {{ page.class }}">
 
   <div class="container">
-    {% include docs-sidebar.html %}
+    {% include_cached docs-sidebar.html %}
```
lena-larionova, line 15: This file will need adjusting before we can cache it. Right now, the first version of the sidebar that loads up is the same one that will persist between versions. 

For example: if you open Enterprise 2.4.x docs, then switch to 2.2.x and try to click a menu item, it opens the 2.4.x version of that link.

This is happening because of page-context assign variables:
https://github.com/benbalter/jekyll-include-cache#one-potential-gotcha

mheap: This was resolved by passing in `kong_version` (even if we don't use it) https://github.com/mheap/docs.konghq.com/commit/3083ed65250a40e954683a155a528edceafc1530#diff-b76293185985e73d1cdbcb33314dad3913b088a48e4fdfa91c6608f0bbe31756R15

lena-larionova:
Just found two more places where caching is still not working quite right:

* Active nav state in versioned guides. The active state is saved for the first topic you click on and then persists no matter where you switch to after.
* Breadcrumbs will probably also need the version passed to the include, as they're currently also caching incorrectly.

mheap:
Both of those issues should now be resolved.

* Active navigation is now added using javascript rather than the `active` class in Jekyll. This is a reasonable solution as we already use JS to expand the accordion for active navigation so there are no new dependencies. If this solution is accepted we can simplify the templates to remove the active logic (which I have not done yet)
* Breadcrumbs are now passed `page.edition` and `page.dir` which is enough to make each breadcrumb rendering unique. They still use the `page.*` variables internally and there is an opportunity to convert those to `include.*` in the future should this PR be accepted

lena-larionova (approved):
Tested all the cached content, looks like it all works now! 

Thanks, once again, for putting in all this work. LGTM!

Doc team: please do not merge until Monday (July 12) due to potential breaking changes. We've tested as much as could think to test, but in case we missed something, let's not tempt fate by merging this on a Friday.

PR #3006, opened 2021-07-02
Add jekyll-cache to makefile clean

lena-larionova:
### Summary
Adding `rm -rf app/.jekyll-cache` to `clean`. 
Also adding comments to explain what each command does so that anyone can understand what's going on, and which one they might want to run.

### Reason
Jekyll-cache should be cleaned out so that tests are more accurate, and will help with troubleshooting.
For the comments, taking a cue from Michael's [recent PR](https://github.com/Kong/docs.konghq.com/pull/3002). We should try to add code comments as much as possible, so that anyone on the team can understand what's going on.

### Testing
To test, pull down the branch and run `make clean && make install`. It should remove dist and jekyll-cache, which will get recreated next time you `make run`.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -19,6 +19,7 @@ clean:
 	-docker-compose stop
 	-docker-compose rm -f
 	-rm -rf dist
+	-rm -rf app/.jekyll-cache
 	-rm yarn.lock
 	-rm -rf node_modules
 	-rm install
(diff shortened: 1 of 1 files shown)


HCloward (approved):
dist and app/.jekyll-cache files removed with make clean && make install and reappeared with make run

PR #3015, opened 2021-07-06
KIC troubleshooting: Turn code phrases into codeblocks

lena-larionova:
### Review
@Bradamant3 

### Summary
Turning code phrases into code blocks and some minor teeny edits.

### Reason
Long code phrases were breaking mobile site.

I wonder if we should add some guidelines around when to use a code phrase vs when to use a code block @HCloward @Bradamant3 

### Testing
https://deploy-preview-3015--kongdocs.netlify.app/kubernetes-ingress-controller/1.3.x/troubleshooting/

### DIFF
--- a/app/kubernetes-ingress-controller/1.3.x/troubleshooting.md
+++ b/app/kubernetes-ingress-controller/1.3.x/troubleshooting.md
@@ -154,9 +154,18 @@ $ kubectl exec test-701078429-s5kca -- curl --cacert /var/run/secrets/kubernetes
 If it is not working, there are two possible reasons:
 
 1. The contents of the tokens are invalid.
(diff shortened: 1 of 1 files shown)


Bradamant3 (approved):
LGTM (although as a side note inquiring minds would like to know, just what makes a k8s installation "non-standard" ... but that's for another day, another PR :D )

lena-larionova:
> LGTM (although as a side note inquiring minds would like to know, just what makes a k8s installation "non-standard" ... but that's for another day, another PR :D )

I'd love to know that too -- I stared at that phrase while making this change thinking, should I try to track it down now? But no, too out of scope, keeping it simple.

PR #3020, opened 2021-07-08
Contributing: Add caution about using editable placeholders

lena-larionova:
### Summary
Adding caution against using editable placeholders for private or sensitive information. 

### Reason
Ran into this issue and realized we need guidelines around it.

### Testing
TBA

### DIFF
--- a/app/contributing/markdown-rules.md
+++ b/app/contributing/markdown-rules.md
@@ -85,7 +85,7 @@ Gateway and Kong Mesh, use the page version variable.
 * **Use `latest` in unversioned docs:** If you're linking to a versioned topic
 from an unversioned topic, use `/latest/` instead of a version name or variable.
 
-### Adding new entries to the ToC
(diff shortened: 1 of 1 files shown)


app/contributing/markdown-rules.md
```diff
@@ -134,6 +134,16 @@ the code, you can use editable placeholders.
 Editable placeholders are only supported in HTML, and are not supported with
 fenced codeblocks.
```
Bradamant3, line 135: While I'm at it ... (realize this isn't part of changed content, but ... I tripped over this wording. Several times, in fact ...)

```suggestion
Editable placeholders require HTML markup. They are not supported in pure markdown (fenced) codeblocks.
```

Bradamant3 (approved):
I _think_ I can approve and you can still commit suggestion if you like? (Otherwise, don't worry about it ...)

PR #3026, opened 2021-07-12
[DOCU-1604] add links section

falondarville:
### Review

@Kong/team-docs 

### Summary

Add to Style Guide:
* Quick fix in Content best practices table
* Add Links section
* Add content to Images section

### Reason

Add info that will make our site more accessible. 

### Testing

Tested locally. 

https://deploy-preview-3026--kongdocs.netlify.app/contributing/style-guide/


Bradamant3 (approved):
Nice! You caught at least one other thing I was gonna push a fix for ("use" and "utilize") ❤️ . Also, in this case I think the "don't" admonition is the best way to document the links issue, because it's still so often the default.

PR #3027, opened 2021-07-12
Include cache fixes

mheap:
### Review

@falondarville 

### Summary
After merging the build improvement PR @falondarville noticed that the blue bar wasn't showing on the homepage. This is due to the sidebar not existing and has been fixed in this PR.

Additionally, the page title generation was being cached and the Algolia search was missing code on the landing page. Both of these are resolved in this PR.

### Reason
Fixes for production and PR #3003 

### Testing
* Blue bar should show up on the landing page
* Page titles should be correct


falondarville (approved):
LGTM

PR #3030, opened 2021-07-13
Fix include_cache issue on plugin pages

mheap:
### Summary

All plugins are showing `google-auth` as the name due to the recent update to use `include_cached`. This PR passes in the parameters directly to fix this issue

Screenshot 2021-07-13 at 11 09 43

### Testing
Visit `/hub/kong-inc/basic-auth/` - you should see 

```
curl -X POST http://{HOST}:8001/services/{SERVICE}/plugins \
    --data "name=basic-auth"  \
```

rather than

```
curl -X POST http://{HOST}:8001/plugins/ \
    --data "name=google-logging" 
```

### DIFF
--- a/app/_layouts/extension.html
+++ b/app/_layouts/extension.html
@@ -233,7 +233,7 @@ breadcrumbs:
         {% endif %}
 
         {% if page.params.examples != false %}
-          {% include_cached hub-examples.html %}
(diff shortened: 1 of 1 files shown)


Bradamant3 (approved):
LGTM

PR #3044, opened 2021-07-22
[DOCU-1620] enable tabbing through Docs menu item

falondarville:
### Review

@Kong/team-docs 

### Summary

* Docs submenu now opens upon "enter"
* Tab through opened menu
* Close menu upon tabbing to Support menu item

### Reason

Improve accessibility for users using keyboards w/o mouse. 

### Testing

Tab through to the main navigation.

### DIFF
--- a/app/_assets/javascripts/navbar.js
+++ b/app/_assets/javascripts/navbar.js
@@ -39,15 +39,13 @@ jQuery(function () {
       $(".with-submenu").toggleClass("submenu-opened");
       $(".navbar-item-docs").setAttribute("aria-hidden", "false");
       return false;
-    } 
-
+    }
     // if user doesn't open Docs submenu, move focus to Support menu item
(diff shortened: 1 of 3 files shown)


lena-larionova:
> Docs submenu now opens upon "enter"

This is great - I tested it, works wonderfully. Also tested tabbing through the whole menu to the next top item, the submenu closes automatically.

My only worry is this: if you just tab and never use the Enter key, it actually still continues to tab through the menu as before. Is there a way to auto-open the menu when tabbing to it instead of needing to press Enter?

falondarville:
It's standard to open the menu upon Enter. See: https://a11y-guidelines.orange.com/en/web/components-examples/dropdown-menu/. However, I realized that if I don't want to enter the submenu, I should go to Support instead of still through the Docs menu

lena-larionova (approved):

PR #3048, opened 2021-07-23
fix(hybrid) correct descrption for PKI mode to validate from configured

fffonion:
CA

### Review

@reviewer


falondarville (approved):
LGTM

PR #3050, opened 2021-07-24
[DOCU-1624] Remove search bar subtitle and increase margin

lena-larionova:
### Summary
Removing subtitle text that was on the landing page initially for old users of the old doc site, which changed in Feb 2021.
Also removing related CSS class.

### Reason
https://konghq.atlassian.net/browse/DOCU-1624

### Testing
TBA


falondarville (approved):
LGTM!

PR #3060, opened 2021-07-28
Remove email form from install pages

lena-larionova:
### Summary
Removing the email form from OSS installation pages (such as https://docs.konghq.com/install/centos/, see form at the bottom of the page with a giant "Notify me" button), and all related includes and classes.

Fixed a broken link to community.

### Reason
Multiple reasons:
1. The form was broken.
2. The same form already exists in the site footer, and it works there.
3. The text for the form was inaccurate: it was telling users that they'll get release notes, but we don't actually have any control of what the users receive. It's a marketing newsletter.
4. It's a marketing form and it really shouldn't be in the body of doc content.

### Testing
Look through the OSS install pages, eg https://deploy-preview-3060--kongdocs.netlify.app/install/centos/ - there should be no email form at the bottom.

Test the form in the footer to make sure that I didn't accidentally remove something related to it as well.


falondarville (approved):
LGTM

PR #3061, opened 2021-07-28
CSS cleanup, part 2

lena-larionova:
### Summary
Continuing CSS cleanup. 
* Removing classes and stylesheets related to deleted javascript from app.js in https://github.com/Kong/docs.konghq.com/pull/3034.
* Removing the following stylesheets:
  * `enterprise.less`: defines enterprise sections. We no longer have these, and instead we use an enterprise directory.
  * `plugins.less`: we use `extension.less`, not these styles.
  * `about.less`, `community.less`: referring to pages that don't exist on the doc site. About and community pages redirect to the marketing site.
  * `spinner.less`: a loading icon, maybe? We definitely don't use this.

### Reason
Part of the larger doc site refactoring project. Removing unused files and classes because they slow down page load times.

### Testing
Tested locally to make sure nothing breaks, but reviewers will need to spot-check.

Link TBA


falondarville (approved):
LGTM

PR #3063, opened 2021-07-28
add underline on hover to main nav and footer

falondarville:
### Review

@lena-larionova 

### Summary

* when fixing underline on hover for the submenu non-clickable headers, the styling stripped from the remaining nav items. I added underlining to non-Docs items
* added underline on hover to footer items
* remove bold when active for Docs menu item

### Reason

Increase accessibility by adding more visible indicator that user is hovering on a link. 

### Testing

Hover over main nav and footer link items.

### DIFF
--- a/app/_assets/stylesheets/base.less
+++ b/app/_assets/stylesheets/base.less
@@ -209,6 +209,12 @@ a {
   }
 }
 
+a.navbar-item {
+  &:hover {
+    text-decoration: underline;
+  }
+}
+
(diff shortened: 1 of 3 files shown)


lena-larionova (approved):
LGTM! I noticed this earlier today and wanted to call it out, but you're already on it.

With the footer, I think it's fine that we diverge from KongHQ in this way, but just be careful making too many changes that'll have it looking different from the main site. Though technically, we're already doing that with the top nav anyway...

PR #3065, opened 2021-07-28
[DOCU-1635] make features cards stack on mobile

falondarville:
### Review

@Kong/team-docs 

### Summary

Previously, the "Key features" and "Ecosystem integrations" cards on the landing pages (on mobile) can only be accessed if the user scrolls right. This makes it difficult for users to understand how to use the page to see the content "hidden". 

The "Key features" and "Ecosystem integrations" cards now stack on smaller screen so that you don't have to scroll right. 

### Reason

* users can see all content without having to guess how to use the scrolling feature
* provides more accessible elements
* makes more sense visually

### Testing

Check out the landing page "Key features" and "Ecosystem integrations" cards on smaller screens and mobile. They will stack now.

### DIFF
--- a/app/_assets/stylesheets/pages/landing-page.less
+++ b/app/_assets/stylesheets/pages/landing-page.less
@@ -360,7 +360,6 @@ header.navbar {
 
   section.docs {
     display: flex;
-    // margin-top: 40px;
 
     section.title-cards {
       display: flex;
@@ -644,6 +643,10 @@ header.navbar {
 
           .card {
             margin-right: 0;
+
+            &:first-child {
+              margin-bottom:24px;
+            }
           }
         }
       }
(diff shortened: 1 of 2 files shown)


lena-larionova (commented):
Nice job getting them to stack.

Can you also adjust the widths so that they have the same margins as the tiles above them? 

Desired state, should match this:
Screen Shot 2021-07-28 at 2 51 41 PM

Current state:
Screen Shot 2021-07-28 at 2 52 26 PM

(don't try to make them match the Kong Mesh and Kong Gateway tiles, those margins are slightly off and also need fixing)

lena-larionova (approved):
LGTM, just need to adjust padding on "Key features" title on mobile so that it matches the sections above it.

PR #3067, opened 2021-07-29
fix z-index

falondarville:
### Review

@lena-larionova 

### Summary

Increase z-index on navbar. 

### Reason

Created menu overlap. 

### Testing

Go to /hub/kong-inc/application-registration/, open the right-side bar menu and scroll up.


lena-larionova (approved):
Looks like this fixes it, LGTM!

PR #3074, opened 2021-07-30
[DOCU-1645] Remove unused top-level directories

lena-larionova:
### Summary
Removing top-level directories that only contain redirect files. We should not be keeping hundreds of empty files around forever just for redirects; redirects should be managed through the `/app/_redirects/` file.

Removing the following directories:
* `app/about`
  * Fixing links to point to `/hub/`, `konghq.com/about-kong-inc/`, or `konghq.com/faqs/` where appropriate
  * Deleting links that are pointing to `/about/ `content that doesn't exist, even on the redirected page. For example, the links from gateway introduction pages linking to "purpose and philosophy", then saying "now you know about API gateways" are inaccurate. The `about` page on konghq.com does not talk about what API gateways are in anything more than extremely broad, one-sentence terms, and linking users off to a different website to show them something not-very-useful is detrimental to the docs user experience.
  * Adding redirects for to the `konghq.com/about-kong-inc/` page
* `app/careers`: also updating the `Makefile` to stop excluding careers in the link checker, as there's nothing to exclude

The following directories did not have any links leading to them, so I was able to simply remove them without any other cleanup:
* `app/download`
* `app/license`
* `app/privacy`
* `app/terms`

### Reason
Part of ongoing doc site cleanup project.

### Testing
TBA


falondarville (approved):
LGTM

PR #3075, opened 2021-07-30
[DOCU-1630] Remove echint

lena-larionova:
### Review
@falondarville 

### Summary
Removing echint.

### Reason
Echint is deprecated and not maintained: https://www.npmjs.com/package/echint
This code linter flags very odd warnings as errors (mostly about having an extra newline at the end of a file, or an indent in the wrong place) and causes builds to fail.

### Testing
Run the build locally and make sure it doesn't break.
Wait for checks to pass on this PR.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -13,7 +13,6 @@
         "browser-sync": "2.27.4",
         "cloudflare": "2.7.0",
         "del": "6.0.0",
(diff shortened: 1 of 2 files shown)


falondarville (approved):
LGTM. Known test response message re: es-abstract

PR #3080, opened 2021-08-03
Prevent `page.*` variables being used in includes

mheap:
### Review
@Kong/team-docs 

I'd recommend looking at the commits one by one by clicking on the hash rather than looking at all changed files. Each commit contains changes to the include, and all callers of that include

### Summary
Fail the build if any `page.*` variables are used in an included file. This will prevent any issues when using `include_cached` when a page level variable is used directly without being passed into the `include_cached` call and is incorrectly cached.

### Reason

This is a **proposed** change. I've refactored a few files to show the required changes for each include, and attached sample output from the detection script.

---

The `include_cached` gem uses any variables passed in to the include as a cache key, enabling us to cache pages based on the values used within the page e.g.:

```ruby
{% include_cached hub-examples.html params=page.params publisher=page.publisher name=page.name %}
```

Without these parameters being passed in, all pages that use `hub-examples.html` will show an identical page, no matter what the value of `page.params` is, due to `include_cached` serving the file from it's cache.

It is possible to pass in these parameters to `include_cached` and then use the `page.*` variables (e.g. `page.params`) to force `include_cached` to generate the page without switching to use `include.params`. This is how we solved the caching issue immediately once it was discovered.

However, this is prone to error as we may use a `page.*` variable in an include without passing it in to the call to `include_cached`, and will receive a cached page unintentionally. This will be tough to spot in a lot of cases.

This PR adds an `include-check.sh` script that can be run locally and within CI to ensure that all values used in includes are passed in as a parameter to `include_cached` rather than depending on `page.*`.

Running `include-check.sh` outputs the following information to help debug:

```
Page variables must not be used in includes.
Pass them in via include_cached instead

Files that currently use page.*:

File: app/_includes/header.html
via:
app/install/index.html
app/_layouts/install.html
app/_layouts/page.html
app/_layouts/extension.html
==========
File: app/_includes/docs-sidebar-item.html
via:
app/_includes/docs-sidebar-item.html
app/_includes/docs-sidebar.html
==========
File: app/_includes/md/enterprise/deploy-license.md
via:
app/enterprise/2.4.x/vitals/vitals-influx-strategy.md
app/enterprise/2.4.x/deployment/licenses/deploy-license.md
==========
File: app/_includes/md/enterprise/download/statsd.md
via:
app/enterprise/0.36-x/admin-api/vitals/vitals-prometheus-strategy.md
app/enterprise/1.5.x/admin-api/vitals/vitals-prometheus-strategy.md
app/enterprise/2.1.x/vitals/vitals-prometheus-strategy.md
app/enterprise/2.2.x/vitals/vitals-prometheus-strategy.md
app/enterprise/0.35-x/admin-api/vitals/vitals-prometheus-strategy.md
app/enterprise/2.3.x/vitals/vitals-prometheus-strategy.md
app/enterprise/1.3-x/admin-api/vitals/vitals-prometheus-strategy.md
app/enterprise/0.34-x/vitals-prometheus-strategy.md
app/enterprise/2.4.x/vitals/vitals-prometheus-strategy.md
==========
```

### Testing
Manual review after updating each page.


app/_includes/docs-sidebar-item.html
```diff
@@ -6,7 +6,7 @@
   {% for item in include.items %}
   {% if item.items %}
   {% assign id = include.id | append: '-' | append: forloop.index %}
-  {% include docs-sidebar-item.html id=id docs_url=include.docs_url items=item.items title=item.text url=item.url absolute_url=item.absolute_url active=item_active %}
+  {% include docs-sidebar-item.html id=id url=include.url docs_url=include.docs_url items=item.items title=item.text url=item.url absolute_url=item.absolute_url active=item_active %}
```
lena-larionova, line 9: This statement is declaring `url` twice: `url=include.url` and `url=item.url`. This is causing multiple nav entries to be flagged as active: https://deploy-preview-3080--kongdocs.netlify.app/enterprise/2.5.x/support-policy/

app/_includes/docs-sidebar.html
```diff
@@ -90,30 +90,30 @@
...
+  {% elsif include.edition == 'contributing' %}
   {% assign docs_url = '/contributing' %}
   {% endif %}
 
   <ul class="accordion-container">
     <input id="accordion-opened" type="text" />
-    {% for nav in page.nav_items %}
-    {% include docs-sidebar-item.html id=forloop.index docs_url=docs_url items=nav.items url=nav.url absolute_url=nav.absolute_url target=item.target icon=nav.icon title=nav.title %}
+    {% for nav in include.nav_items %}
+    {% include docs-sidebar-item.html id=forloop.index url=include.url docs_url=docs_url items=nav.items url=nav.url absolute_url=nav.absolute_url target=item.target icon=nav.icon title=nav.title %}
```
lena-larionova, line 116: Duplicate `url=` entries in this include: `url=include.url` and `url=nav.url`.

lena-larionova (approved):
LGTM!

PR #3084, opened 2021-08-04
[DOCU-1655]Updates datadog plugin doc with correct default value

HCloward:
### Review
@svenwal - I removed the original string and replaced it with 
### Summary
This is to remove the current default value for config.metrics as it is wrong. 
### Reason
[DOCU-1655](https://konghq.atlassian.net/browse/DOCU-1655)
### Testing
[Datadog plugin](https://deploy-preview-3084--kongdocs.netlify.app/hub/kong-inc/datadog/)

### DIFF
--- a/app/_hub/kong-inc/datadog/index.md
+++ b/app/_hub/kong-inc/datadog/index.md
@@ -79,6 +79,8 @@ params:
       datatype: array of record elements
       description: |
         List of Metrics to be logged. Available values are described at [Metrics](#metrics).
+        By default, the plugin logs all available metrics. If you specify an array of metrics,
+        only the listed metrics are logged.
     - name: prefix
       required: true
       default: "`kong`"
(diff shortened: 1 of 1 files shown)


app/_hub/kong-inc/datadog/index.md
```diff
@@ -75,7 +75,7 @@ params:
       description: The port to send data to on the upstream server
     - name: metrics
       required: true
-      default: All metrics are logged
+      default: 
       datatype: array of record elements
       description: |
         List of Metrics to be logged. Available values are described at [Metrics](#metrics).
```
lena-larionova, line 81: If you're leaving out the default (and I understand why, since the default is "everything"), then it might be useful to note in the description here that all metrics are logged by default, unless you specify a list explicitly (if I'm understanding correctly). So, something like:

> By default, the plugin logs all available metrics. If you specify an array of metrics, only the listed metrics are logged.

Need dev confirmation that that's correct though.

HCloward: @svenwal What do you think about Lena's comment above?

lena-larionova (approved):
LGTM!

PR #3085, opened 2021-08-04
Version switcher cache

mheap:
### Review
@lena-larionova 

### Summary
Desired behavior (was previously working): when you're on a doc page (eg https://docs.konghq.com/enterprise/2.1.x/deployment/installation/docker/) and switch to another version of this page using the dropdown in the top left, the same page should load up but with the new URL ( eg https://docs.konghq.com/enterprise/2.4.x/deployment/installation/docker/). Caching breaks this.

Caching is also breaking the link to latest documentation from older versions. Again, see https://docs.konghq.com/enterprise/2.1.x/deployment/installation/docker/ - the note at the top of the page should link to the latest version of this Docker installation, but instead it simply goes to https://docs.konghq.com/enterprise/

### Reason
DOCU-1633

### Testing
Switch versions on the left hand side switcher and it should take you to the same page. Use the link at the top of the main content block to get back to the latest version. It should keep you on the same page

### DIFF
--- a/app/_layouts/docs-v2.html
+++ b/app/_layouts/docs-v2.html
@@ -11,7 +11,7 @@ page.kong_latest.release %}
 <div class="page v2 {{ page.class }}" data-url={{ page.url }}>
 
   <div class="container">
-    {% include_cached docs-sidebar.html kong_version=page.kong_version edition=page.edition %}
+    {% include_cached docs-sidebar.html url=page.url kong_version=page.kong_version edition=page.edition %}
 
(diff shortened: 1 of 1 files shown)


mheap:
This PR has an impact on build time as it prevents caching. This is a fix for the existing behaviour, and a long term solution that does not impact caching (similar to how we do active states in the navigation) is to follow

lena-larionova (approved):
> This PR has an impact on build time as it prevents caching.

I was wondering why the build time was 17 minutes. That's frustrating.

However, problem solved for now. Thanks for the fix!

PR #3090, opened 2021-08-05
fixes version issues in installation docs

HCloward:
### Review
Check to be sure all installation docs point to the correct version number in Docker.
### Summary
I tried to be clever and added a bunch of includes bc so much content in the installation docs is the exact same, except for the key difference in the version number needing to be pulled from Docker. The includes were worthless since each version's needs to differ in that way. So removed the includes and updated the versions to be correct.
### Reason
Someone noticed the latest version of docs pointed to the 2.4 version in Docker. What he didn't note is that in fact all of the 2.x versions were pointing to 2.4 bc of my cleverness described above.
### Testing
Will include Netlify build link when available.


Bradamant3 (approved):
LGTM. Spot-checked at least a dozen and all looked good. Extra props for getting started on an approach we should figure out support for ❤️

PR #3100, opened 2021-08-11
Redirects for the KIC beta

lena-larionova:
### Summary
Add Netlify redirects for KIC beta.

### Reason
KIC beta.

### Testing
TBA


GayleNeumann (approved):
LGTM

PR #3103, opened 2021-08-11
[DOCU-1692] Update site title, name, and description

lena-larionova:
### Review
@falondarville and also @Kong/team-docs 

### Summary
This change will give us better link previews and a more up-to-date website summary for search engines.

**Reviewers:** I did some research on the difference between site name and site title, and came up blank. Site name isn't even a default jekyll variable, it's a custom one that someone added. I'm still not sure why we had a different site name vs site title, so I made them the same and removed the redundancy in link titles. However, if someone knows more about this and could answer _why_ we might have this and whether we might need to come up with a unique site name, I'd be glad to hear about it.

### Reason
Site metadata was very out of date, and our link previews would only ever generate a site description as the link title. 

### Testing
TBA


falondarville (approved):
LGTM

PR #3104, opened 2021-08-11
Fix subscribe button in footer

falondarville:
### Review

@lena-larionova 

### Summary

The subscribe button was supposed to be in the footer, but instead was floating up near or under the top main nav. I fixed this. 

### Reason

It was broken. 

### Testing

https://deploy-preview-3104--kongdocs.netlify.app

Check that the subscribe button appears in the footer next to the email input form.

### DIFF
--- a/app/_assets/stylesheets/footer.less
+++ b/app/_assets/stylesheets/footer.less
@@ -69,9 +69,18 @@
     }
   }
 
+  @media (max-width: 1045px) {
+    #footer-form-button {
+      font-size: 14px;
+    }
+  }
+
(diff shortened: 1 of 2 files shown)


lena-larionova:
@falondarville Doesn't look quite right above 767px and below... 1000-something:

Screen Shot 2021-08-12 at 9 19 13 AM
Screen Shot 2021-08-12 at 9 19 20 AM

The width calc seems to mess with it - if you disable it, the width looks better. Might need a media query though.

lena-larionova (approved):

PR #3109, opened 2021-08-12
chore: Update link-checker action

lena-larionova:
### Review

@reviewer

### Summary
Switching over our link checker action to the new build flow used [here](https://github.com/Kong/docs.konghq.com/blob/main/.github/workflows/run-smoke-tests.yml). Should shorten the run time by a few minutes.

Note that the pull_request trigger is temporary; using it to test this action through this PR. Will remove it if the action works.

### Reason
Attempting to shorten action run time. The docker container also tends to double link-checker run time.

### Testing
Wait for check to pass (or not pass).

### DIFF
--- a/.github/workflows/check-links.yml
+++ b/.github/workflows/check-links.yml
@@ -3,19 +3,52 @@ on:
   workflow_dispatch:
   schedule:
   - cron: "0 0 * * 1" # once every week
+  pull_request:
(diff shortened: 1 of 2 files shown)


lena-larionova:
Will revisit this approach in another way, as this is mostly a bandaid for a problematic link-checker.

PR #3112, opened 2021-08-16
Deleted the duplicated order for ip-restriction

ikeike443:
### Review

@Kong/team-docs 

### Summary

 I checked the order by the below commands:

```
cd /usr/local/share/lua/5.1/kong/plugins/
find . | xargs grep  \.PRIORITY |  sort -k 3 -t " " -n -r | grep ip-
```

Then you can find below:
`./ip-restriction/handler.lua:  PRIORITY = 990,`

### DIFF
--- a/app/enterprise/2.5.x/plugin-development/custom-logic.md
+++ b/app/enterprise/2.5.x/plugin-development/custom-logic.md
@@ -268,7 +268,6 @@ pre-function                | `+inf`
 correlation-id              | 100001
 zipkin                      | 100000
 exit-transformer            | 9999
-ip-restriction              | 3000
 bot-detection               | 2500
 cors                        | 2000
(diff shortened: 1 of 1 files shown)


app/enterprise/2.5.x/plugin-development/custom-logic.md
```diff
@@ -268,7 +268,6 @@ pre-function                | `+inf`
 correlation-id              | 100001
 zipkin                      | 100000
 exit-transformer            | 9999
-ip-restriction              | 3000
```
ikeike443, line 271: this is wrong

lena-larionova (approved):
Thanks for the fix, and LGTM!

I checked the Enterprise-side table as well, looks like it was already correct there.

PR #3116, opened 2021-08-16
[DOCU-536] Add Run in Insomnia button for Konnect

lena-larionova:
### Review
@falondarville 

### Summary
Adding "Run in Insomnia" button, generated with https://insomnia.rest/create-run-button. 

### Reason
To make it easier to use the Konnect API, and to cross-advertise Insomnia.

### Testing
https://deploy-preview-3116--kongdocs.netlify.app/konnect/reference/konnect-api/

### DIFF
--- a/app/konnect/reference/konnect-api.md
+++ b/app/konnect/reference/konnect-api.md
@@ -15,12 +15,18 @@ For a list of all endpoints, see:
 
 You can run {{site.konnect_short_name}} API requests using
 [Insomnia](https://insomnia.rest), Kong's open-source API client. If you have
-Insomnia v2021.4.0 or later installed, click the button below to import the Konnect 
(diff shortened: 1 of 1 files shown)


falondarville (commented):
Text looks good. The collection/document is not importing into my Insomnia application.

lena-larionova:
For future reference: tested Insomnia v2021.3.0, v2021.4.0, and v2021.3.1. Turns out this was a new feature with v2021.4.0.

falondarville (approved):
Confirmed with @develohpanda that the button works pre-2021.4.0 (there must have been something wrong with my setup on 2021.3.0). We don't need the disclaimer but we can recommend they upgrade as we want all users to be on the latest version. Also note that the auto upgrade won't happen unless the user completely closes out of the application.

PR #3127, opened 2021-08-20
[WIP] Splitting commands and responses in KIC guides

lena-larionova:
### Summary
Splitting out commands from responses in three KIC guides: Getting started, Getting started with Istio, and Cert manager.

### Reason
https://konghq.atlassian.net/browse/DOCU-1658 and many other complaints in the past. 
Users get confused as they try to copy a command and run it without looking at every line:
* Often, the examples in the KIC docs have many lines of command (e.g. `kubectl apply` examples) and one line of response, and it's hard to see that. 
* Conversely, sometimes the command is one line and the response/example is many lines, but it looks daunting and complicated at first glance because all are in the same codeblock.

### Testing
TBA


lena-larionova:
Closing as this covers the Istio guide, and that was rewritten entirely anyway.

PR #3130, opened 2021-08-20
Stop wrong banner from appearing in KIC beta version

lena-larionova:
### Summary
Setting KIC beta notification banner to not appear on 2.0.x.

### Reason
Since the beta KIC doc is based on main, it's pulling in the beta notification banner into all pages. Because of that, the beta docs end up with two banners on every page: https://kic-v2-beta--kongdocs.netlify.app/kubernetes-ingress-controller/

### Testing
N/A, but tested locally. Will need to merge into beta branch.


GayleNeumann (approved):
LGTM

PR #3133, opened 2021-08-23
[DOCU-1744] Open-source badge

lena-larionova:
### Review
@falondarville 

### Summary
* Adding a new badge for "open-source only"
* Labelling Performance Testing Framework as OSS
* Adding to layout

### Reason
https://konghq.atlassian.net/browse/DOCU-1744

### Testing
TBA

### DIFF
--- a/app/_assets/stylesheets/badges.less
+++ b/app/_assets/stylesheets/badges.less
@@ -55,6 +55,15 @@
       background-color: @red-700;
       text-align: center;
     }
+
+  &.oss::after {
+      content: "OPEN-SOURCE ONLY";
+      opacity: 0.7;
+      color: white;
+      border-color: #6e30bf;
+      background-color: #6e30bf;
+      text-align: center;
+    }
 }
 
 span.badge {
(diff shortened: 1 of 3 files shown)


HCloward (approved):
Looks good - like purple - wondering though, should we shorten it with the abbreviation for open source - OSS? Also, for Enterprise we only have the one word and leave out only. I see why you include only bc this feature is strictly OSS whereas other OSS features do apply to enterprise, but not sure if it is confusing that it is not explicitly stated for enterprise.

lena-larionova:
> Looks good - like purple - wondering though, should we shorten it with the abbreviation for open source - OSS? 

OSS looked super awkward and short on its own, that's why I went with open-source 😁  but we could make it work with "OSS-only".

> Also, for Enterprise we only have the one word and leave out only. I see why you include only bc this feature is strictly OSS whereas other OSS features do apply to enterprise, but not sure if it is confusing that it is not explicitly stated for enterprise.

I was thinking of the difference this way: Enterprise is a roll-up of everything below it, open-source is the foundation. Calling something "Enterprise" comes with an implicit "only", as it's a paid tier. Calling it "open-source" (without only) makes it sound foundational, and implies that it's included in Enterprise.

PR #3134, opened 2021-08-24
CA rotation in Kong Mesh

jakubdyszkiewicz:
Add info about CA rotation in the upcoming release of Kong Mesh.

### DIFF
--- a/app/_data/docs_nav_mesh_1.4.x.yml
+++ b/app/_data/docs_nav_mesh_1.4.x.yml
@@ -54,3 +54,5 @@
       url: /features/kds-auth
     - text: FIPS support
       url: /features/fips-support
+    - text: Certificate Authority rotation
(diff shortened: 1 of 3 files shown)


Bradamant3:
@jakubdyszkiewicz I pulled down and copyedited directly bc it's faster and ... well, release week. I did a minimal job in the interest of time, but also because I wanted to ask about a different approach. I understand the step-by-step guide model, and it can be useful (although much overused in general on this site), but it feels a bit like overkill here. The yaml is all dupe or near-dupe, and I wonder whether the same result could be achieved with smaller snippets and explanatory text instead of instructional text. BUT ... we're also running out of time, so probably best to :shipit: as is and rework later, if you agree. I'll approve as soon as you give me a 👍  here or on Slack.

PR #3137, opened 2021-08-25
Invalid reference to config,yaml in the documentation

rahulrai-in:
Replaced `config,yaml` with `values.yaml` in document.


HCloward (approved):

lena-larionova (approved):
Nice catch, LGTM! Thanks for your contribution.

PR #3162, opened 2021-08-30
[DOCU-1756][DOCU-1412] Konnect: Improve cert renewal instructions

lena-larionova:
### Summary
* Improving language around cert renewal
* Fixing a step referring to a Docker tab **Generate Certs** button which doesn't exist.
* Splitting the instructions for updating certs into Docker, Linux, and Kubernetes.

Picked
### Reason
https://konghq.atlassian.net/browse/DOCU-1412
https://konghq.atlassian.net/browse/DOCU-1756

For Kubernetes/Helm instructions: I went with the delete-and-replace method for secrets. There were a couple more options that I investigated:

**Option 1:** `kubectl edit secret <secret-name>` - directly edit the secret. 
Con: have to manually decode/encode the string. This is as easy as `echo 'MWYyZDFlMmU2N2Rm' | base64 --decode`, but it adds an extra step.

**Option 2:** Pipe contents of secret to output, edit file, then apply. Con: same as above.

**Option 3:** Delete secret, create new one, replace. 

Went with the third option as it seems like the most secure(?), simple, and involves least extra unfamiliar work outside of kubernetes.

### Testing
Preview: https://deploy-preview-3162--kongdocs.netlify.app/konnect/runtime-manager/renew-certificates/


lena-larionova:
@rainest @shaneutt not sure whom to ping for review here, but the tl;dr for context is: there was recently an issue with cert rotation in Konnect, and one of the outcomes was to revisit this ticket that describes how to renew a cert in a Kubernetes environment.

lena-larionova:
@Kong/team-k8s

rainest (approved):
You can use `stringData` as shown in https://kubernetes.io/docs/concepts/configuration/secret/#basic-authentication-secret to enter strings without encoding. K8S will encode them and convert to `data` for you when you save.

That said, for this I'm kinda inclined to go with creating a new Secret anyway. Multi-line strings in YAML are a bit annoying to work with, and if you already have files it's easier to use the create from file tool anyway.

Approving the proposed version here since the file method is what I prefer.

Bradamant3 (approved):
LGTM

PR #3185, opened 2021-09-06
Fixed the URL

rahulrai-in:
1. The proxy is either available at a URL [Linux](https://docs.konghq.com/konnect/runtime-manager/gateway-runtime-conf/) or an IP address [Kubernetes](https://docs.konghq.com/konnect/runtime-manager/gateway-runtime-kubernetes/)
2. The next segment of the URL must be the route and not the name of the service

### DIFF
--- a/app/konnect/dev-portal/developers/dev-gen-creds.md
+++ b/app/konnect/dev-portal/developers/dev-gen-creds.md
@@ -23,7 +23,7 @@ A credential, or API key, generated in the {{site.konnect_short_name}} Dev Porta
    using your `key-auth` credential:
 
    ```
-   <proxy-url or proxy-ip>/<route>?apikey=<apikey>
+   {PROXY_URL_OR_PROXY_IP}/{ROUTE}?apikey={APIKEY}
    ```
 
 ## Delete a credential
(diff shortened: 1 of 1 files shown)


app/konnect/dev-portal/developers/dev-gen-creds.md
```diff
@@ -23,7 +23,7 @@ A credential, or API key, generated in the {{site.konnect_short_name}} Dev Porta
    using your `key-auth` credential:
 
    ```
-   <proxy-url>/<service>/?apikey=<apikey>
+   <proxy-url or proxy-ip>/<route>?apikey=<apikey>
```
lena-larionova, line 26:
```suggestion
   {PROXY_URL_OR_PROXY_IP}/{ROUTE}?apikey={APIKEY}
```
We're transitioning variables in the docs to this formatting.

 
I think it's also useful to add a real example here:

> For example, the final URL might look something like this:
> 
> ```
> http://localhost:8000/mock/?apikey=mykey
>```

lena-larionova (commented):
Nice catch, thanks for fixing this. I just have a suggestion on how to improve the usability a little bit more.

lena-larionova (approved):

PR #3186, opened 2021-09-07
[DOCU-1808]Updates kong version in download code sample

HCloward:
### Summary
@liyangau noted in docs channel on Slack that this download code sample was pointing to the prior version and not the latest version, 1.7. I have updated the variable to point to the 1 index instead of 0.
### Reason
Want to be sure users are being directed to download the version that matches the version of the doc they are reading.
[DOCU-1808](https://konghq.atlassian.net/browse/DOCU-1808)
### Testing
Will include Netlify link when available.


Bradamant3 (approved):
LGTM

PR #3200, opened 2021-09-09
Update navigation to include Kong Academy link, remove outdated links

rossmcdonald:
### Summary
Kong Academy now supplants both Kong U and the old workshops/labs environment.

### Reason
The Kong U and old Portal links are now outdated and should no longer be publicized.


rossmcdonald:
CC @felderi and @jmaher31 for approval (in case it's needed). You can see what this will look like in a preview [here](https://deploy-preview-3200--kongdocs.netlify.app/) for reference:

image

lena-larionova (approved):
LGTM!

PR #3210, opened 2021-09-13
[DOCU-1821][DOCU-1758][DOCU-1710] Refactor plugin hub stylesheet

lena-larionova:
### Review
@falondarville, but really also all of @Kong/team-docs for style feedback

### Summary
 * Removing the old `hub.css` file and replacing it with `hub-landing-page.less`. 
   * Paring down a 15k line file into 370 lines.
   * Combining some classes, removing unnecessary properties
   * Adding comments to top-level sections to explain what they're styling
   * Fixing mobile responsiveness issues
   * Aligning styles closer to docs.konghq.com landing page
 * Fixing plugin filter to hide cards completely instead of just fading them out; scrolling to top on new filter
 * Removing smoke test for `hub.css`
 * Accessibility fixes for sidebar

### Reason
* Getting rid of a legacy file that was messing up our plugin hub page and causing issues with navbar, footer, and search.
  * Should improve page performance significantly, as the site doesn't have to load a completely different, massive CSS file just for one page
* Adjusting filter as the current behaviour gets constant complaints.

### Testing
**Note:** Close the promo banner at the top of the page to check styles. It's messing with our site in places and needs to be fixed separately.

https://deploy-preview-3210--kongdocs.netlify.app/hub/
Check:
* Filters
* Hover states
* Mobile responsiveness (resize the browser a bunch and make sure that the cards align)
* Top nav and footer - the new CSS file doesn't touch them, which is the point. The previous one was overwriting some styles
* Tab through the left menu, should be accessible
* Colours, alignment, anything that might look out of place

In the code (if you're reviewing that side), check that the content order makes sense to you and that the new CSS file is easy to navigate and understand.


falondarville (approved):
Looks great!! I tested it for accessibility. In the future, we should create a way for a user to be moved to the cards once they have pressed ENTER on a filter. Right now, the functionality is much improved. 

I also looked through the code and really appreciate the thorough code comments. 

https://media.giphy.com/media/tsX3YMWYzDPjAARfeg/giphy.gif

PR #3216, opened 2021-09-14
Introduce a "no-icon" option for admonitions and remove "tip"

lena-larionova:
### Review
@Kong/team-docs 

### Summary
* Add a `no-icon` option for admonitions
* Removing the `tip` class, as we decided sometime back, as a team, to narrow down the options to only three
* Cleaning up any use of the `tip` style (there weren't many)
* Adding usage info to contributing guide

### Reason
Need an iconless option for some situations. For example, when labelling a product "beta" or "tech preview", we don't want the X icon around as it looks like the feature is broken (this just came up for Okta IdP docs). 

### Testing
https://deploy-preview-3216--kongdocs.netlify.app/contributing/markdown-rules/#admonitions

### DIFF
--- a/app/_assets/stylesheets/page.less
+++ b/app/_assets/stylesheets/page.less
@@ -1276,7 +1276,7 @@ Generic Styling for Desktop
       display: inline;
     }
 
-    &::before {
+    &:not(.no-icon)::before {
       content: "\f05a";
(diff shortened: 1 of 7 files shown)


falondarville (approved):
LGTM

PR #3219, opened 2021-09-15
Vault Common Name support

jakubdyszkiewicz:
Add docs about Vault Common Name support. This will land in Kong Mesh 1.4.1

### DIFF
--- a/app/mesh/1.4.x/features/vault.md
+++ b/app/mesh/1.4.x/features/vault.md
@@ -198,7 +198,7 @@ spec:
             namespace: "" # optional
             pki: kmesh-pki-default # name of the configured PKI
             role: dataplane-proxies # name of the role that will be used to generate data plane proxy certificates
-            commonName: '{{ tag "kuma.io/service" }}.mesh' # optional. If set, then common name is added to the certificate. You can use "tag" directive to pick a tag which will be base for common name.
+            commonName: '{{ tag "kuma.io/service" }}.mesh' # optional. If set, then commonName is added to the certificate. You can use "tag" directive to pick a tag which will be base for commonName.
             tls:
               caCert:
                 secret: sec-1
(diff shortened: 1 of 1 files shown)


Bradamant3:
@jakubdyszkiewicz thanks for the heads up! Will create the branch for the 1.4.1 release and retarget this PR.

app/mesh/1.4.x/features/vault.md
```diff
@@ -196,6 +198,7 @@ spec:
             namespace: "" # optional
             pki: kmesh-pki-default # name of the configured PKI
             role: dataplane-proxies # name of the role that will be used to generate data plane proxy certificates
+            commonName: '{{ tag "kuma.io/service" }}.mesh' # optional. If set, then common name is added to the certificate. You can use "tag" directive to pick a tag which will be base for common name.
```
HCloward, line 201: I'm not sure if this should match as it does in the previous comment, but if it is referring to the same thing consistency helps to avoid confusion

```suggestion
            commonName: '{{ tag "kuma.io/service" }}.mesh' # optional. If set, then commonName is added to the certificate. You can use "tag" directive to pick a tag which will be base for commonName.
```

app/mesh/1.4.x/features/vault.md
```diff
@@ -232,6 +235,9 @@ mtls:
         address: https://vault.8200
         agentAddress: "" # optional
         namespace: "" # optional
+        pki: kmesh-pki-default # name of the configured PKI
+        role: dataplane-proxies # name of the role that will be used to generate data plane proxy certificates
+        commonName: '{{ tag "kuma.io/service" }}.mesh' # optional. If set, then common name is added to the certificate. You can use "tag" directive to pick a tag which will be base for common name.
```
HCloward, line 240:
```suggestion
        commonName: '{{ tag "kuma.io/service" }}.mesh' # optional. If set, then commonName is added to the certificate. You can use "tag" directive to pick a tag which will be base for commonName.
```

HCloward (approved):

PR #3220, opened 2021-09-15
KIC: Set Host header when accessing external service

cwurm:
These days, most internet facing services will require a `Host` header in a request. Most importantly, Cloudflare does not allow direct IP access to any services it sits in front of (e.g. try accessing mockbin.org using its IP address `http://172.67.201.247/request` to see what happens).

The Host header is added automatically by browsers and CLI tools when accessing a service directly (e.g. running `curl -v httpbin.org/anything` will set `Host: httpbin.org`). But it is not added when using KIC in combination with external services. 

Coincidence would have it that the Httpbin service used here is one of very few services that will accept requests even without the correct header. But most other services will not - as I painfully discovered today, trying the instructions on this docs page with other services e.g. mockbin.org or mocktainer.io will fail and often with hard to understand error messages.

Therefore the change of adding the Host header into the example `curl` command - while not strictly required in this specific case, it's harmless to still specify it (and sending it is arguably the right thing to do in general) and helps when using the examples as templates (as I tried to do today).


Bradamant3:
@shaneutt do you want to weigh in on this one?

shaneutt (approved):

shaneutt:
Seems fine, as stated its innocuous at worst, but provides some helpful illustration at best if a user is doing some experimentation outside of what the guide specifically provides.

lena-larionova:
@cwurm I think we were all waiting for the 2.0 beta docs to merge before publishing this change. 
Can you also make the change in the newly published 2.0 doc?

PR #3222, opened 2021-09-15
Minor Mesh release (without Kuma)

Bradamant3:
### Review

@Kong/team-docs 

### Summary
Patch release for Mesh only to fix goroutine leak (see changelog item)

### Reason
Release

### Testing
Will post Netlify link. See also version number change in `kong_versions.yml`


falondarville (approved):
LGTM

PR #3227, opened 2021-09-17
Added 2.5.x compatibility

d19dotca:
### Summary
Added 2.5.x compatibility in list.

### Reason
This plugin works correctly on 2.5.0 but is missing from the documentations as a supported version. I believe this is due to confusion earlier on deprecation of this plugin which was a decision that was later reversed as I understand it. Ref: https://konghq.atlassian.net/browse/FT-1695


lena-larionova:
This doesn't hurt to add, but we're working on a less manual way of updating this. @HCloward do you want to merge this change, or would you rather wait until we get the one metadata file method implemented?

PR #3229, opened 2021-09-20
[DOCU-1843] Guide for using defaults with decK

lena-larionova:
### Summary
Documenting the decK defaults feature introduced in 1.8 (https://github.com/Kong/deck/pull/419). 

Going by Harry's responses on https://konghq.atlassian.net/browse/FTI-2356 and looking through the implementation PR in the decK repo, I documented only the core entities. However, based on Fel's last comment on the FTI ticket, it looks like it might be possible to set plugin defaults as well, as long as you know what the default values are? Not sure about this.

### Reason
https://konghq.atlassian.net/browse/DOCU-1843

### Testing
https://deploy-preview-3229--kongdocs.netlify.app/deck/1.8.x/guides/defaults/

### DIFF
--- a/app/deck/1.8.x/guides/defaults.md
+++ b/app/deck/1.8.x/guides/defaults.md
@@ -153,10 +153,8 @@ Summary:
 {% endcapture %}
 {{ deck_diff2 | indent | replace: " </code>", "</code>" }}
 
-    Notice that even though you've made no changes, the response shows a list
-    of new property configurations. This is happening because
-    {{site.base_gateway}} applied defaults and decK is unaware of them, so decK
-    treats them like changes to the configuration.
+    Even though you've made no changes, the response shows a list
+    of new property configurations. The list of new configurations appears because {{site.base_gateway}} applied defaults and decK is unaware of them, so decK treats them like changes to the configuration.
 
 ### Set defaults
 
--- a/app/deck/1.8.x/guides/defaults.md
+++ b/app/deck/1.8.x/guides/defaults.md
@@ -1,10 +1,10 @@
 ---
 title: Set Up Object Defaults
 ---
-Use object defaults to enforce a set of standard values and help avoid
+Use object defaults to enforce a set of standard values and avoid
 repetition in your configuration.
 
-You can set configuration defaults for core {{site.base_gateway}} objects:
+You can set configuration defaults for the following core {{site.base_gateway}} objects:
 - Service
 - Route
 - Upstream
@@ -16,11 +16,11 @@ decK supports setting object defaults both in self-managed
 {:.important}
 > **Important:** This feature has the following limitations:
 * Plugin object defaults are not supported.
-* If an existing property's default value changes in a future Gateway release,
+* If an existing property's default value changes in a future {{site.base_gateway}} release,
 decK has no way of knowing that this change has occured, as its `defaults`
 configuration would overwrite the value in your environment.
 
-## Behavior
+## Object defaults behavior
 
 Defaults get applied to both new and existing objects. If an object has an
 explicit setting for a property, the object-level setting takes precedence over
@@ -30,7 +30,7 @@ the default.
 what the defaults are for each object in the
 [Admin API reference](/gateway-oss/latest/admin-api/), or use the
 [`/schemas`](/gateway-oss/latest/admin-api/#retrieve-entity-schema) endpoint to
-retrieve the latest object schemas for your instance of the Gateway.
+retrieve the latest object schemas for your instance of the {{site.base_gateway}}.
 
 Configuring your own defaults is a good way to keep updated on potential
 breaking changes between versions. If you upgrade {{site.base_gateway}} to a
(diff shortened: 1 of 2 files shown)


app/deck/1.8.x/guides/defaults.md
```diff
@@ -0,0 +1,380 @@
...
+an explicit setting for a property, the object-level setting takes precendence
+over the default.
+
+{{site.base_gateway}} sets some default values for most objects. You can see
+what the defaults are for each object in the
+[Admin API reference](/gateway-oss/latest/admin-api/).
+
+If defaults are not set in the declarative configuration file, any newly
+configured object picks up {{site.base_gateway}}'s defaults and diverges from
+the source configuration file. You can see
```
hbagdi, line 23:
```suggestion
the source configuration file, ultimately resulting in a diff (not good). You can see
```

Something along those lines

app/deck/1.8.x/guides/defaults.md
```diff
@@ -0,0 +1,380 @@
...
+[an example of this](#create-a-file-and-test-without-defaults) in this guide.
+
+decK supports setting object defaults both in self-managed
+{{site.base_gateway}} and with {{site.konnect_saas}}.
+
+{:.important}
+> **Important:** This feature has the following limitations:
+> * Plugin object defaults are not supported.
```
hbagdi, line 31: Correct.

app/deck/1.8.x/guides/defaults.md
```diff
@@ -0,0 +1,380 @@
...
+>* This method of defining defaults is manual and requires upkeep. If
+configuration defaults change in future {{site.base_gateway}} versions, it
+could cause conflicts with the defaults defined in your declarative
+configuration. If this happens, check the latest
+[Gateway changelog](/enterprise/changelog) to find the source of the conflict.
```
hbagdi, line 36: 1. I understand calling this out but it rarely happens because it is a breaking change.
2. This is a good thing. If you upgrade your gateway and your deck diff run shows a diff, you know something has changed and that probably needs your attention. We can't expect a user to read the entire changelog (most don't) but they will look at the CI failure. It is another place to provide some feedback to the user.

The change we could do is that we could have a more positive tone here, describing the (above described) merits of the feature and instill more confidence in the product.
This is a strong opinion, lossely held situation.

app/deck/1.8.x/guides/defaults.md
```diff
@@ -0,0 +1,380 @@
...
+            timeouts: 0
+        threshold: 0
+```
+For all available properties, see the
+[Upstream object](/gateway-oss/latest/admin-api/#upstream-object) documentation.
+
+{% endnavtab %}
+{% navtab Target %}
+Targets have no required default properties. However, you can set custom
+defaults as desired.
```
hbagdi, line 375: Not sure what you mean by "custom".
`Weight` is what we default I think (please recheck)

hbagdi (commented):
Great doc. +1 overall.

app/deck/1.8.x/guides/defaults.md
lena-larionova: Thanks for the pointer to the /schemas endpoint (on the ticket), I didn't know that existed. Checked it,`weight` defaults to 100. I'll adjust.

app/deck/1.8.x/guides/defaults.md
rainest: We don't actually expect a conflict here, correct? I'd expect that even if Kong's default changes, decK will still respect whatever is in your file anyway, e.g. if we change the connect_timeout default from 60s to 90s, and you were and continued using a defaults file with 60s, you'd still get 60s on all services without their own config: decK sets that value explicitly, so Kong does not auto-fill it.

The maintenance burden is that you need to keep your defaults file up to date if you want the stock defaults always. Changing that for an upgrade is not in-line with how Kong would operate in practice though, since changing the Kong default wouldn't effect existing config, only new config, whereas changing the decK default file _would_ affect existing config.

lena-larionova: Travis's point is what I was covering, though not super clearly in the original draft. But I think what we have here are two different points that need to be called out:
- Per @hbagdi's point: Post-gateway upgrade, if a new value appears in a diff, that means the value has a Gateway-side default and you should add it to your config file. This is positive and should be called out in doc.
- Per @rainest's point: However (and this is the limitation/part where users have to be careful), if you have a some default set for a value and the existing default value changes in a new version, then you won't know about it unless you keep an eye on changelogs. This is especially important if the default changed for some security reason, which definitely does happen with kong.conf, at least (maybe less so for gateway entities).

rainest: Yes, I think so. Harry's scenario (new field with a default) is much more common. Changes to existing schema defaults are quite uncommon, but not impossible. I'm not sure how to best explain it since most of them are probably innocuous and you're fine leaving them as-is, so despite the differences between how decK and Kong defaults are applied, most users probably don't need to worry about them.

If we had something significant enough that we felt existing config needed to change, I'd expect that we'd call that out loudly with either a large warning or even a migration to update values, but I can't recall that we've ever done that in the past.

So mine maybe warrants a blurb for "Differences between changes to default updates in Kong and decK" that effectively states "updated Kong defaults apply only to new objects, updated decK defaults apply to existing objects also" and "do I need to keep up with the latest Kong defaults to update my decK defaults accordingly? generally no: you should add new fields but do not need to revise existing ones unless you want to."

rainest (approved):

app/deck/1.8.x/guides/defaults.md
```diff
@@ -0,0 +1,418 @@
...
+http :8001/schemas/targets
+```
+{% endnavtab %}
+{% endnavtabs %}
+
+## See also
+* [Deduplicate plugin configuration](/deck/{{page.kong_version}}/guides/deduplicate-plugin-configuration)
+* [Distributed configuration for Kong Gateway using decK](/deck/{{page.kong_version}}/guides/distributed-configuration)
+* [Using multiple files to store configuration](/deck/{{page.kong_version}}/guides/multi-file-state)
+* Kong Gateway admin API: [`/schemas` endpoint](/gateway-oss/latest/admin-api/#retrieve-entity-schema)
```
HCloward, line 418:
```suggestion
* {{site.base_gateway}} admin API: [`/schemas` endpoint](/gateway-oss/latest/admin-api/#retrieve-entity-schema)
```

HCloward (approved):
Made a couple of suggestions, otherwise looks ok to me.

PR #3237, opened 2021-09-22
decK for DB-less mode in the getting-started-guide

aaronArinder:
### Reason
When using DB-less mode, the admin API (and I'm assuming Kong Manager, but I don't know that to be true) doesn't support creating services to avoid config drift. This should clarify that for other folks going through the guide.


lena-larionova (approved):
LGTM!

You're right about Kong Manager as well :) DB-less mode takes declarative configuration only, which in this case means following the decK path. Thanks for your contribution!

PR #3240, opened 2021-09-23
fix: Literal should be used instead of a method that results in literal

iamrajiv:


lena-larionova:
Thank you for your contribution, but we have decided not to merge it, as this change is a matter of personal preference.

PR #3253, opened 2021-09-24
docs(configuration) add AAAA to dns_order

fffonion:


fffonion:
@Tieske The original PR says it's an unsupported feature, do we want to document it actually?

Tieske:
I think not, becasue it's unsupported. If we add it it should get a note that we do not fully support it, just for testing/experimenting.

fffonion:
@Tieske In that case I'm closing this PR, thanks!

PR #3255, opened 2021-09-24
2.6.x/prometheus request termination

kikito:
In addition to documenting 2.6 changes, this PR also:

* Includes a "new in this version" section in those plugins
* Adds a previously missing version for the request-termination plugin.


app/_hub/kong-inc/request-termination/index.md
```diff
@@ -69,16 +73,31 @@ params:
...
     Once applied, every request (within the configured plugin scope of a Service,
     Route, Consumer, or global) will be immediately terminated by
     sending the configured response.
 
 ---
 
+## New in 2.1.0
+
+- `trigger` config option
+- `echo` config option
```
PaulFischer-Kong, line 95: @Tieske - By tomorrow am (EST) would you mind improving the request-termination docs - maybe even reflect the latest changelog entries.

fffonion:
Not sure if there's anyone being assinged to https://github.com/Kong/kong/commit/8e4c0033e6d2b8b6be0250fa805e8263b189a168#diff-bde7632435763f17816c0dc31c3912aa1462a51c60c881f19d0644401a9469c4 but we would need to document that as well. It's a non-breaking behaviour change.
We can add the new behaviour to sample output of doc for example https://github.com/Kong/docs.konghq.com/pull/3255/files#diff-65e3908f76b1496bf2df357ebeb7aaabcfc912ec7e68b727f8371b12378837a9R207
becomes:
```
kong_upstream_target_health{upstream="<upstream_name>",target="<target>",address="<ip>:<port>",state="healthchecks_off",subsystem="http"} 0
kong_upstream_target_health{upstream="<upstream_name>",target="<target>",address="<ip>:<port>",state="healthy",subsystem="http"} 1
kong_upstream_target_health{upstream="<upstream_name>",target="<target>",address="<ip>:<port>",state="unhealthy",subsystem="http"} 0
kong_upstream_target_health{upstream="<upstream_name>",target="<target>",address="<ip>:<port>",state="dns_error",subsystem="http"} 0
```

app/_hub/kong-inc/request-termination/index.md
```diff
@@ -69,16 +73,31 @@ params:
       default: "`application/json; charset=utf-8`"
       datatype: string
       description: Content type of the raw response configured with `config.body`.
+    - name: trigger
+      required: false
+      default:
+      description: When not set, the plugin always activates. When set to a string, the plugin will activate exclusively on requests containing either a header or a query parameter named like said string.
```
HCloward, line 79:
```suggestion
      description: When not set, the plugin always activates. When set to a string, the plugin will activate exclusively on requests containing either a header or a query parameter that includes the string.
```

app/_hub/kong-inc/request-termination/2.0.x.md
```diff
@@ -0,0 +1,90 @@
...
+    Once applied, every request (within the configured plugin scope of a Service,
+    Route, Consumer, or global) will be immediately terminated by
+    sending the configured response.
+
+---
+
+## New in 2.0.0
+
+There were changes in the Plugin Handler structure and on the Plugins DAO (`load_plugin_schemas`) that make this plugin
+backwards-incompatible if another plugin depended on it. Hence the bump to 2.0.0
```
HCloward, line 82:
```suggestion
backwards-incompatible if another plugin depended on it. 
```

app/_hub/kong-inc/request-termination/index.md
```diff
@@ -69,16 +73,31 @@ params:
...
+    - name: echo
+      required: false
+      default: false
+      description: When set, the plugin will echo a copy of the request back to the client. The main usecase for this is debugging. It can be combined with `trigger` in order to debug requests on live systems without disturbing real traffic
```
HCloward, line 83:
```suggestion
      description: When set, the plugin will echo a copy of the request back to the client. The main usecase for this is debugging. It can be combined with `trigger` in order to debug requests on live systems without disturbing real traffic.
```

app/_hub/kong-inc/request-termination/index.md
kikito: I think "is named like" is better than "includes" here. Explanation:

A header (or a query parameter) has two parts: the name and the value. For example in `My-Header: 14` the string `My-Header` is the name and `14` is the value.

When I say the header is "named like said string" I mean that its name must be literally the same. For example, if `trigger` is set to `Activate-Debug` then this header: `Activate-Debug: true` will be named the same, because its name is the same. The value can be anything, even the empty value 

A header that **contains** the string, on the other hand, could be be `My-Activate-Debug: true`, on which the string of the name contains `Activate-Debug` as a substring) or `My-Something-Other: Activate-Debug`, where it is on its value.

If there is another way to say what I meant in English, I don't know it. Perhaps I should add an example?

app/_hub/kong-inc/request-termination/index.md
kikito: If you are referring to the `trigger` and echo config options, they are explained on the same commit, a bit later (the section you are commenting in only "lists the changes" - a changelog of sorts).

kikito:
I have added docs for that option that @fffonion asked about (thanks!)

And also incorporated or answered to all the feeback from @HCloward and @PaulFischer-Kong .

app/_hub/kong-inc/request-termination/index.md
HCloward: how's that?

app/_hub/kong-inc/request-termination/index.md
PaulFischer-Kong: Got it - thank you. Missed that.

app/_hub/kong-inc/prometheus/index.md
```diff
@@ -78,8 +87,12 @@ dashboard: [https://grafana.com/dashboards/7424](https://grafana.com/dashboards/
 - **Connections**: Various Nginx connection metrics like active, reading,
   writing, and number of accepted connections.
 - **Target Health**: The healthiness status (`healthchecks_off`, `healthy`, `unhealthy`, or `dns_error`) of Targets
+<<<<<<< HEAD
   belonging to a given Upstream.
-- **Dataplane Status**: The last seen timestamp, config hash, and config sync status for
+=======
```
PaulFischer-Kong, line 92: @kikito I think your git messages ` <<<<<<< HEAD ` were merged here in the doc as text- please remove.

HCloward: that was my fault - sorry @kikito

app/_hub/kong-inc/request-termination/index.md
kikito: Good enough! Accepted the change

PR #3272, opened 2021-09-28
Cont Guidelines - Hackathon: rm Issue assignment

falondarville:
Given discussion with the Insomnia team regarding their Issue process, it would be in our (and the community's) best interest to stop assigning Issues to individual contributors (outside of the Kong org). 

### Reasons
* It puts pressure on a single individual to commit to that Issue even if they find it too complex after they start working on it
* It creates potential conflict between contributors vying to work on the same task
* It increases overhead for repo maintainers, who have to keep an eye on whether or not a PR is raised by a specific contributor for that specific Issue originally assigned to them
* Individuals outside of the Kong org also do not have the ability to assign Issues to themselves, so it puts onus on us to go through each Issue and assign it as comments appear


lena-larionova (approved):
Agreed, this is a good call! Thanks for talking with the Insomnia team and getting their advice on this.

PR #3285, opened 2021-09-29
[DOCU-1714] Guidelines for using includes

lena-larionova:
### Summary
Adding guidelines for using includes to the contributing guide, and adding links to the new instructions from both the site and include directory READMEs.

### Reason
Why the contributing guide and not something internal? Because there's no reason to hide these instructions, and if anyone runs into issues, we can send them a link to the guide.

### Testing
https://deploy-preview-3285--kongdocs.netlify.app/contributing/includes/

### DIFF
--- a/README.md
+++ b/README.md
@@ -4,23 +4,23 @@
 
 # KONG's Documentation Website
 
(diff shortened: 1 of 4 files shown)


falondarville (approved):
LGTM, I agree this should be public.

PR #3292, opened 2021-10-01
Update style-guide.md

hughbussell:
I believe that the columns were switched that tell you not to use _appears_


lena-larionova (approved):
Nice catch! Thanks for the fix, LGTM

PR #3294, opened 2021-10-02
[DOCU-1122][DOCU-1094] Migration guide from gateway to Konnect

lena-larionova:
### Summary
Guide for migrating from a self-hosted Kong Gateway setup to Konnect Cloud. 
* Convert object configurations with decK
* Add dev portal documents with decK
* Decide what to do with your data planes, then power down old ones + their datastores, if those exist
* Do all the other things manually using a bunch of other guides

Used decK to test as much as possible; the rest is manual and hard to write concrete steps for. The whole process is mostly "use deck to convert your configuration, then add everything else manually, therefore follow these other guides".

Guide based on personal testing, demos, and Michael Heap's blog: https://konghq.com/blog/migrating-kong-gateway-konnect/. 

### Reason
Need a migration guide between self-hosted and cloud deployments.

### Testing
https://deploy-preview-3294--kongdocs.netlify.app/konnect/migrate/

### DIFF
--- a/app/konnect/migrate.md
+++ b/app/konnect/migrate.md
@@ -171,7 +171,7 @@ applications registered through the Portal, those developers need to create new
 accounts in {{site.konnect_saas}} and register their applications in the new
 location.
     * [Developer registration](/konnect/dev-portal/access-and-approval/dev-reg)
-    * [Enable application registration](/konnect/dev-portal/administrators/app-registration/enable-app-reg):
+    * [Enable application registration](/konnect/dev-portal/applications/enable-app-reg):
     App registration in {{site.konnect_saas}} works through a different
     mechanism than in self-hosted {{site.base_gateway}}. Enable app
     registration on each service that requires it.
--- a/app/konnect/migrate.md
+++ b/app/konnect/migrate.md
@@ -67,17 +67,18 @@ a warning if this occurs.
     ```
 
     {:.note}
-    > You can also use decK with Konnect more securely by storing your password
-    in a file, then either calling it with `--konnect-password-file pass.txt`,
-    or adding it to your decK configuration under the `konnect-password` option.
-    See the [`deck konnect`](/deck/latest/reference/deck_konnect) reference for
-    more information about the flag.
+    > **Note:** You can also use decK with {{site.konnect_saas}} more securely
+    by storing your password in a file, then either calling it with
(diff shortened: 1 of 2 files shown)


rossmcdonald (approved):
This looks great. The only thing I think is missing is a discussion around self-hosted workspaces and how they factor into a migration plan, but that can be addressed later.

lena-larionova:
> This looks great. The only thing I think is missing is a discussion around self-hosted workspaces and how they factor into a migration plan, but that can be addressed later.

@rossmcdonald What is the current approach that you've been using? 
Right now, what comes to mind is either:
* Export config with the workspace flag and combine into one file, then convert to Konnect format. Make sure that all object names are unique; may require lots of work to convert
* Maybe convert files individually? Is this better or worse?

The closest parallel I can think of in Konnect to workspaces is service packages - if your services are small permutations of each other, then maybe we can recommend grouping them as versions? What would make sense here?

rossmcdonald:
@lena-larionova We should review with Product to see what they recommend. Regarding:

> if your services are small permutations of each other, then maybe we can recommend grouping them as versions? What would make sense here?

IMO the main driver behind separating services by workspace is RBAC. I don't think combining services into service packages will suffice with the current set of features, but Product might have another opinion.

app/konnect/migrate.md
```diff
@@ -0,0 +1,190 @@
...
+## Post-migration tasks
+
+See the following docs to set up any additional things you may need:
+
+* **Dev Portal applications and developers:** If you have developers or
+applications registered through the Portal, those developers need to create new
+accounts in {{site.konnect_saas}} and register their applications in the new
+location.
+    * [Developer registration](/konnect/dev-portal/access-and-approval/dev-reg)
+    * [Enable application registration](/konnect/dev-portal/administrators/app-registration/enable-app-reg):
```
falondarville, line 174:
```suggestion
    * [Enable application registration](/konnect/dev-portal/applications/enable-app-reg):
```

Redirect is in place, but this is the new link.

falondarville (approved):
LGTM! Very clear and concise.

lena-larionova:
@rossmcdonald re:

> We should review with Product to see what they recommend

That sounds like a good argument for your original suggestion to push these changes and look for answers about workspaces after, which is what'll I'll do. Tracking it with: https://konghq.atlassian.net/browse/DOCU-1889

PR #3296, opened 2021-10-04
[DOCU-1882]Updates versions for OPA plugin for v2.6.0.0 release

HCloward:
### Summary
For the v2.6.0.0 KG EE release, we have added request path, allowing administrators to set policies on a path basis more easily. 
### Reason
The docs need to be updated to reflect the new version.
### Testing
Will include test link when available.

### DIFF
--- a/app/_hub/kong-inc/opa/0.1.x.md
+++ b/app/_hub/kong-inc/opa/0.1.x.md
@@ -1,7 +1,7 @@
 ---
 name: OPA
 publisher: Kong Inc.
-version: 0.2.x
+version: 0.1.x
 # internal handler v 0.1.0
 
 desc: Authorize requests against Open Policy Agent
@@ -23,8 +23,6 @@ categories:
 kong_version_compatibility:
     enterprise_edition:
       compatible:
-        - 2.6.x
-        - 2.5.x
         - 2.4.x
 
 params:


app/_hub/kong-inc/opa/0.1.x.md
```diff
@@ -0,0 +1,314 @@
+---
+name: OPA
+publisher: Kong Inc.
+version: 0.1.x
+# internal handler v 0.1.0
```
hbagdi, line 5: This was bumped to 0.2.0 internally.

hbagdi: Probably also need to rename the file.

app/_hub/kong-inc/opa/0.1.x.md
```diff
@@ -0,0 +1,314 @@
...
+```
+{
+  "input": {
+    "request": { # details about the request from client to Kong
+      "http": {
+        "host": "example.org", # host header used by the client to make the request
+        "port": "8000",        # port to which the request was made
+        "tls": {},             # TLS details if the request was made on HTTPS and Kong terminated the TLS connection
+        "method": "GET",       # HTTP method used in the request
+        "scheme": "http",      # protocol used to make the request by the client, this can be either `http` or `https`
```
hbagdi, line 189: This was the enhancement in the new version:

```suggestion
        "scheme": "http",      # protocol used to make the request by the client, this can be either `http` or `https`
        "path": "/foo/bar",      # HTTP path used in the request
```

app/_hub/kong-inc/opa/0.1.x.md
HCloward: the index file is the most recent version of the plugin - and this file has the 0.2.x version info in it

hbagdi (approved):

PR #3301, opened 2021-10-05
Site maintenance (security risk): update rack dependency

falondarville:
### Summary

A security update indicated that the `rack` gem should be updated. 

This will require contributors to `make clean` and `make install` to update the gem. 

### Reason

Security. 

### Testing

Click around and see if everything functions as expected. Tested locally.
https://deploy-preview-3301--kongdocs.netlify.app


lena-larionova (approved):
LGTM, the smoke test passed (it uses `rack`) and site built fine.

PR #3302, opened 2021-10-05
[DOCU-1856][DOCU-1893] Accessible badges

lena-larionova:
### Summary
* Adding links to badges that appear in page titles, linking off to pricing page
  * Adding `aria-label` for links, following WGAC guidelines: https://www.w3.org/WAI/WCAG21/Techniques/aria/ARIA8.html
  * I considered adding tooltips but that's more of a nice-to-have and doesn't add much for accessibility
* Removing Konnect Enterprise subscription banner in favor of badges in plugins
* Replacing banners in Getting Started Guide with either badges, explanations, or simplified them so that badges aren't needed
* Labelling getting started guide dev portal and RBAC pages as enterprise

### Reason
Badges are not accessible. If someone is visually impaired or uses a screenreader, the badge gets passed over completely and they have no way of knowing it exists.

For plugins: request from Gayle for jq plugin, as we have too many banners. 

### Testing
Tested locally on plugins and getting started guide; also tested with WAVE plugin.

Check any Plus or Enterprise plugin, making sure to tab over to the badge:
https://deploy-preview-3302--kongdocs.netlify.app/hub/kong-inc/application-registration/

Sample getting started guide page:
https://deploy-preview-3302--kongdocs.netlify.app/getting-started-guide/2.6.x/manage-teams/

### DIFF
--- a/app/_assets/stylesheets/badges.less
+++ b/app/_assets/stylesheets/badges.less
@@ -1,19 +1,19 @@
 .badge {
   &:after {
-    display: inline-flex;
-    font-weight: 500;
-    font-size: 12px;
-    vertical-align: middle;
-    text-align: center;
-    max-width: 200px;
-    line-height: 14px;
-    height: 14px;
-    width: auto;
-    padding: 3px 10px;
-    font-family: @family-display;
-    border-radius: 10px;
-    margin-left: 7px;
-  }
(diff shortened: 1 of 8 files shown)


falondarville (approved):
LGTM!! LOVE this :)

PR #3321, opened 2021-10-14
Updating the admin api route path handingly strategy table

andrewgkew:
### Summary

1. Fixed a markdown syntax error that was breaking the table
2. Added some more info to table to make it clearer when you have a request path or not
3. Tested every case to ensure was correct

### Reason

Was walking through the logic with a client and found it slightly confusing, and also noticed the table broken in the official docs

### Testing

Tested this using a kong instance and setting up a service and route and calling the proxy api to test each combination in the table. Have also shared this with the client (new table) and they are happier that it makes more sense and values match what they have seen too.

### DIFF
--- a/app/enterprise/2.5.x/admin-api/index.md
+++ b/app/enterprise/2.5.x/admin-api/index.md
@@ -1442,7 +1442,7 @@ the concatenated path will be `/sre`.
 Both versions of the algorithm detect "double slashes" when combining paths, replacing them by single
 slashes.
 
-The following table shows all combinations of path handling version, strip path and request:
+The following table shows the possible combinations of path handling version, strip path, and request:
 
 | `service.path` | `route.path` | `request` |`route.strip_path` | `route.path_handling` | request path | upstream path |
 |----------------|--------------|-----------|-------------------|-----------------------|--------------|---------------|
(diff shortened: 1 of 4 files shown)


andrewgkew:
@lena-larionova - I realize that this change has been made on 2.5.x so if you need me to merge it into 2.6.x then let me know

app/enterprise/2.5.x/admin-api/index.md
```diff
@@ -1442,16 +1442,26 @@ the concatenated path will be `/sre`.
...
-|----------------|--------------|--------------------|-----------------------|--------------|---------------|
-| `/s`           | `/fv0`       | `false`            | `v0`                  | `/fv0req`    | `/s/fv0req`   |
-| `/s`           | `/fv1`       | `false`            | `v1`                  | `/fv1req`    | `/sfv1req`    |
-| `/s`           | `/tv0`       | `true`             | `v0`                  | `/tv0req`    | `/s/req`      |
-| `/s`           | `/tv1`       | `true`             | `v1`                  | `/tv1req`    | `/sreq`       |
-| `/s`           | `/fv0/`      | `false`            | `v0`                  | `/fv0/req`   | `/s/fv0/req`  |
-| `/s`           | `/fv1/`      | `false`            | `v1`                  | `/fv1/req`   | `/sfv1/req`   |
-| `/s`           | `/tv0/`      | `true`             | `v0`                  | `/tv0/req`   | `/s/req`      |
-| `/s`           | `/tv1/`      | `true`             | `v1`                  | `/tv1/req    | `/sreq`       |
+The following table shows all combinations of path handling version, strip path and request
```
lena-larionova, line 1445:
```suggestion
The following table shows all combinations of path handling version, strip path and request:
```

PR #3326, opened 2021-10-15
Update using-tcpingress.md

sigreen:
### Summary
Setup port forwarding for minikube running on macos

### Reason
Minikube does not enable an external LoadBalancer service out of the box, hence proxy address is blank when the ingress rules are created.

### Testing
Please follow the steps I listed on the doc to test this feature.

### DIFF
--- a/app/kubernetes-ingress-controller/2.0.x/guides/using-tcpingress.md
+++ b/app/kubernetes-ingress-controller/2.0.x/guides/using-tcpingress.md
@@ -39,6 +39,8 @@ Please follow one of the
 If everything is setup correctly, making a request to Kong should return
 HTTP 404 Not Found.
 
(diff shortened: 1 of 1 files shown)


app/kubernetes-ingress-controller/2.0.x/guides/using-tcpingress.md
```diff
@@ -39,6 +39,8 @@ Please follow one of the
 If everything is setup correctly, making a request to Kong should return
 HTTP 404 Not Found.
 
+> **Note**: If you are running the example using Minikube on macos, you may need to run `minikube tunnel` in a separate terminal window.  This will allow external LoadBalancer which is not enabled out of the box.
```
lena-larionova, line 42:
```suggestion
> **Note**: If you are running the example using Minikube on MacOS, you may need 
to run [`minikube tunnel`](https://minikube.sigs.k8s.io/docs/handbook/accessing/#loadbalancer-access)
in a separate terminal window.  This exposes LoadBalancer services 
externally, which is not enabled by default.
```

PR #3340, opened 2021-10-20
Version support: Add entry for 2.6

lena-larionova:
### Summary
Adding 2.6 to the version support policy table.

### Reason
2.6 went out 10/14.

Does not need to be run by PM; PM has already set a documented support period for 2.x, and the only number that should change is the release date.

### Testing
TBA


falondarville (approved):
LGTM

PR #3373, opened 2021-11-11
doc(router) add docs for TLS passthrough

fffonion:
### Summary
Add doc for TLS passthrough routing, a new feature being added to 2.7.

### Reason
The original PR is at https://github.com/Kong/kong/pull/6757.

### DIFF
--- a/app/gateway/2.7.x/admin-api/index.md
+++ b/app/gateway/2.7.x/admin-api/index.md
@@ -1659,8 +1659,11 @@ following attributes must be set:
 * For `https`, at least one of `methods`, `hosts`, `headers`, `paths` or `snis`;
 * For `tcp`, at least one of `sources` or `destinations`;
 * For `tls`, at least one of `sources`, `destinations` or `snis`;
+* For `tls_passthrough`, must set `snis`;
 * For `grpc`, at least one of `hosts`, `headers` or `paths`;
 * For `grpcs`, at least one of `hosts`, `headers`, `paths` or `snis`.
+
+A route can't have both `tls` and `tls_passthrough` in protocols at same time.
 
 #### Path handling algorithms
 
--- a/app/gateway/2.7.x/reference/proxy.md
+++ b/app/gateway/2.7.x/reference/proxy.md
@@ -1325,6 +1325,33 @@ This means all of the below setup are supported in this mode:
 protocol list are supported. This list can be found in their respective documentation
 on [Kong Hub](https://docs.konghq.com/hub/).
 
+[Back to top](#introduction)
+
+## Proxy TLS passthrough traffic
+
+Starting with version 2.7, Kong supports proxying a TLS request without terminating or known
+as SNI proxy.
+
+Kong will not decrypt the incoming TLS traffic, but use the connecting client's
+TLS SNI server name extension to find the matching Route and Service and forward the
+complete TLS request to upstream.
+
+In this mode, request must be sent to a port that has `ssl` flag in its `stream_listen` directive.
+Aside from defining `stream_listen`, appropriate Route
+object with protocol types of `tls_passthrough` and `snis` field should be created, the corresponding
+Service object must have protocol of `tcp`. Different from proxying TLS traffic,
+SNI and Certificate entity is not required to be set and will not be used.
+
+Routes are not allowed to match `tls` and `tls_passthrough` protocols at same time; but same
+SNI is allowed to match `tls` and `tls_passthrough` in different Routes. It's possible to set
+Route to `tls_passthrough` and Service to `tls`, in this mode the connection to upstream will
+be TLS encrypted twice.
+
+**Note:** To allow plugins run in this mode, it must have protocols field explictly
+include `tls_passthrough`.
+
+[Back to top](#introduction)
+
 ## Conclusion
 
 Through this guide, we hope you gained knowledge of the underlying proxying
--- a/app/gateway/2.7.x/reference/proxy.md
+++ b/app/gateway/2.7.x/reference/proxy.md
@@ -1329,25 +1329,31 @@ on [Kong Hub](https://docs.konghq.com/hub/).
 
 ## Proxy TLS passthrough traffic
 
-Starting with version 2.7, Kong supports proxying a TLS request without terminating or known
+{{site.base_gateway}} supports proxying a TLS request without terminating or known
 as SNI proxy.
 
-Kong will not decrypt the incoming TLS traffic, but use the connecting client's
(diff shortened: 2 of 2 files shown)


lena-larionova:
@fffonion As of this week, we have moved over all gateway docs into `/gateway/`. Can you please update your branch and make these changes in the `/gateway/2.7.x/` directory instead?

fffonion:
@lena-larionova Just did : )

app/gateway/2.7.x/reference/proxy.md
```diff
@@ -1325,6 +1325,33 @@ This means all of the below setup are supported in this mode:
...
+
+Kong will not decrypt the incoming TLS traffic, but use the connecting client's
+TLS SNI server name extension to find the matching Route and Service and forward the
+complete TLS request to upstream.
+
+In this mode, request must be sent to a port that has `ssl` flag in its `stream_listen` directive.
+Aside from defining `stream_listen`, appropriate Route
+object with protocol types of `tls_passthrough` and `snis` field should be created, the corresponding
+Service object must have protocol of `tcp`. Different from proxying TLS traffic,
+SNI and Certificate entity is not required to be set and will not be used.
```
lena-larionova, line 1343:
```suggestion
In this mode, you need to:
* Create a Route object with the protocol `tls_passthrough`, and the 
`snis` field set to one or more SNIs.
* Set the corresponding Service object's protocol to `tcp`. 
* Send requests to a port that has the `ssl` flag in its `stream_listen` 
directive.

Separate SNI and Certificate entities aren't required and won't be used.
```

I moved the `stream_listen` point to the end of the list, as it seems like this is the order of operations - create Service & Route, then send requests.

app/gateway/2.7.x/reference/proxy.md
```diff
@@ -1325,6 +1325,33 @@ This means all of the below setup are supported in this mode:
...
+
+Routes are not allowed to match `tls` and `tls_passthrough` protocols at same time; but same
+SNI is allowed to match `tls` and `tls_passthrough` in different Routes. It's possible to set
+Route to `tls_passthrough` and Service to `tls`, in this mode the connection to upstream will
+be TLS encrypted twice.
```
lena-larionova, line 1348:
```suggestion
Routes are not allowed to match `tls` and `tls_passthrough` protocols at same time. 
However, the same SNI is allowed to match `tls` and `tls_passthrough` in different 
Routes. 

It's also possible to set Route to `tls_passthrough` and Service to `tls`. In this 
mode, the connection to the upstream will be TLS-encrypted twice.
```

app/gateway/2.7.x/reference/proxy.md
```diff
@@ -1325,6 +1325,33 @@ This means all of the below setup are supported in this mode:
...
+
+**Note:** To allow plugins run in this mode, it must have protocols field explictly
+include `tls_passthrough`.
```
lena-larionova, line 1351:
```suggestion
{:.note}
> **Note:**  To run any plugins in this mode, the plugin's `protocols` field must
include `tls_passthrough`.
```

Question: does every plugin that currently supports TLS also support TLS passthrough?

app/gateway/2.7.x/admin-api/index.md
```diff
@@ -1659,9 +1659,12 @@ following attributes must be set:
 * For `https`, at least one of `methods`, `hosts`, `headers`, `paths` or `snis`;
 * For `tcp`, at least one of `sources` or `destinations`;
 * For `tls`, at least one of `sources`, `destinations` or `snis`;
+* For `tls_passthrough`, must set `snis`;
```
lena-larionova, line 1662: This file is generated from the code repo, please make the changes here instead: https://github.com/Kong/kong/blob/master/autodoc/admin-api/data/admin-api.lua#L857

app/gateway/2.7.x/reference/proxy.md
```diff
@@ -1325,6 +1325,33 @@ This means all of the below setup are supported in this mode:
 protocol list are supported. This list can be found in their respective documentation
 on [Kong Hub](https://docs.konghq.com/hub/).
 
+[Back to top](#introduction)
+
+## Proxy TLS passthrough traffic
+
+Starting with version 2.7, Kong supports proxying a TLS request without terminating or known
+as SNI proxy.
```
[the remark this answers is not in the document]
fffonion: The segment is bit confusing here, I was trying to say `proxying a TLS request without terminating` this feature is
also known as "SNI proxy", may need some help on reworkding here.

app/gateway/2.7.x/reference/proxy.md
fffonion: In most cases I see, plugins that support stream subsystem either declare it supports "all protocols", which include tls_passthrough and tls for the same time, but I seldomly see a plugin declare it only supports `tls`. And this paragraph is noting that for the later case, plugin author will need to manually add `tls_passthrough` otherwise the plugin won't run.

app/gateway/2.7.x/admin-api/index.md
fffonion: Opened https://github.com/Kong/kong/pull/8099

app/gateway/2.7.x/reference/proxy.md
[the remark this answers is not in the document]
lena-larionova: Took a stab at rephrasing, let me know if that sounds correct to you (see suggestion above).

app/gateway/2.7.x/reference/proxy.md
lena-larionova: We currently list out the protocols that our plugins support in the plugin docs. Does each plugin that supports TLS also need to have `tls_passthrough` added to the list? For example, see: https://docs.konghq.com/hub/kong-inc/acme/

fffonion: Not really, I don't acme plugin supports tcp/tls at first, since it's declared to support http/https/grpc/grpcs only (https://github.com/Kong/kong/blob/master/kong/plugins/acme/schema.lua#L58).
Only if a plugin declares it supports all protocols like `file-log` like
https://github.com/Kong/kong/blob/master/kong/plugins/file-log/schema.lua#L7 we should add the
`tls_passthrough` to its list.

lena-larionova: Manually maintained for now, unfortunately :( We're working on changing that for Q1/Q2 next year though. I'll grep the plugin files to locate all `typedefs.protocols` and add them there; will ping you on the PR when it's ready.

dndx (approved):

PR #3378, opened 2021-11-17
doc(kic) remove CREATE from Secret hook

rainest:
### Summary
Remove CREATE Secret from the admission webhook watched actions. Due to 
changes in implementation, this is no longer used. We only check Secret 
UPDATEs.

### Reason
This is tied to the changes in https://github.com/Kong/kubernetes-ingress-controller/pull/2015. It should not be merged until that is actually released (not just merged). This will remain a draft until we're ready to release.


rainest:
Superseded by https://github.com/Kong/docs.konghq.com/pull/3518

PR #3380, opened 2021-11-17
Fix "self-hosted" terminology

lena-larionova:
### Summary
* Changing instances of "self-hosted" to "self-managed"
* Removing some "self-hosted" labels from the landing page altogether, as they are adding unnecessary complexity

### Reason
"Self-hosted" is not the correct term, it should be "self-managed". This has been set for a while, but we've never gone through to docs to fix this.

### Testing
Main changes are on the landing page: https://deploy-preview-3380--kongdocs.netlify.app/


falondarville (approved):
LGTM

PR #3387, opened 2021-11-18
Chore/adjust installation instructions

hutchic:
### Summary
Our installation instructions are kind of part way between the old OSS install instructions and the old enterprise installation instructions. For example we're doing `rpm --eval` in spots that don't do anything

The table of contents is a good example of the change

Before:
image

After:
image

### Reason
simplifying docs so folks don't need to choose their release distribution

### Testing
do a `docker run -it --rm x:y` and run through the installation instructions ommitting any `sudo`'s

Before
image

After
image


lena-larionova:
Added nested tabs for EE/OSS. 

For RHEL, there were a bunch of copy-pasta errors - please check that I fixed them correctly.

hutchic (commented):
I validated the installation instructions by running them in docker containers and feel good with this change 👍

lena-larionova (approved):
LGTM, thanks for putting this together!

PR #3390, opened 2021-11-18
[DOCU-2038] Improvements to the Gateway navigation

lena-larionova:
### Summary

* Move the Admin API to the top of the Reference section. It was pointed out that it’s one of our most frequently used docs and should be easier to find.
* Removing "Develop" section, since you don't actually develop anything in the Gateway itself (except maybe custom plugins). This section was a placeholder from the start of our project and shouldn't have been left in. Dev Portal doesn't actually belong here - these aren’t docs for developing, they’re docs for configuring the Dev Portal. 
* Moving Dev Portal and GraphQL to Configure section.
* Adding entries for Prometheus and Zipkin plugins in the Monitor section.

### Reason
Requests/feedback following single-sourcing of the Gateway doc. See https://konghq.atlassian.net/browse/DOCU-2038 for detail.

### Testing
See left-side navigation on https://deploy-preview-3390--kongdocs.netlify.app/gateway/


GayleNeumann (approved):
LGTM

PR #3391, opened 2021-11-19
doc(hybrid) add warning for intermediate proxy

rainest:
### Summary
Add a warning to the TLS mode section instructing users to use PKI mode if they place an intermediate TLS-aware proxy between CP and DP nodes.

### Reason
Hybrid shared mTLS forcibly sets SNI to `kong_clustering`. Intermediate TLS proxies that use SNI to route upstream will almost certainly not recognize this and will misroute traffic or present the wrong certificate. PKI mode allows explicitly setting SNI to the CP hostname, which should be recognized if the intermediate is properly configured.

We do not currently set SNI by default in PKI mode (we probably should), so this instructs users to set that config value also.

### Testing
Await Netlify to see if I broke the warning box.

### DIFF
--- a/app/gateway/2.6.x/plan-and-deploy/hybrid-mode/hybrid-mode-setup.md
+++ b/app/gateway/2.6.x/plan-and-deploy/hybrid-mode/hybrid-mode-setup.md
@@ -31,6 +31,13 @@ by both CP and DP nodes.
 (CA). Kong validates both sides by checking if they are from the same CA. This
 eliminates the risks associated with transporting private keys.
 
+{:.warning}
+> **Warning:** if you have a TLS-aware proxy between the DP and CP nodes, you
+must use PKI mode and set `cluster_server_name` to the CP hostname in
+kong.conf. Shared mode uses a non-standard value for TLS server name
+indication, and this will confuse TLS-aware proxies that rely on SNI to route
+traffic.
+
 For a breakdown of the properties used by these modes, see the
 [configuration reference](#configuration-reference).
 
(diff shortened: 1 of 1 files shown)


MosAmokhtari (approved):

app/gateway/2.6.x/plan-and-deploy/hybrid-mode/hybrid-mode-setup.md
```diff
@@ -31,6 +31,13 @@ by both CP and DP nodes.
 (CA). Kong validates both sides by checking if they are from the same CA. This
 eliminates the risks associated with transporting private keys.
 
+{:.warning}
+> **Warning:** if you have a TLS-aware proxy between the DP and CP nodes, you
```
lena-larionova, line 35:
```suggestion
> **Warning:** If you have a TLS-aware proxy between the DP and CP nodes, you
```

app/gateway/2.6.x/plan-and-deploy/hybrid-mode/hybrid-mode-setup.md
```diff
@@ -31,6 +31,13 @@ by both CP and DP nodes.
...
+must use PKI mode and set `cluster_server_name` to the CP hostname in
+kong.conf. Shared mode uses a non-standard value for TLS server name
```
lena-larionova, line 37:
```suggestion
`kong.conf`. Do not use shared mode, as it uses a non-standard value for TLS server name
```

lena-larionova (approved):
Minor edits, otherwise LGTM!

PR #3392, opened 2021-11-19
docs(mesh/rbac) add note about LOCALHOST_IS_ADMIN for example

michaelbeaumont:
Otherwise the example doesn't show what it's supposed to and the final request won't be denied.


Bradamant3 (approved):
LGTM&TYSM (line edits to come in general collection of fixes)

PR #3394, opened 2021-11-19
Remove extra correlation-id entry

lena-larionova:
### Summary
Removing one extra line from the plugin priority table.

### Reason
The OSS/Free mode table should have two entries for correlation ID, but the Enterprise table should only have one. Accidentally added two entries to both.

### Testing
TBA


falondarville (approved):
LGTM

PR #3401, opened 2021-11-22
[TDX-1631] Dev Portal: Add showExtensions option

falondarville:
### Summary

This change applies to self-managed Dev Portal. In "Working with Templates", I added a note that you can add the property `showExtensions` and set it to `true` to show custom properties. 

### Reason

Closes [TDX-1631](https://konghq.atlassian.net/browse/TDX-1631) (not part of 2.7)

### Testing

https://deploy-preview-3401--kongdocs.netlify.app/gateway/2.6.x/developer-portal/working-with-templates/#show-custom-properties

### DIFF
--- a/app/gateway/2.6.x/developer-portal/working-with-templates.md
+++ b/app/gateway/2.6.x/developer-portal/working-with-templates.md
@@ -20,6 +20,10 @@ You may use the following tags in templates:
 * `{# comments #}` everything between `{#` and `#}` is considered to be commented out (i.e., not outputted or executed).
 {% endraw %}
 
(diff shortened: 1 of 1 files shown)


lena-larionova (approved):
LGTM!

davidma415:
I'm just curious if we should make a note that it's already in the spec-renderer.html, but they just have to change it to true. Besides that, looks good!

falondarville:
> I'm just curious if we should make a note that it's already in the spec-renderer.html, but they just have to change it to true. Besides that, looks good!

Yes, I will add that information!

PR #3404, opened 2021-11-23
Update docker.md

rkazak:
Added DB less example for EE edition.  Also no mention here about license setup, or the differences between docker on OSX, as this looks like it's written from the linux point of view? Also maybe we need the same for windows? ( I do not have a windows system so I cannot test that ).


falondarville:
@rkazak so sorry, I messed up the merge from fork when I was making formatting edits. I have cherry-picked onto a new branch. Thank you so much for testing and for submitting this PR. 

Reference: https://github.com/Kong/docs.konghq.com/pull/3419

PR #3414, opened 2021-11-23
Incorrect YAML indentation

johnfitzpatrick:
### Summary
Incorrect YAML indentation in example

### Reason
Should be the following

```
    $ CSR=$(cat backend-owner.csr | base64 | tr -d "\n") && echo "apiVersion: certificates.k8s.io/v1
    kind: CertificateSigningRequest
    metadata:
      name: backend-owner
    spec:
      request: $CSR
      signerName: kubernetes.io/kube-apiserver-client
      usages:
      - client auth" | kubectl apply -f -
```


falondarville (approved):
LGTM, thank you!

PR #3416, opened 2021-11-23
Generate combined gateway versions

lena-larionova:
### Summary
Add list of `site.data.kong_versions_gateway` versions to the plugin doc. Currently, the doc isn't picking up 2.6.x because it's only generating from the old split CE/EE lists.

### Reason
Currently, the doc isn't picking up 2.6.x because it's only generating from the old split CE/EE lists. 
For example, this plugin is compatible with 2.6.x but the live doc has no versions in the compatibility list (bottom of right sidebar): https://docs.konghq.com/hub/kong-inc/jq/, even though the [source file](https://github.com/Kong/docs.konghq.com/blame/main/app/_hub/kong-inc/jq/index.md#L37) includes the version.

### Testing
All Kong plugins should now have 2.6.x in the sidebar. For example, see https://deploy-preview-3416--kongdocs.netlify.app/hub/kong-inc/jq/, or https://deploy-preview-3416--kongdocs.netlify.app/hub/kong-inc/grpc-web/ for both OSS and EE entries.

### DIFF
--- a/app/_layouts/extension.html
+++ b/app/_layouts/extension.html
@@ -87,12 +87,10 @@ breadcrumbs:
 {% if page.kong_version_compatibility.community_edition %}
   {% assign compat = true %}
   {% assign ce_compatible = page.kong_version_compatibility.community_edition.compatible %}
-  {% assign ce_incompatible = page.kong_version_compatibility.community_edition.incompatible %}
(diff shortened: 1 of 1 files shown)


falondarville (approved):
LGTM

PR #3420, opened 2021-11-24
docs(installation): nuke amazonlinux 1 and jessie install links

hutchic:
### Summary

Amazon Linux 2 was released in 2017 and Amazon Linux 1 EOL including an extention was December 30, 2020

Debian Stretch was released in 2017 and Jessie was EOL was June 30, 2020 (ELTS is 2022).

### Reason

This doesn't change that we release packages for these releases and folks will be able to interprit where to locate them
based on the installation instructions it simply doesn't actively document how to download / install them.

Arguably we should stop doing new releases to EOL releases and only to patch / security but that's a product question I won't explore right now


lena-larionova (approved):
LGTM! Checked the remaining links, everything works.

PR #3421, opened 2021-11-24
chore(installation): pin the apt and yum repository version explicitely

hutchic:
### Summary

Install an explicit version when installing from our yum and apt repository for verbosity and anti-failure

### Reason
More verbose is better. Folks that are fluent in yum / apt will know the package version isn't explicitely required but having it by default is better than not.

Ensures that the customer knows exactly what package is being installed


lena-larionova (approved):
LGTM, versions all print correctly in preview.

PR #3441, opened 2021-12-01
docs(kmesh) expiration and revocation of cp token

jakubdyszkiewicz:
### Summary

Add information about expiration and revocation


app/mesh/1.4.x/features/kds-auth.md
```diff
@@ -161,9 +161,43 @@ Verify the remote control plane is connected with authentication by looking at t
...
+type: GlobalSecret
+name: control-plane-token-revocations
+data: {{ revocations }}" | kumactl apply --var revocations=$(echo '0e120ec9-6b42-495d-9758-07b59fe86fb9' | base64) -f -
+```
+{% endnavtab %}
+{% endnavtabs %}
+
+## Rotate signing key
+
+If a signing key is compromised, you must rotate it and all the tokens.
```
mheap, line 200: If a signing key is comprimised, would that automatically invalidate all tokens once rotated, or do you also have to add all of those tokens to the revocation list?

jakubdyszkiewicz: Once you remove the compromised signing key, all tokens signed with this signing key are no longer valid so you don't have to manually add them to the revocation list.

See `Remove the old signing key` - `All new connections to the global control plane now require tokens signed with the new signing key`

Bradamant3 (approved):
LGTM, will edit lightly after merge

PR #3446, opened 2021-12-02
[DOCU-2045] Banner for older versions of plugins

lena-larionova:
### Summary
Add a banner to older plugin versions. 
Removing non-existent `header_caption` field, it hasn't (ever?) been in use.

Note: The yellow is a bit different form our normal yellow banner tone. It's the new Kongponents yellow; I'll be updating all docs to use the new tones in a future PR.

### Reason
See ticket; we have no visual distinction for older plugin docs, so it's easy to accidentally end up on an old page.

### Testing
Open any kong-inc plugin doc and check latest version vs older version, for example:
* Latest: https://deploy-preview-3446--kongdocs.netlify.app/hub/kong-inc/application-registration/
* Older (see banner, check link): https://deploy-preview-3446--kongdocs.netlify.app/hub/kong-inc/application-registration/1.0.x.html

### DIFF
--- a/app/_assets/stylesheets/pages/extension.less
+++ b/app/_assets/stylesheets/pages/extension.less
@@ -322,3 +322,18 @@
     }
   }
 }
+
+.old-version-banner {
+  padding: 10px 20px;
+  display: block;
+  background-color: #fff3d8;
+  border: none;
+  border-radius: 4px;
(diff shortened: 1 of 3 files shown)


GayleNeumann (approved):
Nice! LGTM

PR #3447, opened 2021-12-02
[DOCU-2046] Fix PDF gen (post single-sourcing)

lena-larionova:
### Summary
* Adding `gateway` to list of mapped paths
* Setting versioned URLs for old `/enterprise/` and `/gateway-oss/` landing pages
This should fix PDF generation.

### Reason
Fixes two different errors:
* https://github.com/Kong/docs.konghq.com/runs/4257695434?check_suite_focus=true#step:10:18
  Didn't add `gateway` to list of versions, so it didn't recognize the value
  
* https://github.com/Kong/docs.konghq.com/runs/4257745507?check_suite_focus=true#step:10:17
  The `/enterprise/` and `/gateway-oss/` aliases no longer exist without versions, because the latest version is `/gateway/`. On a live doc, this would be a redirect; on a local build (or a PDF build), it reads as a broken link, so the page can't be found.

### Testing
N/A, have to merge and then test.

### DIFF
--- a/app/_data/docs_nav_ce_2.4.x.yml
+++ b/app/_data/docs_nav_ce_2.4.x.yml
@@ -1,7 +1,7 @@
 # Generated via autodoc-nav
 - title: Introduction
   icon: /assets/images/icons/documentation/icn-flag.svg
-  url: /gateway-oss/
+  url: /gateway-oss/2.4.x/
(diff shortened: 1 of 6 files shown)


GayleNeumann (approved):
LGTM

PR #3448, opened 2021-12-02
[DOCU-2071] Fix 404 plus related things

Bradamant3:
### Summary
Fix 404 link and related tasks

### Reason
Single-sourcing followup fixes, https://konghq.atlassian.net/browse/DOCU-2071

### Testing
Tested locally, will add Netlify link when ready


app/_redirects
```diff
@@ -1,6 +1,7 @@
-# FREE TRIALS
+# FREE TRIALS, GENERAL LICENSING
 /free-trial/guide/                    https://konghq.com/install
-
+/gateway/plan-and-deploy/licenses     https://konghq.com/pricing/
```
lena-larionova, line 3:
```suggestion
/gateway/plan-and-deploy/licenses     /gateway/latest/plan-and-deploy/licenses
```
This topic does still exist, and there's more info on it than just the plans&comparison: https://docs.konghq.com/gateway/2.6.x/plan-and-deploy/licenses/

lena-larionova (commented):
Can you also remove the plans page from the Konnect Platform nav? 
https://deploy-preview-3448--kongdocs.netlify.app/konnect-platform/

Bradamant3:
welp, more errors in initial commits than either of us realized :-( . PTAL once again @lena-larionova

The original link fix should have been to clean up the path (`latest` was missing, that was all). Managed to miss that one altogether until now, plus the things you caught. Removed the redirect for the 404 fix bc looks as though that was syntax error introduced with single-sourcing, therefore redirect not needed. Can add back in if you think we need.

lena-larionova (approved):
LGTM! Agree with the latest edits - more useful to have a link to licenses, and we don't need a redirect for a syntax error.

PR #3459, opened 2021-12-04
[DOCU-2079] Gateway request count report

lena-larionova:
### Summary

Documentation for the `/license/report` endpoint.
* Adding entries into the existing Admin API `/licenses/` doc: reference entry with breakdown of the fields and what they mean, and an entry in the example doc.
  * The example doc entry seems arbitrary - it's the exact same information, just formatted differently. I don't know that it provides much value, but don't want to address that in this PR - it's a bigger problem.
  * Switched the order of a couple of other entries to place the two "List" actions together.

### Reason
Documentation for https://github.com/Kong/kong-ee/pull/2761.

### Testing
[Reference](https://deploy-preview-3459--kongdocs.netlify.app/gateway/2.7.x/admin-api/licenses/reference/)
[Examples](https://deploy-preview-3459--kongdocs.netlify.app/gateway/2.7.x/admin-api/licenses/examples/)


lena-larionova:
@locao I've added a doc draft for your request count feature for 2.7, but I'm concerned: I added the doc to the existing [Licenses Admin API page](https://deploy-preview-3459--kongdocs.netlify.app/gateway/2.7.x/admin-api/licenses/reference/), but the endpoint you've added is off of `/license/`, not `/licenses/`. Does it need to have its own doc page? Is the similarity of those endpoint names going to be confusing?

locao:
> Does it need to have its own doc page? 

It does have its own doc page: https://deploy-preview-3459--kongdocs.netlify.app/gateway/2.7.x/plan-and-deploy/licenses/report/. I will add the changes to this page, sorry for overlooking it.

lena-larionova:
@locao it seems odd that we didn't previously have a page for this endpoint in the Admin API doc. I think we should still keep it where I added, if you agree? As well as the page that you updated.

app/gateway/2.7.x/admin-api/licenses/reference.md
```diff
@@ -161,41 +184,95 @@ HTTP 200 OK
...
-  "updated_at": 1500508800
-}
-```
+## Generate a report
 
-### Delete a License
+Generate a report on the Kong Gateway instance to gather monthly usage data.
 
-**Endpoint**
+<div class="endpoint get">/licenses/report</div>
```
locao, line 205:
```suggestion
<div class="endpoint get">/license/report</div>
```

locao: I think it is not super-odd to not have this documented before, this endpoint was used only when Kong representatives asked customers to send their usage report if I understood correctly. But as this is available we should document it and I agree with you, let's keep it with the other license manipulation docs (with this small fix).

locao (changes requested):

locao (approved):

GayleNeumann (approved):
LGTM

PR #3461, opened 2021-12-06
Kong Mesh: restricted RBAC on zone CP

jakubdyszkiewicz:
RBAC on Zone CP will be restricted by default in the next minor version of Kong Mesh


Bradamant3 (approved):
LGTM

PR #3470, opened 2021-12-07
Autodocs update

mashapedeployment:


lena-larionova:
For anyone looking at this PR: I reopened the test PR, as all of the content that's needed is already here. Needed to make a bunch of tweaks (we didn't test properly when single-sourcing); will make most of these changes to the code after 2.7 goes out.

PR #3472, opened 2021-12-08
[DOCU-2082] Enterprise install & download instructions for Debian

lena-larionova:
### Summary
Adding download and install instructions to the Debian doc for the `kong-enterprise-edition` package.

### Reason
Support for Debian in 2.7.

### Testing
Preview: https://deploy-preview-3472--kongdocs.netlify.app/gateway/2.7.x/install-and-run/debian/

### DIFF
--- a/app/gateway/2.7.x/install-and-run/debian.md
+++ b/app/gateway/2.7.x/install-and-run/debian.md
@@ -5,7 +5,9 @@ title: Install Kong Gateway on Debian
 > Download the latest {{page.kong_version}} package for Debian:
 >
 > * **Kong Gateway**:
-> [**10 Buster**]({{ site.links.download }}/gateway-2.x-debian-buster/pool/all/k/kong-enterprise-edition/kong-enterprise-edition_{{page.kong_versions[page.version-index].ee-version}}_amd64.deb){:.install-link}
+> [**8 Jessie**]({{ site.links.download }}/gateway-2.x-debian-jessie/pool/all/k/kong-enterprise-edition/kong-enterprise-edition_{{page.kong_versions[page.version-index].ee-version}}_amd64.deb){:.install-link},
+> [**9 Stretch**]({{ site.links.download }}/gateway-2.x-debian-stretch/pool/all/k/kong-enterprise-edition/kong-enterprise-edition_{{page.kong_versions[page.version-index].ee-version}}_amd64.deb){:.install-link},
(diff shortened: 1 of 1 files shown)


app/gateway/2.7.x/install-and-run/debian.md
```diff
@@ -19,12 +25,15 @@ badge: oss
...
+The {{site.base_gateway}} software is governed by the
+[Kong Software License Agreement](https://konghq.com/kongsoftwarelicense/).
 {{site.ce_product_name}} is licensed under an
 [Apache 2.0 license](https://github.com/Kong/kong/blob/master/LICENSE).
 
 ## Prerequisites
 
-You have a supported system with root or [root-equivalent](/gateway/{{page.kong_version}}/plan-and-deploy/kong-user) access.
+* A supported system with root or [root-equivalent](/gateway/{{page.kong_version}}/plan-and-deploy/kong-user) access.
+* (Enterprise only) A `license.json` file from Kong.
```
flrgh, line 37: Minor note: in order to run the installation commands below, there's a prereq that these packages are installed:

* `curl`
* `lsb-release`
* `apt-transport-https` (if installing the apt repository)

They are insanely common packages, so most folks should have them installed already (though the base images from docker hub did not), and in my opinion it is reasonable to assume that the average Kong user will be able to figure out very quickly how to install them if they are missing.

I'm probably 50/50 on if these needs to be included as an extra step. I'll leave it up to you on what balance to strike between terse and verbose.

flrgh (approved):
:+1: looks good to me! I added a note about extra packages that might be needed at install time, but no changes needed there unless you feel like it.

app/gateway/2.7.x/install-and-run/debian.md
flrgh: If you _do_ decide to add this, it should be inserted as step 1:

### Package

1. Install dependencies

```
sudo apt-get install curl lsb-release
```


### APT Repository

1. Install dependencies

```
sudo apt-get install curl lsb-release apt-transport-https
```

lena-larionova: Good callout. If anything, we should probably add these to _all_ of our install docs - one of the outstanding tasks we have is to actually write useful prereqs, and you just did some of the work for us 😄

lena-larionova: I'll include it as prerequisites instead of adding a step. Otherwise we can easily get into muddled territory about what we should be documenting and what we shouldn't.

falondarville (approved):
LGTM

PR #3473, opened 2021-12-08
Update docker.md to start a container with Kong Gateway

hazem-elsayed:
the PG hostname should point to the PG_Host_Name as it is confusing to say "DATABASE_NAME"

### Summary
while running the Kong Gateway it is supposed to read the correct hostname been configured at the previous steps and not the DB name


mheap (approved):

mheap:
@lena-larionova One to port to 2.7 too?

Issue #3487, opened 2021-12-14
404 on link selection!


rkazak:
### Where is the problem?

https://docs.konghq.com/gateway/2.6.x/get-started/comprehensive/dev-portal/

### What happened?

Selecting the link below ( "Managing Access" ) returns a 404. 

```
https://docs.konghq.com/gateway/2.6.x/developer-portal/administration/
```

### What did you expect to happen?

get the requested page.

### Code of Conduct and Community Expectations

- [X] I agree to follow this project's Code of Conduct
- [X] I agree to abide by the Community Expectations

rkazak:
This also happens with the link "Configuring the Dev Portal",
```
https://docs.konghq.com/gateway/2.6.x/developer-portal/configuration/
```

falondarville:
Thank you, I will look into this!

PR #3490, opened 2021-12-14
[DOCU-2101] Add Dev Portal redirects

falondarville:
### Summary

Closes [3487](https://github.com/Kong/docs.konghq.com/issues/3487)

Add two redirects for links that go to 404 from this page: https://docs.konghq.com/gateway/2.6.x/get-started/comprehensive/dev-portal/

Note that Administration and Configuration are menu items that are clicked to open sub-menus and not pages themselves. So I linked instead to a page within those sections.


lena-larionova:
For these, we should fix the links, not just add redirects. We have control over the source of the broken links, so we shouldn't leave them sitting around.

lena-larionova (approved):

PR #3498, opened 2021-12-16
[bug fix] Let right ToC wrap instead of cutting off

lena-larionova:
### Summary
Fixing nav overflow issue.

### Reason
Nav gets cut off if the text is too long and has no whitespace breaks:

Screen Shot 2021-12-15 at 5 50 34 PM

### Testing
Check that nav entries wrap instead of cutting off: https://deploy-preview-3498--kongdocs.netlify.app/gateway/2.6.x/pdk/kong.service.response/


falondarville (approved):
LGTM. This is awesome!

PR #3503, opened 2021-12-17
fix: Load cookie policy only on prod

lena-larionova:
### Summary
Disabling cookie policy in local (`development`) and `preview` environments. 

### Reason
OneTrust only applies to the `konghq.com` domain, so it doesn't register in dev and preview environments, and the policy banner loads every time you load a page. This is unnecessary and also very annoying when testing.

### Testing
No cookie policy banner in the preview: https://deploy-preview-3503--kongdocs.netlify.app/

If we ever need to test something with the cookie policy banner in the future, will need to comment out `{% if jekyll.environment == "production" %}` and `{% endif %}` tags in the `head.html` include.

### DIFF
--- a/app/_includes/head.html
+++ b/app/_includes/head.html
@@ -6,7 +6,8 @@ site.title }}{% endcapture %} {% else %} {% capture page_title %}{% if page.path
 {% endunless %}
 
 <head>
-  
+
+  {% if jekyll.environment == "production" %}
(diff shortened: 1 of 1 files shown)


falondarville (commented):
LGTM. Great idea

falondarville (approved):

PR #3508, opened 2021-12-18
docs(oidc-group-mapping) explain group claim value format and include an example

liyangau:
### Summary
In order to use auto role assigning, users need to know the format of the group claim value.

### Reason
There is no example on the doc at the moment.


app/gateway/2.7.x/configure/auth/kong-manager/oidc-mapping.md
```diff
@@ -37,9 +37,13 @@ to identify which admin value and role name to map from the IdP to {{site.base_g
...
-The `authenticated_groups_claim` value specifies which IdP claim should be assigned to the
-specified {{site.base_gateway}} user. The exact value depends on your IdP -- for example, Okta 
-configures claims for `groups`, and other IdPs might configure them as `roles`.
+The `authenticated_groups_claim` value specifies which IdP claim should be used to assign {{site.base_gateway}} roles to the
+specified {{site.base_gateway}} admin. 
+
+- The group claim name depends on your IdP -- for example, Okta configures claims for `groups`, and other IdPs might configure them as `roles`. 
+- The group claim value need to follow the format `<workspace_name>:<role_name>` for dynamic role assigning. 
+
+For example, if we set `"authenticated_groups_claim": ["groups"]` and `groups:["default:super-admin"]` is returned on admins' id tokens, these admins gets `super-admin` role in `default` workspace.
```
Bradamant3, line 46: @liyangau I think there's some confusion here (which I've also helped create in the original). The _value_ of `authenticated_groups_claim` needs to be something like `groups` (for Okta) or `roles` for a different IdP.

But if I'm understanding correctly, the `<workspace_name>:<role_name>` format is required in the IdP, not in the config for Kong Manager. I ignored this fact, and your changes don't make it quite clear. I can edit for clarity, but can you verify my understanding first? Thanks so much, and apologies for the original confusion!

liyangau: Hi Jennifer,

Thanks for taking a look. Your understanding is correct, `<workspace_name>:<role_name>` needs to be the value set on the IdP side. 

I though ```groups:["default:super-admin"]` is returned on admins' id tokens```  +  `The group claim value need to follow the format <workspace_name>:<role_name> for dynamic role assigning.`  are sufficient to suggests that `<workspace_name>:<role_name>` need to be set on IdP side and returned in `id_token` but maybe it is not clear enough?

I am happy for you to change any wording to make it clearer.

The main purposed of this PR is to let the customer know what they need to set on the IdP side in order to use dynamic role assigning. 

Thank you.

app/gateway/2.7.x/configure/auth/kong-manager/oidc-mapping.md
```diff
@@ -37,9 +37,13 @@ to identify which admin value and role name to map from the IdP to {{site.base_g
...
 The username and password are required for the user to log into the IdP.
 
-The `authenticated_groups_claim` value specifies which IdP claim should be assigned to the
-specified {{site.base_gateway}} user. The exact value depends on your IdP -- for example, Okta 
-configures claims for `groups`, and other IdPs might configure them as `roles`.
+The `authenticated_groups_claim` value specifies which IdP claim should be used to assign {{site.base_gateway}} roles to the
+specified {{site.base_gateway}} admin. 
+
+- The group claim name depends on your IdP -- for example, Okta configures claims for `groups`, and other IdPs might configure them as `roles`. 
+- The group claim value need to follow the format `<workspace_name>:<role_name>` for dynamic role assigning.
```
Bradamant3, line 44:
```suggestion
This value depends on your IdP -- for example, Okta configures claims for `groups`, and another IdP might configure them as `roles`.

In the Idp, the claim must follow the format `<workspace_name>:<role_name>`.
```

app/gateway/2.7.x/configure/auth/kong-manager/oidc-mapping.md
Bradamant3, line 46:
```suggestion
For example, if `"authenticated_groups_claim": ["groups"]` is specified, and in the IdP `groups:["default:super-admin"]` is specified, the administrators specified in `admin_claim` are assigned to the super-admin role in the default {{site.base_gateway}} workspace. 
```

app/gateway/2.7.x/configure/auth/kong-manager/oidc-mapping.md
liyangau: When we are talking about payload of a JWT token, we refer a `key:value` pair as a claim. 

So to make it clearer, I chose the wording as `<claim_name>:<claim_value>`.

What customer need to do is to make sure they map `authenticated_groups_claim` to the right `claim_name` and the `claim_value` must follow the format of `<workspace_name>:<role_name>`.

IMO it is not clear enough to say the ``claim must follow the format of `<workspace_name>:<role_name>` ``, it is the `claim_value` need to follow this format. 

Also both claim and value depend on what users have on IdP (not just the value).

For example, in keycloak I can use **kong-group** as the claim name for groups membership in my application.

app/gateway/2.7.x/configure/auth/kong-manager/oidc-mapping.md
liyangau: I wrote `returned on admins' id tokens` because I want the users to understand the mapping only works if their ID token has this `claim:value` pair. (currently we are using `/userinfo` but whatever is on `/userinfo` should be listed on `id_token` anyway)

Only set `groups:["default:super-admin"]` on IdP does not mean `groups:["default:super-admin"]` will be returned on token or `/userinfo`. If this value is not found, the mapping will fail.

Bradamant3: Yeah, I get this. But it's too much detail on the IdP side, when this page is about the Kong Manager config. Kong users need to check the JWT on the IdP side, but they shouldn't have to do it on the Kong Manager side and if they do, we should probably have a troubleshooting page for this in the KB instead of putting it in the docs. Make sense? If the `claim:value` pair doesn't show up in the token, there are non-Kong issues to deal with ... And if we need to explain at this level of detail, it probably belongs on the OIDC config reference page, not here.

liyangau: Interesting point. Do we have a  troubleshooting page or config reference page for OIDC at the moment? If we have, I think it makes sense to put it there.

`If the claim:value pair doesn't show up in the token, there are non-Kong issues to deal with ...` I agree it is not Kong's problem. However when customer can't use it successfully, they will say it is Kong's problem.

The reason I am writing this with so much detail because we know what customer's pain points are. You will be surprised how many cases we got are related to Kong manager OIDC integration. (Some of them are directly related to claim not in token) IMO there is nothing wrong to provide a bit more detail just to help customer understand what they need to do.

Bradamant3: I am so sorry it's taken me so long to get back to this PR. This extra information is super helpful, and I understand a lot better. I've put it back in, reworded for simplicity and clarity (I hope). Maybe Kong users are as familiar with IdPs and JWTs as you are? But in my experience elsewhere it's not always the case, so I tried to make things simpler.

Bradamant3 (approved):
LGTM 🤞

mheap:
Link checker is returning a HTTP 400 from Netlify. I can't reproduce, but manual checks show that it's fine. :shipit:

PR #3519, opened 2022-01-04
ACL Doc - Fix error on doc from probably copy/past

cmourizard:
### Summary
Multiple part of the document are wrong probably from copy/past of another documentation


lena-larionova (approved):
LGTM, nice catch. Thanks for fixing!

PR #3520, opened 2022-01-04
Remove OSS-only label from Debian tile

lena-larionova:
### Summary
Removing "OSS-only" badge from Debian install tile.

### Reason
As of Gateway 2.7, support for Enterprise installations now exists and is documented.

### Testing
See Debian tile: https://deploy-preview-3520--kongdocs.netlify.app/gateway/2.7.x/install-and-run/


Bradamant3 (approved):
LGTM

PR #3522, opened 2022-01-05
OIDC plugin: Fix typo in param name

lena-larionova:
### Summary
Fixing typo.

### Reason
Typo, should be `config.scopes_required` instead of `scores`. Reported on Slack.


Bradamant3 (approved):
LGTM

PR #3526, opened 2022-01-06
Archive 0.2-0.12 versions of OSS gateway docs

lena-larionova:
### Summary
Removing versions with very low viewership from the live docs site. 
Adding links to the version dropdown to take users to the archive so that it's still easy to find.

Screen Shot 2022-01-06 at 3 57 08 PM

(No, the 0.31 for Enterprise is not a typo 😄 )

### Reason
* **Less clutter:** old content should always have an archive date
* **Why not archive old Enterprise versions?** We can't guarantee that a customer isn't on one of those, so until we get an updated support policy, the safe bet is to keep those up.
* **Why leave 0.13.x and 0.14.x?**
  *  0.14.x is the biggest jump in readership from older 0.x versions (0.13.x has around 9k pageviews in the past year, 0.14.x has 25k. For comparison, 2.6.x has 170k views over the past year, and the release has only been out since October).
  * 0.13.x is mostly hedging. I could be convinced to remove it as well. I'm leaving it up because plugins functioned differently before 0.14.x, and we still refer to this change in many plugin docs (for example, https://docs.konghq.com/hub/kong-inc/basic-auth/2.1-x.html).

### Testing

[Kong Gateway, check version menu](https://deploy-preview-3526--kongdocs.netlify.app/gateway/)
[Kong Gateway OSS (old split doc), check version menu](https://deploy-preview-3526--kongdocs.netlify.app/gateway-oss/2.5.x/)
[Archive](https://github.com/Kong/docs.konghq.com/tree/archive-0.x-ce/archive/gateway-oss)

**Redirects (sample links, check other moved versions as well):**

Note that all redirects will take you to a 404, because the archive is only on this branch currently. Make sure it goes to the correct hypothetical folder (check against the [archive folder](https://github.com/Kong/docs.konghq.com/tree/archive-0.x-ce/archive/gateway-oss) on this branch)

https://deploy-preview-3526--kongdocs.netlify.app/gateway-oss/0.2.x
https://deploy-preview-3526--kongdocs.netlify.app/0.12.x


falondarville (approved):
LGTM, I'm so excited for this change

PR #3532, opened 2022-01-07
Link checker

mheap:
### Summary
Check for broken links on any pages that were created or edited

### Reason
To prevent broken links making it on to the site

### Testing
Add a broken link to a page and raise a PR

### DIFF
--- a/.github/workflows/check-links.yml
+++ b/.github/workflows/check-links.yml
@@ -13,7 +13,8 @@ jobs:
       #  with:
       #    site_name: "kongdocs"
       #    max_timeout: 600
-      - run: |
+      - name: List changed files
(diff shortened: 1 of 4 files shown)


lena-larionova (approved):
This is amazing and very necessary. Thank you for setting this up!

Looks like you ended up using the same package we were using before, broken-link-checker, which means minimal cleanup. I think the only place that traces of the old implementation remain are in the Makefile: https://github.com/mheap/docs.konghq.com/blob/link-checker/Makefile#L56-L63
Should probably clean those out (unless you'd rather make those changes with the sitewide link-checker setup in a later PR).

LGTM, and thanks a ton!

PR #3543, opened 2022-01-11
docs(plugins) add datatheorem to plugin hub

fj-dt:
### Summary
Hello Kong team,
I am a backend developer from [Data Theorem](https://www.datatheorem.com/), and we would like to publish our API Secure product in your Kong Plugin Hub.

### Testing
Tested on my machine using `make run`.


lena-larionova:
Thanks for your submission, but at the moment, we’re not accepting third-party plugins unless they’re from trusted technical partners. We don't have a timeline right now for when that will change.

PR #3554, opened 2022-01-14
[DOCU-2138] Fix multi-line copy to clipboard

mheap:
### Summary
The existing replacement regex didn't support multiple line replacement as
it was missing the `m` modifier.

Multi-line replacement is required when there are multiple commands in a single code block

### Reason
Fix copy to clipboard on pages such as https://docs.konghq.com/gateway/2.7.x/reference/clustering/ (search for `node A`)

### Testing
Try copying the `node A` code block to your clipboard. There will be no `$` present in the review app, but the second command will be prefixed with a `$` on production

### DIFF
--- a/app/_assets/javascripts/copy-code-snippet-support.js
+++ b/app/_assets/javascripts/copy-code-snippet-support.js
@@ -60,7 +60,7 @@ jQuery(function () {
         snippet
           .find("code")
           .text()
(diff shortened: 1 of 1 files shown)


lena-larionova (approved):
Tested the linked page: copied with the copy-code button, ran the copied command through terminal (no `$` character present), and it ran the two commands in order. LGTM!

PR #3561, opened 2022-01-18
[DOCU-1859] Deactivated orgs in Konnect

lena-larionova:
### Summary

Adding a doc that explains how an org might get deactivated and what the results of deactivation are.

Note that all of this doc is _external_. Therefore, all information in this doc applies to any average person using Konnect and not an internal developer. 

### Reason
Need to be transparent about our account/org management. Also, ticket has been around for a while.

### Testing
Doc preview: https://deploy-preview-3561--kongdocs.netlify.app/konnect/org-management/deactivation/

### DIFF
--- a/app/konnect/org-management/deactivation.md
+++ b/app/konnect/org-management/deactivation.md
@@ -3,12 +3,16 @@ title: Account and Organization Deactivation
 no_version: true
 ---
 
-A {{site.konnect_saas}} organization is automatically deactivated after 30
+A free {{site.konnect_saas}} organization is automatically deactivated after 30
 days of inactivity.
 
 Your organization is considered inactive when:
 * There is no user login into the organization during the past 30 days.
-* There are no API requests in either the current or the previous billing cycle.
+* There are no API requests in either the current or the previous billing cycle
+(30 day increments).
+
+Organizations in Plus or Enterprise tiers are exempt, and are never deactivated
+automatically.
 
 ## What happens if an organization is deactivated?
 
@@ -36,12 +40,3 @@ There is currently no way to manually deactivate an organization.
 
 If you need to reactivate an organization or unlock an email for use with
 another organization, reach out to [Kong Support](https://support.konghq.com/).
-
-
-_Oustanding questions:_
-- _Is a deactivated org ever deleted permanently? Can a user request that an
-org be deleted permanently? How would they do that?_
-- _How long do users have after deactivation happens to recover their org? Is there
-any situation where data might be deleted automatically?_
-- _Recovering an email attached to an account_
-- _Can you deactivate org manually?_
--- a/app/konnect/org-management/deactivation.md
+++ b/app/konnect/org-management/deactivation.md
@@ -12,7 +12,8 @@ Your organization is considered inactive when:
 (30 day increments).
 
 Organizations in Plus or Enterprise tiers are exempt, and are never deactivated
-automatically.
+automatically. To close a Plus or Enterprise account, you can
+request deactivation through [Kong Support](https://support.konghq.com/).
 
 ## What happens if an organization is deactivated?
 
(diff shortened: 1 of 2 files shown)


app/konnect/org-management/deactivation.md
```diff
@@ -0,0 +1,47 @@
+---
+title: Account and Organization Deactivation
+no_version: true
+---
+
+A {{site.konnect_saas}} organization is automatically deactivated after 30
+days of inactivity.
+
+Your organization is considered inactive when:
```
divyagadila, line 9: we de-activate account only if its free account. If its Plus or enterprise, we do not de-activate account due to in-activity.

divyagadila (changes requested):

app/konnect/org-management/deactivation.md
lena-larionova: Does this point still apply?
> There are no API requests in either the current or the previous billing cycle.

Ie, does an Enterprise or Plus account get deactivated in this situation? If it doesn't get deactivated, is there some interim state, eg dormant, where the user is not billed?

divyagadila: As far as I know, weekly task that disables organizations do not modify any orgs which belong to Plus or enterprise tier.

lena-larionova: So then the point "There are no API requests in either the current or the previous billing cycle" does not apply at all? The term "billing cycle" is confusing if we only use it to refer to free accounts.

divyagadila: This is still one of the criteria to determine inactivity.

lena-larionova:
@divyagadila Added some clarifying language based on your feedback. Can you also take a look at the following questions?

- Is a deactivated org ever deleted permanently? Can a user request that an
org be deleted permanently? How would they do that?
- How long do users have after deactivation happens to recover their org? Is there
any situation where data might be deleted automatically?
- Is it possible to recover an email attached to an account?
Can a Konnect user deactivate an org manually (self-service)? I wasn't able to find a way to do so through the UI or the Konnect API, so the draft says you can't.

divyagadila:
> I answered below questions to the best of my knowledge, @kinman : adding Karen for fact checking
Is a deactivated org ever deleted permanently? Can a user request that an
org be deleted permanently? How would they do that?
Divya: we only remove ControlPlane for a given org, we do not remove org/user details.
How long do users have after deactivation happens to recover their org? Is there
any situation where data might be deleted automatically?
Divya: No timeline. just provisioned controlPlane is removed when account is de-activated.
Is it possible to recover an email attached to an account?
Divya: we dont have support for this yet, here is ticket to add support to recover email so user can re-use/register with same email again: https://konghq.atlassian.net/browse/KHCP-2046. 
Can a Konnect user deactivate an org manually (self-service)? I wasn't able to find a way to do so through the UI or the Konnect API, so the draft says you can't.
Divya: I don't think user can de-activate account

app/konnect/org-management/deactivation.md
```diff
@@ -0,0 +1,42 @@
...
+* The email associated with the organization is locked and can't be used to
+create another {{site.konnect_saas}} account.
+
+If you have any runtimes registered through the Runtime Manager, they won't be
+stopped by {{site.konnect_saas}}. They will no longer proxy data, but the
+instances will keep running until you stop them manually.
+
+## Deactivate or reactivate an organization
+
+There is currently no way to manually deactivate an organization.
```
kinman, line 39:
```suggestion
To deactivate an organization that you registered, please contact [Kong Support](https://support.konghq.com/).
```

I suggest this rather than leaving the user at a dead end. If it is phrased this way because Customer Support doesn't want to field these requests, feel free to leave as is.

lena-larionova: I haven't heard complaints from support yet, and we have links to contact them all over the Konnect docs 😄 I'll add this.

falondarville (approved):
LGTM, clear and concise. Edit: misspelled concise.

PR #3562, opened 2022-01-19
Ensure there's a review:* label on all PRs

mheap:
### Summary
Fail the build if there is not a `review:*` label applied to the PR

### Reason

When raising a pull request, it's useful to indicate what type of review you're looking for from the team. To help with this, we've added three labels that can be applied:

- `review:copyedit`: Request for writer review.
- `review:general`: Review for general accuracy and presentation. Does the doc work? Does it output correctly?
- `review:tech`: Request for technical review from an SME.

At least one of these labels must be applied to a PR or the build will fail

### Testing
This PR will fail the build until the `review:tech` label is applied

### DIFF
--- a/.github/workflows/require-review-label.yml
+++ b/.github/workflows/require-review-label.yml
@@ -1,0 +1,13 @@
+name: Require Review Label
+on:
+  pull_request:
+    types: [opened, labeled, unlabeled]
+jobs:
+  label:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: mheap/github-action-required-labels@v1
+        with:
+          mode: minimum
+          count: 1
+          labels: "review:tech, review:copyedit, review:general, review:autogen"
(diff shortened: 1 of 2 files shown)


mheap:
image

image

image

image

lena-larionova (approved):
One typo, otherwise LGTM! Appreciate the addition to the README :)

lena-larionova:
Had a thought just now: we're definitely going to have exceptions, mainly from bot PRs and autodocs. What's the better approach - automatically set a review label on those PRs, or configure this action to ignore them?

mheap:
In this case I'd automatically set the label. Which label would you use for autodocs? I'm assuming `review:tech` for dependency upgrades

lena-larionova:
> I'm assuming `review:tech` for dependency upgrades

This makes sense to me.

PR #3564, opened 2022-01-19
Improve Gateway EE upgrade guide and fix wrong versions

lena-larionova:
### Summary
* Fixing wrong versions in OSS doc from 2.4.x to 2.7.x
* Updating all versions in EE upgrade doc for 2.7.x
* Removing incremental upgrade requirement, as this isn't actually necessary at all.
* Attempt to clarify/simplify the path - if minor upgrade, move on. If major upgrade, look over the changelogs for breaking changes first, then otherwise move on.
* Make it clearer that you have three main path options:
  * Upgrade your Postgres DB directly with `migrations up`/`migrations finish`
  * Upgrade your Cassandra DB with the same method as PG, or do a database dump first to reduce downtime
  * Install on a fresh datastore, only migrating the `kong.conf` file.

### Reason
Many confused users, internally and externally. A few issues called out on Slack and in calls. 

### Testing
Doc preview: https://deploy-preview-3564--kongdocs.netlify.app/gateway/2.7.x/install-and-run/upgrade-enterprise/

### DIFF
--- a/app/gateway/2.7.x/install-and-run/upgrade-enterprise.md
+++ b/app/gateway/2.7.x/install-and-run/upgrade-enterprise.md
@@ -64,12 +64,12 @@ to version 2.1.x before continuing.
 ### Hybrid mode considerations
 
 {:.important}
-> **Important:** If you are currently running in [hybrid mode](/enterprise/{{page.kong_version}}/deployment/hybrid-mode/),
+> **Important:** If you are currently running in [hybrid mode](/gateway/{{page.kong_version}}/plan-and-deploy/hybrid-mode/),
 upgrade the control plane first, and then the data planes.
 
 * If you are currently running 2.7.x in classic (traditional)
   mode and want to run in hybrid mode instead, follow the hybrid mode
-  [installation instructions](/gateway/{{page.kong_version}}/plan-and-deploy/hybrid-mode-setup/)
+  [installation instructions](/gateway/{{page.kong_version}}/plan-and-deploy/hybrid-mode/hybrid-mode-setup/)
   after running the migration.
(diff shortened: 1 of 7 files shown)


app/gateway/2.7.x/install-and-run/upgrade-enterprise.md
```diff
@@ -20,68 +20,60 @@ distinction between major, minor, and patch versions. The upgrade
...
 ### Hybrid mode considerations
 
 {:.important}
-> **Important:** If you are currently running in [hybrid mode](/enterprise/{{page.kong_version}}/deployment/hybrid-mode/),
-upgrade the Control Plane first, and then the Data Planes.
+> **Important:** If you are currently running in [hybrid mode](/gateway/{{page.kong_version}}/plan-and-deploy/hybrid-mode/),
+upgrade the control plane first, and then the data planes.
 
-* If you are currently running 2.6.x in classic (traditional)
+* If you are currently running 2.7.x in classic (traditional)
```
locao, line 70: It seems to me that "2.6" was right in this context. This phrase was meant for users who are using traditional mode, want to upgrade and start using hybrid mode at the same time, right?

locao (approved):
I left one comment about an entry that I'm not sure, it may work either way. Other than that, everything is right now.

Thank you, this was probably a source of a lot of confusion!

app/gateway/2.7.x/install-and-run/upgrade-enterprise.md
lena-larionova: Nice catch, you're right. I'll change this one back.

mheap (approved):

PR #3567, opened 2022-01-20
[DOCU-157][DOCU-399][DOCU-1450] Remove deprecated api entity

lena-larionova:
### Summary
Scrub the `/apis` endpoint from the plugin docs and the latest version of Gateway docs. 

Question for reviewers: should I backport these edits? That is, should I scrub this endpoint from all supported versions of Gateway, at least? 

### Reason
The API entity was replaced with Service in CE 0.13/EE 0.32 - that is, very very long ago, pretty much at the start of Kong's lifetime. We have many languishing tickets mentioning this deprecated endpoint.

### Testing
TBA

### DIFF
--- a/app/_hub/kong-inc/acme/index.md
+++ b/app/_hub/kong-inc/acme/index.md
@@ -19,7 +19,6 @@ kong_version_compatibility:
       - 2.7.x
 params:
   name: acme
-  api_id: false
   service_id: false
   route_id: false
   consumer_id: false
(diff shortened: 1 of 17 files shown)


falondarville (approved):
LGTM, worth scrubbing the `$` from these? Or should we do that separately?

lena-larionova:
> worth scrubbing the `$` from these? Or should we do that separately?

It wouldn't hurt. I'll do that.

PR #3577, opened 2022-01-24
Fix positioning of `name` datatype for plugins

lena-larionova:
### Summary
Fix typo on closing table cell tag (`</td>`). 

### Reason
Closing tag is in the wrong place, so the `Type: string` line ends up outside the table:
Screen Shot 2022-01-24 at 9 54 29 AM

### Testing
TBA
Check any plugin


falondarville (approved):
LGTM, spot checked three plugins.

PR #3582, opened 2022-01-25
[DOCU-1720] MVP Embedded Spotlight API documentation

falondarville:
This is the MVP for embedding Spotlight API documentation for the Portal API. The scope of this MVP was getting the API embedded into the docs site. This required minor styling to fix some minor issues with H1, and left and right padding around the second and third columns.

Access the documentation via Gateway > Configure > Kong Dev Portal > Kong API.

Screen Shot 2022-01-27 at 2 37 40 PM

Link: https://deploy-preview-3582--kongdocs.netlify.app/gateway/2.7.x/developer-portal/portal-api/

Some considerations for the future:
- This addition increases build time as we are adding a lot of files. Let's consider keeping API documentation on a separate repo and pulling in the output in to the docs site through a webhook. 
- Please also note that this solution is not mobile-friendly.
- I resorted to using `!important` to overwrite styling in some places as some elements are declared multiple times. Let's do a refactor of the CSS to clean it up.


mheap:
This is looking good.

Three pieces of feedback:

I think we should hide the `Try It Out` section as it won't work without configuring CORS

image

```
.TryItPanel {
  display: none;
}
```

Secondly, when the announcement bar is present the `Overview` link gets cut off

image

Finally, we're missing a redirect from `/files-api` to `/portal-api`

Once these are resolved I think we're good to ship

mheap (changes requested):
This is looking good.

Three pieces of feedback:

I think we should hide the `Try It Out` section as it won't work without configuring CORS

image

```
.TryItPanel {
  display: none;
}
```

Secondly, when the announcement bar is present the `Overview` link gets cut off

image

Finally, we're missing a redirect from `/files-api` to `/portal-api`

Once these are resolved I think we're good to ship

lena-larionova:
> I think we should hide the Try It Out section as it won't work without configuring CORS
Makes sense to me for an MVP.

It would be really awesome if we _could_ have something that sends sample requests, but I'm guessing that would be complicated to set up properly. This actually also appeared on a wishlist of doc improvements from support just this week:

> The apigee "live docs" were great. You can call endpoints in your instance from the doc pages: https://cloud.google.com/apigee/docs/reference/apis/apigee/rest/v1/organizations/create

So - future enhancement for sure, but agree that it makes sense to leave out for now.

lena-larionova (approved):
Did a working session together trying to track down some of the styling issues, and I think this covers most of the real obvious stuff. The "Overview" link was oddly difficult to fix.

LGTM for a v1!

Some things to do for after merge:
* Add or expand descriptions for each request.
* Switch to using an api.html layout file instead of the default so we have better control over the styling. 
* Switch to an `/api` top level location in the docs site directory instead of nesting it under a gateway version. We didn't do it for this iteration because it requires some finicky metadata changes.

mheap (approved):
There's one small bug, but not enough to block merge.

When the banner is present I see the "Dev Portal" header in the sidebar. When I collapse the banner it disappears.

All the other changes LGTM

lena-larionova:
> When the banner is present I see the "Dev Portal" header in the sidebar. When I collapse the banner it disappears.

+1 on that - I should've added it to the list. Because this page is not using a dedicated layout file, the control over styling is limited. We went back and forth on this bug and decided it would be fine for a first iteration to leave it as is, since it doesn't block any links.

PR #3584, opened 2022-01-25
Naming and word choice page for style guide

lena-larionova:
### Summary
* Moving terminology to its own page,  and adding more context +info to each name.
* Adding guidelines for bias-free terms
* Adding guidelines for other software (internal and external) terms and names
* Started a short list of other words not to use, and what to use instead. This needs to be expanded on.

### Reason
This info is currently missing from our style guide.

### Testing
https://deploy-preview-3584--kongdocs.netlify.app/contributing/word-choice/


falondarville (approved):
LGTM, the only thing that's standing out for me is don't use via. What's the reasoning for that?

PR #3587, opened 2022-01-26
Toggle all nav tabs with the same title

mheap:
### Summary
Switch all navtabs when a tab is selected

### Reason
Makes it easier for people to choose a configuration option and use it through an entire guide

### Testing
Try out /hub/kong-inc/basic-auth/ on the review app

### DIFF
--- a/app/_assets/javascripts/app.js
+++ b/app/_assets/javascripts/app.js
@@ -468,7 +468,26 @@ $("a[data-filter]").on("keypress", function(e) {
   const navtabs = $("div[data-navtab-id]");
 
   navtabs.on("click", function () {
-    const navtabTitle = $(this);
+     activateNavTab($(this))
+  });
+
+  navtabs.on("keypress", function(e) {
+    if (e.keyCode === 13) {
+      activateNavTab($(this))
+    }
+  });
+
+  function activateNavTab(navtabTitle) {
+    // Toggle all nav tabs that match this title
+    const text = navtabTitle.text();
+    const search = $(".navtab-title").filter(function () {
+      return $(this).text().toLowerCase().indexOf(text.toLowerCase()) >= 0;
+    }).each(function(k,v){
+      activateSingleNavTab($(v));
+    });
+  }
+
+  function activateSingleNavTab(navtabTitle){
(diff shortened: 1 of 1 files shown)


lena-larionova (commented):
Looks like we'll need to be diligent with making sure the titles are consistent - for example, in the plugin docs, some sections have "Kong Manager" and others have "Manager". But we should be doing that _anyway_, so this is a good reminder for it.

I notice some jumpiness though, which I assume is happening because of the size of the content in each tab. But it feels especially painful if you flip though a later section on the page, eg: https://deploy-preview-3587--kongdocs.netlify.app/hub/kong-inc/basic-auth/#enable-the-plugin-globally

Test by toggling the Admin API tab to Kubernetes and to others. It jumps the positioning on the page to top vs bottom. Can it be normalized to always focus on the tab as the top of page?

mheap:
@lena-larionova I implemented normalisation and it felt off to me. When the tab was centered on the page and didn't move (e.g. the first tab), having it scroll was weird.

I've added some logic that will scroll to the selected tab if it is no longer visible on the screen when a tab is selected.

I toyed with the idea of making it scroll if the position on the page moves by more than 200px, but chose to go with the visibility one. Let me know if you'd like me to push a commit with this implementation so that you can try it out

lena-larionova:
> I implemented normalisation and it felt off to me. When the tab was centered on the page and didn't move (e.g. the first tab), having it scroll was weird.

Is that the current behavior in the preview? Because that's looking weird to me. Maybe we don't want to do gradual scroll? That makes it more obvious that something strange happened.

> I toyed with the idea of making it scroll if the position on the page moves by more than 200px

I'm not sure that'll make a big difference, but maybe?

mheap:
Yep, the current behaviour is to scroll in to view if the content would otherwise be off the screen. I can update the branch to jump to the content immediately rather than smooth scrolling

mheap:
@lena-larionova Give it 10 minutes for the review app to deploy then try the new "jump-to" implementation. It feels better to me

lena-larionova (approved):
I love it - much better. Thanks for all the work on this! LGTM

PR #3589, opened 2022-01-27
Ensure sidebar + top nav module list is consistent

mheap:
### Summary
Followup on #3566 

Now that the sidebar / top nav are consistent, let's add a test to make sure it never gets out of sync again.

I also added a missing `s` from the top navigation that the test highlighted.

### Reason
One less thing for people to have to remember. The build will fail if we forget to change one of the two sections

### Testing
The tests will pass


lena-larionova (approved):
Looks good - we'll also have to make sure to update the spec if class names or IDs ever change, but that would also come up in a failed check.

PR #3591, opened 2022-01-27
Update require-review-label.yml

mheap:
### Summary
When a label is added _after_ a PR is opened, the original commit will still be marked as failing. This is due to the action we use reading `event.json` rather than using the API (see https://github.com/mheap/github-action-required-labels/issues/14)

This PR adds a `GITHUB_TOKEN` which allows us to add a label then re-run the job to get a green tick

### Reason
Misleading build failures are bad. We can fix them

### Testing
This PR will pass


lena-larionova (approved):
Thanks for this update! I've had to manually rerun the check for a few people already 😅

PR #3592, opened 2022-01-27
Release: Gateway patches 2.7.1.0 and 2.6.0.3

lena-larionova:
### Summary
* Changelog entries for 2.7.1.0 and 2.6.0.3.
* Adding a config reference doc entry for the new parameter `cluster_max_payload`.
  * Note to reviewers: the text for this entry is unedited and getting added as-is from the `kong.conf.default` file. Not changing it to avoid conflicts with autodoc generation; will need to be edited in the source file.

Also note that the wiki lists this change for 2.6.0.3:
> Fix The garbage can icon overlaps the View link which causes users to click Delete instead of View

However, this change already went in with 2.6.0.1 (https://docs.konghq.com/gateway/changelog/#2601) - was there a regression, did it not actually go into 2.6.0.1, or is it just mislabelled?

### Reason
Patch releases.

### Testing
https://deploy-preview-3592--kongdocs.netlify.app/gateway/changelog/

### DIFF
--- a/app/gateway/changelog.md
+++ b/app/gateway/changelog.md
@@ -10,11 +10,16 @@ no_version: true
 ### Features
 
 #### Enterprise
-* You can now configure [`cluster_max_payload`](/gateway/latest/reference/configuration/#cluster-max-payload)
+* You can now configure [`cluster_max_payload`](/gateway/latest/reference/configuration/#cluster_max_payload)
 for hybrid mode deployments. This configuration option sets the maximum payload
 size allowed to be sent across from the control plane to the data plane. If your
 environment has large configurations that generate `payload too big` errors
 and don't get applied to the data planes, use this setting to adjust the limit.
+* When using PKI for certificate verification in hybrid mode, you can now 
+configure a list Common Names allowed to connect to a control plane with the
+[`cluster_allowed_common_names`](/gateway/latest/reference/configuration/#cluster_allowed_common_names)
+option. If not set, only data planes with the same parent domain as the control
+plane cert are allowed.
 
 ### Fixes
 
--- a/app/gateway/changelog.md
+++ b/app/gateway/changelog.md
@@ -15,7 +15,7 @@ for hybrid mode deployments. This configuration option sets the maximum payload
 size allowed to be sent across from the control plane to the data plane. If your
 environment has large configurations that generate `payload too big` errors
 and don't get applied to the data planes, use this setting to adjust the limit.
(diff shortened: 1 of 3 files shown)


app/gateway/changelog.md
```diff
@@ -196,6 +243,9 @@ Consumer entity, and the username would remain locked. For example, if the
 admin was created in workspace `dev` and deleted from workspace `QA`, this
 issue would occur.
 
+* Phone home metrics are now sent over TLS, meaning that any analytics data
+on Kong Gateway usage now travels through an encrypted connection.
```
lena-larionova, line 247: Note that I also added this change to the 2.7.0.0 changelog, since it got missed.

tyler-ball:
@lena-larionova 

> Also note that the wiki lists this change for 2.6.0.3:
> > Fix The garbage can icon overlaps the View link which causes users to click Delete instead of View

> However, this change already went in with 2.6.0.1 (https://docs.konghq.com/gateway/changelog/#2601) - was there a regression, did it not actually go into 2.6.0.1, or is it just mislabelled?

That change did not make it into 2.6.0.1 and will be released for the first time in 2.6.0.3

tyler-ball (approved):

app/gateway/changelog.md
```diff
@@ -4,6 +4,58 @@ no_search: true
...
+  value by adding a nil value check after performing `ngx.shared.dict` operations.
+  * Fixed a timer leak that caused the timers to be exhausted and failed to
+  start any other timers used by Kong, showing the error `too many pending timers`.
+
+    Before, the plugin used one timer for each namespace maintenance process,
+    increasing timer usage on instances with a large number of rate limiting
+    namespaces. Now, it uses a single timer for all namespace maintenance.
+
+* [Rate Limiting Advanced](/hub/kong-inc/rate-limiting-advanced) (`rate-limiting-advanced`)
+  * Fixed a 500 error which occurred when consumer groups were enforced but no
```
falondarville, line 50:
```suggestion
  * Fixed a 500 error that occurred when consumer groups were enforced but no
```

falondarville (approved):
LGTM, one minor grammar suggestion

PaulFischer-Kong (approved):

tyler-ball:
@lena-larionova We need to remove the following line from the release notes:
```
* Fixed a performance issue with Kong Manager, which occurred when admins had
access to multiple workspaces.
```
That is the fix for FTI-3077, which is having some test issues and will not make it out in the 2.7.1.0 release

PR #3595, opened 2022-01-28
Don't check 'Edit this page' links on forks

mheap:
### Summary
Skip the "Edit this page" links when checking PRs from forks

### Reason
The build process does not have access to the upstream repo name, so we cannot dynamically generate the correct URL for PRs from forks. This leads to every fork PR failing the build.

This PR will exclude the edit link when a PR comes from a fork

### Testing
I've tested on this PR


lena-larionova (approved):
LGTM, checked build - links are definitely still broken (as expected), but the broken link check passes (also as expected).

PR #3596, opened 2022-01-28
Sidebar rendering speed improvements

mheap:
### Summary
Reduce duplication during sidebar navigation item rendering.

The `include-check.sh` script previously added was used to make sure we don't depend on any `page.*` level variables in includes. Passing them in to the `include_cached` function breaks the cache for each permutation of `page.*` input

### Reason
Save some build minutes on Netlify, make life better for people running the docs locally

Original build time:

image

After switching to `include_cached` for the sidebar:

image

### Testing
This needs extensive testing. Check the sidebar on different products _and_ different versions.

I'm 95% confident with this change thanks to the `include-check.sh` work

### DIFF
--- a/app/_includes/breadcrumbs.html
+++ b/app/_includes/breadcrumbs.html
@@ -5,7 +5,7 @@
       <i class="sidebar-toggle"></i>
     {% endif %}
     <li class="breadcrumb-item">
-      {% if include.edition =='enterprise' or include.edition =='konnect' or include.edition =='konnect-platform' or include.edition == 'mesh' or page.edition == 'gateway' %}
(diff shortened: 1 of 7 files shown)


lena-larionova (approved):
Tried this out both through the netlify preview and locally, haven't run into any issues. 
* Tested both versioned and unversioned docs (checked a few pages in every doc set/product we have), looked through older versions, and checked the plugin hub for good measure (which shouldn't be affected anyway, but just in case).
* Tested making changes to the sidebar locally and saving while the build is running - it works, changes are made, and the caching does not cause any issues. Flipping back and forth between versions after an edit is also not an issue.

For reference, my local build time went from 1.5-2.5 minutes before this change, down to about 50 seconds after this change. With a sidebar edit, that build time goes up to just over 1 minute.

Based on all that, LGTM! Couldn't find any issues.

PR #3598, opened 2022-01-28
Remove scroll-to-top button from mobile

lena-larionova:
### Summary
Hiding the scroll-to-top button on mobile, as it shouldn't be used there. 

### Reason
Scrolling to the top on a phone or tablet is built into the device and we're unnecessarily crowding the screen with this button.

### Testing
https://deploy-preview-3598--kongdocs.netlify.app/gateway/ (or any page): open the link on mobile, or resize your browser to 800px width or less, and make sure the scroll-to-top button doesn't appear.


falondarville (approved):
LGTM

PR #3603, opened 2022-01-31
Change upstream name in getting started guide

### LINKED ISSUE #3601: Conflict occurred when creating an Upstream.
### Where is the problem?

https://docs.konghq.com/gateway/2.7.x/get-started/comprehensive/load-balancing/

### What happened?

When we ran the following curl command written in the Configure Upstream Services section, we received the following 409 conflict status responses.

```
❯ curl -i -X POST http://localhost:8001/upstreams \
--data name=upstream 
HTTP/1.1 409 Conflict
Date: Fri, 28 Jan 2022 03:51:20 GMT
Content-Type: application/json; charset=utf-8
Connection: keep-alive
Access-Control-Allow-Origin: *
Content-Length: 139
X-Kong-Admin-Latency: 73
Server: kong/2.7.0

{"fields":{"name":"upstream"},"name":"unique constraint violation","message":"UNIQUE violation detected on '{name=\"upstream\"}'","code":5}
```

Kong Version: 2.7.0

We changed the value of `data` arguments from `upstream` to `upstream-service` and ran the curl again, and it worked.
We think the same problem probably occurs when using Kong Manager or decK(YAML).

### What did you expect to happen?

We think the problem occurs because Upstream named `upstream` already exists.
Therefore we think we change the value of `data` arguments, it will work correctly.
(For example: change from `upstream` to `upstream-service`)

If there are no problems, we will create Pull Request.

### Code of Conduct and Community Expectations

- [X] I agree to follow this project's Code of Conduct
- [X] I agree to abide by the Community Expectations

lena-larionova:
### Summary
Renaming the upstream in the getting started guide from `upstream` to `example_upstream` to avoid using reserved entity names and causing issues.

### Reason
Fixes #3601 (submitter closed issue before I pushed a fix, but it's still an issue.)

Did this in all versions because it has the potential to break the guide's flow entirely.

### Testing
https://deploy-preview-3603--kongdocs.netlify.app/gateway/2.7.x/get-started/comprehensive/load-balancing/

### DIFF
--- a/app/gateway/2.6.x/get-started/comprehensive/load-balancing.md
+++ b/app/gateway/2.6.x/get-started/comprehensive/load-balancing.md
@@ -30,7 +30,7 @@ In this section, you will create an Upstream named `upstream` and add two target
 1. Access your Kong Manager instance and your **default** workspace.
 2. Go to **API Gateway** > **Upstreams**.
 3. Click **New Upstream**.
-4. For this example, enter `upstream` in the **Name** field.
+4. For this example, enter `example_upstream` in the **Name** field.
 5. Scroll down and click **Create**.
 6. On the Upstreams page, find the new upstream service and click **View**.
 7. Scroll down and click **New Target**.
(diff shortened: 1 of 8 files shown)


falondarville (approved):
LGTM

PR #3607, opened 2022-02-01
Gulpfile updates

mheap:
### Summary

1. Speed up the build by removing the implicit call to `clean`.
2. Delete the HTML size report that we don't use
3. Remove remaining references to GH Pages

### Reason
This is best explained with a 70 second video

https://user-images.githubusercontent.com/59130/151998634-aded07c0-6de8-41b9-b8b8-bb3929f7f77d.mp4

### Testing
Run `gulp` locally

### DIFF
--- a/gulpfile.js
+++ b/gulpfile.js
@@ -119,9 +119,7 @@ function jekyll(cb) {
     profile = "-dev";
   }
 
-  var command =
-    `bundle exec jekyll build --config jekyll${profile}.yml --profile --destination ` +
-    paths.dist;
+  var command = `bundle exec jekyll build --config jekyll${profile}.yml --profile`
 
   childProcess.exec(command, function (err, stdout, stderr) {
     log(stdout);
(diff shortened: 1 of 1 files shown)


lena-larionova:
Not sure what's going on, but I pulled this branch down to test and the file watcher isn't working for me. From what I can tell, nothing's being sent to `dist`?

mheap:
The initial build time won't decrease, but subsequent build times should (so long as you don't delete the `dist` folder)

lena-larionova:
> The initial build time won't decrease, but subsequent build times should (so long as you don't delete the dist folder)

Didn't delete the folder, subsequent build times are still all 55 seconds.

lena-larionova (approved):
For posterity: after a bunch of testing together, it turned out the issue was being caused by something in my local clone of the repo. With a fresh clone, builds are now down to 8-10 seconds. 🎉 

LGTM!

PR #3608, opened 2022-02-01
[DOCU-2114] Fix Docker install instructions

lena-larionova:
### Summary

Based on the confusion caused by editable placeholders and a few other changes, fixing the Docker install instructions for Gateway:
- Removing editable placeholders and replacing with hardcoded values. If a user needs to change any values, they can change them.
- Adding descriptions for parameters, so that if they want to customize the install, they know what each param does
- Cleaning up order of instructions and chunking up the DB-less mode section
- Cleaning up DB-less mode text
- FIxing `migrations` command
- Removing step to tag docker image:
  - We were tagging the OSS `kong` image  as `kong`, which is redundant
  - We shouldn't use `kong-ee` anymore. The Enterprise image name is `kong/kong-gateway`, which is fairly clear as it is and is different from the OSS image.

**Explanations for a couple of specific changes:**
- Why remove `KONG_CASSANDRA_CONTACT_POINTS`? 
  - Cassandra support is deprecated and will be removed, so we shouldn't be using Cassandra in our examples. Also, the examples configure a Postgres database, and it's really strange that there's one param that configures something for Cassandra.
-  Why set `KONG_ADMIN_GUI_URL` to `localhost` instead of a placeholder? 
   - `localhost` really is a placeholder. By using this value instead of something like `<hostname>`, people can copy+paste more easily. It can also be confusing to figure out what your hostname is; this provides a default that they can adjust.

**Tested everything that I could:**
- The database-backed instructions (both OSS and Enterprise) work as documented, and you can copy-paste the content with no changes if you want, and it'll all work. 
- Had trouble testing the db-less mode section as documented, as it doesn't really work on MacOS (known limitation based on volume locations). If someone else can test that, that would be ideal.

### Reason
The installation doc was broken and we were getting lots of reports about it.

### Testing
https://deploy-preview-3608--kongdocs.netlify.app/gateway/2.7.x/install-and-run/docker/
Run through the instructions, if possible, and install Gateway.

### DIFF
--- a/app/gateway/2.6.x/install-and-run/docker.md
+++ b/app/gateway/2.6.x/install-and-run/docker.md
@@ -226,7 +226,7 @@ to help you set up and enhance your first Service.
 In particular, right after installation you might want to:
 * [Create a service and a route](/gateway/{{page.kong_version}}/get-started/comprehensive/expose-services)
 * [Configure a plugin](/gateway/{{page.kong_version}}/get-started/comprehensive/protect-services)
-* [Secure your Services with authentication](/gateway/{{page.kong_version}}/get-started/comprehensive/secure-services)
+* [Secure your services with authentication](/gateway/{{page.kong_version}}/get-started/comprehensive/secure-services)
 * [Load balance traffic across targets](/gateway/{{page.kong_version}}/get-started/comprehensive/load-balancing)
 
 ### Clean up containers
@@ -241,7 +241,6 @@ docker container rm kong-gateway
 docker container rm kong-database
 docker network rm kong-net
 ```
-
 
 ## Install Kong Gateway in DB-less mode
 
@@ -293,13 +292,13 @@ backed up by a Redis cluster).
     This guide assumes you named the file `kong.yml`.
 
 1.  Save your declarative configuration locally, and note the filepath.
-For this example, we're using `/kong/declarative/kong.yml`.
 
 ### Start Kong Gateway in DB-less mode
 
 {% include_cached /md/admin-listen.md desc='long' %}
 
-1. Run the following command to start a container with {{site.base_gateway}}.
+1. From the same directory where you just created the `kong.yml` file,
+run the following command to start a container with {{site.base_gateway}}:
 
 {% capture start_container %}
 {% navtabs codeblock %}
@@ -307,7 +306,7 @@ For this example, we're using `/kong/declarative/kong.yml`.
 ```sh
 docker run -d --name kong-dbless \
  --network=kong-net \
- -v "$(pwd)"/kong/declarative:/kong/declarative/ \
+ -v "$(pwd):/kong/declarative/" \
  -e "KONG_DATABASE=off" \
  -e "KONG_DECLARATIVE_CONFIG=/kong/declarative/kong.yml" \
  -e "KONG_PROXY_ACCESS_LOG=/dev/stdout" \
@@ -331,7 +330,7 @@ docker run -d --name kong-dbless \
 ```sh
 docker run -d --name kong-dbless \
  --network=kong-net \
- -v "$(pwd)"/kong/declarative:/kong/declarative/ \
+ -v "$(pwd):/kong/declarative/" \
  -e "KONG_DATABASE=off" \
  -e "KONG_DECLARATIVE_CONFIG=/kong/declarative/kong.yml" \
  -e "KONG_PROXY_ACCESS_LOG=/dev/stdout" \
@@ -353,7 +352,7 @@ docker run -d --name kong-dbless \
     Where:
     * `--name` and `--network`: The name of the container to create,
     and the Docker network it communicates on.
-    * `-v "$(pwd)"/path/to/source:/path/to/target/`: Mount a directory on your
+    * `-v $(pwd):/path/to/target/`: Mount the current directory on your
     local filesystem to a directory in the Docker container. This makes the
     `kong.yml` file visible from the Docker container.
     * [`KONG_DATABASE`](/gateway/{{page.kong_version}}/reference/configuration/#database):
@@ -361,7 +360,7 @@ docker run -d --name kong-dbless \
     backing database for configuration storage.
     * [`KONG_DECLARATIVE_CONFIG`](/gateway/{{page.kong_version}}/reference/configuration/#declarative_config):
     The path to a declarative configuration file inside the container.
-    This path should match the path that you're mapping with `-v`.
+    This path should match the target path that you're mapping with `-v`.
     * All [`_LOG`](/gateway/{{page.kong_version}}/reference/configuration/#general-section)
     parameters: set filepaths for the logs to output to, or use the values in
     the example to  print messages and errors to `stdout` and `stderr`.
@@ -396,7 +395,7 @@ to help you set up and enhance your first Service.
 If you use the sample `kong.yml` in this guide, you already have a Service and
 a Route configured. Here are a few more things to check out:
 * [Configure a plugin](/gateway/{{page.kong_version}}/get-started/comprehensive/protect-services)
-* [Secure your Services with authentication](/gateway/{{page.kong_version}}/get-started/comprehensive/secure-services)
+* [Secure your services with authentication](/gateway/{{page.kong_version}}/get-started/comprehensive/secure-services)
 * [Load balance traffic across targets](/gateway/{{page.kong_version}}/get-started/comprehensive/load-balancing)
 
 ### Clean up containers
@@ -406,9 +405,7 @@ can clean them up using the following commands:
 
 ```
 docker kill kong-dbless
-docker kill kong-database
 docker container rm kong-dbless
-docker container rm kong-database
 docker network rm kong-net
 ```
 
--- a/app/gateway/2.6.x/install-and-run/docker.md
+++ b/app/gateway/2.6.x/install-and-run/docker.md
@@ -175,7 +175,7 @@ docker run -d --name kong \
  -e "KONG_DATABASE=postgres" \
  -e "KONG_PG_HOST=kong-database" \
  -e "KONG_PG_USER=kong" \
- -e "KONG_PG_PASSWORD=kong" \
+ -e "KONG_PG_PASSWORD=kongpass" \
  -e "KONG_PROXY_ACCESS_LOG=/dev/stdout" \
  -e "KONG_ADMIN_ACCESS_LOG=/dev/stdout" \
  -e "KONG_PROXY_ERROR_LOG=/dev/stderr" \
--- a/app/gateway/2.6.x/install-and-run/docker.md
+++ b/app/gateway/2.6.x/install-and-run/docker.md
@@ -35,28 +35,16 @@ The {{site.base_gateway}} software is governed by the
 * A Docker-enabled system with proper Docker access
 * (Enterprise only) A `license.json` file from Kong
 
-## Pull the Kong Gateway image
+Choose a path to install Kong Gateway:
+* [With a database](#install-kong-gateway-with-a-database): Use a database to
+store Kong entity configurations. Can use the Admin API or declarative
+configuration files to configure Kong.
+* [Without a database (DB-less mode)](#install-kong-gateway-in-db-less-mode):
+Store Kong configuration in-memory on the node. In this mode, the Admin API is
+read only, and you have to manage Kong using declarative configuration.
 
-Pull the Docker image:
-
-{% navtabs codeblock %}
-{% navtab Kong Gateway %}
-```bash
-docker pull kong/kong-gateway:{{page.kong_versions[page.version-index].ee-version}}-alpine
-```
-{% endnavtab %}
-{% navtab Kong Gateway (OSS) %}
-```bash
-docker pull kong:{{page.kong_versions[page.version-index].ce-version}}-alpine
-```
-{% endnavtab %}
-{% endnavtabs %}
-
-You should now have your {{site.base_gateway}} image locally.
-
-Next, choose a path and install Kong Gateway:
-* [With a database](#install-kong-gateway-with-a-database)
-* [Without a database (DB-less mode)](#install-kong-gateway-in-db-less-mode)
+If this is your first time trying out Kong Gateway, we recommend installing it
+with a database.
 
 ## Install Kong Gateway with a database
 
@@ -105,7 +93,7 @@ docker run --rm --network=kong-net \
  -e "KONG_PG_HOST=kong-database" \
  -e "KONG_PG_PASSWORD=kongpass" \
  -e "KONG_PASSWORD=test" \
-kong/kong-gateway kong migrations bootstrap
+kong/kong-gateway:{{page.kong_versions[page.version-index].ee-version}}-alpine kong migrations bootstrap
 ```
 {% endnavtab %}
 {% navtab Kong Gateway (OSS) %}
@@ -114,7 +102,7 @@ docker run --rm --network=kong-net \
  -e "KONG_DATABASE=postgres" \
  -e "KONG_PG_HOST=kong-database" \
  -e "KONG_PG_PASSWORD=kongpass" \
-kong kong migrations bootstrap
+kong:{{page.kong_versions[page.version-index].ce-version}}-alpine kong migrations bootstrap
 ```
 {% endnavtab %}
 {% endnavtabs %}
@@ -122,19 +110,19 @@ kong kong migrations bootstrap
 {{ migrations | indent | replace: " </code>", "</code>" }}
 
     Where:
-    * [`KONG_DATABASE`](/gateway/{{page.kong_version}}/reference/configuration/#database)
-     is the database type, or `off` if setting up a data plane for [hybrid mode](/gateway/{{page.kong_version}}/plan-and-deploy/hybrid-mode/hybrid-mode-setup).
-    * [`KONG_PG_HOST`](/gateway/{{page.kong_version}}/reference/configuration/#postgres-settings)
-    is the name of the Postgres Docker container that is communicating over the
+    * [`KONG_DATABASE`](/gateway/{{page.kong_version}}/reference/configuration/#database):
+     Specifies the type of database that Kong is using.
+    * [`KONG_PG_HOST`](/gateway/{{page.kong_version}}/reference/configuration/#postgres-settings):
+    The name of the Postgres Docker container that is communicating over the
     `kong-net` network, from the previous step.
-    * [`KONG_PG_PASSWORD`](/gateway/{{page.kong_version}}/reference/configuration/#postgres-settings)
+    * [`KONG_PG_PASSWORD`](/gateway/{{page.kong_version}}/reference/configuration/#postgres-settings):
     The password that you set when bringing up the Postgres container in the
     previous step.
     * `KONG_PASSWORD` (Enterprise only): The default password for the admin
     super user for Kong Gateway.
-    * `kong/kong-gateway kong migrations bootstrap` or `kong kong migrations bootstrap`:
-    In order, this is the Kong Gateway container tag, followed by the command
-    to Kong to prepare the Postgres database.
+    * `{IMAGE-NAME:TAG} kong migrations bootstrap`:
+    In order, this is the Kong Gateway container name and tag, followed by the
+    command to Kong to prepare the Postgres database.
 
 ### Start Kong Gateway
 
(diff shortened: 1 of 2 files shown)


falondarville (approved):
LGTM! I ran through the Gateway install (with a database) and everything went smoothly. All of the text "Where:" looks good too.

app/gateway/2.6.x/install-and-run/docker.md
```diff
@@ -57,53 +51,91 @@ docker pull kong:{{page.kong_versions[page.version-index].ce-version}}-alpine
...
 {% endnavtab %}
 {% endnavtabs %}
-{% endcapture %}
-{{ pull_image | indent | replace: " </code>", "</code>" }}
 
-   You should now have your {{site.base_gateway}} image locally.
+You should now have your {{site.base_gateway}} image locally.
 
-2. Tag the image:
+Next, choose a path and install Kong Gateway:
```
mheap, line 57: Why would I choose one over the other?

app/gateway/2.6.x/install-and-run/docker.md
```diff
@@ -57,53 +51,91 @@ docker pull kong:{{page.kong_versions[page.version-index].ce-version}}-alpine
...
+kong kong migrations bootstrap
+```
+{% endnavtab %}
+{% endnavtabs %}
 {% endcapture %}
 {{ migrations | indent | replace: " </code>", "</code>" }}
 
+    Where:
+    * [`KONG_DATABASE`](/gateway/{{page.kong_version}}/reference/configuration/#database)
+     is the database type, or `off` if setting up a data plane for [hybrid mode](/gateway/{{page.kong_version}}/plan-and-deploy/hybrid-mode/hybrid-mode-setup).
```
mheap, line 126: This section is about setting up Kong with a database. I'd avoid mentioning hybrid mode at all

app/gateway/2.6.x/install-and-run/docker.md
```diff
@@ -57,53 +51,91 @@ docker pull kong:{{page.kong_versions[page.version-index].ce-version}}-alpine
...
+    * [`KONG_PG_HOST`](/gateway/{{page.kong_version}}/reference/configuration/#postgres-settings)
+    is the name of the Postgres Docker container that is communicating over the
+    `kong-net` network, from the previous step.
+    * [`KONG_PG_PASSWORD`](/gateway/{{page.kong_version}}/reference/configuration/#postgres-settings)
+    The password that you set when bringing up the Postgres container in the
+    previous step.
+    * `KONG_PASSWORD` (Enterprise only): The default password for the admin
+    super user for Kong Gateway.
+    * `kong/kong-gateway kong migrations bootstrap` or `kong kong migrations bootstrap`:
```
mheap, line 135: We can remove the image name here as they're not expected to copy/paste from the explanation. Showing both may lead to questions about when to use `kong` or `kong/kong-gateway`

app/gateway/2.6.x/install-and-run/docker.md
```diff
@@ -57,53 +51,91 @@ docker pull kong:{{page.kong_versions[page.version-index].ce-version}}-alpine
...
+docker run --rm --network=kong-net \
  -e "KONG_DATABASE=postgres" \
- -e "KONG_PG_HOST=<div contenteditable="true">{PG_CONTAINER_NAME}</div>" \
- -e "KONG_PG_PASSWORD=<div contenteditable="true">{CONTAINER_PASSWORD}</div>" \
- -e "KONG_PASSWORD=<div contenteditable="true">{PASSWORD}</div>" \
- <div contenteditable="true">{TAG_NAME}</div> <div contenteditable="true">{DATABASE_NAME}</div> migrations bootstrap</code></pre>
+ -e "KONG_PG_HOST=kong-database" \
+ -e "KONG_PG_PASSWORD=kongpass" \
+ -e "KONG_PASSWORD=test" \
+kong/kong-gateway kong migrations bootstrap
```
mheap, line 108: Using `kong/kong-gateway` is the same as `kong/kong-gateway:latest` which may not match `docker pull kong/kong-gateway:{{page.kong_versions[page.version-index].ee-version}}-alpine` from above

Output:

```
docker run --rm --network=kong-net \
  -e "KONG_DATABASE=postgres" \
  -e "KONG_PG_HOST=kong-database" \
  -e "KONG_PG_PASSWORD=kongpass" \
  -e "KONG_PASSWORD=test" \
 kong/kong-gateway kong migrations bootstrap

Unable to find image 'kong/kong-gateway:latest' locally
latest: Pulling from kong/kong-gateway
Digest: sha256:8d69f2ad968154c4751410a5f4c3cfe51bc5c325c9e5a5f47d711afeb05d2c2d
```

app/gateway/2.6.x/install-and-run/docker.md
```diff
@@ -151,58 +185,83 @@ docker pull kong:{{page.kong_versions[page.version-index].ce-version}}-alpine
...
 {% endnavtabs %}
 {% endcapture %}
 {{ start_container | indent | replace: " </code>", "</code>" }}
 
-   {:.note}
-   > **Note**: The `HOSTNAME` for `KONG_PORTAL_GUI_HOST` should be preceded by a protocol, for example, `http://`.
-
-2. Verify your installation:
+    Where:
+    * `--name` and `--network`: The tag of the {{site.base_gateway}} image that
```
mheap, line 196: `--name` is the name of the container to create. This can be anything.

The `kong/kong-gateway` entry at the end of the command is the image name and tag

app/gateway/2.6.x/install-and-run/docker.md
```diff
@@ -151,58 +185,83 @@ docker pull kong:{{page.kong_versions[page.version-index].ce-version}}-alpine
...
+    you're using, and the Docker network it communicates on.
+    * [`KONG_DATABASE`](/gateway/{{page.kong_version}}/reference/configuration/#database):
+    Specifies whether this node connects directly to a database.
```
mheap, line 199:
```suggestion
    Specifies the type of database that Kong is using
```

app/gateway/2.6.x/install-and-run/docker.md
```diff
@@ -132,16 +165,17 @@ docker pull kong:{{page.kong_versions[page.version-index].ce-version}}-alpine
  -p 8445:8445 \
  -p 8003:8003 \
  -p 8004:8004 \
- kong-ee</code></pre></div>
+ kong/kong-gateway
```
mheap, line 168: Missing the tag version

app/gateway/2.6.x/install-and-run/docker.md
```diff
@@ -151,58 +185,83 @@ docker pull kong:{{page.kong_versions[page.version-index].ce-version}}-alpine
  -p 8443:8443 \
  -p 127.0.0.1:8001:8001 \
  -p 127.0.0.1:8444:8444 \
- kong:latest</code></pre></div>
+ kong:latest
```
mheap, line 188: Should read the tag version from config

app/gateway/2.6.x/install-and-run/docker.md
```diff
@@ -151,58 +185,83 @@ docker pull kong:{{page.kong_versions[page.version-index].ce-version}}-alpine
...
+    ```sh
+    curl -i -X GET --url http://localhost:8001/services
+    ```
 
     You should receive a `200` status code.
 
-3. (Not available in OSS) Verify that Kong Manager is running by accessing it using the URL specified in `KONG_ADMIN_GUI_URL`:
-
-      <pre><code>http://<div contenteditable="true">{HOSTNAME}</div>:8002</code></pre>
+1. (Not available in OSS) Verify that Kong Manager is running by accessing it using the URL specified in `KONG_ADMIN_GUI_URL`:
```
mheap, line 226: If I have a working docker install with a DB here, the next step is to configure a service (link to quickstart?)

I don't want to see the db-less instructions as I already have a gateway

Maybe move `Next Steps` in to this section + the db-less section and duplicate

mheap: I'd also love instructions on how to stop the containers I just started + remove the network

```bash
docker kill kong-gateway
docker kill kong-database
docker container rm kong-gateway
docker container rm kong-database
docker network rm kong-net
```

app/gateway/2.6.x/install-and-run/docker.md
```diff
@@ -151,58 +185,83 @@ docker pull kong:{{page.kong_versions[page.version-index].ce-version}}-alpine
...
+    * `--name` and `--network`: The tag of the {{site.base_gateway}} image that
+    you're using, and the Docker network it communicates on.
+    * [`KONG_DATABASE`](/gateway/{{page.kong_version}}/reference/configuration/#database):
+    Specifies whether this node connects directly to a database.
+    * [`KONG_PG_HOST`](/gateway/{{page.kong_version}}/reference/configuration/#postgres-settings):
+    The name of the Postgres Docker container that is communicating over the
+    `kong-net` network.
+    * [`KONG_PG_USER` and `KONG_PG_PASSWORD`](/gateway/{{page.kong_version}}/reference/configuration/#postgres-settings):
+     The Postgres username and password. Kong Gateway needs the login information
+     to store configurations in this database.
```
mheap, line 205:
```suggestion
     to store configuration data in the `KONG_PG_HOST` database
```

app/gateway/2.6.x/install-and-run/docker.md
```diff
@@ -220,32 +279,41 @@ The steps involved in starting Kong in [DB-less mode](/gateway/{{page.kong_versi
     ]
     ```
 
-    Notice the `MountPoint` entry. We will need that path in the next step.
+    Notice the `MountPoint` entry. You will need that path in the next step.
+
+### Prepare your configuration file
 
-3. Prepare your declarative configuration file:
+1. Prepare your declarative configuration file in `.yml` or `.json` format.
```
mheap, line 286: Can we add an example declarative config here so that people don't need to think?

```yaml
_format_version: "2.1"
_transform: true

services:
- name: my-service
  url: https://example.com
  routes:
  - name: my-route
    paths:
    - /
```

app/gateway/2.6.x/install-and-run/docker.md
```diff
@@ -220,32 +279,41 @@ The steps involved in starting Kong in [DB-less mode](/gateway/{{page.kong_versi
...
+{% include_cached /md/admin-listen.md desc='long' %}
+
+1. Run the following command to start a container with {{site.base_gateway}}.
 
-   Although it's possible to start the Kong container with `KONG_DATABASE=off`, it is usually
-   desirable to also include the declarative configuration file as a parameter via the
-   `KONG_DECLARATIVE_CONFIG` variable name. To do this, we need to make the file
-   "visible" from within the container. Use the `-v` flag, which maps
-   the Docker volume to the `/usr/local/kong/declarative` folder in the container.
+   Although it's possible to start the Kong container with just
```
mheap, line 304: Too much detail here imo. Don't explain `KONG_DECLARATIVE_CONFIG` is optional. They won't be successful without it, so we can gloss over the detail here and add more context if they're reading the db-less guide

app/gateway/2.6.x/install-and-run/docker.md
```diff
@@ -280,27 +350,53 @@ The steps involved in starting Kong in [DB-less mode](/gateway/{{page.kong_versi
...
 {% endnavtab %}
 {% endnavtabs %}
 {% endcapture %}
 {{ start_container | indent | replace: " </code>", "</code>" }}
 
-5. Verify that {{site.base_gateway}} is running:
-
-    <pre><code>curl -i http://<div contenteditable="true">{HOSTNAME}</div>:8001</code></pre>
+    Where:
+    * `--name` and `--network`: The tag of the {{site.base_gateway}} image that
```
mheap, line 361: As above -`--name` is a container name, not image name

app/gateway/2.6.x/install-and-run/docker.md
```diff
@@ -280,27 +350,53 @@ The steps involved in starting Kong in [DB-less mode](/gateway/{{page.kong_versi
...
+    you're using, and the Docker network it communicates on.
+    * `-v kong-vol:/usr/local/kong/declarative`: The name of the Docker volume
```
mheap, line 363: This needs updating to `-v $PWD:/usr/local/kong/declarative`

app/gateway/2.7.x/install-and-run/docker.md
```diff
@@ -153,61 +187,79 @@ docker pull kong:{{page.kong_versions[page.version-index].ce-version}}-alpine
...
-> **Important**: Running {{site.base_gateway}} on Docker in DB-less mode is
-supported for both Enterprise and OSS images. However, there is currently no
-documentation for setting up an Enterprise image in DB-less mode, so adjust the
-following steps as needed for an Enterprise deploy.
+## Install Kong Gateway in DB-less mode
 
 The steps involved in starting Kong in [DB-less mode](/gateway/{{page.kong_version}}/reference/db-less-and-declarative-config) are the following:
 
-1. Create a Docker network:
+### Set up a Docker volume
```
mheap, line 237: This title talks about `volume` then the first step is to create a network, which is confusing to me

app/gateway/2.7.x/install-and-run/docker.md
```diff
@@ -153,61 +187,79 @@ docker pull kong:{{page.kong_versions[page.version-index].ce-version}}-alpine
...
-    precaution in case you want to add other things in the future (like a rate-limiting plugin
+    precaution in case you want to add other things in the future (like a Rate Limiting plugin
     backed up by a Redis cluster).
 
-2. Create a Docker volume:
-
-    For the purposes of this guide, a [Docker Volume][Docker Volume] is a folder inside the host machine which
-    can be mapped into a folder in the container. Volumes have a name. In this case we're going
-    to name ours `kong-vol`
+1. Create a Docker volume:
```
mheap, line 252: We don't need to create a volume. Docker supports mounting existing folders. We can remove all references to `docker volume` and use `-v $PWD:/path/to/folder` in the `docker` command to mount the current directory

mheap (changes requested):
I've left comments on 2.6, all of which apply to 2.7 too.

We show how to get up and running with `docker` commands. Would it make sense to provide `docker-compose` configurations too? Getting started can be as simple as:

```bash
# DB
curl http://whatever-we-like.com/kong/compose-db.yml > docker-compose.yml
docker-compose up
```

Or for DB-less:

```bash
curl http://whatever-we-like.com/kong/kong-declarative.yml > kong.yml
curl http://whatever-we-like.com/kong/kong-dbless.yml > docker-compose.yml
docker-compose up
```

app/gateway/2.6.x/install-and-run/docker.md
lena-larionova: Hmm. I didn't realize. There's two ways to go about this, then: either bring back the step to tag the image that you pulled, or include the version here. 
- Tagging would make each codeblock cleaner, but adds another step
- Using `kong/kong-gateway:<version>` makes for a longer codeblock but removes the need to pull and tag the image initially.

app/gateway/2.6.x/install-and-run/docker.md
lena-larionova: My goal here was to clear up the confusion around `kong kong` - the question about what this means and why it sometimes works and sometimes doesn't has floated around the internet for a long time. So the explanation is useful, but I can leave a placeholder for the image name, eg `<image-name> kong migrations bootstrap`

app/gateway/2.6.x/install-and-run/docker.md
mheap: Use `kong/kong-gateway:version`. Retagging local images isn't a common workflow

lena-larionova: After looking at your other comments, looks like tag version makes things simpler. I'll go with that.

In this case: does it mean that they don't actually need to pull the image before going through the rest of the steps?

app/gateway/2.6.x/install-and-run/docker.md
lena-larionova: This question is much bigger on the inside, and really we should be linking them to a comparison of database-backed vs db-less deployment options (which... currently doesn't exist? We have a very top-level one that Shane asked us to remove because it wasn't super accurate, and no one's written one since.) 

In lieu of a proper breakdown, I'll try to come up with a one sentence summary of each option here.

app/gateway/2.6.x/install-and-run/docker.md
mheap: Yep! The first time they use it the image will be pulled automatically

app/gateway/2.6.x/install-and-run/docker.md
mheap: If the next step is to follow the getting started guide which needs the admin API to be writable, maybe we just add "We recommend trying out Kong with a DB if this is your first time trying Kong"

app/gateway/2.7.x/install-and-run/docker.md
lena-larionova: I couldn't get this working with the syntax you provided, but this is what worked:

`-v "$(pwd)"/source:/target/ ` 

So, I did: 
`-v "$(pwd)"/declarative:/declarative/ ` 

And pointing to `/usr/local` caused permissions errors, which is why I just went with `/declarative/`. Do we want to assume that users will have the knowledge to work out any permissions issues, or does it seem fine to provide a sample path like this?

app/gateway/2.6.x/install-and-run/docker.md
lena-larionova: The getting started guide actually provides yaml config snippets and gives instructions to apply them using decK - so this isn't a limitation.

app/gateway/2.6.x/install-and-run/docker.md
```diff
@@ -289,34 +271,45 @@ The steps involved in starting Kong in [DB-less mode](/gateway/{{page.kong_versi
...
+        strip_path: true
+    ```
+
     This guide assumes you named the file `kong.yml`.
 
-1.  Save your declarative configuration file inside the `MountPoint` path
-mentioned in the previous step. In the case of this
-guide, that would be `/var/lib/docker/volumes/kong-vol/_data/kong.yml`.
+1.  Save your declarative configuration locally, and note the filepath.
+For this example, we're using `/kong/declarative/kong.yml`.
```
mheap, line 296: You don't need to note the filepath here. The `$(pwd)` uses the current directory, and `/kong/declarative` is the path inside the container that we mount to.

What's important is that they run the `docker run` command in the same directory as the `kong.yml` that they just created

lena-larionova:
@mheap Applied all of your feedback, but used slightly different mount syntax for db-less. Let me know if that works.

Re: docker-compose: for the scope of this PR and the sake of getting accurate information out to users, I'd rather move that to a follow-up task. I just ran through what it would take to add a section based on our [docker compose repo](https://github.com/Kong/docker-kong/tree/master/compose), and it would require figuring out which env variables to edit/set either in prep on my system, creating a secret, and... possibly something else that's not currently obvious to me.

app/gateway/2.6.x/install-and-run/docker.md
```diff
@@ -394,14 +387,33 @@ docker run -d --name kong \
...
 [Docker Volume]: https://docs.docker.com/storage/volumes/
 
+### Get started with Kong Gateway
+
+Now that you have a running Gateway instance, Kong provides a series of
+[getting started guides](/gateway/{{page.kong_version}}/get-started/comprehensive/)
+to help you set up and enhance your first Service.
+
+If you use the sample `kong.yml` in this guide, you already have a Service and
+a Route configured. Here are a few more things to check out:
```
mheap, line 397: Should we mention that they need to choose the `Using decK (YAML)` tab once they follow a link?

app/gateway/2.6.x/install-and-run/docker.md
```diff
@@ -394,14 +387,33 @@ docker run -d --name kong \
...
+* [Load balance traffic across targets](/gateway/{{page.kong_version}}/get-started/comprehensive/load-balancing)
+
+### Clean up containers
+
+If you're done testing Kong Gateway and no longer need the containers, you
+can clean them up using the following commands:
+
+```
+docker kill kong-dbless
+docker kill kong-database
```
mheap, line 409: We shouldn't have a database in `dbless`

mheap (changes requested):
Looking awesome. Just a few things to clear up

app/gateway/2.6.x/install-and-run/docker.md
lena-larionova: Ack copy-pasta error

app/gateway/2.6.x/install-and-run/docker.md
mheap: Scrap that, I have a PR open to automatically select the correct tab based on the URL at https://github.com/Kong/docs.konghq.com/pull/3616

lena-larionova: Have a commit ready to go for when the tab PR gets merged. I'll push + rebase after.

mheap (approved):

PR #3610, opened 2022-02-01
[DOCU-1698] External link icons + update font awesome

lena-larionova:
### Summary
* Adding external link icons. 
  * Went with the pure CSS route; can also do this with JS, if reviewers think that would be better.
* Updating font-awesome using their latest docs; the version we were using didn't have a lot of the newer icons, including an external link icon. Nowadays, they provide a script to add to your site instead of referencing a versioned stylesheet. This way, it should stay up-to-date.

**Exceptions:** No link icons on buttons, on badges, or in the footer. For buttons and badges, this is a stylistic choice; for the footer, it's for consistency with the konghq.com site. 

**@ Reviewers:** Should these icons also be left out of the navbar, for consistency with our other sites? Additionally, there's some jumpiness because of the way the page loads, with text loading first and then icons. It's mainly obvious in the navbar.

### Reason
Accessibility guidelines state that external links, especially any links that open in a new tab, must have some warning. For consistency, I've applied the icon to all external links.

A subsequent PR should also add hidden labels on the links to label them external, and notify screenreaders that this link will open in a new tab/window.

Sources:
Guidelines for links: https://www.w3.org/TR/WCAG20-TECHS
Failure conditions: https://www.w3.org/TR/WCAG20-TECHS/F22.html

https://accessibility.oit.ncsu.edu/it-accessibility-at-nc-state/developers/accessibility-handbook/mouse-and-keyboard-events/links/links-and-new-windows/

### Testing
https://deploy-preview-3610--kongdocs.netlify.app/ - scroll down on the landing page, links to Insomnia and Kuma are labelled with the icon
Look through docs overall for instances of the icon.

### DIFF
--- a/app/_assets/stylesheets/pages/docs.less
+++ b/app/_assets/stylesheets/pages/docs.less
@@ -759,3 +759,37 @@
     width: 100%;
   }
 }
+
+// External link icons in both left and right sidebars
+.docs-toc, .docs-sidebar, .landing-page {
+  a:not([href^='https://konghq.com']):not([href^='https://docs.konghq.com']):not([href^='#']):not([href^='/']):not([data-filter]):after {
+      font-family: 'FontAwesome';
+      content: " \f35d";
+      margin-left: 0.4rem;
+      font-size: 12px;
+  }
+}
+
+// External icons in content body, excluding badges
+.content {
+  a:not(.badge):not([href^='https://docs.konghq.com']):not([href^='#']):not([href^='/']):not([data-filter]):after {
+      font-family: 'FontAwesome';
+      content: " \f35d";
+      margin-left: 0.2rem;
+      margin-right: 0.2rem;
+      vertical-align: middle;
+      font-size: 12px;
+      font-weight: lighter;
+  }
+}
+
+// External link icons in top site navbar
+.navbar-items {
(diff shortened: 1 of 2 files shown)


mheap:
In general this looks good.

Changes that look off visually to me and I'd exclude them from having the icon added:

Edit/Report page links:
image

Top URL bar
image

lena-larionova:
Agree on the top menu bar links, and I have no issues removing the github link icons either.

For some specifics though: if we're removing link icons from the navbar, does it also make sense to remove them from the dropdowns? I like having them for Insomnia and Kuma, but I don't know if that's too inconsistent now.

Screen Shot 2022-02-02 at 11 46 32 AM

The other two places where they show up in sidebars are version dropdown and the very rare occurrence in the left nav:
Screen Shot 2022-02-02 at 11 46 44 AM
Screen Shot 2022-02-02 at 11 46 15 AM

mheap:
I like having them for Insomnia and Kuma.

It makes sense to have them in the version drop down for archived versions.

Why is the changelog an external link?

lena-larionova:
> Why is the changelog an external link?

All of our open-source project changelogs are currently external links, as the changelogs still live on github for the various projects. We've talked about moving them over but engineering teams aren't super keen on that, so we'd have to duplicate the information.

lena-larionova:
Also removed from install doc banners to avoid some really painful styling:

Screen Shot 2022-02-02 at 12 42 49 PM

mheap (approved):

PR #3611, opened 2022-02-01
Adjust phrasing for Dev Portal OIDC changelog entry

lena-larionova:
### Summary
Adjusting phrasing for the changelog entry for Dev Portal auth with OIDC in 2.7.

### Reason
Issue reported in Slack. Changelog entry was misleading.
https://kongstrong.slack.com/archives/CDSTDSG9J/p1643636506353189

Explanation from Nate Bailey: 

> The flow for developer portal registration changed for OIDC. The user now authenticates with the IDP first and the email from the claims is used to pre populate the registration form to prevent a mismatch. This is different from the autogeneration of accounts that was added to the Manager. The Portal cannot support this because there are dynamic custom fields required for developers that we cannot get from the claims.

### Testing
https://deploy-preview-3611--kongdocs.netlify.app/gateway/changelog/#dev-portal

### DIFF
--- a/app/gateway/changelog.md
+++ b/app/gateway/changelog.md
@@ -96,7 +96,7 @@ and use the `/consumer_groups` endpoint to manage the groups.
 #### Dev Portal
 * The Dev Portal API now supports `sort_by={attribute}` and `sort_desc`
 query parameters for sorted list results.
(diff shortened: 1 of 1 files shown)


falondarville (approved):
LGTM

mheap (approved):

PR #3614, opened 2022-02-02
Use quoted struct tags for golang configuration

FreekingDean:
### Summary
This adds quotes to the Golang struct tags.

### Reason
The current Golang config struct has a minor error in that the struct tags are missing. This would cause an issue when encoding/decoding the json configs. Example: https://go.dev/play/p/2hCSyXzXWgR

### Testing
[Here is an example of it in action](https://go.dev/play/p/2hCSyXzXWgR)


FreekingDean:
I am unable to add labels, but I'd assume this would be a technical review :).

mheap (approved):
Thanks for catching this 🙌

PR #3616, opened 2022-02-02
Allow selecting NavTab using the URL

mheap:
### Summary
Allows you to specify the nav tab to activate via the URL

### Reason
So that we can automatically show declarative config when linking to a page from a DB-less context

### Testing
Visit /gateway/2.7.x/get-started/comprehensive/protect-services/?tab=Using%20decK%20(YAML)


mheap:
Potential followup PR:

When someone clicks a tab, update the URL to contain this information so that if they share it the recipient sees exactly the same view

lena-larionova (approved):
LGTM, tested a few different pages, including plugin hub

Can you add info about how to use this to our contributing guide? 
https://docs.konghq.com/contributing/markdown-rules/#tabs

Just needs a couple of lines on how to link to a particular tab set.

You can also add it under "Links" instead of tabs - honestly, I'm not 100% sure which one makes more sense to me.

lena-larionova:
Wait, one thing: this doesn't work with anchor links. 
This works:

https://deploy-preview-3616--kongdocs.netlify.app/gateway/2.7.x/get-started/comprehensive/protect-services/?tab=Using%20decK%20(YAML)

This works:

https://deploy-preview-3616--kongdocs.netlify.app/gateway/2.7.x/get-started/comprehensive/protect-services/#validate-rate-limiting/

This doesn't work:

https://deploy-preview-3616--kongdocs.netlify.app/gateway/2.7.x/get-started/comprehensive/protect-services/#validate-rate-limiting/?tab=Using%20decK%20(YAML)

mheap:
@lena-larionova The `?tab=` has to come before the `#` as it’s a query parameter and is part of the URL rather than a URL fragment

lena-larionova:
> @lena-larionova The `?tab=` has to come before the `#` as it’s a query parameter and is part of the URL rather than a URL fragment

Aha - that would be the kind of useful info to add the contributing guide 😄

mheap:
Thanks for catching some edge cases. I've added some info to the contributing guide and made the following changes:

* The value for `tab` is now a url-friendly version of the tab title
* When there are multiple sets of tabs on a page with different titles, the default tab is now activated on any navtabs that don't match the search
* Support has been added for selecting multiple tabs based on title

PR #3620, opened 2022-02-03
Target deck tabs for db-less getting started guide links

lena-larionova:
### Summary
Follow up on #3608 .

Didn't get the commit in before PR was merged.

### Reason
Better linking.

### Testing
TBA


lena-larionova:
> The behaviour changed a little. Instead of the exact text the `tab` value should now be `?tab=use-deck-yaml`

That's what I get for just pushing a change I prepped last night and not checking again.

mheap (approved):

PR #3621, opened 2022-02-03
doc(kic) add version 2.2

rainest:
Add KIC 2.2 and the new Gateway API guide.

### Summary
Adds KIC 2.2. Non-metadata changes are in the version reference, CLI args, and Gateway API guide.

### Reason
KIC 2.2 will be released. Hold until https://github.com/Kong/kubernetes-ingress-controller/issues/2231 is complete.

### Testing
Netlify review. Gateway steps verified independently.

PR #3625, opened 2022-02-04
Re-add missing alpha and beta banners

lena-larionova:
### Summary
* Add conditional alpha and beta banners to main docs layout
* Update alpha and beta banners on plugins layout
* Create a badge for alpha/tech preview 
  * Went with the name "tech preview" as that's what Konnect is using instead of alpha, but added (alpha quality) in brackets because our open-source projects are more used to this terminology. "Alpha" is also used in decK flags.

### Reason
We're missing `alpha` and `beta` tag handling logic from the `docs-v2.html` layout, although it does exist in the `extension.html` layout. Noted in https://github.com/Kong/docs.konghq.com/pull/3621#issuecomment-1029444084.

### Testing

Tested by applying `alpha: true` and `beta: true` to a few random pages.

Screen Shot 2022-02-03 at 7 21 41 PM

Screen Shot 2022-02-03 at 7 22 01 PM

### DIFF
--- a/app/_assets/stylesheets/badges.less
+++ b/app/_assets/stylesheets/badges.less
@@ -52,6 +52,14 @@
       text-align: center;
     }
 
+  &.alpha::after {
+      content: "TECH PREVIEW";
+      color: white;
+      border-color: @steel-400;
+      background-color: @steel-400;
+      text-align: center;
+    }
+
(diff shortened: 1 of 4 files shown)


falondarville (approved):
LGTM

PR #3635, opened 2022-02-08
Fix issues with archive

lena-larionova:
### Summary
* Bring back accidentally-archived data files for Gateway OSS 0.13.x-1.2.x versions.
* Delete duplicate 2.6.x files from archive

### Reason
Bringing back the nav files as they never should've been moved to the archive. The versions are still up on the site, but because the navs were removed, the docs for those versions are almost completely inaccessible. 

Removing the 2.6.x archived docs as they were temporary, duplicate backups for the single-sourcing project. There is no need for these backups; we have git history and previous versions, in case we ever need to track down something that got lost in the single-sourcing move.

### Testing
Open any version of the Gateway OSS docs from 0.13.x to 1.2.x and check that the nav exists: 
https://deploy-preview-3635--kongdocs.netlify.app/gateway-oss/0.13.x/

Compare to the current broken version, where there is no nav:
https://docs.konghq.com/gateway-oss/0.13.x/


falondarville (approved):
LGTM. Verified that the menu populates.

PR #3682, opened 2022-02-16
[DOCU-493] response trans adv plugin: missing params

falondarville:
### Summary

Add missing parameters to Response Transformed Advanced plugin.

### Reason

Addresses [DOCU-493](https://konghq.atlassian.net/browse/DOCU-493)

### Testing

Requires an engineer to ensure the definition for `config.rename.if_status` is accurate. **EDIT**: Shane provided a correction and I've incorporated it. I've also removed `tags`.

https://deploy-preview-3682--kongdocs.netlify.app/hub/kong-inc/response-transformer-advanced/#parameters

### DIFF
--- a/app/_hub/kong-inc/response-transformer-advanced/index.md
+++ b/app/_hub/kong-inc/response-transformer-advanced/index.md
@@ -54,10 +54,6 @@ params:
   konnect_examples: false
   dbless_compatible: 'yes'
   config:
-    - name: tags
-      required: false
-      datatype: array of string elements
-      description: Set of strings for grouping and filtering.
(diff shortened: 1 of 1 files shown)


eskibars:
Chatted with @falondarville over Slack.  This is the 1 parameter we need to add.  The documentation for the parameter though should be the same as the others `config.*.if_status` parameters, e.g. that it chooses when to apply.  That is, it should say:
> List of response status codes or status code ranges to which the transformation will apply. Empty means all response codes.

lena-larionova (approved):
LGTM

PR #3684, opened 2022-02-16
[DOCU-1618] RM deprecated Target endpoints

falondarville:
### Summary

All the following from EE, CE, and Gateway (single sourced) 2.1.x and newer.

1. Remove endpoints that meet the following form:

```
/targets
/targets/{targets}
/targets/{targets}/upstream
```

2. Remove `run_on`
3. For list targets and add targets, remove `host:port`

### Reason

Addresses [DOCU-1618](https://konghq.atlassian.net/browse/DOCU-1618)

### Testing

Applied to 2.1.x and newer. Here's the latest from our single sourced Gateway docs:

**Note that the indenting was there to begin with - I didn't modify the styling. 

https://deploy-preview-3684--kongdocs.netlify.app/gateway/2.7.x/admin-api/
https://deploy-preview-3684--kongdocs.netlify.app/gateway/2.6.x/admin-api/

### DIFF
--- a/app/enterprise/2.1.x/admin-api/index.md
+++ b/app/enterprise/2.1.x/admin-api/index.md
@@ -2595,15 +2595,46 @@ HTTP 200 OK
 
 ### Retrieve Upstream
 
-##### Retrieve Upstream
-
 <div class="endpoint get">/upstreams/{name or id}</div>
 
 Attributes | Description
 ---:| ---
 `name or id`<br>**required** | The unique identifier **or** the name of the Upstream to retrieve.
 
+*Response*
+
+```
+HTTP 200 OK
+```
+
+```json
+{{ page.upstream_json }}
+```
+
 ---
+
+### Update Upstream
+
+<div class="endpoint patch">/upstreams/{name or id}</div>
+
+Attributes | Description
+---:| ---
+`name or id`<br>**required** | The unique identifier **or** the name of the Upstream to update.
+
+*Request Body*
+
+{{ page.upstream_body }}
+
+
+*Response*
+
+```
+HTTP 200 OK
+```
+
+```json
+{{ page.upstream_json }}
+```
 
 ### Update Or Create Upstream
 
@@ -2645,13 +2676,17 @@ See POST and PATCH responses.
 
 ### Delete Upstream
 
-##### Delete Upstream
-
 <div class="endpoint delete">/upstreams/{name or id}</div>
 
 Attributes | Description
 ---:| ---
 `name or id`<br>**required** | The unique identifier **or** the name of the Upstream to delete.
+
+*Response*
+
+```
+HTTP 204 No Content
+```
 
 ---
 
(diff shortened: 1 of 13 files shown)


lena-larionova (approved):
LGTM, checked through /enterprise/, /gateway-oss/, and /gateway/. Don't see any /target endpoints anywhere other than under /upstreams. 

This is a change that will need a parallel update in the Kong/kong repo as well: 
https://github.com/Kong/kong/blob/master/autodoc/admin-api/data/admin-api.lua#L1604

There's some weird behavior there - it looks like they tried to label these endpoints with `skip: true`, but it isn't working? Will need more digging to figure out what's wrong, and why those endpoints still get generated into docs.

falondarville:
Thanks @lena-larionova. To confirm, should I just remove these endpoints from the most recent version of the docs, or do we regenerate all versions every time (meaning I would remove the endpoints from all versions as per this PR)?

lena-larionova:
> Thanks @lena-larionova. To confirm, should I just remove these endpoints from the most recent version of the docs, or do we regenerate all versions every time (meaning I would remove the endpoints from all versions as per this PR)?

What you've done in this PR is good, it's ready to go. I would recommend merging `main` into this PR first and copying the change to 2.8 as well, and then merge this PR.

My concern going forward is the kong repo side, which seems to have tried to disable the endpoints but it didn't work. That needs investigation.

falondarville:
Roger that. Let's create a separate task for investigating why `skip: true` is not working as expected. 

Will update for 2.8.x.

Re: my question, it looks also like all versions are generated from a single origin so the update only needs to be made in one file.

falondarville:
@kikito, we're trying to remove just the following endpoints from the Kong repo:

```
/targets
/targets/{targets}
/targets/{targets}/upstream
```

Looking at these lines, it looks like removing 1620 may remove endpoints that we need to keep. 
https://github.com/Kong/kong/blob/master/autodoc/admin-api/data/admin-api.lua#L1620

Would you be able to weight in on how we should delete these endpoints without deleting anything else inadvertently? 

`skip = true` is not working as expected.

kikito:
@falondarville Hi, I gave this a look.

> Looking at these lines, it looks like removing 1620 may remove endpoints that we need to keep.

1620 has moved since you first pointed at it. I think you were hinting at removing the following lines:

```
      ["/targets"] = {
        -- This is not using `skip = true` because
        -- we want the sections for GETting targets and POSTing targets to appear,
        -- but we don't want them to appear using `GET /targets` and `POST /targets`.
        -- Instead, we want the section itself to appear, but only the endpoints
        -- generated via foreign keys (`GET /upstreams/:upstreams/targets` and
        -- `POST /upstreams/:upstream/targets`)
        endpoint = false,
      },
      ["/targets/:targets"] = {
        skip = true,
      },
      ["/targets/:targets/upstreams"] = {
        skip = true,
      },
```

I think removing any of those will have the opposite effect - it will make more of the endpoints associated with the targets appear. The `skip = true` and `endpoint = false` lines make (or should make) those endpoints not appear.

I tried regenerating the autodocs locally in the current master and the `endpoint = false` line made the code throw an error, though. Perhaps that's what you were referring to when you said that 

> `skip = true` is not working as expected.

So I added a fix for the error I found here: https://github.com/Kong/kong/pull/8516

PR #3687, opened 2022-02-18
doc(external-plugins) use pip3 and mention API reference

fffonion:
Use pip3 (Python 3.x) in instruction and reference to the API reference link.


mheap:
@fffonion Is the PDK Python3 only?

fffonion:
@mheap It's not, technically it runs on both Py2 and Py3, but as Python2 being deprecated and pip for python3 are named differently (`pip3`), I made the change in our doc to use pip3.

mheap (approved):

PR #3688, opened 2022-02-18
docs(mesh): fix windows installation docs

bartsmykla:
### Summary

Change `Kuma` to `Kong Mesh` as well as the name of env var
from `KUMA_LICENSE_PATH` to `KMESH_LICENSE_PATH`

### Reason

Found out that the docs are wrong

### Testing

It's small PR so looking at other parts of the installation where we are using proper env var should be enough, but if not, we can always test the instructions on windows machine

### DIFF
--- a/app/mesh/1.5.x/installation/windows.md
+++ b/app/mesh/1.5.x/installation/windows.md
@@ -16,9 +16,9 @@ Tested on Windows 10 and Windows Server 2019.
 {:.note}
 > **Note**: Transparent proxying is not supported on Windows.
 
-## 1. Download Kuma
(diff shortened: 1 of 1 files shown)


parkanzky (approved):

lena-larionova (commented):
It might be better to move this PR to the release/mesh-1.6 branch, that way you can add these fixes to both 1.5 and 1.6.
FYI @lahabana

bartsmykla:
According to your suggestions, moved here: https://github.com/Kong/docs.konghq.com/pull/3703

PR #3691, opened 2022-02-18
[DOCU-2192] Forward proxy mTLS support

lena-larionova:
### Summary
* Added `http_proxy_host`, `http_proxy_port`, `https_proxy_host`, and
`https_proxy_port` configuration parameters for mTLS support. 
* Labelling `proxy_host` and `proxy_port` as deprecated.
* Fixed previous plugin version, it was labelled incorrectly. See https://github.com/Kong/kong-ee/blob/master/plugins-ee/forward-proxy/CHANGELOG.md. 

This doc likely needs more detail. What else needs to be configured for an mTLS connection? 

### Reason
Gateway 2.8 includes changes to the forward proxy plugin.
Eng PR: https://github.com/Kong/kong-ee/pull/2950

### Testing
Doc preview: https://deploy-preview-3691--kongdocs.netlify.app/hub/kong-inc/forward-proxy/

### DIFF
--- a/app/_data/extensions/kong-inc/forward-proxy/versions.yml
+++ b/app/_data/extensions/kong-inc/forward-proxy/versions.yml
@@ -1,4 +1,5 @@
-- release: 1.0.6
+- release: 1.2.x
+- release: 1.1.x
 - release: 1.0.5
 - release: 0.36-x
 - release: 0.35-x
--- a/app/_hub/kong-inc/forward-proxy/1.1.x.md
+++ b/app/_hub/kong-inc/forward-proxy/1.1.x.md
@@ -1,7 +1,7 @@
 ---
 name: Forward Proxy Advanced
 publisher: Kong Inc.
-version: 1.2.x
+version: 1.1.x
 desc: Allows Kong to connect to intermediary transparent HTTP proxies
 description: |
   The Forward Proxy plugin allows Kong to connect to intermediary transparent
@@ -16,7 +16,7 @@ categories:
 kong_version_compatibility:
   enterprise_edition:
     compatible:
-      - 2.8.x
+      - 2.7.x
 
 params:
   name: forward-proxy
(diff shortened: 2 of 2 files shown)


pintsized:
The mTLS feature is designed to work in the same way it does without the plugin. Previously this wasn't possible for technical reasons.

Specifically, if a user configures a `service` with a `certificate` entity (cert / key pair), Kong will use them to as part of a mutual TLS handshake, which can be used to identify the client, potentially authenticating the request.

But if they added the `forward-proxy` plugin, it is supposed to behave in the same way end-to-end, it's just that this is tunnelled over a forwarding proxy server. To do this, it has to use a different mechanism under the hood, which did not support mTLS until now.

So the change is that, unlike before, any configured client certificates will be forwarded to the service just like in regular Kong, even though the request is actually going through a plugin and via a forwarding proxy server. There is no new configuration for this, it's just that previously the certificates would not be utilised if configured.

Just shout if you need any further details.

lena-larionova:
> There is no new configuration for this, it's just that previously the certificates would not be utilised if configured.

Just to clarify: you mean nothing new beyond the named `http_` and `https_` configuration options, correct? Or does the plugin only have the `proxy_host` and `proxy_port` options from the user perspective, until 3.0, where they will get replaced with... the new http and https options?

pintsized:
> > There is no new configuration for this, it's just that previously the certificates would not be utilised if configured.
> 
> Just to clarify: you mean nothing new beyond the named `http_` and `https_` configuration options, correct? Or does the plugin only have the `proxy_host` and `proxy_port` options from the user perspective, until 3.0, where they will get replaced with... the new http and https options?

Sorry, yes there is nothing new beyond those configuration options. They are not directly related to the mTLS feature itself, except to the extent that mTLS does require a working https proxy configuration. But to confirm, they are available in 2.8, with the old ones deprecated and marked for removal in 3.0.

I meant only that the mTLS feature does not have any configuration of its own, it uses the certificate entities just like Kong would without the proxy.

GayleNeumann (approved):

PR #3696, opened 2022-02-22
Add backporting tools

mheap:
### Summary
These tools are not officially supported. They were built as a proof of concept and shared in case they're useful to others.

They allow you to backport uncommitted changes to old versions.

Here's how you'd backport changes made in `deck/1.11.x/index.md` to all versions since (and including) `1.9.x`:

```bash
node run.js --oldest 1.9.x --source ../app/deck/1.11.x/index.md
```

You need to use `run.js` once per edited file. Run `git diff` to get the name of the file to pass in.

To backport multiple changes in a single command, you can use a loop:

```bash
OLDEST=2.6.x FILES=../app/gateway; \
for i in $(git diff --name-only $FILES); do node run.js --oldest $OLDEST --source $i; done
```

### Reason
To make backporting changes easier across versions.

### Testing

Check the README.md in this PR and try it locally

Here's a video too:

https://user-images.githubusercontent.com/59130/155161215-4a9df5b5-75e7-4965-9a04-bcab7d623072.mp4

### DIFF
--- a/backporting/README.md
+++ b/backporting/README.md
@@ -1,0 +1,20 @@
+# Backporting Tools
+
+These tools are not officially supported. They were built as a proof of concept and shared in case they're useful to others.
+
+They allow you to backport uncommitted changes to old versions.
+
+Here's how you'd backport changes made in `deck/1.11.x/index.md` to all versions since (and including) `1.9.x`:
+
+```bash
+node run.js --oldest 1.9.x --source ../app/deck/1.11.x/index.md
+```
+
+You need to use `run.js` once per edited file. Run `git diff` to get the name of the file to pass in.
+
+To backport multiple changes in a single command, you can use a loop:
+
+```bash
+OLDEST=2.6.x FILES=../app/gateway; \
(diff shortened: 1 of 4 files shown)


lahabana:
Sweet!!! is the next step is to have a non published "dev" or "next release" and not have to deal with branches anymore?

lahabana:
Also while talking about this for kumahq/kuma-website I was thinking that this should be a github actions rather than a script.
Ideally I'd like to mark a PR with label 'backport-x.y" and a github action executes to add a commit with the backport to the existing PR. This makes it easy to review and reviewers can decide to take the fix to other versions without commiter's interaction.

mheap:
Making this PR driven is on my list, but not for the first iteration. I was thinking about adding slash command support, e.g.

```
/backport min=1.2.x max=1.6.x
```

This would keep the label set nice and small on a repo

RE: `next` branches, we tend to create a `release/x.y.z` branch as the target. We can do that early, or create a `next/mesh` branch for WIP changes if we don't know the next version number

mheap:
Closing this in favour of single sourcing docs across versions

PR #3703, opened 2022-02-23
docs(mesh): fix windows installation docs

bartsmykla:
### Summary

Change `Kuma` to `Kong Mesh` as well as the name of env var
from `KUMA_LICENSE_PATH` to `KMESH_LICENSE_PATH`

### Reason

Found out that the docs are wrong

### Testing

It's small PR so looking at other parts of the installation where we are using proper env var should be enough, but if not, we can always test the instructions on windows machine

### DIFF
--- a/app/mesh/1.5.x/installation/windows.md
+++ b/app/mesh/1.5.x/installation/windows.md
@@ -16,9 +16,9 @@ Tested on Windows 10 and Windows Server 2019.
 {:.note}
 > **Note**: Transparent proxying is not supported on Windows.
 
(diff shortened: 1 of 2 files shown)


lahabana:
@lena-larionova would love a review on today please :)

lahabana:
The links are broken because the release is not released yet

lena-larionova (approved):

lena-larionova:
Checked the links manually, they should work when this is live.

lahabana:
@lena-larionova Thx!

PR #3719, opened 2022-02-28
Update index.md to include details about JWK cache duration

ItsKaleb:
Documentation should indicate how long discovery information remains cached ~~from the created_at time.
Note to Kong engineer, please update PR with required time period value or provide relevant config value name for cache duration.~~

### Summary
Include note in Discovery Cache section around how long the discovery information is cached for

### Reason
Provide visibility over cache behaviour for discovery information

### Testing
N/A


lena-larionova:
@bungle or @flrgh, can you please verify that this information is correct?

app/_hub/kong-inc/openid-connect/index.md
```diff
@@ -1612,7 +1612,7 @@ The OpenID Connect plugin extends the [Kong Admin API][admin] with a few endpoin
 ### Discovery Cache
 
 When configuring the plugin using `config.issuer`, the plugin will store the fetched discovery
-information to the Kong database, or in the worker memory with Db-less.
+information to the Kong database, or in the worker memory with Db-less. Discovery information remains cached until the OIDC plugin is enabled, disabled or updated.
```
lena-larionova, line 1615:
```suggestion
information to the Kong database, or in the worker memory with DB-less. Discovery information remains cached until the OIDC plugin is enabled, disabled, or updated.
```

ItsKaleb:
I have had further update from Kong Support around OIDC behaviour.
So far what I've been told is that:
* The cache is only refreshed for each individual OIDC plugin when that plugin is enabled/disabled/updated in some way.
* The cache does not remove any expired JWK information, meaning that old public keys would still be tried when validating JWT signatures
* The cache does not poll for new JWK information, meaning if a JWK is rolled on the token issuer, the OIDC will not attempt to find a new key in the event that a JWT's signature does not match any existing cached key.

If this can all be verified and updated in the documentation, that would be appreciated. This is quite important information to know, as the implications (at least in my organisation's use case) would mean we could have widespread issues when rolling JWK information on our identity providers.

ItsKaleb:
There is also a feature request raised (FTI-3215) to update the behaviour of the plugin to routinely poll for changes in the JWK information.

lena-larionova:
@flrgh, can you please take a look at https://github.com/Kong/docs.konghq.com/pull/3719#issuecomment-1056021996 and verify if this is correct?

ItsKaleb:
Hi @lena-larionova, @flrgh 
Checking in to see if there's been a chance to confirm what I was told by the Kong Support Engineer?

Thanks

flrgh:
@lena-larionova @ItsKaleb sorry for the delay in response--guess this one slipped through my inbox.

I am not at all familiar with the `openid-connect` plugin code. @bungle any chance you know the answer to this off the top of your head? If not, just drop an :-1: on this comment or something, and I'll read through the code to get an answer for y'all.

flrgh:
@lena-larionova @ItsKaleb sorry for the delay in response--guess this one slipped through my inbox.

I am not at all familiar with the `openid-connect` plugin code. @bungle any chance you know the answer to this off the top of your head? If not, just drop an :-1: on this comment or something, and I'll read through the code to get an answer for y'all.

cloudjumpercat:
> The discovery info is cached in shared memory using Kong's core cache mechanism, so it's probably more accurate to say that the cache persists until Kong is restarted _or_ until [db_cache_ttl](https://docs.konghq.com/gateway/latest/reference/configuration/#db_cache_ttl) seconds have elapsed. Disabling/enabling/updating the plugin while Kong is running won't do anything to the cache.
> 
> @bungle can you please confirm?

@bungle Could you take a look at Michael's comment when you have a chance? Thanks!

ItsKaleb:
> The discovery info is cached in shared memory using Kong's core cache mechanism, so it's probably more accurate to say that the cache persists until Kong is restarted _or_ until [db_cache_ttl](https://docs.konghq.com/gateway/latest/reference/configuration/#db_cache_ttl) seconds have elapsed. Disabling/enabling/updating the plugin while Kong is running won't do anything to the cache.

So what I'm interested in, in this case, is how to have a bit more confidence in the behaviour of Kong when the IdP used by an OpenId Connect plugin rolls its public signing keys. Most IdPs use a cross-over period, where both the old and new public signing keys are presented for a short while. If the Kong cache doesn't expire during this period, the new signing key won't be picked up, and it's possible that the cross-over period will be missed entirely. This would result in Kong's cache retaining an invalid signing key and failing to validate presented tokens that were issued under the new key (which in turn may cause disruption to services proxied by Kong).

Will there be a way to ensure that IdPs are polled to check for key material that isn't in the Kong cache? Or is the only option here to use a shorter db_cache_ttl value (which will impact multiple behaviours). Maybe this is becoming more of a support question than a discussion around documentation, though if the answer lies in the value of the db_cache_ttl setting, perhaps a best-practice/recommendation could be included in the docs for what db_cache_ttl should be set to?

ItsKaleb:
> There is also a feature request raised (FTI-3215) to update the behaviour of the plugin to routinely poll for changes in the JWK information.

Further to my previous post, has there been any updates to the status of this FTI ticket?

cloudjumpercat:
Just an update on this ticket: I'm following up with engineering to see if we can get confirmation on this comment: https://github.com/Kong/docs.konghq.com/pull/3719#issuecomment-1145501701

app/_hub/kong-inc/openid-connect/index.md
```diff
@@ -1612,7 +1612,7 @@ The OpenID Connect plugin extends the [Kong Admin API][admin] with a few endpoin
 ### Discovery Cache
 
 When configuring the plugin using `config.issuer`, the plugin will store the fetched discovery
-information to the Kong database, or in the worker memory with Db-less.
+information to the Kong database, or in the worker memory with DB-less. Discovery information remains cached until the OIDC plugin is enabled, disabled, or updated.
```
bungle, line 1615: I think this is more complicated than this. The save triggers rediscovery. But rediscovery is also controlled by `config.rediscovery_lifetime` (which limits how often `rediscovery` actually happens). Not only that. Rediscovery may also happen when plugin is presented with a token which key is not known.

bungle:
> I have had further update from Kong Support around OIDC behaviour. So far what I've been told is that:
> 
> * The cache is only refreshed for each individual OIDC plugin when that plugin is enabled/disabled/updated in some way.

Where you did get that? When plugin cannot find a new key, it will try rediscovering to find one.

> * The cache does not remove any expired JWK information, meaning that old public keys would still be tried when validating JWT signatures

I think previous are kept for one rotation and then they are gone.

> * The cache does not poll for new JWK information, meaning if a JWK is rolled on the token issuer, the OIDC will not attempt to find a new key in the event that a JWT's signature does not match any existing cached key.

It will attempt to find a new key.

cloudjumpercat:
Hi @ItsKaleb ! Aapo responded to your questions [here](https://github.com/Kong/docs.konghq.com/pull/3719#issuecomment-1385320847) if you wouldn't mind taking a look.

ItsKaleb:
Hi @cloudjumpercat, @bungle.

Unfortunately, this is no longer fresh in my mind. However, upon reviewing the details in the original support case (00026495), I believe that the original concern came about from the discovery of behaviour and a concurring article that OIDC plugin cache data was not refreshed unless the plugin was toggled or a DELETE endpoint was used to manually clear the cache.

This thread was more around confirming the actual behaviour and ensuring that it was articulated in the docs.

The testing done by the Kong support team-member indicated that the JWK cache was not refreshed automatically, however given that test was almost a year ago, it's possible it has been changed since and probably needs re-testing.

Again though, this isn't so much about what the behaviour should be and more around ensuring that the current behaviour is documented to allow proper planning and expectations when rolling IDP signing keys.

Thanks.

app/_hub/kong-inc/openid-connect/index.md
cloudjumpercat, line 1615:
```suggestion
information to the Kong database, or in the worker memory with DB-less. Discovery information is cached except in the following circumstances:
* The plugin can't find a new key, so it retries discovering one.
* JWK information is retained for one rotation before being removed.
* The cache does not poll for new JWK information. This means if a JWK is rolled on the token issuer, the OIDC attempts to find a new key (based on `config.rediscovery_lifetime`) when it cannot find a key. 
* In traditional mode, you can manually delete the issuer cache in the database.
```
@ItsKaleb @bungle Would this work/is this accurate?

ItsKaleb: Hi @cloudjumpercat 

Those changes sound pretty clear to me, though just on the last one, is the cache deletion endpoint not available in Hybrid mode?

Also, as it happens, this conversation became much more relevant this week as my company had a production incident relating to rolling the certificates and the OIDC plugin. I have a new case open with Kong Support to investigate further (00033131). 

Just for visibility, we had the config.rediscovery_lifetime set at 300 seconds, however due to a DNS change our Kong servers were no longer able to reach our IDP. As best as I have been able to tell, it looks like the cache refresh/rediscovery process was failing silently and tokens were being validated based on a very stale cache. I will see what further investigation yields with the support team, but I may have some more edits to suggest based on potential findings.

Thanks.

cloudjumpercat: Hi @ItsKaleb ! I'm not sure about the cache deletion endpoint in hybrid mode, I was just going off this comment from @bungle: "On traditional, it is possible to manually delete the issuer cache in db." 

> I may have some more edits to suggest based on potential findings.

Sounds good! I'll leave this PR open.

ItsKaleb: Hi @cloudjumpercat, I've just committed some new updates to the description for rediscovery_lifetime based on feedback from the active support case. It sounds like perhaps there's some contention around the role of this config field? I've included the response from Kong Support in my commit description. Happy to hear if either yourself or @bungle have further comments around that.
Thanks.

ItsKaleb:
Just towards my last commit, I can understand what's being said about the role of the rediscovery_lifetime value, though it feels a bit un-intuitive. When I hear 'rediscovery lifetime', that makes me think that it's a timer telling Kong that it should rediscover JWK info every _n_ seconds, which would seem to concur more with what @bungle has been indicating above. 

Whereas from the Kong Support response, it sounds like the value is more to do with avoiding high volumes of discovery requests to an IDP by adding a cooldown on the rediscovery attempt. Perhaps rediscovery_cooldown is a more apt name? 

That being said, both functionalities (automated regular rediscoveries and rediscovery cooldown) seem like important features to have...

cloudjumpercat:
> Just towards my last commit, I can understand what's being said about the role of the rediscovery_lifetime value, though it feels a bit un-intuitive. When I hear 'rediscovery lifetime', that makes me think that it's a timer telling Kong that it should rediscover JWK info every _n_ seconds, which would seem to concur more with what @bungle has been indicating above.
> 
> Whereas from the Kong Support response, it sounds like the value is more to do with avoiding high volumes of discovery requests to an IDP by adding a cooldown on the rediscovery attempt. Perhaps rediscovery_cooldown is a more apt name?
> 
> That being said, both functionalities (automated regular rediscoveries and rediscovery cooldown) seem like important features to have...

@ItsKaleb Thanks for making that commit! If you want to recommend changing that parameter name, you could submit a PR to the kong-ee repo and edit this file: https://github.com/Kong/kong-ee/blob/4e7cb5db2b5beb04954a88416dc97182c9392d7a/plugins-ee/openid-connect/kong/plugins/openid-connect/schema.lua#L105

ItsKaleb:
Thanks.
I don't think I'd go as far as to say the variable _should_ be renamed, was more calling that out as a conversation around if my understanding is correct of it being more of a cooldown timer. I've always found I can get a bit nit-picky over naming, so not wanting to derail things here 😄. Plus, renaming config fields always ends up being less trivial than expected, so I don't want to cause headaches there.

Also, while I'd love to help out with raising a PR or investigating plugin behavior in that repo, I believe it is restricted access as it's the enterprise components for Kong.

ItsKaleb:
Just to provide some visibility, here is a summary so far on what has been discussed in support ticket 00033131:
* The OIDC plugin does not periodically poll for new JWK info
* JWK info is discovered 'just in time' when a token requiring a new JWK is presented
* The rediscovery_lifetime field is more of a rediscovery cooldown - the number of seconds that will be waited before retrying a discovery attempt in the event no suitable JWK is found initially
* The OIDC plugin is able to validate tokens for multiple (past) issuers, based on the issuer cache
* The issuer cache should be cleared periodically to maintain predictable behaviour. It does not have any automatic expiry or TTL

app/_hub/kong-inc/openid-connect/index.md
```diff
@@ -233,9 +233,12 @@ params:
       default: 30
       datatype: integer
       description: |
-        Specifies how often (in seconds) the plugin completes a re-discovery.
+        Specifies how long (in seconds) the plugin waits between discovery attempts. Discovery is still triggered on an as-needed basis.
         > The re-discovery usually happens when the plugin cannot find a key for verifying
-        > the signature.
+        > the signature. For example, if a token is presented for which Kong does not have a
+        > JWK cached, it will poll the discovery endpoint for new JWK data. If that discovery
+        > attempt does not yeild a JWK that can validate the token, Kong will wait the specified
```
bungle, line 240:
```suggestion
        > attempt does not yield a JWK that can validate the token, Kong will wait the specified
```

ItsKaleb:
Hi,
I think between the discussion here and over on the support ticket, I am more confident with the understanding around the discovery cache behavior. I've updated the PR with a few more details based on our conversations. If those look accurate and are satisfactory for merging, I'm happy to have this PR closed out.
Thanks.

cloudjumpercat:
@bungle Could you check out the most recent commit and if everything looks good, add your approval? Then we can get this merged :)

bungle (approved):

cloudjumpercat (approved):
LGTM! @ItsKaleb Thanks for submitting these changes and for sticking with this update!

PR #3720, opened 2022-02-28
docs(acme) add version 0.4.0 and add new config option

fffonion:
### Summary
Add ACME plugin 0.4.0 docs.

Note: The new version of plugin is released together with Kong/EE 2.8.x, but we didn't
bump the version in code, this causes slight confusion. But as we have monorepo for bundled plugins at both OSS and
EE, does it make sense to start not having plugin versions and use the Kong version to inidicate version of plugin?


lena-larionova:
> But as we have monorepo for bundled plugins at both OSS and
> EE, does it make sense to start not having plugin versions and use the Kong version to inidicate version of plugin?

That's the plan for the future, hopefully starting with 3.0 (fingers crossed).

lena-larionova:
@fffonion can you please retarget this PR to the `release/gateway-2.8` branch?

lena-larionova:
Did a bit of chore/cleanup:
* Added 0.4.x to the nav
* Fixed filename for 0.3.0 -> 0.3.x
* Added a changelog to call out what Gateway version this came out in

lena-larionova (approved):

fffonion:
Thank you @lena-larionova for taking care of this one!

PR #3738, opened 2022-03-03
Fix broken link in footer

lena-larionova:
### Summary
Fixing a broken link to konghq.com/gateway/. This used to be a redirect but no longer works.

### Reason
Every link check will fail until this is fixed because the footer is on every page.

### Testing
Link check passed.


GayleNeumann (approved):
LGTM

PR #3739, opened 2022-03-03
[DOCU-2217] Better license info in installation guides

lena-larionova:
### Summary
* Add license env variable to docker installation instructions.
  *  If the user wants to, they can simply exclude the variable. If they have no license and leave the setting in, the container still starts with no issues, just without a license.
  * Added links to license docs for other license application options

* Improve license step for distros + a bit of cleanup 
  * Instead of linking to the "Deploy license" topic, included the instructions on the page, so that users have a clear choice depending on their environment.
  * Removed editable placeholders; a few had editable password fields, which isn't secure, and otherwise, we're avoiding using editable placeholders until we overhaul them.
  * Short descriptions for Manager and Dev Portal sections to call out the need for a database.
There's still a lot that needs to be cleaned up in this file, but I had to draw the line somewhere.

Did not do:
* Any changes to the kubernetes/helm/openshift topics, as their license info is already complete afaik.
* Did not add a license step to the MacOS install, as we only document the steps for open-source.

### Reason
Users are struggling to find Enterprise license info.

### Testing
Tested the Docker instructions. Also tested applying a license via API, just in case.

[Docker](https://deploy-preview-3739--kongdocs.netlify.app/gateway/2.8.x/install-and-run/docker/)
[RHEL](https://deploy-preview-3739--kongdocs.netlify.app/gateway/2.8.x/install-and-run/rhel/)
All other distros (Amazon Linux, CentOS, Ubuntu, and Debian) use the same include file, so the output is the same as RHEL.

### DIFF
--- a/app/gateway/2.8.x/install-and-run/docker.md
+++ b/app/gateway/2.8.x/install-and-run/docker.md
@@ -130,6 +130,21 @@ kong:{{page.kong_versions[page.version-index].ce-version}}-alpine kong migration
 
 {% include_cached /md/admin-listen.md desc='long' %}
 
+1. (Optional) If you have an Enterprise license for {{site.base_gateway}},
+export the license key to a variable:
+
+    The license data must contain straight quotes to be considered valid JSON
+    (`'` and `"`, not `’` or `“`).
+
+    {:.note}
+    > **Note:**
+    The following license is only an example. You must use the following format,
+    but provide your own content.
+
+    ```bash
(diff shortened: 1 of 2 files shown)


app/gateway/2.8.x/install-and-run/docker.md
```diff
@@ -147,6 +162,7 @@ docker run -d --name kong-gateway \
  -e "KONG_ADMIN_ERROR_LOG=/dev/stderr" \
  -e "KONG_ADMIN_LISTEN=0.0.0.0:8001" \
  -e "KONG_ADMIN_GUI_URL=http://localhost:8002" \
+ -e "KONG_LICENSE_DATA=${KONG_LICENSE_DATA}" \
```
mheap, line 165: I _think_ you can provide `-e KONG_LICENSE_DATA` and it will read from the environment by default, ignoring it if there is no value available

mheap (approved):

app/gateway/2.8.x/install-and-run/docker.md
lena-larionova: Yes, you're totally right. And I realize that you also commented this in an earlier conversation now, which I missed. Tested, works, updating the doc.

PR #3740, opened 2022-03-03
[DOCU-348] Admin API: rm del service for route

falondarville:
### Summary

Remove Admin API **Delete Service Associated to a Specific Route** endpoint from versions 1.3.x onward.

### Reason

Addresses [DOCU-348](https://konghq.atlassian.net/browse/DOCU-348)

### Testing

Latest:
https://deploy-preview-3740--kongdocs.netlify.app/gateway/2.8.x/admin-api/#delete-service

### DIFF
--- a/app/enterprise/1.3-x/admin-api/index.md
+++ b/app/enterprise/1.3-x/admin-api/index.md
@@ -1106,15 +1106,6 @@ Attributes | Description
 `name or id`<br>**required** | The unique identifier **or** the name of the Service to delete.
 
 
-##### Delete Service Associated to a Specific Route
-
-<div class="endpoint delete">/routes/{route name or id}/service</div>
-
-Attributes | Description
----:| ---
-`route name or id`<br>**required** | The unique identifier **or** the name of the Route associated to the Service to be deleted.
-
-
 *Response*
 
 ```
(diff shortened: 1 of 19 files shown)


lena-larionova (commented):
This is good for the docs side, but this is yet another thing that will need an update in the code repo as well. 

I did a bit of digging in the code repo and here's what I found:
Looks like you can specify a request type to ignore for a specific endpoint, and that works. For example:
https://github.com/Kong/kong/blob/master/autodoc/admin-api/data/admin-api.lua#L1234-L1237

```
      ["/plugins/:plugins/route"] = {
        DELETE = {
          endpoint = false,
        }
```
If you look at our docs, there is no entry for deleting a route via plugins, so we know it works:
https://docs.konghq.com/gateway/2.8.x/admin-api/, search the page for `/plugins/{plugin id}/route`


So you'll want to do the same thing for this endpoint (`["/routes/:routes/service"]`) and add it to the routes section, right above route entity fields for consistency:
https://github.com/Kong/kong/blob/master/autodoc/admin-api/data/admin-api.lua#L930

falondarville:
Are you saying we should add `DELETE = false` rather than delete it outright?

lena-larionova (approved):
LGTM for the docs repo side!

lena-larionova:
> Are you saying we should add DELETE = false rather than delete it outright?

We should do both. The work you did in this PR was necessary, since we don't regenerate old content. To avoid the same issue going forward, we should add `DELETE = false` for this endpoint in the code repo.

PR #3744, opened 2022-03-04
Playwright

mheap:
### Summary
Switch from `rspec` to `playwright` for testing.

### Reason
We're consolidating on JavaScript for testing tools, and switching away from `rspec/rack` to `playwright/netlify dev` allows us to test redirects, headers etc in addition to content

### Testing
See `run-smoke-tests.yml` in GitHub Actions


mheap:
Merged to continue adding tests (e.g. mesh latest_version header check)

PR #3745, opened 2022-03-04
Fixed /licenses/report endpoint to be per license creation date not m…

d19dotca:
### Summary
Corrected the data for the /license/report endpoint (it is not monthly but per license creation date), as well as updated the time stamps and version accordingly to be a bit more recent.

### Reason
This was confusing a customer who had different output from the example, and wasn't seeing monthly reports as it stated. It turns out this is per license creation date to current date (or until next license date listed), not a monthly total.

### Testing
Install a license file in a Kong Gateway and run the /license/report endpoint. Even if it's an old license file it'll show data from the license creation date.

### DIFF
--- a/app/gateway/2.8.x/admin-api/licenses/reference.md
+++ b/app/gateway/2.8.x/admin-api/licenses/reference.md
@@ -35,7 +35,7 @@ HTTP 200 OK
 {
   "data": [
     {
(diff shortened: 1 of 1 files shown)


lena-larionova:
> This was confusing a customer who had different output from the example, and wasn't seeing monthly reports as it stated. It turns out this is per license creation date to current date (or until next license date listed), not a monthly total.

License reporting by month was introduced in 2.7 (see changelog: https://github.com/Kong/docs.konghq.com/blob/main/app/gateway/changelog.md#enterprise-6). What version of Kong Gateway is your client running?

d19dotca:
Ah good catch! The customer was running 2.2, and I had tried in my 2.6 environment, so I guess that explains part of it. We had to go off the documentation for 2.7/2.8 as it was completely missing for the /license/report endpoint in 2.6 and lower in the docs. So perhaps my changes could be pushed to 2.6 and down to 2.2 or even lower possibly?

lena-larionova:
It's unfortunately only documented here for 2.6: 
konghq.com/gateway/2.6.x/plan-and-deploy/licenses/report/ 
And here for 2.3-2.5: 
https://docs.konghq.com/enterprise/2.5.x/deployment/licenses/report/

We can backport your changes to 2.3, but before that, the `/licenses` Admin API doc and endpoints didn't exist. `/license/reports` was the only endpoint that did exist at that time.

mheap:
@d19dotca @lena-larionova Should we proceed with this PR or close it?

lena-larionova:
I think we can close this out. The changes made in the PR don't apply to 2.8, as they were made with 2.6 in mind. The task now, I think, is to make sure that ` /license/report` appears in the /license API reference doc in versions prior to 2.7. I can do that separately.

PR #3760, opened 2022-03-11
Fix typo in allowing-multiple-authentication-methods.md

pastine:
### Summary
I believe that there is a typo in the docs, the following sentence
> Next, we add both Key Auth and Basic Auth plugins to our consumer, and set the anonymous fallback to the consumer we created earlier.

Should be changed for
> Next, we add both Key Auth and Basic Auth plugins to our service, and set the anonymous fallback to the consumer we created earlier.

### Reason
The documentation is saying that we are adding the plugins to the consumer, but the plugins are being added to the service.

### Testing
No need, it's just a documentation typo fix.


pastine:
It seems that I cannot add a label to the PR. I'm not sure why. It should be `review:copyedit` I think

lena-larionova (approved):

PR #3779, opened 2022-03-23
Fixing all traces of old "remote" against "zone"

svenwal:
This document has been very outdated referencing the word "remote" when we have renamed this to "zone" a while back. This includes a long list of variable names which made all copy&paste examples fail


mheap:
@svenwal Do you know which version this change was made in? Then we can backport the changes to the relevant versions

svenwal:
Kuma 1.2 / Kong Mesh 1.3 is when we moved the naming from "Remote" to "Zone"

mheap (approved):

PR #3783, opened 2022-03-23
Update Gateway 2.8.x support dates

mheap:
### Summary
Update Gateway 2.8 release dates to be 12 months later

### Reason
2.8 is the last minor version before the major 3.0 release, so we're adding an additional support period

### Testing
Waiting for review app


eskibars:
LGTM as well.  Thanks

PR #3787, opened 2022-03-25
Corrected the install URL for EE on Ubuntu

d19dotca:
Please see https://kongstrong.slack.com/archives/C0DSXDXV1/p1648243358762779 for more information.

### Summary
The Enterprise tab was pointing to the OSS version of the file, and had the wrong naming convention as well. These are corrected now with this PR.

### Reason
The URLs were incorrect earlier, pointing to a 404 Not Found file instead as the URL wasn't accurate.

### Testing
Can follow the fully formed URLs to test. Was a little difficult for me to see in Preview mode as it didn't contain the full URL, but I believe the changes I made are all that's needed (1: pointing to kong-enterprise-edition in the directory structure, and 2: replacing _amd64 with _all).

### DIFF
--- a/app/gateway/2.8.x/install-and-run/ubuntu.md
+++ b/app/gateway/2.8.x/install-and-run/ubuntu.md
@@ -51,7 +51,7 @@ Install {{site.base_gateway}} on Ubuntu from the command line.
 {% navtabs codeblock %}
 {% navtab Kong Gateway %}
 ```bash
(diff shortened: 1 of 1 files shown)


lena-larionova (changes requested):
The generated URL looks correct for Ubuntu:
https://deploy-preview-3787--kongdocs.netlify.app/gateway/2.8.x/install-and-run/ubuntu/

It looks like the same thing happened with Debian, can you please fix that page as well? https://deploy-preview-3787--kongdocs.netlify.app/gateway/2.8.x/install-and-run/debian/

lena-larionova (approved):
LGTM for this change. Fixing Debian in another PR, since the error is in multiple versions.

PR #3794, opened 2022-03-29
Added note about a feature being PostreSQL-applicable only

d19dotca:
### Summary
Added a note about a feature being PostgreSQL only (i.e. does not apply to Cassandra users).

### Reason
Seems this feature is applicable to PostgreSQL only, based on comments in https://github.com/Kong/kong-ee/pull/2839.

### DIFF
--- a/app/gateway/changelog.md
+++ b/app/gateway/changelog.md
@@ -10,7 +10,7 @@ no_version: true
 
 #### Enterprise
 
-* Improved tables in Kong Manager:
+* Improved tables in Kong Manager: (for PostgreSQL-backed instances only)
   * Click on a table row to access the entry instead of using the
   old **View** icon.
   * Search and filter tables through the **Filters** dropdown, which is located


lena-larionova:
We have [deprecated Cassandra support in 2.7](https://docs.konghq.com/gateway/changelog/#deprecated-1)(ie, no more new features, eventually won't support it at all in 3.x). So starting from that point, shouldn't everything new be assumed to _not_ be supported in Cassandra? We could pretty much say the same thing for all new features in 2.8 - Cassandra-backed? You won't get the new stuff.

mheap:
I think that's a fair assumption, but we should also explicitly call it out to reduce the number of support questions raised

mheap (approved):

d19dotca:
> We have [deprecated Cassandra support in 2.7](https://docs.konghq.com/gateway/changelog/#deprecated-1)(ie, no more new features, eventually won't support it at all in 3.x). So starting from that point, shouldn't everything new be assumed to _not_ be supported in Cassandra? We could pretty much say the same thing for all new features in 2.8 - Cassandra-backed? You won't get the new stuff.

Agreed, however it raises some support requests while customer's don't quite understand why something isn't working as they expect it to, and seems easy to avoid with a quick note. If we're able to add notes periodically until 3.0 or something, that'd be helpful to both customers and CRE. :-) 

For context, this whole thing was discovered because of a customer seeing that filtering / searching broke completely and we narrowed it down to their use of Cassandra. We found the feature in GitHub and it was clear they didn't intend to support Cassandra for it, which is understandable, however by implementing the new feature they inadvertently broke the existing filtering for Cassandra users. I filed an FTI in Jira for that and it's been resolved already by the looks of it: https://konghq.atlassian.net/browse/FTI-3276

PR #3798, opened 2022-03-30
Add konnect_icon helper

mheap:
### Summary
Adds a new helper to show inline Konnect icons

Usage: `{% konnect_icon runtimes %}`

Output:
```
![#{@icon} icon](/assets/images/icons/konnect/icn-#{@icon}.svg){:.inline .konnect-icn .no-image-expand}
```

### Reason
Add a single point of configuration for how inline icons look in the Konnect docs

### Testing
Add `{% konnect_icon runtimes %}` to a page and inspect the HTML. You'll see the `alt` text if the icon doesn't exist

### DIFF
--- a/app/_plugins/konnect_icon_tag.rb
+++ b/app/_plugins/konnect_icon_tag.rb
@@ -1,0 +1,15 @@
+module Jekyll
+  class KonnectIconTag < Liquid::Tag
+
+    def initialize(tag_name, icon, tokens)
+      super
+      @icon = icon.strip
+    end
+
+    def render(context)
+      "![#{@icon} icon](/assets/images/icons/konnect/icn-#{@icon}.svg){:.inline .konnect-icn .no-image-expand}"
+    end
+  end
+end
+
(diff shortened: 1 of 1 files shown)


app/_plugins/konnect_icon_tag.rb
```diff
@@ -0,0 +1,15 @@
+module Jekyll
+  class KonnectIconTag < Liquid::Tag
+
+    def initialize(tag_name, icon, tokens)
+      super
+      @icon = icon
+    end
+
+    def render(context)
+      "![#{@icon} icon](/assets/images/icons/konnect/icn-#{@icon}.svg){:.inline .konnect-icn .no-image-expand}"
```
lena-larionova, line 10: Looks like this is adding an extra space in the filename, so it's generating a broken image:

Screen Shot 2022-03-30 at 12 45 05 PM

https://deploy-preview-3798--kongdocs.netlify.app/konnect/

PR #3817, opened 2022-04-05
Fix consumers groups rate limiting overrides attributes

etienne-napoleone:
### Summary

Fixes consumers groups rate limiting overrides body attributes.

https://docs.konghq.com/gateway/2.7.x/admin-api/consumer-groups/reference/#configure-rate-limiting-for-a-consumer-group

### Reason

They do not mention the keys being nested in a config dict.

incorrect body payload:

```json
{
  "window_size": [60],
  "limit": [200]
}
```

correct body payload:

```json
{
  "config": {
    "window_size": [60],
    "limit": [200]
  }
}
```

### Testing

verify by sending a PUT to the admin endpoint with both payloads mentioned above

### DIFF
--- a/app/gateway/2.7.x/admin-api/consumer-groups/reference.md
+++ b/app/gateway/2.7.x/admin-api/consumer-groups/reference.md
@@ -377,10 +377,10 @@ Attribute                             | Description
 
 Attribute     | Description  
 ---------:    | --------
-`limit`<br>*required*       | An array of one or more requests-per-window limits to apply. There must be a matching number of window limits and sizes specified.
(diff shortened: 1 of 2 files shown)


etienne-napoleone:
only tested on 2.7.x btw, thanks for applying to other relevant versions

lena-larionova (approved):
Thanks for catching this, LGTM! Fix is definitely correct, I've used this endpoint many times myself 😄 . I applied the fix to 2.8.x as well (feature didn't exist before 2.7, so no backporting needed).

PR #3819, opened 2022-04-07
fix(kong-mesh): fix support policy

lahabana:
### Summary
Missing support policy for recent versions of KM


mheap (approved):
Merging due to customer question, but I have a follow up question.

@lahabana Can we have one row for each version from 1.0 to 1.6?

PR #3820, opened 2022-04-07
Improve formatting of API endpoint responses

austinkelleher:
### Summary

Small formatting improvement of API endpoint responses

### Reason

Because the new format is more consistent and easier to read 😬 

### Testing

No functional changes were made.


mheap (approved):
Thanks for the contribution 🙌

PR #3829, opened 2022-04-11
chore(client): update example for kong.client.get_forwared_ip

shettyh:


lena-larionova (approved):

This was fixed in the docs in gateway 2.7.x; will backport the fix to all other instances of this error after merging this PR.

PR #3840, opened 2022-04-13
Update kuma.io links to match new site structure

lena-larionova:
### Summary
Updating all old `/documentation/` links to new navigation for the Kuma site.

Ran a search for all links to the kuma.io site and updated everything that didn't match the new structure.

### Reason
Kuma site navigation was reworked in the previous version but links were never updated, and they don't have redirects on their site - so we're getting tons of 404s: https://github.com/Kong/docs.konghq.com/runs/5998975783?check_suite_focus=true

### Testing
The link checker should pass.

Edit: it passed!


lena-larionova:
Tests all passed - looks like there was some weirdness with gh server during the original build check, so there's a false failed check on the commit. If you look at the checks themselves though, everything passed successfully.

mheap (approved):
I re-ran the check and it passed on the commit too. Thanks for catching this

PR #3842, opened 2022-04-14
Update helm install commands

philnichol:
### Summary

Unless I have a very funky local setup, the command `helm chart values x/x` does not work. Using `helm show values x/x` does work though.

```
$ helm version
version.BuildInfo{Version:"v3.7.1", GitCommit:"1d11fcb5d3f3bf00dbe6fe31b8412839a96b3dc4", GitTreeState:"clean", GoVersion:"go1.17.2"}
$ helm chart values kong-mesh/kong-mesh
Error: unknown command "chart" for "helm"
Run 'helm --help' for usage.
```

### Reason

I was unable to add the label, but it would be `review:copyedit` 😄

### DIFF
--- a/app/mesh/1.6.x/installation/helm.md
+++ b/app/mesh/1.6.x/installation/helm.md
@@ -69,8 +69,8 @@ suggest `kong-mesh-system`.
     deployment, but there are more advanced [deployment modes](https://kuma.io/docs/latest/introduction/deployments/)
     like _multi-zone_.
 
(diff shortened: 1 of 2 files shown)


lena-larionova (approved):
Thanks for catching this - you're totally right, `helm chart values` isn't a thing.

For anyone looking at this in the future, here's the reference doc for `helm show values`: https://helm.sh/docs/helm/helm_show_values/

PR #3845, opened 2022-04-15
Release: decK 1.12

lena-larionova:
### Summary
Create decK v1.12 doc. Using single-source method, so building from src/deck instead of a versioned folder.

No new content in this PR, only prepping the release branch. Will be adding changes to the CLI ref later. 

### Reason

Upcoming decK release. 
https://konghq.atlassian.net/browse/DOCU-2267

### Testing
Netlify preview:
https://deploy-preview-3845--kongdocs.netlify.app/deck/
https://deploy-preview-3845--kongdocs.netlify.app/deck/installation
Tested by building locally.

Please check preview carefully, since this will be the first time we use the `src` folder method in production.

PR #3853, opened 2022-04-18
Add `version` to single sourcing

mheap:
### Summary
Make `page.version` and `page.release` available in single sourced content

### Reason
We need the concrete version number (e.g. `1.2.0`) rather than the release (`1.2.x`) in some places. This PR makes it available

### Testing
I've tested on #3845 and it worked as expected

### DIFF
--- a/app/_plugins/generators/single_source_generator.rb
+++ b/app/_plugins/generators/single_source_generator.rb
@@ -3,18 +3,30 @@
 module SingleSource
   class Generator < Jekyll::Generator
     priority :highest
-    def generate(site)
+    def generate(site) # rubocop:disable Metrics/AbcSize
+      # Load versions data file
+      @kong_versions = SafeYAML.load(File.read('app/_data/kong_versions.yml'))
+
+      # Generate pages
       Dir.glob('app/_data/docs_nav_*.yml').each do |f|
         data = SafeYAML.load(File.read(f))
         next unless data.is_a?(Hash) && data['generate']
 
         # Assume that the whole file should be treated as generated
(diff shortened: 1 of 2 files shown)


lena-larionova (commented):
Tested locally, worked as expected with a full single-sourced doc (`assume_generated: true`). 

With `assume_generated: false`, I had some odd behavior. Not sure if related to this PR or a general issue. When setting that value and setting `generate: true` both at the top level and on a specific item, it never picked up the item. 

```yaml
generate: true
assume_generated: false
product: deck
release: 1.11.x
items:
  - title: Introduction
    icon: /assets/images/icons/documentation/icn-flag.svg
    url: /deck/
    absolute_url: true
    items:
      - text: Terminology
         url: /terminology
        # This setting is ignored. Instead of reading /src/deck/terminology.md, it still reads /app/deck/1.11.x/terminology.md. 
        generate: true
```
Ran through this in a bunch of different ways (setting `src`, deleting the file in /app/deck/1.11.x/terminology.md, trying on other files, trying a file with no content) , and cleared the cache every time. Still the same result: with 'assume_generated: false`, it ignores `generate: true` on individual entries.

mheap:
Good catch. This was a bug from the initial implementation where the code used `generated` and the docs used `generate` and `generated` interchangeable. I've standardised on `generate: true` and the above should now work

lena-larionova (approved):
Unfortunately, I'm still running into the same error. However - I don't think it's anything in this PR. I've tested by building from `main` as well, so I don't think it was just introduced now. 

LGTM for this one; something to investigate for a follow-up.

PR #3861, opened 2022-04-19
Autodoc source_url

lena-larionova:
### Summary
Apply `source_url` to autogenerated files.
* All versions of decK that had autogen files
* All single-sourced versions of the Gateway doc. @mheap wasn't sure if I should backport further - is it worth it to go all the way back? Where do we stop?

Parallel PR in kong repo: https://github.com/Kong/kong/pull/8695
Parallel PR in decK repo: https://github.com/Kong/deck/pull/650

### Reason
To make sure that, when an autogenerated file is edited, the change triggers a warning; see https://github.com/Kong/docs.konghq.com/pull/3841. 

### Testing
Make sure the check triggers on this PR. Netlify preview won't show anything, since this isn't a content change.


lena-larionova:
See comment - it worked! Removed the label, since I have a parallel PR open to update the source files as well.

mheap:
> All single-sourced versions of the Gateway doc. @mheap wasn't sure if I should backport further - is it worth it to go all the way back? Where do we stop?

Single sourced only makes sense to me. We rarely (never?) regenerate older files

mheap (approved):

PR #3869, opened 2022-04-20
Fix missing version-index on /latest/ pages

mheap:
### Summary
Fix missing `version-index` so that the version shows correctly on the install pages

Current prod:

image

With fix:

image

### Testing
https://deploy-preview-3869--kongdocs.netlify.app/gateway/latest/install-and-run/rhel/


Guaris (approved):
LGTM

Screen Shot 2022-04-20 at 2 34 16 PM

PR #3884, opened 2022-04-27
Update URL when using in-page navigation

mheap:
### Summary
Update the address bar + add a history entry when using in-page navigation

### Reason
Fix the “click anchor link, highlight address bar, copy URL from there” muscle memory

### Testing
Try using any sidebar navigation in the review app

image


lena-larionova (approved):
LGTM, tested on main site and plugins.

PR #3889, opened 2022-04-28
Fix proxy-cache config.memory.dictionary_name default value

mheap:
### Summary
Both proxy-caching and proxy-caching-adv plugin say the default value of config.memory.dictionary_name is kong_cache. But it is not correct, please replace it with kong_db_cache in below URL

### Reason
DOCU-2286

### Testing
Review app


Guaris:
I found this value here: 
https://deploy-preview-3889--kongdocs.netlify.app/hub/kong-inc/graphql-proxy-cache-advanced/#strategies

It wasn't listed in the JIRA ticket, but I checked the plugin in Gateway and the default value is `kong_db_cache`

mheap:
Thanks for the catch @Guaris. I think is is good to approve + merge whenever you're happy

PR #3890, opened 2022-04-28
Docu 2281 select gateway version

mheap:
### Summary
Add support for pinning Plugin version to Gateway version, but does **not** update the `versions.yml` file for any plugins other than Basic Auth (which will need updating with actual values. The current values are just a demo)

### Reason
We're switching to pinning the plugin version to the same as the Gateway release. This PR adds all of the code required to make this happen.

### Testing
Check out Basic Auth (I'll add a link) and review the docs to see if they make sense


mheap:
Failing build will be resolved once #3891 is merged and this is rebased

Guaris (approved):

PR #3899, opened 2022-04-30
Update using-kongclusterplugin-resource.md

rkazak:
Removed the '$' which prevented easy cut and paste, differentiated between command and results (  wonder if there is a better way to do that ).  Also an interesting thing I use a mac which defaults to zsh now, unfortunately I wanted to comment the commands but zsh in native mode does not allow comments in the interactive shell ( that was news to me. ) so wonder if we should be recommending `bash` as the execution environment - then we could use comments to aid documentation. Additionally if people like me save terminal session logs the comments will be there to when we cut and paste the commands into a term.

I could not find the template for commands by numbers that seem to appear in some pages, that may also be a thing to include to show the steps?

### DIFF
--- a/app/kubernetes-ingress-controller/2.3.x/guides/using-kongclusterplugin-resource.md
+++ b/app/kubernetes-ingress-controller/2.3.x/guides/using-kongclusterplugin-resource.md
@@ -23,7 +23,10 @@ If everything is setup correctly, making a request to Kong should return
 HTTP 404 Not Found.
 
 ```bash
-$ curl -i $PROXY_IP
+curl -i $PROXY_IP
+```
+results
+```
 HTTP/1.1 404 Not Found
 Content-Type: application/json; charset=utf-8
 Connection: keep-alive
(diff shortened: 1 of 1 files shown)


rkazak:
Here is my shell testing fyi

```
Test2:~ rkazak$ zsh
Test2.local >date # test
date: illegal time format
usage: date [-jnRu] [-r seconds|file] [-v[+|-]val[ymwdHMS]]
            [-I[date | hours | minutes | seconds]]
            [-f fmt date | [[[mm]dd]HH]MM[[cc]yy][.ss]] [+format]
```

```
Test2.local >bash

The default interactive shell is now zsh.
To update your account to use zsh, please run `chsh -s /bin/zsh`.
For more details, please visit https://support.apple.com/kb/HT208050.
Test2.local >date # test
Sat Apr 30 10:36:03 PDT 2022
Test2.local >
```

mheap:
Closing due to inactivity

rkazak:
Sorry - The idea was to separate the output from the commands being executed.

mheap (approved):

PR #3906, opened 2022-05-05
Update Plugin compatibility in Konnect and adjust badges

lena-larionova:
### Summary
* Updating plugin compatibility with Konnect
  * Note for app reg plugin to clarify its relationship with Konnect, as this is one that trips people up a lot  
* Making sure each plugin has both all necessary badges to match the ones on the plugin hub landing page 
  *  Had to move the badges down below the plugin title to fit for some of the longer ones + better alignment. There is a lot of room to redesign this in the future (Angel and I brainstormed some ideas), but for now, we simply need to make sure the info is where it needs to be.
  * Removed the duplicate entry with a badge from the plugin sidebar and rephrased the Konnect compatibility row to make it parallel with the "Db-less compatible?" entry
* Updated the plugin compatibility table and switched the icons to unicode while I was there

### Reason

Announcement from PM (Hayden): The following plugins are no longer available in Konnect cloud (konnect.konghq.com), starting today.

* degraphql
* graphql-rate-limiting-advanced
* vault-auth
* key-auth-enc

Additionally, at the request of PM (quite a while ago), plugins should be consistently labelled with their correct tiers. They are already labelled this way on the plugin hub landing page, but not so on the individual plugin pages. 

For example, compare the badges on the tile for the JWT plugin:

Screen Shot 2022-05-04 at 9 24 40 PM

With the plugin's own page: 
Screen Shot 2022-05-04 at 9 25 02 PM

### Testing

* Check some plugins on the [Plugin Hub landing page ](https://deploy-preview-3906--kongdocs.netlify.app/hub/) and make sure that their tile badges match the badges on their actual pages, eg:
https://deploy-preview-3906--kongdocs.netlify.app/hub/kong-inc/jwt-signer/
  
  Screen Shot 2022-05-04 at 9 38 16 PM
  
  It's not the prettiest design in the world, but it gets the point across.

* Sidebar for any plugin should have no badges and should have a "Konnect Cloud compatible?" row, eg:
   https://deploy-preview-3906--kongdocs.netlify.app/hub/kong-inc/mtls-auth/

  Screen Shot 2022-05-04 at 9 42 43 PM
  
* [Plugin compatibility table with unicode icons and updates to compat details](https://deploy-preview-3906--kongdocs.netlify.app/konnect-platform/compatibility/plugins/)

* [Note in app registration plugin doc](https://deploy-preview-3906--kongdocs.netlify.app/hub/kong-inc/application-registration/)


lena-larionova:
Fixed the broken icons - looks like we weren't generating anything for the `integration` plugin type. 

Also hid the "Plugin version" row for any plugins that don't have versions.  

Re:

> these do not contain the "Konnect Compatible" row. Some of them are 3rd party apps, I wasn't sure if not adding the field on these was intended.

Third-party plugins should not have the "Konnect Compatible" row, we currently don't provide a self-service way to use third-party/custom plugins in Konnect at all. We don't really know if they're compatible or not; the user has to reach out to support if they want to use a custom plugin in Konnect, and support has to test. 

Of the plugins in your list, these are the two that caught my eye:
https://deploy-preview-3906--kongdocs.netlify.app/hub/kong-inc/correlation-id/ - this one looks fine to me, can you double-check?
https://deploy-preview-3906--kongdocs.netlify.app/hub/kong-inc/kong-terraform-aws/ - this plugin is a weird case of originally being developed by an external contributor but now the repo is under Kong's name. It need further review, I don't really know what to do with it at the moment.

Guaris (approved):
lgtm

PR #3909, opened 2022-05-06
Move mode badges to the beginning of list items

nrichers:
### Summary
This bulleted list explains modes and subscriptions. Since the badges for these get introduced here, it might look cleaner to put them at the beginning of the list items rather than weaving them into the sentence structure.

### Reason
Just exploring your docs and GitHub workflow.


lena-larionova (approved):
I like the idea, great suggestion! LGTM

PR #3925, opened 2022-05-11
Fix Ubuntu installation instructions

mheap:
### Summary
Fix Ubuntu install instructions

### Reason
I tried to install Kong based on the current instructions and it didn't work.

### Testing
Follow the instructions on Ubuntu

### DIFF
--- a/app/gateway/2.8.x/install-and-run/ubuntu.md
+++ b/app/gateway/2.8.x/install-and-run/ubuntu.md
@@ -55,18 +55,11 @@ curl -Lo kong-{{page.kong_versions[page.version-index].ce-version}}.amd64.deb "{
 ```bash
 sudo dpkg -i kong-enterprise-edition-{{page.kong_versions[page.version-index].ee-version}}.all.deb
 ```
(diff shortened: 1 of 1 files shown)


Guaris:
There is a small dependency issue when installing kong with the latest version of Ubuntu Jammy.

```
dpkg: dependency problems prevent configuration of kong-enterprise-edition:
 kong-enterprise-edition depends on zlib1g-dev; however:
  Package zlib1g-dev is not installed.
```
 
after installing zlib1g-dev it installs just fine. If you broke the package manager while trying to install Kong though you have to fix it before you can install that dependency.

Guaris:
> @Guaris Interesting, I didn't hit that issue running Jammy on AWS, but I have seen it before on other versions. It's not a Jammy specific issue.
> 
> How would you recommend proceeding? Add to docs or leave it out and let people follow the error message?

I have a gut feeling that if someone has a working Ubuntu installation with other packages, they wont run into this issue. I added a note to address dependency issues if someone is approaching this doc with a brand new ubuntu install.

Guaris (approved):

PR #3928, opened 2022-05-12
Seo sitemap fixes

mheap:
### Summary
Remove URLs that were unexpectedly present in the sitemap.xml file despite having `noindex` set

### Reason
Google's getting confused about whether it should index the pages we list (we shouldn't)

### Testing

The following URLs show up in https://docs.konghq.com/sitemap.xml but not the preview sitemap

* https://docs.konghq.com/gateway/2.6.x/configure/auth/kong-manager/oidc/
* https://docs.konghq.com/mesh/1.6.x/
* https://docs.konghq.com/mesh/1.1.x/overview/

### DIFF
--- a/app/_plugins/generators/canonical_url_generator.rb
+++ b/app/_plugins/generators/canonical_url_generator.rb
@@ -40,7 +40,7 @@ module CanonicalUrl
         unless versioned_product?(url_segments[0])
           all_pages[page.url] = {
             'url' => page.url,
-            'sitemap' => page.data['seo_noindex'] ? false : true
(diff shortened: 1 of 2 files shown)


lena-larionova (approved):
LGTM, sitemap no longer includes those URLs.

This raises a different issue though: we really shouldn't have situations like docs.konghq.com/konnect-platform/compatibility, where we have nested URLs (docs.konghq.com/konnect-platform/compatibility/plugins) but the top level one isn't a real thing. Something to note for a future nav audit.

PR #3930, opened 2022-05-12
Add OpenAPI file for Konnect

mheap:
### Summary
Add OpenAPI reference for the Konnect API

### Reason
So that customers can automate their infrastructure instead of using the UI

### Testing
Try creating/listing/deleting a runtime group through the API


mheap:
APIs are not launching as part of Helium

PR #3932, opened 2022-05-12
Remove deprecated blacklist/whitelist params and single-source plugins

lena-larionova:
### Summary
Testing out single-sourcing while updating three plugins to remove the old `blacklist` and `whitelist` parameters.

* Using `delegate_versions` in each plugin's version file to tell the plugin to pick up all supported gateway versions, which means 2.x-3.x. Using this instead of a selected version list because otherwise it would be something like: 3.0.x, 2.8.x, 2.1.x, 2.0.x - and we'd immediately get questions about "why isn't this plugin supported in 2.6.x?". 
  * The other option I considered was having a version range, eg: 3.0.x, 2.1.x-2.8.x, 2.0.x - however, our URLs are generated based on these versions, so there would be a very unwieldy URL with a range in it.
 
*  Removing the `blacklist` and `whitelist` deprecation warnings from 3.0.x and moving the info to the changelog only.
   * Why remove? When a param is removed (not just deprecated) from the code, the docs should do the same. If we still have users on very old versions, they can reference the changelog.
 
* Set conditionals on each changelog's output. If someone is looking at 2.1.x, they'll only see the changes in+before that version; if they're looking at 3.0.x, they'll see all changes since the start of the changelog.
* Deleting old versions and single-sourcing old content into `_index.md`

### Reason
https://konghq.atlassian.net/browse/DOCU-2296

### Testing
You'll need to check at least the following versions of each plugin: 3.0.x, 2.8.x, 2.1.x, and 2.0.x. There are slight differences in the conditionals between them, and this is the first time we're trying this method.

https://deploy-preview-3932--kongdocs.netlify.app/hub/kong-inc/ip-restriction/
https://deploy-preview-3932--kongdocs.netlify.app/hub/kong-inc/bot-detection/
https://deploy-preview-3932--kongdocs.netlify.app/hub/kong-inc/acl/


app/_hub/kong-inc/ip-restriction/_index.md
```diff
@@ -71,42 +66,108 @@ params:
...
+
+```bash
+$ curl -X POST http://kong:8001/services/{service}/plugins \
+    --data "name=ip-restriction" \
+    --data "config.allow=127.0.0.0/24" \
+    --data "config.deny=127.0.0.1"
+```
+{% endif_plugin_version %}
+
+{% if_plugin_version lte:2.0.x %}
```
mheap, line 141:
```suggestion
 {% if_plugin_version eq:2.0.x %}
```

Not a major change, but given we don't support anything older than 2.0 I think this indicates intent better

app/_hub/kong-inc/bot-detection/_index.md
```diff
@@ -20,18 +20,7 @@ kong_version_compatibility:
       - 2.2.x
       - 2.1.x
       - 2.0.x
-      - 1.5.x
```
mheap, line 23: Version compatibility is hidden when `strategy: matrix`. Can we remove the entire `kong_version_compatibility` entryfrom these plugins?

app/_hub/kong-inc/bot-detection/_index.md
```diff
@@ -1,7 +1,7 @@
 ---
 name: Bot Detection
 publisher: Kong Inc.
-version: 1.0.0
+version: 3.0.x
```
mheap, line 4: I _think_ version can be removed when using single sourcing as it gets injected by the generator

mheap (changes requested):

app/_hub/kong-inc/bot-detection/_index.md
lena-larionova: I want to, but the plugin doc layout and the `hub_cards.html` include both use `kong_version_compatibility` to determine whether a plugin is available in CE or EE gateway versions. So, short answer - yes, we can remove, but not without a bit of rework and preferably not in this PR.

mheap (approved):

PR #3960, opened 2022-05-20
[DOCU-2349] GraphQL server path

lena-larionova:
### Summary
* Add new `graphql_server_path` parameter
* Set up version file so that we can single-source this plugin going forward
* Some grammar, formatting, and style cleanup

### Reason
https://konghq.atlassian.net/browse/DOCU-2349

### Testing
https://deploy-preview-3960--kongdocs.netlify.app/hub/kong-inc/degraphql/

### DIFF
--- a/app/_hub/kong-inc/degraphql/_index.md
+++ b/app/_hub/kong-inc/degraphql/_index.md
@@ -1,7 +1,6 @@
 ---
 name: DeGraphQL
 publisher: Kong Inc.
-version: 0.1.0
 desc: Transform a GraphQL upstream into a REST API
 description: |
   This plugin transforms a GraphQL upstream into a traditional endpoint by mapping URIs into GraphQL queries.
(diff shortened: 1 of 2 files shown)


app/_hub/kong-inc/degraphql/_index.md
```diff
@@ -27,113 +26,127 @@ params:
   name: degraphql
   service_id: true
   dbless_compatible: 'yes'
-  config: null
+  config:
+    - name: graphql_server_path
+      required: false
+      default: "/graphql"
+      datatype: string
+      description: The path to the GraphQL server.
```
lena-larionova, line 34: For reviewers: I don't know if there's much more to say about the server path param. Maybe an example? But even then, there's already one in the "default" field.

app/_hub/kong-inc/degraphql/_index.md
```diff
@@ -27,113 +26,127 @@ params:
   name: degraphql
   service_id: true
   dbless_compatible: 'yes'
-  config: null
+  config:
+    - name: graphql_server_path
+      required: false
```
lena-larionova, line 31: I set this as "false" even though the code has it as required. It's been pointed out that entries with a default value that are labelled as required are confusing to users - they're required to set the value, but the value is set? This makes no sense.

Tieske: I think we better stay consistent. A value being required means you cannot set it explicitly to a JSON `null` value. 

I agree it’s not intuitive, but it certainly makes sense.

aboudreault: +1 to preserve `required: true`. But yeah... totally agree with you Lena. we should probably how to document this in a more intuitive way :thinking:

lena-larionova: > we should probably check how to document this in a more intuitive way 🤔

We'll give this bit some more thought. For now, I'll set it to `required: true`.

app/_hub/kong-inc/degraphql/_index.md
```diff
@@ -27,113 +26,127 @@ params:
...
-### 2. Configure the Plugin on the Service
+```bash
+curl -X POST http://localhost:8001/services \
+  --data name="github" \
+  --data url="https://api.github.com"
 
-The plugin takes over the service. From this point on, the service represents
-our REST api and not the graphql endpoint itself. It will return a 404 Not Found
-if no DeGraphQL routes have been configured.
+curl -X POST http://localhost:8001/services/github/routes \
```
mheap, line 54: Why do we configure a route if degraphql takes over the service later?

app/_hub/kong-inc/degraphql/_index.md
```diff
@@ -27,113 +26,127 @@ params:
...
+```bash
+curl -X POST http://localhost:8001/services/github/plugins \
+  --data name="degraphql"
+```
 
-  ```bash
-  $ curl -X POST http://localhost:8001/services/github/degraphql/routes \
-    --data uri="/me" \
-    --data query="query { viewer { login } }"
+The plugin takes over the Service. From this point on, the Service represents
```
mheap, line 67: > The plugin takes over the Service

I'm not sure what this is telling me. I _think_ it's something like:

```suggestion
Enabling the degraphql plugin will disable all existing routes in your service. From this point on, the Service represents
```

mheap (dismissed):

app/_hub/kong-inc/degraphql/_index.md
lena-larionova: @aboudreault would you know the answer to this?

app/_hub/kong-inc/degraphql/_index.md
Tieske: More like this:
Enabling the degraphql plugin will disable the regular service, instead the plugin will now build the path and graphql query to hit the Graphql service with.

Tieske (approved):
thx @lena-larionova !

PR #3967, opened 2022-05-24
Fix PDF Generation for plugins + docs

mheap:
### Summary
PDF generation is broken in two ways:

1. The switch from `index` to `_index` for plugins prevented PDF generation for the default version of plugins
2. The addition of the OpenAPI renderer prevented that page from being printed to PDF

Some `prettier` printing also made it in to the PR, but I don't think it's a bad thing

### Reason
Fixing the existing tooling

### Testing

Run this locally:

```
cd pdf-generation
npm ci

# Build a plugin PDF
KONG_PLUGIN_NAME=basic-auth node run.js

# Build a docs PDF
KONG_DOC_VERSIONS=gateway_2.8.x node run.js
```

### DIFF
--- a/pdf-generation/create-pdf.js
+++ b/pdf-generation/create-pdf.js
@@ -1,51 +1,74 @@
-const puppeteer = require('puppeteer')
-const PDFMerger = require('pdf-merger-js')
+const puppeteer = require("puppeteer");
+const PDFMerger = require("pdf-merger-js");
 
-async function createPDF (name, urls) {
-  const merger = new PDFMerger()
+async function createPDF(name, urls) {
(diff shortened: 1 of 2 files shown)


Guaris (approved):
**issue (non-blocking)**: Numbered steps dont render. Marking non-blocking because I believe it is unrelated to this fix. 
Screen Shot 2022-05-24 at 4 58 34 PM
Screen Shot 2022-05-24 at 4 58 34 PM

mheap:
Numbers not rendering is unrelated to this fix, and PDF generation is infrequent enough that we should not invest time fixing it imo

PR #3974, opened 2022-05-25
Single source deck 

Guaris:
### Summary
Single sources every doc in the deck section. 

## Test
https://deploy-preview-3974--kongdocs.netlify.app/deck/

The biggest difference can be seen in the CLI reference


mheap:
@Guaris I've fixed the issue with the index pages not rendering. I made some URL logic changes (`/deck/latest/` is canonical rather than `/deck/`) which I avoided previously to reduce risk, but were required to make this work. I've tested and everything seems to be working.

mheap (approved):

PR #3987, opened 2022-05-30
Adding Vale to CI/CD

Guaris:
### Summary
This PR adds vale. 

### Reason
We need vale to enforce consistency in an automated way. 

### Testing
locally: 
`vale TARGET_FILE` 

Github: 
modify a file and make a commit. 

I've deleted all of our other tests in this PR. When this is ready to be merged, I can make a new PR with those included.


.github/workflows/vale_linter.yml
```diff
@@ -0,0 +1,18 @@
+name: textlinting
```
mheap, line 1: GH Action workflow filenames are `kebab-case` rather than `snake_case`

.github/styles/kong/OxfordComma.yml
```diff
@@ -0,0 +1,12 @@
+# Warning: OxfordComma
```
mheap, line 1: Do filenames need to be `TitleCase`? Case sensitivity can cause issues on some filesystems so we tend towards all lowercase with `-` separators

.github/styles/kong/Terms.yml
```diff
@@ -0,0 +1,8 @@
+extends: substitution
+message: Use '%s' instead of '%s'.
+level: error
+ignorecase: true
+swap:
+  kong: Kong
+  k8s: Kubernetes
+
```
mheap, line 8: Should we extend this to capitalise all products (e.g. Konnect)?

Even better, should we prevent direct usage of Konnect and suggest the config variable for that product?

mheap (changes requested):
This will be a great addition.

Should we choose a base ruleset to start from, or build one ourselves?

mheap:
Updates based on a call yesterday:

* Run on all modified files
* When adding a new rule, the person adding it must ensure that the `src` directory passes all checks

Could we also get some docs in the `docs` folder, including instructions on how to install and run Vale locally (bonus points if there are instructions on how to integrate with VSCode too)

Guaris:
> Updates based on a call yesterday:
> 
> * Run on all modified files
> * When adding a new rule, the person adding it must ensure that the `src` directory passes all checks
> 
> Could we also get some docs in the `docs` folder, including instructions on how to install and run Vale locally (bonus points if there are instructions on how to integrate with VSCode too)

Added all of this information to the doc.

.github/styles/kong/Terms.yml
```diff
@@ -0,0 +1,17 @@
+extends: substitution
+message: Use '%s' instead of '%s'.
+level: error
+ignorecase: false
```
lena-larionova, line 4: I wonder if we should have two substitution lists: one that ignores case, and one that's specifically there to fix case. I can't get the `[Bb]` syntax to work currently - I don't know that it's supposed to work in a key?

mheap:
Is there a way to stop duplication when an item is missing from the dictionary, but is present in substitutions? e.g.
image

---

Should all content within backticks be ignored? e.g. `file.yml` will currently throw a `kong.Spelling` error "Did you really mean 'yml'?". I _think_ this would be sensible

---

Can we add a block ignore that we can use to exclude specific sections? In this case, parts of the `docs/vale.md` file where we show what _not_ to do. Something like the following:

```
<!-- VALE_IGNORE -->
 Proper nouns that must be capitalized, or written in a specific way can be handled using a combination of `dictionary.txt` and `terms.yml`. In the case of a word like "k8s", where you want to suggest every instance of k8s be changed to "Kubernetes" in `terms.yml` create a rule like this: 
<!-- END_VALE_IGNORE -->
```

Without a block ignore, we'll get a warning that `k8s` should be `Kubernetes` every time we modify that docs file.

docs/vale.md
```diff
@@ -153,15 +150,21 @@ TokenIgnores = {%.*?%}, \
 *  `{{.*?}}`: Ignores instances where we use `{{ example }}` and the pattern **inside** of the brackets. 
 
 
+## Ignoring text 
+
+Vale can be forced to ignore sections of text using the following syntax:
+<!-- vale off -->
```
mheap, line 156: This doesn't show on the page. Does it need to be inside a code block?

Guaris: Nice catch, yes.

mheap (approved):

PR #3990, opened 2022-06-01
Combine plugin examples into tabs for better UX

lena-larionova:
### Summary
An attempt at combining the plugin config examples into something that's hopefully easier to navigate/understand.

You can set the `parent` class on a group of `navtabs` to surround it with a border. This makes it clearer that the tabs contain tabs.

Previously:

Screen Shot 2022-05-20 at 1 33 23 PM

Reworked:

https://user-images.githubusercontent.com/54370747/171506991-09cd8f27-2fcb-4394-80a8-0038d7b2cd1a.mov

### Reason
Most recently, the issue was raised by Rick. However, this issue has been called out many times in the past: we have too many similar examples and they cover a huge portion of the plugin page while being mostly the same. 

Feedback from Rick: to a new user, it looks like each section has to be done in order (enable on a service, then on a route, etc). It's not immediately clear that a) these instructions are independent from one another, and b) these are typical examples only, and not necessarily the best/only way to configure a plugin.

https://konghq.atlassian.net/browse/DOCU-2350

### Testing
Look at any kong plugin, eg https://deploy-preview-3990--kongdocs.netlify.app/hub/kong-inc/key-auth/

### DIFF
--- a/app/_includes/hub-examples.html
+++ b/app/_includes/hub-examples.html
@@ -115,7 +115,7 @@ endif %}{% endif_plugin_version %}{% endfor
 
 {% capture plugin_config_for_service %}
 
-The following examples provide some typical configuration for enabling
+The following examples provide some typical configurations for enabling
 the <code>{{include.params.name}}</code> plugin on a
 [service](/gateway/latest/admin-api/#service-object).
 
@@ -250,7 +250,7 @@ is not already prefilled.{% unless
 {% if include.params.route_id %}
 {% capture plugin_config_for_route %}
 
-The following examples provide some typical configuration for enabling
+The following examples provide some typical configurations for enabling
 the <code>{{include.params.name}}</code> plugin on a
 [route](/gateway/latest/admin-api/#route-object).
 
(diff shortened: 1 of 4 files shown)


app/_includes/hub-examples.html
```diff
@@ -250,12 +249,15 @@
 
 {% if include.params.route_id %}
 {% capture plugin_config_for_route %}
-### Enable the plugin on a route
+
+The following examples provide some typical configuration for enabling
```
Guaris, line 253:
```suggestion
The following examples provide some typical configurations for enabling
```

app/_includes/hub-examples.html
```diff
@@ -383,12 +382,15 @@
 
 {% if include.params.consumer_id %}
 {% capture plugin_config_for_consumer %}
-### Enabling the plugin on a consumer
+
+The following examples provide some typical configuration for enabling
```
Guaris, line 386:
```suggestion
The following examples provide some typical configurations for enabling
```

app/_includes/hub-examples.html
```diff
@@ -495,17 +495,18 @@
...
 
 {% capture plugin_global_config %}
-### Enable the plugin globally
 
 A plugin which is not associated to any service, route, or consumer is
 considered _global_, and will be run on every request. Read the
 [Plugin Reference](/gateway/latest/admin-api/#add-plugin) and the [Plugin Precedence](/gateway/latest/admin-api/#precedence)
 sections for more information.
 
+The following examples provide some typical configuration for enabling
```
Guaris, line 504:
```suggestion
The following examples provide some typical configurations for enabling
```

Guaris (changes requested):
**issue (grammar):** One small nitpick configuration -> configurations.

PR #3993, opened 2022-06-06
Show only EE plugins selector

mheap:
### Summary
Add a toggle for showing all/only plugins for a specific tier.

It doesn't work well with `plus` - I need to do more research to see if we have any plus-only plugins. 

https://user-images.githubusercontent.com/59130/172216560-8f9d9107-7e92-4c48-b8a6-97cd7ee76109.mp4

### Reason
Requested feature

### Testing
Visit /hub/ in the review app


lena-larionova:
I wonder if "plus" isn't working because it's a two-part selector that's also excluding certain plugins:

```'
{% if extn.plus or extn.kong_version_compatibility.community_edition.compatible != nil %}
  {% unless extn.cloud == false %}plus {% endunless %}
{ % endif %}
```

Whereas enterprise has two different selectors, where the two `if` statements aren't dependent on each other. :
```
{% if extn.kong_version_compatibility.enterprise_edition.compatible != nil %}ee-compat {% endif %} 
{%  if extn.enterprise %}enterprise  {% endif %}
```

`ee-compat`: pulls all compatible plugins, including open-source
`enterprise`: enterprise-level, not open-source

lena-larionova:
Screen Shot 2022-06-07 at 9 32 26 AM

If you select either the "konnect" or "third-party" filter, you get this message. I want to say that we don't want a banner there at all, but what do you think?

app/_assets/javascripts/app.js
```diff
@@ -336,16 +336,84 @@ jQuery(function () {
...
+
+  var tierMap = {
+    "plus": "Plus",
+    "ee-compat": "Enterprise"
+  }
+
+  var content = allOnlyToggle.find("blockquote");
+
+  var other = pluginsAllOrOnly == "all" ? "only" : "all";
+  content.html("Showing <strong>"+pluginsAllOrOnly+"</strong> plugins available on the " + tierMap[target] + " tier. <u>Switch to show <strong>"+other+"</strong> " + tierMap[target] + " plugins?</u>")
```
lena-larionova, line 369: For Enterprise plugins: we're kind of stuck in awkward place, where the on-prem Enterprise tier includes all Plus and Enterprise plugins. Since there's no separate Plus tier for Kong Gateway alone, the filter that selects Enterprise _only_ isn't accurate for that group. A couple of approaches we could take:

* Have two Enterprise filters: Konnect Enterprise and Gateway Enterprise
* Label the "Subscription tiers" section as "Konnect subscription tiers" only
* Have two links in the Enterprise banner, something like: 

  >  Showing plugins only available on the Konnect Enterprise tier. Switch to show Kong Gateway Enterprise plugins, or all Enterprise plugins?

Now that I look again, I'm not really a fan of options 2 and 3. So maybe we need another filter.

mheap: I've added a second filter when you open up `Enterprise` to enable showing plus too. It's not ideal, but I prefer it there to splitting on-prem and cloud enterprise in to separate filters at this point. I'd like to keep the changes to a minimum

Guaris:
I'm struggling with the different states. 

>Showing all plugins available on the Enterprise tier. Show Enterprise-only plugins?

Is this showing me **Enterprise Only** + ** free Tier** with the implication that the enterprise tier has access to free plugins? 

> Showing plugins only available on the Enterprise tier. Show all Enterprise plugins?

When I click to through to this state, it is clear that these are exclusively the Enterprise-only plugins. 

> Showing plugins only available on the Enterprise tier. Show all Enterprise plugins?
> If you run Kong Gateway on your own infrastructure, you also have access to all Plus plugins. Click Here to show them

Then this combined message state has an option to show me Plus + Enterprise but only in the case that I'm using my own infrastructure.

If I am understanding this correctly then we may be showing too many states. 

What if we kept these three states?
> Only Plus
> Only Enterprise 
> Enterprise + Plus with the banner that says people using their own infra have access to all of the plus plugins? 


and got rid of these two states? 
> All Plugins available to plus
> All Plugins available to Enterprise 

Because the two "all" states if I'm understanding correctly contain tier + free so maybe that is something that needs to be expressed within the free section? "Free plugins are available across all tiers"

mheap:
We have three tiers available:

Free = Free only
Plus = Free + Plus (only available on cloud)
Enterprise = Free + Plus + Enterprise (cloud + on-prem)

> Is this showing me Enterprise Only + ** free Tier** with the implication that the enterprise tier has access to free plugins?

Yes, Enterprise Only + Free Tier (_and_ Plus). Enterprise have access to **all** plugins

> and got rid of these two states?

These states currently exist, and the change is intended to be additive rather than changing existing behaviour. There is a use case where Free/Plus/Enterprise customers want to see all plugins that they have access to _on a single screen_.

Guaris:
> Yes, Enterprise Only + Free Tier (_and_ Plus). Enterprise have access to **all** plugins

Then could the orange banner be changed to
> Kong Enterprise  users have access to all Plugins click Here to show them
Or is the infrastructure piece important?

mheap:
We can't change it to all plugins, as we don't show free. This is related to Lena's comment above:

> For Enterprise plugins: we're kind of stuck in awkward place, where the on-prem Enterprise tier includes all Plus and Enterprise plugins

So Enterprise on-prem is both Plus and Enterprise, whereas Enterprise on Cloud is just Enterprise - if we're showing the plugins only added by that tier. Enterprise gets access to all plugins when viewing everything they have access to, rather than just a specific tier

app/_assets/javascripts/app.js
```diff
@@ -336,16 +336,113 @@ jQuery(function () {
...
+
+  allOnlyToggle.show();
+
+  if (!allOnlyToggle.length){
+    allOnlyToggle = $('<div id="plugin-banner" class="content" style="padding:0"><blockquote class="note"></blockquote></div>');
+    hubCards.prepend(allOnlyToggle)
+  }
+
+  if (!eePlusBanner.length){
+    eePlusBanner = $('<div id="ee-plus-banner" class="content" style="padding:0"><blockquote class="important">If you run Kong Gateway on your own infrastructure, you also have access to all <strong>Plus</strong> plugins. <u id="show-plus-ee">Click Here</u> to show them</blockquote></div>');
```
lena-larionova, line 367:
```suggestion
    eePlusBanner = $('<div id="ee-plus-banner" class="content" style="padding:0"><blockquote class="important">If you run Kong Gateway on your own infrastructure, you also have access to all <strong>Plus</strong> plugins. <u id="show-plus-ee">Show Plus plugins?</u> </blockquote></div>');
```
Minor suggestion for consistency, and we should avoid using "click here".

lena-larionova (approved):
One minor edit, otherwise LGTM.

PR #3989, opened 2022-05-31
chore: release kong mesh 1.8.0 and 1.7.1

bartsmykla:
### Summary
Prepare `release/mesh-1.8` branch for the real release with changelog updated etc.

- [x] ~~blocked by: https://github.com/kumahq/kuma-website/pull/837~~

### Reason
n/a

### Testing

n/a

PR #3997, opened 2022-06-07
docs(*): added logic that detect arch for kong-mesh

lukidzi:
### Summary

Detect arch type of the machine.

### Reason
Since `kong-mesh` 1.8 we support arm64, and we didn't detect the arch type.

### Testing
`./installer.sh | VERSION=1.8.0-rc1  sh - `


lahabana:
Superseeded by: #3989

PR #4015, opened 2022-06-14
docs(km) note Kong Manager perf issue in traditional mode

mayocream:
### Summary

Add a note for the Kong Manager perf issue in traditional mode.

### Reason

We are telling the clients that using Kong Manager in traditional mode may have an impact on the overall performance of the Kong.

INTF-2900

### DIFF
--- a/app/gateway/2.8.x/index.md
+++ b/app/gateway/2.8.x/index.md
@@ -18,9 +18,9 @@ With {{site.base_gateway}}, users can:
 * Create a thriving API developer ecosystem
 * Proactively identify API-related anomalies and threats
 * Secure and govern APIs/services, and improve API visibility across the
-entire organization
+entire organization ss
 
-## Extending the Kong Gateway
+## Extending the {{site.base_gateway}}
 
 {{site.base_gateway}} is a Lua application running in Nginx. {{site.base_gateway}}
 is distributed along with [OpenResty](https://openresty.org/), which is a bundle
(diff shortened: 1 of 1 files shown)


mheap:
@mayocream Why is this only an issue in traditional mode as opposed to Hybrid mode?

mayocream:
> @mayocream Why is this only an issue in traditional mode as opposed to Hybrid mode?

I might not express it clearly; the Kong Manager (Lua code in EE) consumes CPU, I/O resources to calculate and handle frontend user operations, if the client runs in the traditional mode (not Hybrid), this would increase the Kong Proxy's P99 latency.

But in the Hybrid mode, the Kong Manager only runs in the CP node, not impact on the DP performance.

mayocream:
> @Guaris I wouldn't look for this information in the Hybrid mode docs. It'd work well in a Kong Manger section if we have one

We had a [Kong Manager section (<= 2.5.x)](https://github.com/Kong/docs.konghq.com/tree/main/app/enterprise/2.5.x/kong-manager), but got removed in the latest docs :( 

@Kong/team-km will probably start a conversation internally about this.

Guaris:
Understood, then this is a good place for now.

Guaris (approved):

mheap:
@Guaris Could we issue a followup PR explaining why proxy performance may be impacted in traditional mode and why it's not an issue in hybrid mode?

PR #4023, opened 2022-06-15
[DOCU-2341] Improve Okta issuer URL instructions

lena-larionova:
### Summary
* Clarify exactly which issuer URL to use for Okta SSO configuration. 
* Create a troubleshooting topic for account/org issues - we have two pieces of Okta config advice that are buried in the config topic itself. Pulling them into a troubleshooting topic for _hopefully_ easier findability.

### Reason
https://konghq.atlassian.net/browse/DOCU-2341

### Testing
https://deploy-preview-4023--kongdocs.netlify.app/konnect/org-management/okta-idp/#set-up-konnect
https://deploy-preview-4023--kongdocs.netlify.app/konnect/org-management/troubleshoot/

### DIFF
--- a/app/konnect/org-management/okta-idp.md
+++ b/app/konnect/org-management/okta-idp.md
@@ -163,13 +163,13 @@ values are present.
     something like this:
 
         ```
-        https://example-account.okta.com/oauth2/default
+        https://example.okta.com/oauth2/default
         ```
         Where `default` is the name or ID of the authorization server.
 
         {:.note}
         > Note: Do not use the issuer URI from your application's settings. That
-        URI is incomplete: `https://example-account.okta.com`.
+        URI is incomplete: `https://example.okta.com`.
 
 1. Paste the issuer URI from Okta into the **Issuer URI** field in {{site.konnect_short_name}}.
 1. Copy and paste the **Client ID** and **Client Secret** from your Okta
--- a/app/konnect/org-management/okta-idp.md
+++ b/app/konnect/org-management/okta-idp.md
@@ -83,7 +83,7 @@ claims to extract that information.
 
 1. From the left menu, select **Security > API**.
 
-1. Select the Custom Authorization Server that you want to configure.
+1. Select the authorization server that you want to configure.
 
 1. Go to the Claims tab.
 
@@ -157,8 +157,21 @@ values are present.
 
     Refer back to your Okta application to fill in the following fields.
 
-1. Copy the **Issuer URI** from your Okta custom authorization server, then paste it into
-the **Issuer URI** field in {{site.konnect_short_name}}. <!-- found in general settings in Okta app -->
+1. Locate your issuer URI in Okta.
+    1. Go to **Security** > **API**.
+    1. Copy the issuer URI for your authorization server. It should look
+    something like this:
+
+        ```
+        https://example-account.okta.com/oauth2/default
+        ```
+        Where `default` is the name or ID of the authorization server.
+
+        {:.note}
+        > Note: Do not use the issuer URI from your application's settings. That
+        URI is incomplete: `https://example-account.okta.com`.
+
+1. Paste the issuer URI from Okta into the **Issuer URI** field in {{site.konnect_short_name}}.
 1. Copy and paste the **Client ID** and **Client Secret** from your Okta
 application into {{site.konnect_saas}}.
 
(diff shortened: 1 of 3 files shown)


app/konnect/org-management/troubleshoot.md
```diff
@@ -0,0 +1,43 @@
...
+groups from third-party applications (for example, Google groups).
+
+An Okta administrator needs to duplicate the third-party groups
+and re-create them directly in Okta. They can do this by exporting the group
+in CSV format, then importing the CSV file into Okta to populate the new group.
+
+### Named cookie not present
+
+#### Problem
+If the issuer URL is incorrect or incomplete, you may get the following error
```
adamdehaven, line 25: question: should this say “issuer URI” instead of “URL” to match the instructions?

lena-larionova: Yep, that would be a typo.

app/konnect/org-management/okta-idp.md
```diff
@@ -157,8 +157,21 @@ values are present.
...
 
-1. Copy the **Issuer URI** from your Okta custom authorization server, then paste it into
-the **Issuer URI** field in {{site.konnect_short_name}}. <!-- found in general settings in Okta app -->
+1. Locate your issuer URI in Okta.
+    1. Go to **Security** > **API**.
+    1. Copy the issuer URI for your authorization server. It should look
+    something like this:
+
+        ```
+        https://example-account.okta.com/oauth2/default
```
Guaris, line 166:
```suggestion
        https://EXAMPLE_ACCOUNT.okta.com/oauth2/DEFAULT
```

Guaris: okay feel free to ignore this because I think it looks ugly but we have to figure out what to do in these cases

app/konnect/org-management/troubleshoot.md
```diff
@@ -0,0 +1,43 @@
...
+
+### Groups claim is empty when Okta is the IdP
+
+#### Problem
+In the token payload, the `groups` claim is configured correctly but still
+appears empty, or it includes some groups but not all.
+
+#### Solution
+This issue might happen if the authorization server is pulling in additional
+groups from third-party applications (for example, Google groups).
```
Guaris, line 16:
```suggestion
groups from third-party applications, for example, Google groups.
```

app/konnect/org-management/troubleshoot.md
```diff
@@ -0,0 +1,43 @@
...
+
+An Okta administrator needs to duplicate the third-party groups
+and re-create them directly in Okta. They can do this by exporting the group
+in CSV format, then importing the CSV file into Okta to populate the new group.
+
+### Named cookie not present
+
+#### Problem
+If the issuer URI is incorrect or incomplete, you may get the following error
```
Guaris, line 25:
```suggestion
If the issuer URI is incorrect or incomplete, you will get the following error
```

app/konnect/org-management/troubleshoot.md
```diff
@@ -0,0 +1,43 @@
...
+
+#### Solution
+The issuer URI must be in the following format, where `default` is
+the name or ID of the authorization server:
+
+```
+https://example-account.okta.com/oauth2/default
+```
+
+You can find this URI in your Okta developer account, under **Security** > **API**.
```
Guaris, line 43: Would it be possible to link that API section where the URI is?

Guaris (changes requested):

app/konnect/org-management/okta-idp.md
adamdehaven: What about this? 
```suggestion
        https://{example-account}.okta.com/oauth2/default
```

app/konnect/org-management/troubleshoot.md
adamdehaven: I think "may" is correct here, (or possibly change to `may likely`) as I'm sure Okta will change that message at some point in the future

app/konnect/org-management/okta-idp.md
lena-larionova: But then we're back to the issue of using `{ }` where it might be confusing. 

I think what we're running up against here is fake URL vs placeholder. Even the Google style guide makes that distinction:

https://developers.google.com/style/examples

Gitlab has a good section as well (much smaller): https://docs.gitlab.com/ee/development/documentation/styleguide/#fake-urls

In URLs especially, that tends to be the prevailing style.

adamdehaven: makes sense to me. I definitely wouldn't change `/default` to `/DEFAULT` as a lot of times, that is in fact the correct value if they only have a single app set up

Guaris (approved):

PR #4034, opened 2022-06-17
Removed 'path' as option for admin_gui_url

d19dotca:
### Summary
Removed the path part of the schema presented for the admin_gui_url property.

### Reason
While the schema technically allows for the path to be set and won't throw errors, it doesn't appear to actually do anything other than the host and port part. I believe this is (possibly) confirmed too by FTI-976 which hasn't been implemented yet and is a feature request for allowing customers to set their own root paths for Kong Manager.

### Testing
Please consider double-checking this though in case I'm overlooking something, but I believe this change is correct based on the FTI and my own testing.

review:tech


d19dotca:
Hi @Guaris - Good to know - thanks! I just created the PR for that one now too instead.

PR #4051, opened 2022-06-23
key-auth-enc plugin: Remove non-existent ttl parameter

lena-larionova:
### Summary
Removing a setting that isn't available in the plugin.

### Reason
Parameter doesn't exist.
 
Reported on Slack, see: https://kongstrong.slack.com/archives/C0DSXDXV1/p1655994419162749

### Testing
https://docs.konghq.com/hub/kong-inc/key-auth-enc/


cloudjumpercat (approved):
LGTM - ttl parameter didn't display in the netlify preview

PR #4058, opened 2022-06-27
Fix GitHub edit link for single sourced pages

mheap:
### Summary
Fix "edit this page" link on single sourced pages and add some tests to ensure that we don't break them again.

### Reason
Reported by @lena-larionova

### Testing
Edit link is broken on https://docs.konghq.com/deck/latest/

Should be fixed on [LINK]


Guaris (approved):
lgtm

PR #4062, opened 2022-06-27
Vale: Adjust rules for curly brackets

lena-larionova:
### Summary
Adding rules to ignore all content in curly brackets:

Previous rules weren't ignoring the following:
* `{:.class-name}`
* `{#header-short-name}`

### Testing
Check locally.

### DIFF
--- a/.github/styles/kong/Terms.yml
+++ b/.github/styles/kong/Terms.yml
@@ -12,7 +12,7 @@ swap:
   dbless: DB-less
   file name: filename
   hashicorp: HashiCorp
-  kong: Kong
+  'kong(?: )': Kong
   # '(?:kubernetes|k8s)': Kubernetes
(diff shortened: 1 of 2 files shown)


lena-larionova:
Not working quite as wanted on Konnect branch yet.

.github/styles/kong/Terms.yml
```diff
@@ -12,7 +12,7 @@ swap:
   dbless: DB-less
   file name: filename
   hashicorp: HashiCorp
-  kong: Kong
+  '(?:# )kong(?: )': Kong
   # '(?:kubernetes|k8s)': Kubernetes
```
lena-larionova, line 16: We should be able to uncomment the kubernetes/k8s line, as long as we add the same ignores as for `kong`.

PR #4055, opened 2022-06-24
Fix `layout` type in Working with Templates

cloudjumpercat:
Changed `layout` type from boolean to string.

### Summary
Changed `layout` type from boolean to string. Decided to only change the 3.0.x version.

### Reason
[DOCU-1428](https://konghq.atlassian.net/browse/DOCU-1428)

PR #4063, opened 2022-06-27
Update `layout` item for 2.6.x-2.8.x for Working with Templates

cloudjumpercat:
Updated the `layout` config item type.

### Summary

Updated the `layout` config item type.

### Reason

DOCU-1428 (part 2 of the changes, the other changes are being added in #4055 )
### Testing

Check v2.6.x, v2.7.x, and v2.8.x to make sure the `layout` item displays correctly.


Guaris:
There is a vale issue happening here that we've seen in another PR. I'll resolve this tomorrow.

lena-larionova (approved):
LGTM!

cloudjumpercat:
@lena-larionova Could you re-review the vale issues I fixed? There's a few of them that I opted to escape and I wanted to make sure that was the best choice. Also, on line 462 of Working with Templates, the original word was "stringified" so I just wanted to make sure my revision sounds correct.

lena-larionova (approved):
LGTM - the rewrite of the "stringify" term makes sense to me - I like that you rewrote it into standard language. There's no reason to use the term that way if you're not specifically referring to a function name.

And I agree with your choice to turn Vale off on some of these sections with weird headers. This is a known problem topic that we haven't had time to rework (why are those even headers??), so it doesn't make sense to try and wrangle Vale to work with something that shouldn't even be that way.

PR #4086, opened 2022-07-04
Fix broken link in https://docs.konghq.com/mesh/latest/features/cert-manager/

mmorel-35:
### Summary
Fix broken link in https://docs.konghq.com/mesh/latest/features/cert-manager/

### Reason
The redirection is not working properly

Signed-off-by: Matthieu MOREL <mmorel-35@users.noreply.github.com>


lena-larionova (approved):

PR #4100, opened 2022-07-07
[fix] Vale ignore terms values in anchor tags

Guaris:
This change is necessary to fix an issue with Vale.

Currently, anchor tags that contain term values are not ignored. 

So

```
# H1 Title {#H1Anchor} 

```

Would be ignored but 

```
# H1 Title {#konnect} 
```
would trigger the `terms.yml` check. It is possible to use {{site.konnect_short_name}} in anchor tags, but I think this is a better option.


lena-larionova (approved):
LGTM, checked locally.

PR #4115, opened 2022-07-12
Add support for moving URLs and setting canonical URL correctly

mheap:
### Summary
Add the ability to track canonical URLs when the URL of a page changes

### Reason
When we update the IA for 3.0 URLs will change. We don't want duplicate content penalties

### Testing
Visit /gateway-oss/2.5.x/configuration/ and view source

The canonical tag should point to `/gateway/latest/reference/configuration/` which is the new URL for that page

### DIFF
--- a/docs/seo.md
+++ b/docs/seo.md
@@ -1,0 +1,28 @@
+# SEO
+
+The canonical URL for each page is automatically set by the `canonical_url_generator` plugin. It works by looking for a URL that matches the current page, but where the version is higher than the current URL. As an example:
+
+`/gateway/2.3.x/configuration` would have a canonical URL of `/gateway/2.5.x/configuration`. `2.5.x` is the last version in which this URL existed.
+
+In `2.6.x` the page moved to `/gateway/2.6.x/reference/configuration`, which means the canonical URL would be `/gateway/2.8.x/reference/configuration`.
+
+_However_, we have a special `/latest/` URL which is always the latest version, so the canonical URL for the 2.6.x link above would actually be ``/gateway/latest/reference/configuration`.
+
+## Tracking file renames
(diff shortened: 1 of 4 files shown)


docs/seo.md
```diff
@@ -0,0 +1,28 @@
+# SEO
+
+The canonical URL for each page is automatically set by the `canonical_url_generator` plugin. It works by looking for a URL that matches the current page, but where the version is higher than the current URL. As an example:
+
+`/gateway/2.3.x/configuration` would have a canonical URL of `/gateway/2.5.x/configuration`. `2.5.x` is the last version in which this URL existed.
+
+In `2.6.x` the page moved to `/gateway/2.6.x/reference/configuration`, which means the canonical URL would be `/gateway/2.8.x/reference/configuration`.
+
+_However_, we have a special `/latest/` URL which is always the latest version, so the canonical URL for the 2.6.x link above would actually be ``/gateway/latest/reference/configuration`.
```
lena-larionova, line 9:
```suggestion
1. `/gateway/2.3.x/configuration` would have a canonical URL of `/gateway/2.5.x/configuration`. `2.5.x` is the last version in which this URL existed.

2. In `2.6.x` the page moved to `/gateway/2.6.x/reference/configuration`, which means the canonical URL would normally be `/gateway/2.8.x/reference/configuration`.

3. _However_, we have a special `/latest/` URL which is always the latest version, so the canonical URL for the 2.6.x link above would actually be `/gateway/latest/reference/configuration`.
```
Just sticking it into steps to make it easier to follow + typo.

lena-larionova (approved):
LGTM!

Guaris (approved):

PR #4118, opened 2022-07-12
Use canonical URL rather than page_exists on old versions

mheap:
### Summary
Use the page's `canonical_url` rather than `page_exists` when linking to new versions of documentation.

This is functionally equivalent in most cases, but allows us to link to new versions when changing the URL thanks to #4115 

### Testing
I've tested locally by enabling the "latest version" box on gateway-oss 2.5.x (not added to this PR)

This shows the URL pointing to the new configuration guide as expected

image


lena-larionova (approved):
LGTM, tested this locally by switching up the static gateway link in the old split versions banner to the canonical URL. Will need to rephrase that banner in the old split docs if we ever tackle 2.5.x, but we don't need to worry about that yet.

PR #4127, opened 2022-07-15
doc(conf) add declarative_config_encryption_mode in kong.conf

fffonion:
FT-2613

### Summary
- Add the new EE option `declarative_config_encryption_mode`.

### Reason
https://github.com/Kong/kong-ee/pull/3312

### Testing

### DIFF
--- a/src/gateway/reference/configuration.md
+++ b/src/gateway/reference/configuration.md
@@ -1729,6 +1729,18 @@ The declarative configuration as a string
 
 ---
 
+#### declarative_config_encryption_mode
+{:.badge .enterprise}
+
+Set encryption of the declarative config mapped file on filesystem. This configuration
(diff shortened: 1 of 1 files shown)


mheap:
@Guaris @lena-larionova This is an interesting one for Vale. The content comes from a generated file. In this instance, should the autodocs source contain our docs-specific variables, or should we exclude autogenerated files from this substitution?

lena-larionova:
and to follow-up - @fffonion , can you please make these changes in the source file (https://github.com/Kong/kong-ee/blob/master/kong.conf.default)  instead of here? We'll generate the doc on the docs site from there.

fffonion:
@lena-larionova , the config is already in kong.conf.default. In that case I'm closing this as no actions is needed. Thanks for the input on this on!

PR #4129, opened 2022-07-16
Guide `ingress-class` instead of `ingress.class` for Knative

YaSuenag:
### Summary

Guide `ingress-class` instead of `ingress.class` for Knative Ingress.

### Reason

https://github.com/Kong/kubernetes-ingress-controller/issues/2344 introduced to use `ingress-class` instead of `ingress.class` for Knative Ingress, however Kong document hasn't been changed.

### Testing

I tested it on K8s v1.24 + Knative v1.5.0 with Kong v2.8 + Ingress Controller v2.5 .

If we use `ingress.class` in above versions, it does not work (endpoint address is not assigned to K8s service for ksvc.)

### DIFF
--- a/app/_data/docs_nav_kic_2.2.x.yml
+++ b/app/_data/docs_nav_kic_2.2.x.yml
@@ -111,6 +111,8 @@
       url: /guides/preserve-client-ip
     - text: Using Gateway API
       url: /guides/using-gateway-api
+    - text: Using Kong with Knative
+      url: /guides/using-kong-with-knative
 - title: References
   icon: /assets/images/icons/documentation/icn-kubernetes-color.svg
   items:
--- a/app/_data/docs_nav_kic_2.3.x.yml
+++ b/app/_data/docs_nav_kic_2.3.x.yml
@@ -115,6 +115,8 @@ items:
         url: /guides/preserve-client-ip
       - text: Using Gateway API
         url: /guides/using-gateway-api
+      - text: Using Kong with Knative
+        url: /guides/using-kong-with-knative
   - title: References
     icon: /assets/images/icons/documentation/icn-kubernetes-color.svg
     items:
--- a/app/_data/docs_nav_kic_2.4.x.yml
+++ b/app/_data/docs_nav_kic_2.4.x.yml
@@ -115,6 +115,8 @@ items:
         url: /guides/preserve-client-ip
       - text: Using Gateway API
         url: /guides/using-gateway-api
+      - text: Using Kong with Knative
+        url: /guides/using-kong-with-knative
   - title: References
     icon: /assets/images/icons/documentation/icn-kubernetes-color.svg
     items:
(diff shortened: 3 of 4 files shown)


cloudjumpercat:
@YaSuenag Thanks for the feedback! I've got a few questions:

- I was trying to view the doc you changed in the Netlify preview, but I couldn't find it. Do you have the URL to the published doc on docs.konghq.com? 
- I saw a few other places where `ingress.class` is used ([here](https://deploy-preview-4129--kongdocs.netlify.app/kubernetes-ingress-controller/2.4.x/references/annotations/#kubernetesioingressclass), for example). Should all of those be updated as well?
- On that same page I previously listed, there's a note that says "Kubernetes versions after 1.18 introduced the new ingressClassName field to the Ingress spec and [deprecated the kubernetes.io/ingress.class annotation](https://kubernetes.io/docs/concepts/services-networking/ingress/#deprecated-annotation). Ingress resources should now use the ingressClassName field. Kong resources (KongConsumer, TCPIngress, etc.) still use the kubernetes.io/ingress.class annotation." Should `ingress.class` be replaced with `ingressClassName` instead?

YaSuenag:
> * I was trying to view the doc you changed in the Netlify preview, but I couldn't find it. Do you have the URL to the published doc on docs.konghq.com?

I could see the doc on following URL few days ago at least, however now it is 404...  
https://docs.konghq.com/kubernetes-ingress-controller/latest/guides/using-kong-with-knative/

> * I saw a few other places where `ingress.class` is used ([here](https://deploy-preview-4129--kongdocs.netlify.app/kubernetes-ingress-controller/2.4.x/references/annotations/#kubernetesioingressclass), for example). Should all of those be updated as well?
> * On that same page I previously listed, there's a note that says "Kubernetes versions after 1.18 introduced the new ingressClassName field to the Ingress spec and [deprecated the kubernetes.io/ingress.class annotation](https://kubernetes.io/docs/concepts/services-networking/ingress/#deprecated-annotation). Ingress resources should now use the ingressClassName field. Kong resources (KongConsumer, TCPIngress, etc.) still use the kubernetes.io/ingress.class annotation." Should `ingress.class` be replaced with `ingressClassName` instead?

I'm not sure it affects to K8s networking - https://github.com/Kong/kubernetes-ingress-controller/issues/2344 seems to affect Knative Serving at least. So I attempt to update doc for Knative only.

lena-larionova:
> I could see the doc on following URL few days ago at least, however now it is 404...
https://docs.konghq.com/kubernetes-ingress-controller/latest/guides/using-kong-with-knative/

Looks like it was always missing from the navigation, but when we switched to building from `src`, which uses the nav to generate files, it became more obvious that it got dropped. 

We can fix this by adding a nav entry for Knative in 2.3.x and 2.4.x, in the guides section: 
* https://github.com/YaSuenag/docs.konghq.com/blob/knative-doc/app/_data/docs_nav_kic_2.4.x.yml
* https://github.com/YaSuenag/docs.konghq.com/blob/knative-doc/app/_data/docs_nav_kic_2.3.x.yml

In 2.2.x and earlier, it works fine: https://docs.konghq.com/kubernetes-ingress-controller/2.2.x/guides/using-kong-with-knative/

cloudjumpercat:
@lena-larionova I looked in the 2.2.x kic nav file for where this page is located in the nav, and it wasn't listed anywhere (unless I'm missing something). Do we want this page to appear in the nav or should it be unlisted? And if it should be unlisted, how do I do that?

cloudjumpercat (approved):

PR #4141, opened 2022-07-20
Fix migration code example

vynhart:
### Summary
In the example of migration file, it used `postgresql` key while it should be `postgres` instead.


lena-larionova (approved):
Nice catch, thanks for the fix. LGTM

(Verified by checking against some pre-built plugins too, eg https://github.com/Kong/kong/blob/master/kong/plugins/acl/migrations/000_base_acl.lua)

PR #4143, opened 2022-07-20
Further improve the plugin template

lena-larionova:
### Summary
Adding full steps for creating a new plugin doc (icon, versions.yml) to the template.

### Reason
The info is hard to track down and scattered (it exists in some internal docs that aren't kept up-to-date). It's simpler to add this info to the template instead, which is what new plugin authors are already looking at. Often, we will have new plugin submissions where the person fully followed the template instructions and didn't bother looking anywhere else, so the icon and versions file are missing.

Need this for doc team as well.

### Testing

On branch, shouldn't be published to live site.

@mheap I ran into an issue with this template, caused by single-sourcing. The plugin doc generator actually creates a published/live doc for the template as well, under [/hub/_init/my-extension/](https://deploy-preview-4143--kongdocs.netlify.app/hub/_init/my-extension/). Can we add an ignore rule for the `app/_hub/_init/` directory? If we do that, we can also add a sample `versions.yml` which users can similarly duplicate.

### DIFF
--- a/app/_hub/_init/my-extension/_index.md
+++ b/app/_hub/_init/my-extension/_index.md
@@ -14,7 +14,7 @@
 # releases: # Fill in the Gateway version that this plugin is being added in.
 #   - 3.0.x
 #
-# 3. Add an icon for the plugin to app/_assets/images/icons/hub.
+# 3. Add a 64x64px icon for the plugin to app/_assets/images/icons/hub.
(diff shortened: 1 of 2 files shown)


cloudjumpercat (approved):
LGTM

mheap (approved):

PR #4188, opened 2022-08-05
fixed client certificate updating issue FTI-4078

liverpool8056:
### Summary
Fixed client certificate updating issue that the certificate in core_cache should be invalidation when updating, and the corresponding cache item for the new certificate should be used.


mheap:
It looks as though this change was reverted in https://github.com/Kong/kong-ee/pull/3574 and will not be in 2.8.1.3

PR #4247, opened 2022-08-15
docs: fix issues with plugin-ordering examples

jschmid1:
Signed-off-by: Joshua Schmid <jaiks@posteo.de>

rate-limiting's `limit_by` defaults to `consumer`. In this case we do not have a consumer configured as this is not a supported case with plugin-ordering.

### DIFF
--- a/src/gateway/kong-enterprise/plugin-ordering/get-started.md
+++ b/src/gateway/kong-enterprise/plugin-ordering/get-started.md
@@ -30,6 +30,7 @@ curl -i -X POST http://<admin-hostname>:8001/plugins \
   --data name=rate-limiting \
   --data config.minute=5 \
   --data config.policy=local \
+  --data config.limit_by=ip \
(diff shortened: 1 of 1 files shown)


mheap:
@jschmid1 Are you saying that we don't support consumers at all with plugin ordering, or that rate limiting by consumer doesn't make sense if it runs before `key-auth`?

jschmid1:
> @jschmid1 Are you saying that we don't support consumers at all with plugin ordering, or that rate limiting by consumer doesn't make sense if it runs before `key-auth`?

Consumer scoping and plugin reordering is a potential pitfall. If you order a consumer scoped plugin before an authentication plugin it looses the ability to act on `$authenticated_user` (such as in the example of rate-limiting where the rate-limit is actually determined by the consumer, which isn't determined yet as the auth plugin context wasn't executed)

lena-larionova (approved):

PR #4260, opened 2022-08-17
DOCU-2423: Adds StatsD monitoring page under Kong in Production

rspurgeon:
### Summary
Adds StatsD monitoring page under Kong in Production

### Reason
Helps the user with a how-to guide to quickly run Kong GW, StatsD, and StatsD plugin
https://konghq.atlassian.net/browse/DOCU-2423

### Testing
built and viewed docs locally

### DIFF
--- a/src/gateway/kong-production/monitoring/statsd.md
+++ b/src/gateway/kong-production/monitoring/statsd.md
@@ -9,35 +9,22 @@ text based statistics data published by applications.
 
 This guide will help you setup a test {{site.base_gateway}} and
 StatsD service. Then you will generate sample requests to {{site.base_gateway}} and
-observe the collected monitoring data.
+observe the collected monitoring data. 
 
 ### Prerequisites
-This guide assumes the following tools are installed locally:
-* [Docker](https://docs.docker.com/get-docker/) is used to run {{site.base_gateway}}, the supporting database, and StatsD locally. 
-* [curl](https://curl.se/) is used to send requests to {{site.base_gateway}}. `curl` is pre-installed on most systems.
-* [Netcat](http://netcat.sourceforge.net/) installed as `nc` on the `PATH`. `nc` is used to send requests 
-  to the StatsD management interface. `nc` is pre-installed on many systems.
-
-### Configure StatsD monitoring
-
-1. Run the following command to start {{site.base_gateway}} using Docker:
-
+* A test {{site.base_gateway}} instance is installed using the following script:
    ```sh
    curl -Ls get.konghq.com/quickstart | sh -s
    ```
+   This script installs the test instance and a mock service that is used in this guide. 
+   {:.note}
+   > **Note:** This guide's instructions use a test {{site.base_gateway}} instance to demonstrate how the StatsD plugin can be used to collect metrics. If you want to use an existing {{site.base_gateway}} instance for this guide, you must modify the connection information in the commands.
+* [Docker](https://docs.docker.com/get-docker/) is installed locally. It is used to run {{site.base_gateway}}, the supporting database, and StatsD locally. 
+* [curl](https://curl.se/) is installed locally. It is used to send requests to {{site.base_gateway}}. `curl` is pre-installed on most systems.
+* [Netcat](http://netcat.sourceforge.net/) is installed locally as `nc` on the `PATH`. `nc` is used to send requests 
+  to the StatsD management interface. `nc` is pre-installed on many systems.
 
-   You should see the following message when {{site.base_gateway}} is ready:
-
-   ```text
-   ✔ Kong is ready!
-   ```
-
-   The script also installs a mock service to make testing easier. You will use this 
-   service later to generate metrics data. You can send a sample request with:
-
-   ```sh
-   curl localhost:8000/mock/requests
-   ```
+### Configure StatsD monitoring
 
 1. Run a StatsD container to capture monitoring data:
 
--- a/src/gateway/kong-production/monitoring/statsd.md
+++ b/src/gateway/kong-production/monitoring/statsd.md
@@ -23,7 +23,7 @@ observe the collected monitoring data.
 
    {:.note}
       > This step is optional if you wish to use an existing {{site.base_gateway}} installation. When using an existing
-        {{site.base_gateway}}, you will need to modify the remaining commands as appropriate to account for network
+        {{site.base_gateway}}, you will need to modify the commands to account for network
         connectivity and installed {{site.base_gateway}} services and routes.
 
    ```sh
--- a/src/gateway/kong-production/monitoring/statsd.md
+++ b/src/gateway/kong-production/monitoring/statsd.md
@@ -12,13 +12,13 @@ StatsD service. Then you will generate sample requests to {{site.base_gateway}} 
 observe the collected monitoring data.
 
 ### Prerequisites
-This guide assumes each of the following tools are installed locally. 
-* [Docker](https://docs.docker.com/get-docker/) is used to run Kong, the supporting database, and StatsD locally. 
-* [curl](https://curl.se/) is used to send requests to the gateway. Most systems come with `curl` pre-installed.
+This guide assumes the following tools are installed locally:
+* [Docker](https://docs.docker.com/get-docker/) is used to run {{site.base_gateway}}, the supporting database, and StatsD locally. 
+* [curl](https://curl.se/) is used to send requests to {{site.base_gateway}}. `curl` is pre-installed on most systems.
 * [Netcat](http://netcat.sourceforge.net/) installed as `nc` on the `PATH`. `nc` is used to send requests 
-  to the StatsD management interface. Many systems come with `nc` pre-installed.
+  to the StatsD management interface. `nc` is pre-installed on many systems.
 
-### Guide
+### Configure StatsD monitoring
 
 1. Run the following command to start {{site.base_gateway}} using Docker:
 
@@ -32,7 +32,7 @@ This guide assumes each of the following tools are installed locally.
    ✔ Kong is ready!
    ```
 
-   The script has also installed a mock service to make testing easier. You will use this 
+   The script also installs a mock service to make testing easier. You will use this 
    service later to generate metrics data. You can send a sample request with:
 
    ```sh
@@ -59,9 +59,9 @@ This guide assumes each of the following tools are installed locally.
 
    You should receive a JSON response with the details of the installed plugin.
 
-1. Generate sample traffic to the mock service. This will allow you to observe 
-   metrics generated from the StatsD plugin. The following command will generate 60 
-   requests over 1 minute. Run the following in a new terminal:
+1. Generate sample traffic to the mock service. This allows you to observe 
+   metrics generated from the StatsD plugin. The following command generates 60 
+   requests over one minute. Run the following in a new terminal:
 
    ```sh
    for _ in {1..60}; do {curl localhost:8000/mock/request; sleep 1; } done
@@ -87,12 +87,12 @@ This guide assumes each of the following tools are installed locally.
    END
    ```
 
-The [Kong Hub StatsD Plugin](/hub/kong-inc/statsd/) 
-contains the full documentation on the plugin usage and configuration
+See the [StatsD plugin](/hub/kong-inc/statsd/) 
+documentation for more information about how to use and configure the plugin.
 
-### Cleanup
+### Remove the test StatsD service 
 
-Once you are done experimenting with StatsD and {{site.base_gateway}}, use the following
+Once you are done experimenting with StatsD and {{site.base_gateway}}, you can use the following
 commands to stop and remove the services created in this guide:
 
 ```sh
@@ -100,9 +100,9 @@ docker stop kong-quickstart-statsd
 curl -Ls get.konghq.com/quickstart | sh -s -- -d
 ```
 
-### More Information
-* [How to monitor with Prometheus](/gateway/{{page.gateway_version}}/kong-production/monitoring/prometheus/) 
-provides a guide to using [Prometheus](https://prometheus.io/docs/introduction/overview/) for monitoring with the 
-[{{site.base_gateway}} Plugin](/hub/kong-inc/prometheus/)
+### More information
+* [How to monitor with Prometheus](/gateway/{{page.kong_version}}/kong-production/monitoring/prometheus/) 
+describes how to use [Prometheus](https://prometheus.io/docs/introduction/overview/) to monitor {{site.base_gateway}} using the  
+[Prometheus plugin](/hub/kong-inc/prometheus/).
 * See the [Tracing API Reference](/gateway/{{page.kong_version}}/kong-production/tracing/api/) for information
-on {{site.base_gateway}}'s tracing capabilites
+about {{site.base_gateway}}'s tracing capabilities.
--- a/src/gateway/kong-production/monitoring/statsd.md
+++ b/src/gateway/kong-production/monitoring/statsd.md
@@ -12,19 +12,23 @@ StatsD service. Then you will generate sample requests to {{site.base_gateway}} 
 observe the collected monitoring data. 
 
 ### Prerequisites
-* A test {{site.base_gateway}} instance is installed using the following script:
-   ```sh
-   curl -Ls get.konghq.com/quickstart | sh -s -- -m
-   ```
-   The `-m` flag instructs the script to install a mock service that is used in this guide to generate sample metrics.
-   {:.note}
-   > **Note:** This guide's instructions use a test {{site.base_gateway}} instance to demonstrate how the StatsD plugin can be used to collect metrics. If you want to use an existing {{site.base_gateway}} instance for this guide, you must modify the connection information in the commands.
 * [Docker](https://docs.docker.com/get-docker/) is installed locally. It is used to run {{site.base_gateway}}, the supporting database, and StatsD locally. 
+
+   If you will be using an existing {{site.base_gateway}} instance instead of the test instance, you may need configure Docker to ensure features like network connectivity function correctly.  
 * [curl](https://curl.se/) is installed locally. It is used to send requests to {{site.base_gateway}}. `curl` is pre-installed on most systems.
 * [Netcat](http://netcat.sourceforge.net/) is installed locally as `nc` on the `PATH`. `nc` is used to send requests 
   to the StatsD management interface. `nc` is pre-installed on many systems.
 
 ### Configure StatsD monitoring
+
+{:.note}
+   > **Note:** This guide's instructions use a test {{site.base_gateway}} instance to demonstrate how the StatsD plugin can be used to collect metrics. If you want to use an existing {{site.base_gateway}} instance for this guide, you must modify the connection information in the commands.
+
+1. Optional: Install {{site.base_gateway}} using the following script:
+   ```sh
+   curl -Ls get.konghq.com/quickstart | sh -s -- -m
+   ```
+   The `-m` flag instructs the script to install a mock service that is used in this guide to generate sample metrics.
 
 1. Run a StatsD container to capture monitoring data:
 
@@ -77,7 +81,7 @@ observe the collected monitoring data.
 See the [StatsD plugin](/hub/kong-inc/statsd/) 
 documentation for more information about how to use and configure the plugin.
 
-### Remove the test StatsD service 
+### Clean up the test instance 
 
 Once you are done experimenting with StatsD and {{site.base_gateway}}, you can use the following
 commands to stop and remove the services created in this guide:
--- a/src/gateway/kong-production/monitoring/statsd.md
+++ b/src/gateway/kong-production/monitoring/statsd.md
@@ -4,7 +4,7 @@ content_type: how-to
 ---
 
 [StatsD](https://github.com/statsd/statsd) is a network daemon that collects
-and aggreates performance metrics by listening on the network for simple 
+and aggreates performance metrics by listening on the network for 
 text based statistics data, published by applications. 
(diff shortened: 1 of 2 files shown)


src/gateway/kong-production/monitoring/statsd.md
```diff
@@ -1,6 +1,108 @@
...
+
+```sh
+docker stop kong-quickstart-statsd
+curl -Ls get.konghq.com/quickstart | sh -s -- -d
+```
+
+### More Information
+* [How to monitor with Prometheus](/gateway/{{page.gateway_version}}/kong-production/monitoring/prometheus/) 
+provides a guide to using [Prometheus](https://prometheus.io/docs/introduction/overview/) for monitoring with the 
+[{{site.base_gateway}} Plugin](/hub/kong-inc/prometheus/)
```
cloudjumpercat, line 106:
```suggestion
* [How to monitor with Prometheus](/gateway/{{page.kong_version}}/kong-production/monitoring/prometheus/) 
describes how to use [Prometheus](https://prometheus.io/docs/introduction/overview/) to monitor {{site.base_gateway}} using the  
[Prometheus plugin](/hub/kong-inc/prometheus/).
```

src/gateway/kong-production/monitoring/statsd.md
```diff
@@ -1,6 +1,108 @@
...
+
+Once you are done experimenting with StatsD and {{site.base_gateway}}, use the following
+commands to stop and remove the services created in this guide:
+
+```sh
+docker stop kong-quickstart-statsd
+curl -Ls get.konghq.com/quickstart | sh -s -- -d
+```
+
+### More Information
```
cloudjumpercat, line 103:
```suggestion
### More information
```

src/gateway/kong-production/monitoring/statsd.md
```diff
@@ -1,6 +1,108 @@
...
+* [How to monitor with Prometheus](/gateway/{{page.gateway_version}}/kong-production/monitoring/prometheus/) 
+provides a guide to using [Prometheus](https://prometheus.io/docs/introduction/overview/) for monitoring with the 
+[{{site.base_gateway}} Plugin](/hub/kong-inc/prometheus/)
+* See the [Tracing API Reference](/gateway/{{page.kong_version}}/kong-production/tracing/api/) for information
+on {{site.base_gateway}}'s tracing capabilites
```
cloudjumpercat, line 108:
```suggestion
* See the [Tracing API Reference](/gateway/{{page.kong_version}}/kong-production/tracing/api/) for information
about {{site.base_gateway}}'s tracing capabilities.
```

src/gateway/kong-production/monitoring/statsd.md
```diff
@@ -1,6 +1,108 @@
...
+[StatsD](https://github.com/statsd/statsd) is a network daemon that collects
+and aggreates performance metrics by listening on the network for simple 
+text based statistics data published by applications. 
+
+This guide will help you setup a test {{site.base_gateway}} and
+StatsD service. Then you will generate sample requests to {{site.base_gateway}} and
+observe the collected monitoring data.
+
+### Prerequisites
+This guide assumes each of the following tools are installed locally.
```
cloudjumpercat, line 15:
```suggestion
This guide assumes the following tools are installed locally:
```

src/gateway/kong-production/monitoring/statsd.md
```diff
@@ -1,6 +1,108 @@
...
+and aggreates performance metrics by listening on the network for simple 
+text based statistics data published by applications. 
+
+This guide will help you setup a test {{site.base_gateway}} and
+StatsD service. Then you will generate sample requests to {{site.base_gateway}} and
+observe the collected monitoring data.
+
+### Prerequisites
+This guide assumes each of the following tools are installed locally. 
+* [Docker](https://docs.docker.com/get-docker/) is used to run Kong, the supporting database, and StatsD locally.
```
cloudjumpercat, line 16:
```suggestion
* [Docker](https://docs.docker.com/get-docker/) is used to run {{site.base_gateway}}, the supporting database, and StatsD locally. 
```

src/gateway/kong-production/monitoring/statsd.md
```diff
@@ -1,6 +1,108 @@
...
+text based statistics data published by applications. 
+
+This guide will help you setup a test {{site.base_gateway}} and
+StatsD service. Then you will generate sample requests to {{site.base_gateway}} and
+observe the collected monitoring data.
+
+### Prerequisites
+This guide assumes each of the following tools are installed locally. 
+* [Docker](https://docs.docker.com/get-docker/) is used to run Kong, the supporting database, and StatsD locally. 
+* [curl](https://curl.se/) is used to send requests to the gateway. Most systems come with `curl` pre-installed.
```
cloudjumpercat, line 17:
```suggestion
* [curl](https://curl.se/) is used to send requests to {{site.base_gateway}}. `curl` is pre-installed on most systems.
```

src/gateway/kong-production/monitoring/statsd.md
```diff
@@ -1,6 +1,108 @@
...
+* [Netcat](http://netcat.sourceforge.net/) installed as `nc` on the `PATH`. `nc` is used to send requests 
+  to the StatsD management interface. Many systems come with `nc` pre-installed.
```
cloudjumpercat, line 19:
```suggestion
* [Netcat](http://netcat.sourceforge.net/) installed as `nc` on the `PATH`. `nc` is used to send requests 
  to the StatsD management interface. `nc` is pre-installed on many systems.
```

src/gateway/kong-production/monitoring/statsd.md
```diff
@@ -1,6 +1,108 @@
...
+
+### Guide
```
cloudjumpercat, line 21:
```suggestion
### Configure StatsD monitoring
```

src/gateway/kong-production/monitoring/statsd.md
```diff
@@ -1,6 +1,108 @@
...
+   curl -Ls get.konghq.com/quickstart | sh -s
+   ```
+
+   You should see the following message when {{site.base_gateway}} is ready:
+
+   ```text
+   ✔ Kong is ready!
+   ```
+
+   The script has also installed a mock service to make testing easier. You will use this
```
cloudjumpercat, line 35:
```suggestion
   The script also installs a mock service to make testing easier. You will use this 
```

src/gateway/kong-production/monitoring/statsd.md
```diff
@@ -1,6 +1,108 @@
...
+       --data "name=statsd"  \
+       --data "config.host=kong-quickstart-statsd" \
+       --data "config.port=8125"
+   ```
+
+   You should receive a JSON response with the details of the installed plugin.
+
+1. Generate sample traffic to the mock service. This will allow you to observe 
+   metrics generated from the StatsD plugin. The following command will generate 60 
+   requests over 1 minute. Run the following in a new terminal:
```
cloudjumpercat, line 64:
```suggestion
1. Generate sample traffic to the mock service. This allows you to observe 
   metrics generated from the StatsD plugin. The following command generates 60 
   requests over one minute. Run the following in a new terminal:
```

src/gateway/kong-production/monitoring/statsd.md
```diff
@@ -1,6 +1,108 @@
...
+     'statsd.metrics_received': 56,
+     'kong.mock.request.count': 7,
+     'kong.mock.request.status.200': 7,
+     'kong.mock.request.status.total': 7
+   }
+   END
+   ```
+
+The [Kong Hub StatsD Plugin](/hub/kong-inc/statsd/) 
+contains the full documentation on the plugin usage and configuration
```
cloudjumpercat, line 91:
```suggestion
See the [StatsD plugin](/hub/kong-inc/statsd/) 
documentation for more information about how to use and configure the plugin.
```

src/gateway/kong-production/monitoring/statsd.md
```diff
@@ -1,6 +1,108 @@
...
+
+### Cleanup
```
cloudjumpercat, line 93:
```suggestion
### Remove the test StatsD service 
```

src/gateway/kong-production/monitoring/statsd.md
```diff
@@ -1,6 +1,108 @@
...
+
+Once you are done experimenting with StatsD and {{site.base_gateway}}, use the following
+commands to stop and remove the services created in this guide:
```
cloudjumpercat, line 96:
```suggestion
Once you are done experimenting with StatsD and {{site.base_gateway}}, you can use the following
commands to stop and remove the services created in this guide:
```

src/gateway/kong-production/monitoring/statsd.md
```diff
@@ -1,6 +1,108 @@
...
+### Prerequisites
+This guide assumes each of the following tools are installed locally. 
+* [Docker](https://docs.docker.com/get-docker/) is used to run Kong, the supporting database, and StatsD locally. 
+* [curl](https://curl.se/) is used to send requests to the gateway. Most systems come with `curl` pre-installed.
+* [Netcat](http://netcat.sourceforge.net/) installed as `nc` on the `PATH`. `nc` is used to send requests 
+  to the StatsD management interface. Many systems come with `nc` pre-installed.
+
+### Guide
+
+1. Run the following command to start {{site.base_gateway}} using Docker:
```
cloudjumpercat, line 23: @rspurgeon What happens if you already have Gateway installed and you try to run this script? I was kind of thinking that the audience of this doc might already have Gateway installed. If so, would it be better to tell them which routes/services/etc. to configure so they can use this guide? Or we could provide the option like:

1. Do one of the following:
    - If you don't have Kong installed, run this quickstart script:
    - If you've already installed Kong, make sure you add these services/routes that we'll be using for this guide:

cloudjumpercat (changes requested):

src/gateway/kong-production/monitoring/statsd.md
rspurgeon: The quickstart script is designed to rebuild the test gateway if it already exists.

IMO we should just give the user the necessary instructions inline of page.  It's a single line of code to get the test gateway, so i think it's a distraction to redirect them to another page.  The script also sets up the mock service which aids in the later instructions, otherwise we have to write confusing "if you have this do this if you don't have this do that" kind of copy.

src/gateway/kong-production/monitoring/statsd.md
rspurgeon: The cleanup does more than just remove the statsd service, it cleans up the gateway, database, and docker network as well

src/gateway/kong-production/monitoring/statsd.md
lena-larionova: I think Diana's idea is a good one. There's nothing wrong with having the script as a prerequisite, or providing the alternative to add a list of routes and services that they will be using in this guide. 

While the quickstart script and the entities it creates are great resources, sticking this in as an absolute requirement gates the guide for anyone with an existing installation. 

> otherwise we have to write confusing "if you have this do this if you don't have this do that" kind of copy.

We can write this once, in the prereq, and then the rest of the guide can just assume they have the routes/services.

> The quickstart script is designed to rebuild the test gateway if it already exists.

That seems like a bad idea, if the person isn't expecting it. If we don't give them the option to add this to their existing installation, then we're losing potential users.

As much as I like the idea of "every page is page one", starting with an installation of kong gateway for a guide like this is going too far. We have to remember that our audience is varied, and users of all stages in their Kong journey could be wanting to do something like set up statsd - by adding a step to install gateway anew, we're essentially saying "you can't actually do this in your installation, you have to start over".

rspurgeon: In my experience this kind of soup-to-nuts quickstart documentation is best for getting up and running with concepts, otherwise the documentation becomes littered with conditionals for the reader (click this link if you have this or do this if you don't).  Documentation pages that are simple, complete, and to the point is the best in my personal opinion.

The task at hand is to learn about the StatsD plugin and the concepts around it.  As a potential reader of this documentation, I would assume I could apply the techniques to my own installation, or else why would it exist at all? 

This is why people use Docker for development, ephemeral things to test that are easily disposed and recreated. The script gives them a quick test environment to work with, named in a particular way that can be disposed of with no cost.  They should have no expectation about it that would surprise them.  

When testing new concepts, it's less common to want to add to an existing service, one that is likely already serving another purpose. You spin up something new to learn the objective, take the education and then apply it. I wouldn't expect lose any potential users over this workflow.  

All that said, I think some dialog at the beginning of the document that says, "If you don't have a Kong Gateway, run this script" and keep the script instruction _in line of the page_ would be fine.

Then we also need to say something like... "If you're using an existing gateway, modify the connection information in the following command."  Otherwise we need to use a variable or something which is currently represented in the doc as `localhost:8000-1`.   As Diana also mentioned, we'd also need to add something to inform the user to add a new route and service to match the document, or tell them to modify the commands as needed.

In my personal opinion, it adds a lot of extra cognitive overhead for the reader who is very likely learning this on a test gateway they will create. Regardless, please feel free to make the desired changes to the document as the team desires to get it complete for the release.

src/gateway/kong-production/monitoring/statsd.md
```diff
@@ -1,6 +1,95 @@
...
+observe the collected monitoring data. 
+
+### Prerequisites
+* A test {{site.base_gateway}} instance is installed using the following script:
+   ```sh
+   curl -Ls get.konghq.com/quickstart | sh -s -- -m
+   ```
+   The `-m` flag instructs the script to install a mock service that is used in this guide to generate sample metrics.
+   {:.note}
+   > **Note:** This guide's instructions use a test {{site.base_gateway}} instance to demonstrate how the StatsD plugin can be used to collect metrics. If you want to use an existing {{site.base_gateway}} instance for this guide, you must modify the connection information in the commands.
```
cloudjumpercat, line 21: @rspurgeon @lena-larionova I'm not sure why GitHub removed our conversation, but I added this and was wondering what you think. Should we provide additional information in the note about what services/routes/etc. they should add to an existing instance, or would users know enough to know what to modify in the commands?

src/gateway/kong-production/monitoring/statsd.md
```diff
@@ -1,6 +1,108 @@
...
+     'kong.mock.request.status.200': 7,
+     'kong.mock.request.status.total': 7
+   }
+   END
+   ```
+
+See the [StatsD plugin](/hub/kong-inc/statsd/) 
+documentation for more information about how to use and configure the plugin.
+
+### Remove the test StatsD service
```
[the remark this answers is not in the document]
rspurgeon: I still think "cleanup" here is the best word to encompass what the steps do.

I think we might be debating the difference between a how-to guide and a reference document.  I might be mis-informed about the goal of this page, but I don't view this as the reference material for StatsD. I view this as a conceptual How-to with the goal to get the user to understand that Monitoring with StatsD is possible and what are the various concepts and components involved.

We have Docker as a prereq here, and we instruct the user to run StatsD in Docker. If they are going to connect that to a "Production" gateway (or any Gateway not in the same Docker netwrok), there are other factors we might have to document.  Network connectivity for example might not work "OOTB" from StatsD to GW when it's running in a container.  I don't think in the area of conceptual learning, people are / should be connecting to production services and I'm not sure we should solve for that case here.

src/gateway/kong-production/monitoring/statsd.md
rspurgeon: I'm in favor of this.  I do think if someone is going to be connecting StatsD, from inside Docker, to an existing Gateway, they likely need to have enough knowledge to do that without a lot of very specific guidance from them.  Their gateway could be anywhere, so we can't give them exact guidance.

The only other change I might make is, instead of it being a note after the user runs the script, but this guidance before the script and tell them it's optional and why.  So i'd move this up to the instructions for Step 1

rspurgeon: I didn't notice that this step got moved to Prerequisites.  IMO this isn't a prereq, it's a step in the guide.  We may choose to make that step optional, but I think prereqs are tools that must be installed or other system setups that are required to proceed.    If this were a pre-req, it would need to come after Docker or curl becuase it uses both of those tools.  I still very much prefer that this step be part of the guide with text to make it optional if desired.

rspurgeon:
One other consideration for this guide is that StatsD collects data broadcast over UDP, and with the GW and StatsD running in the same network this works (the guide currently uses a named Docker network). If the user runs StatsD and GW in a different network (one where UDP is not properly propagating across the networks), the StatsD service likely will not properly collect data.  The StatsD Advanced plugin supports tcp communications, but that is EE.

I'm not sure if this should be documented here, in the StatsD plugin reference, or left to the StatsD documentation. If this page is meant to walk the user through setting up StatsD and seeing the data collection work, we should consider this technical aspect.

src/gateway/kong-production/monitoring/statsd.md
```diff
@@ -1,6 +1,99 @@
...
+text based statistics data published by applications. 
+
+This guide will help you setup a test {{site.base_gateway}} and
+StatsD service. Then you will generate sample requests to {{site.base_gateway}} and
+observe the collected monitoring data. 
+
+### Prerequisites
+* [Docker](https://docs.docker.com/get-docker/) is installed locally. It is used to run {{site.base_gateway}}, the supporting database, and StatsD locally. 
+
+   If you will be using an existing {{site.base_gateway}} instance instead of the test instance, you may need configure Docker to ensure features like network connectivity function correctly.
```
cloudjumpercat, line 17:
```suggestion
   If you will be using an existing {{site.base_gateway}} instance instead of the test instance, you may need configure Docker to ensure Docker and {{site.base_gateway}} are running on the same network.  
```
@rspurgeon For your comment about UDP, would something like this work?

src/gateway/kong-production/monitoring/statsd.md
```diff
@@ -1,6 +1,99 @@
...
+* [curl](https://curl.se/) is installed locally. It is used to send requests to {{site.base_gateway}}. `curl` is pre-installed on most systems.
+* [Netcat](http://netcat.sourceforge.net/) is installed locally as `nc` on the `PATH`. `nc` is used to send requests 
+  to the StatsD management interface. `nc` is pre-installed on many systems.
+
+### Configure StatsD monitoring
+
+{:.note}
+   > **Note:** This guide's instructions use a test {{site.base_gateway}} instance to demonstrate how the StatsD plugin can be used to collect metrics. If you want to use an existing {{site.base_gateway}} instance for this guide, you must modify the connection information in the commands.
+
+1. Optional: Install {{site.base_gateway}} using the following script:
```
cloudjumpercat, line 27: @rspurgeon I moved the note and script back to the steps. Does this look good to you?

src/gateway/kong-production/monitoring/statsd.md
```diff
@@ -1,6 +1,99 @@
...
+     'kong.mock.request.status.200': 7,
+     'kong.mock.request.status.total': 7
+   }
+   END
+   ```
+
+See the [StatsD plugin](/hub/kong-inc/statsd/) 
+documentation for more information about how to use and configure the plugin.
+
+### Clean up the test instance
```
cloudjumpercat, line 84: @rspurgeon Would this heading work for you? I wanted to add more to "clean up" just so users can know what the section will cover at a glance.

src/gateway/kong-production/monitoring/statsd.md
rspurgeon: @cloudjumpercat I've just pushed a revision, please take a look.  I think since we're making the Docker based GW optional, we have to be vague about the instructions for the user wishing to use their own.  UDP networking rules are not easily described here, but it won't be accurate to say that the StatsD in Docker and the existing GW have to be on the "same network" because StatsD is ran in a new Docker network.

In fact, I just realized, the StatsD docker run command is not going to work OOTB if the user does not use the script because it uses a new network which won't exist (--network=kong-quickstart-net), so the user will need to navigate this with our intentionally vague instructions here.

src/gateway/kong-production/monitoring/statsd.md
rspurgeon: @cloudjumpercat  I've just pushed a revision, please take a look

src/gateway/kong-production/monitoring/statsd.md
rspurgeon: @cloudjumpercat I pushed a revision, please take a look.  I personally think just "Clean up" is sufficient, but if you want additional text "test instance" isn't exactly accurate, because the instructions inform the user to cleanup the test kong gateway, database and network along with the StatsD service.  I modified to be vague and say "running software"

Guaris (approved):

rspurgeon:
@Guaris @cloudjumpercat LGTM

src/gateway/kong-production/monitoring/statsd.md
```diff
@@ -1,6 +1,106 @@
...
+
+### Prerequisites
+* [Docker](https://docs.docker.com/get-docker/) is used to run StatsD and supporting services locally. 
+* [curl](https://curl.se/) is used to send requests to {{site.base_gateway}}. `curl` is pre-installed on most systems.
+* [Netcat](http://netcat.sourceforge.net/) is installed as `nc` on the system `PATH`. `nc` is used to send requests 
+  to the StatsD management interface. `nc` is pre-installed on many systems.
+
+### Configure StatsD monitoring
+
+1. Install a test {{site.base_gateway}}:
```
Guaris, line 22:
```suggestion
1. Install {{site.base_gateway}}:
```

Guaris (approved):

PR #4262, opened 2022-08-17
Fix vault bundles description

cloudjumpercat:
### Summary

Updated vaults description based on the following feedback:

> In 2.8.1.3 vault bundles are turned on by default and the out of the box memory consumption increase is immediately visible. I think we should update our docs to point this out and show how to turn off vault bundles if customer is not using secrets-mgmt.

I also updated the kong.conf file in kong/kong-ee: https://github.com/Kong/kong-ee/pull/3635 

I didn't think we'd also need to update [this page](https://docs.konghq.com/gateway/latest/plan-and-deploy/security/secrets-management/#beta-and-general-availability-phases), but let me know if you think we should.

### DIFF
--- a/app/gateway/2.8.x/reference/configuration.md
+++ b/app/gateway/2.8.x/reference/configuration.md
@@ -530,7 +530,7 @@ Comma-separated list of vaults this node should load. Vaults are `bundled` by de
 The specified name(s) will be substituted as such in the Lua namespace:
 `kong.vaults.{name}.*`.
 
-To disable vault bundles, set `vaults` to `off`.
+In version 2.8.1.3, enabling vaults can result in increased memory usage. To conserve memory usage, you can disable vault bundles by setting `vaults` to `off`.
 
 **Default:** `bundled`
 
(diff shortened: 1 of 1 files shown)


mheap:
@cloudjumpercat Should we mention increased memory usage in 2.8.1.3 s as a reason to turn vaults off if you're not using them?

app/gateway/2.8.x/reference/configuration.md
```diff
@@ -530,6 +530,8 @@ Comma-separated list of vaults this node should load. Vaults are `bundled` by de
 The specified name(s) will be substituted as such in the Lua namespace:
 `kong.vaults.{name}.*`.
 
+In version 2.8.1.3, enabling vaults can result in increased memory usage. To conserve memory usage, you can disable vault bundles by setting `vaults` to `off`.
```
lena-larionova, line 533: Q: Is this true _only_ in 2.8.1.3, or from that version onward?

acgoldsmith (approved):
LGTM w/ conclusion that memory usage increase is only expected to happen in 2.8.1.3.

PR #4268, opened 2022-08-18
[DOCU-2459] Gateway icons

lena-larionova:
### Summary
Updating the icons in the new nav to be unique for each section.
* I used the icon we used to use for Dev Portal for the "Develop Custom Plugins" section. It's not an official dev portal icon, it just signifies development - so I think that should be alright.
* Used icons that we already have for most things, had to adjust the learning icon to work for the nav (it's not in use anywhere else). 
* The previous icon used for getting started was a timer, which is supposed to denote getting started quickly. I'm not sure that it really applies (or ever did). The alternative is the learning icon. What do we think?

  Screen Shot 2022-08-18 at 10 13 23 AM Screen Shot 2022-08-18 at 10 14 41 AM

* Fixed flag icon so it doesn't get blurry (img width is set to 20px, but icon width was 19px, so it was getting a tiny bit stretched).

### Reason
https://konghq.atlassian.net/browse/DOCU-2459

### Testing
https://deploy-preview-4268--kongdocs.netlify.app/gateway/latest/


cloudjumpercat (commented):
Everything looks good to me! I think I'd slightly prefer the "learning" icon to the timer icon for Getting Started.

Guaris:
I agree with Diana otherwise LGTM.

acgoldsmith (approved):
LGTM +1 for the learning over timer :)

PR #4275, opened 2022-08-19
add missing single quote

Hayk-S:
### Summary
This PR contains a fix for `acl` route creation command by adding a single quote at the end of the curl command

### Reason
I have noticed that the current example is not working with a missing single quote.

### Testing
With the fix in this PR, the curl works as expected and I am able to create the route.


lena-larionova (approved):
LGTM, nice catch. Just added a word from the topic to our dictionary to fix the failing check.

PR #4276, opened 2022-08-19
Fix PDF generation for single sourced content

mheap:
### Summary
PDF generation is broken for single sourced content

### Reason
To be able to build Mesh PDFs

### Testing

```
cd pdf-generation
npm ci
KONG_DOC_VERSIONS=mesh_1.8.x node run.js
```

### DIFF
--- a/pdf-generation/build-urls.js
+++ b/pdf-generation/build-urls.js
@@ -11,7 +11,13 @@ function extractUrls (input) {
 }
 
 module.exports = function (input) {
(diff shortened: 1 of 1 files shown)


lena-larionova:
it's failing at 
```
Adding http://localhost:3000/mesh/1.8.x/features/windows/
```
 I'm guessing that's because the topic doesn't exist and it's a link only (`generate: false`)

Guaris (approved):

PR #4278, opened 2022-08-19
[DOCU-2432] Kong Manager overview

lena-larionova:
### Summary
* Added a screenshot (from 3.0 test instance) breaking down the Kong Manager interface
  * Went with an individual workspace vs dashboard for all workspaces because then you get to see the side menu, so it's more useful. The overall dashboard looks almost the same for both.
* More details about what you can do in KM
* Table corresponding to screenshot
  * I went with a table over an ordered list here so that I could break the sections up into menus. If I did that with ordered lists, each list automatically starts with 1 again, which defeats the purpose.

### Reason
https://konghq.atlassian.net/browse/DOCU-2432

### Testing
Use Netlify preview


acgoldsmith (approved):
LGTM & I really like this method of explaining the UI elements!!

PR #4282, opened 2022-08-22
fix the service name in graphql quickstart

samugi:
### Summary
The service name was first defined as `graphql-service` but then referred to as `example-service` in the rest of the guide.


lena-larionova (approved):
LGTM, thanks for the fix!

PR #4327, opened 2022-08-25
[Fix] Remove broken image from Rbac guide 

Guaris:
### Summary
Removes a broken image from the RBAC guide. 

### Reason
Reported in Slack

### Testing
/gateway/latest/configure/auth/rbac/


acgoldsmith (approved):
LGTM

PR #4342, opened 2022-08-27
Moved urls redirects

mheap:
### Summary
Add redirects for any entries in `moved_urls.yml`

### Reason
Any moved URLs need an entry in `moved_urls.yml` so that canonical URLs and the "latest version" links work. We can reuse that data to redirect any `/latest/` URLs.

### Testing
Visit `/gateway/latest/install-and-run/rhel/` and it should redirect

Resolves DOCU-2483

### DIFF
--- a/app/_plugins/generators/alias_generator.rb
+++ b/app/_plugins/generators/alias_generator.rb
@@ -1,5 +1,6 @@
 # frozen_string_literal: true
 
+require 'yaml'
 # Alias Generator for Posts using Netlify _redirects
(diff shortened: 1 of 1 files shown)


Guaris:
should you point it to 3.0? or the navbar updates branch?

Guaris:
> should you point it to 3.0? or the navbar updates branch?

I figured it out, this is supposed to point to 3.0

Guaris (approved):

PR #4364, opened 2022-08-30
Update support policy table for 3.0.x.x

lena-larionova:
### Summary
Added an entry for 3.0 in the support policy table.
Followed the 24 month period for support, then another 12 months for end of sunset support, based on the standard set with 2.1.x.

### Reason
Needs updating with release.

### Testing
Netlify


acgoldsmith (approved):
LTGM

PR #4392, opened 2022-09-05
Clarify "unsupported" in installation table

mheap:
### Summary
Switch from "Local only" to "Runs, but no official support available"

### Reason
We have customers running production workloads on some "unsupported" deployments. This PR makes it clearer that they are valid deployment targets, but support is not available

### Testing
/gateway/latest/install/


cloudjumpercat (approved):
LGTM!

PR #4411, opened 2022-09-08
Rename source_url to source_code for plugins

lena-larionova:
### Summary
`source_url` - > `source_code` for plugin docs.

### Reason
`source_url` is reused in two places: for the plugins, to specify the source code repo link; and for autodocs, to set a source link and kick off the autogen content check. Because of this, the plugin source_url has been creating false positives on the check, making it fail incorrectly.

### Testing
Autogen content check should pass.

Netlify: check that the "Code" link still appears in plugin sidebars, for plugins that have source_code set.


Guaris (approved):
Checked and LGTM

PR #4419, opened 2022-09-09
Removing reference to Verbose in the CLI

d19dotca:
### Summary
In testing, there does not appear to be a verbose option, and in the Portal CLI GitHub there is no reference to verbose either.
Reference: https://github.com/Kong/kong-portal-cli

### Reason
Because the docs included mention of verbose mode which doesn't seem to exist in Portal CLI.


cloudjumpercat (approved):
LGTM!

PR #4422, opened 2022-09-09
We switched our convenience tags from Alpine base to Debian base

tyler-ball:
### Summary
☝️ 

Switching to slim images should be a bigger change for consumers, but we should still mention this.

### DIFF
--- a/app/gateway/changelog.md
+++ b/app/gateway/changelog.md
@@ -56,7 +56,7 @@ This ensures that false positive vulnerabilities don't get flagged during securi
 
     If you want to retain or add other dependencies, you can [build custom Kong Docker images](/gateway/3.0.x/install/docker/build-custom-images).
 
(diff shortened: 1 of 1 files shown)


app/gateway/changelog.md
```diff
@@ -56,6 +56,8 @@ This ensures that false positive vulnerabilities don't get flagged during securi
 
     If you want to retain or add other dependencies, you can [build custom Kong Docker images](/gateway/3.0.x/install/docker/build-custom-images).
 
+* The base os for our convenience docker tags (EG, `latest`, `3.0.0.0`, `3.0`) has switched from Alpine to Debian
```
lena-larionova, line 59:
```suggestion
* The base OS for our convenience docker tags (for example, `latest`, `3.0.0.0`, `3.0`) has switched from Alpine to Debian.
```

lena-larionova (approved):

PR #4437, opened 2022-09-12
Add review label to workflow

lena-larionova:
### Summary
Adding the new `review:sme` label to label-checking workflow.

### Reason
Added a new label to split the job of `review:tech` into two. `review:tech` now means internal docs platform changes, whereas `review:sme` is asking for external review (engineering, PM).

### Testing
Putting the label on this PR to test. If it passes, then it worked.
Edit: yep, it worked.


cloudjumpercat (approved):
LGTM!

PR #4439, opened 2022-09-12
[DOCU-2228] UI screenshot

Guaris:
### Summary
Adds one screenshot to the introduction page for Konnect. 

### Reason
[docu-2228](https://konghq.atlassian.net/browse/DOCU-2228)


lena-larionova (changes requested):
I don't think we should add a completely empty screenshot to the landing page. If we add an image here, it needs to have more oomph - illustrate the possibilities, have a bunch of services set up.

Q: Why is this image in `_assets/images/docs/konnect/azure`? It doesn't seem to have anything to do with Azure.

Issue #4250, opened 2022-08-15
Incorrect information provided re minikube


rkazak:
### Where is the problem?

https://docs.konghq.com/kubernetes-ingress-controller/2.5.x/deployment/kong-enterprise/

### What happened?

The Note mentions this

"If you are running Minikube, you will not get an external IP address."

which is not strictly true, as a correctly configured minikube env will.

```
| minikube4 | docker    | docker  | 192.168.49.2  | 8443 | v1.24.3  | Running |     1 | *      |

Test2 /Users/rkazak/minikube/ex1/dev >kubectl get services -n kong
NAME                      TYPE           CLUSTER-IP      EXTERNAL-IP     PORT(S)                      AGE
kong-admin                LoadBalancer   10.96.161.232   192.168.49.10   80:31421/TCP                 2m31s
kong-manager              LoadBalancer   10.96.39.96     192.168.49.11   80:32470/TCP                 2m31s
kong-proxy                LoadBalancer   10.96.226.161   192.168.49.12   80:32405/TCP,443:32508/TCP   2m30s
kong-validation-webhook   ClusterIP      10.106.90.207   <none>          443/TCP                      2m30s
```

### What did you expect to happen?

We should remove that sentence.  Not sure how much we want to document minikube usage, or just point to appropriate pages.

### Code of Conduct and Community Expectations

- [X] I agree to follow this project's Code of Conduct
- [X] I agree to abide by the Community Expectations

mheap:
@rkazak Are you using MetalLB or similar?

afzal442:
/assign
I tjnk, the above experimentation is done on Minikube cluster without MetalB. If you have large memory set during `minikube start ..`, then you will have everything worked.

Rajakavitha1:
Hi @lena-larionova  can we now mark this issue as closed?

lena-larionova:
@Rajakavitha1 absolutely, thanks for addressing it.

PR #4475, opened 2022-09-18
fix: Removed incorrect information provided re minikube

afzal442:
Signed-off-by: afzal442 <afzal442@gmail.com>

### Summary

Closes: #4250 
### Reason

When you have `minikube start --memory` specifying the memory required, everything will spin up in the clluster.

### DIFF
--- a/src/kubernetes-ingress-controller/deployment/kong-enterprise.md
+++ b/src/kubernetes-ingress-controller/deployment/kong-enterprise.md
@@ -87,8 +87,7 @@ postgres                  ClusterIP      10.63.241.104   <none>          5432/TC
 > Note: Depending on the Kubernetes distribution you are using, you might or might
 not see an external IP assigned to the three LoadBalancer type services. Please see
 your provider's guide on obtaining an IP address for a Kubernetes Service of
-type `LoadBalancer`. If you are running Minikube, you will not get an
-external IP address.
+type `LoadBalancer`.
 
 ### Setup Kong Manager
 


src/kubernetes-ingress-controller/deployment/kong-enterprise.md
```diff
@@ -87,8 +87,7 @@ postgres                  ClusterIP      10.63.241.104   <none>          5432/TC
 > Note: Depending on the Kubernetes distribution you are using, you might or might
 not see an external IP assigned to the three LoadBalancer type services. Please see
 your provider's guide on obtaining an IP address for a Kubernetes Service of
-type `LoadBalancer`. If you are running Minikube, you will not get an
-external IP address.
+type `LoadBalancer`.
```
mheap, line 90: There are some cases where you will not get an external IP address using Minikube.

Can we add more information about how to get an external IP address when using Minikube?

mheap (changes requested):

src/kubernetes-ingress-controller/deployment/kong-enterprise.md
afzal442: External IP is always going to be None, unless you deploy the Service as type: LoadBalancer. If you do that, you have to run minikube tunnel to get access to the service.
Ref: https://github.com/kubernetes/minikube/issues/3966

WDYT?
// @mheap

mheap: The page seems accurate as-is. If we want to add more minikube-specific instructions, feel free to raise a PR with those changes

PR #4482, opened 2022-09-20
update ingress example to networking/v1 version

randmonkey:
### Summary

update examples of `ingress` resource in plugin hub to `networking.k8s.io/v1`

### Reason

`Ingress` with version `networking.k8s.io/v1beta1` is removed in k8s 1.22 and later, so the examples of `Ingress` should be updated.


Guaris:
Thank you!

PR #4484, opened 2022-09-20
Fixed the "allowlist" typo.

felderi:
There was a typo "allowlist".  Fixed to add a space between allow and list.


lena-larionova:
Hey Fel! Thanks for opening the PR, but the term is actually "allowlist", one word - same with "denylist". I agree with you that it does look a bit strange, but it's the correct terminology.

PR #4491, opened 2022-09-21
Change "control plane" to "data planes"

lena-larionova:
### Summary
Clarifying info about DP access through firewalls. Previously, the doc was saying that the CP needs access, which is incorrect.

### Reason
See https://kongstrong.slack.com/archives/CQK8J4VN3/p1663685608861879?thread_ts=1663682193.400399&cid=CQK8J4VN3.

### Testing
Netlify.

### DIFF
--- a/app/konnect/network.md
+++ b/app/konnect/network.md
@@ -35,13 +35,17 @@ want to use port `3001` for the proxy, map `3001:8000`.
 
 ## Hostnames
 
-Add the following hostnames to the allowlist to give data planes access through the firewall:
+Data planes initiate the connection to the {{site.konnect_short_name}} control plane.
+They require access through firewalls to communicate with the control plane.
+
+To let data planes request and receive configuration, and send telemetry data,
+add the following hostnames to the firewall allowlist:
 
 * `cloud.konghq.com`: The {{site.konnect_short_name}} platform.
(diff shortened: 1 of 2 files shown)


app/konnect/network.md
```diff
@@ -35,14 +35,17 @@ want to use port `3001` for the proxy, map `3001:8000`.
 
 ## Hostnames
 
-Add the following hostnames to the allowlist to give the
-{{site.konnect_short_name}} control plane access through the firewall:
+Data planes initiate the connection to the {{site.konnect_short_name}} control plane.
```
rosskukulinski, line 38: In the context of Konnect, we still refer to Runtime instance, not Data plane, right?

lena-larionova: This brings back the whole runtime instance vs data plane terminology discussion. Let me take that to the docs team and get back to you.

lena-larionova: Closing the loop: after a chat with the team, agreed that it makes sense to use "runtime instance" in the Konnect context in most situations, including the one you called out here. 

The "data plane" term will still be used, but in a more generic way. Based on that criteria, the "Network Resiliency and Availability" topic remains as-is, using the "data plane" terminology.

Guaris (approved):

PR #4492, opened 2022-09-21
Kong Quickstart: Pipes quickstart to script to bash instead of sh

rspurgeon:
### Summary

Kat Morgan and I discovered on Ubuntu that when piping the script to sh (which is dash on ubuntu) the shell does not respect the bash shebang and the script syntax is invalid. Modifying the instructions to pass the script to bash works on Ubuntu and MacOS.

### Reason
The instructions do not work on Ubuntu

### Testing
Tested code locally on Ubuntu and MacOS.  Did not test the docs build itself as the change is trivial

### DIFF
--- a/src/gateway/get-started/index.md
+++ b/src/gateway/get-started/index.md
@@ -39,7 +39,7 @@ This script uses Docker to run {{site.base_gateway}} and a [PostgreSQL](https://
 1. Run {{site.base_gateway}} with the `quickstart` script:
 
    ```sh
-   curl -Ls get.konghq.com/quickstart | sh -s
+   curl -Ls get.konghq.com/quickstart | bash -s
    ```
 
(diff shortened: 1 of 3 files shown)


usrbinkat (approved):
/LGTM

Tested the following on ubuntu with success
```bash
curl -Ls get.konghq.com/quickstart | bash -s
```

rspurgeon:
> Unrelated to this change, but should we be providing HTTPS urls e.g. `https://get.konghq.com/quickstart` rather than curling HTTP into bash (which could be MITM'd)

Great catch and point, I will file a fix.

rspurgeon:
@mheap Branch https://github.com/Kong/docs.konghq.com/tree/get-kong-quickstart-https has the pending change for `https` (based on this PR)

mheap (approved):

PR #4493, opened 2022-09-21
Add ability to provide k8s-specific examples

mheap:
### Summary
Add the ability to add k8s-specific examples in the plugin hub

### Reason
Some values need specifying inline in the YAML. The existing examples assumed that `curl` was being used

### Testing
See the Kubernetes tab in /hub/kong-inc/exit-transformer/#example-config


cloudjumpercat (approved):
LGTM!

PR #4503, opened 2022-09-23
doc(kic) add 2.7 docs

rainest:
### Summary
Adds KIC 2.7 with a revised Kong 3.x upgrade guide.

Updates the 2.6 compatibility information based on what we've learned about older EOL Kubernetes versions.

This should not be merged automatically. We will merge it once the release is actually out.

### Reason
2.7 is effectively 2.6.1, but our fix required a breaking change, so it's getting a minor version bump instead of a patch release. The upgrade guide refers to 2.7 as the first version that supports Kong 3.x now, which isn't technically true, but is close enough to true because of 2.6's incomplete support. We don't want to maintain separate versions of the guide for 2.6 and 2.7 because the recommended configuration for 2.6 isn't actually possible.

### Testing
https://deploy-preview-4503--kongdocs.netlify.app/kubernetes-ingress-controller/latest/guides/upgrade-kong-3x/
https://deploy-preview-4503--kongdocs.netlify.app/kubernetes-ingress-controller/latest/references/version-compatibility/

### DIFF
--- a/src/kubernetes-ingress-controller/guides/upgrade-kong-3x.md
+++ b/src/kubernetes-ingress-controller/guides/upgrade-kong-3x.md
@@ -23,7 +23,7 @@ upgrade strategy based on the following examples.
 
 ### Update CRDs
 
-KIC 2.6 includes {{site.base_gateway}} 3.x
+KIC 2.7 includes {{site.base_gateway}} 3.x
 compatibility changes to the controller CRDs. Helm does not update CRDs automatically. You must apply them manually
 before upgrading KIC:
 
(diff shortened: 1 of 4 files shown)


randmonkey (approved):

src/kubernetes-ingress-controller/guides/upgrade-kong-3x.md
```diff
@@ -55,26 +55,28 @@ helm upgrade ${YOUR_RELEASE_NAME} kong/kong \
...
 manual updates to Ingress configuration. In Kong 2.x, Kong applied a heuristic based
 on the presence of special characters in a route path to determine if a path
-was a regular expression. This heuristic was imperfect and Kong 3.x has
-removed it, instead requiring that any regular expression begin with a `~`
-prefix.
+was a regular expression. This heuristic was imperfect and Kong 3.x has removed
+it, instead requiring that any regular expression begin with a `~` prefix.
+Ingress does not allow paths that begin with any character other than `/`,
+however, so Ingress rules with a regular expression path must begin with `/~`
+instead.
```
cloudjumpercat, line 79:
```suggestion
was a regular expression. This heuristic was imperfect and {{site.base_gateway}} 3.x removed
it, and instead requires that any regular expression begins with a `~` prefix.
Ingress does not allow paths that begin with any character other than `/`,  so Ingress rules with a regular expression path must begin with `/~`
instead.
```

src/kubernetes-ingress-controller/guides/upgrade-kong-3x.md
```diff
@@ -86,22 +88,35 @@ to 3.x without updating configuration.
...
-The new 3.x paths are also incompatible with 2.x. Adding the `~` character to
+The new 3.x paths are also incompatible with 2.x. Adding the `/~` prefix to
 Ingress rules directly breaks incremental upgrades. To smooth the migration
-process and allow users to update rules gradually, KIC 2.6 includes the
+process and allow users to update rules gradually, KIC 2.7 includes the
 `enableLegacyRegexDetection` option to continue applying the 2.x regular
 expression heuristic on KIC's end.
 
-If the following rules are met, KIC will insert the `~` prefix:
+If the following sets of rules are met, KIC will create a Kong route path with
```
cloudjumpercat, line 97:
```suggestion
If the following sets of rules are met, KIC creates a Kong route path with
```

cloudjumpercat (changes requested):

cloudjumpercat (approved):
LGTM!

PR #4510, opened 2022-09-26
fix(admin-api): fix wrong api path

vm-001:
### Summary

Fix the wrong API path of "Set Target Address As Unhealthy"

### Reason

The correct API path of "Set Target Address As Unhealthy" is `/upstreams/{upstream name or id}/targets/{target or id}/{address}/unhealthy`


Guaris:
Matches this:
https://github.com/Kong/kong/blob/34e2ec189db51d77fea6175c2daab7973e3f2416/autodoc/admin-api/data/admin-api.lua#L1957

LGTM

Guaris (approved):

PR #4520, opened 2022-09-26
[fix]vale passing on /src/

Guaris:
### Summary
Some new things going on here: 
- Updated vale to the latest version. 
- `auto-ignore.txt` - this is a new ignore dictionary it's used for terms that are in automatically generated docs that we can't easily change. 
- `kongterms.yml` - this is a copy of terms.yml with only kong terms in it. scoped to `text` (this does nothing but I wanted to be explicit so that I can build specific rules for other scopes later). 
- terms.yml  - removed kong terms, added the `custom:true` flag. 
- spelling.yml - added the filter `'\W'` to get rid of the apostrophe issues we were having. 
- vale.ini - added `IgnoredClasses = navtab-titles`. 
- vale should be passing on the `/src/` directory. 

### Reason
It had to be done. 

### Testing
run vale locally.


cloudjumpercat (commented):
So far, everything looks good to me. I tested a nav doc, several docs in /src, Konnect docs, and KIC/Mesh docs. But also, I tested 8 different, random docs and got no errors, warnings, or suggestions, which makes me slightly suspicious. I tested the KIC and Mesh docs because I thought for sure there would be spelling errors in there that Vale would flag, but I guess not? 

Ex. "CRD" is listed in kic/guides/upgrade.md and that wasn't flagged even though we only have a rule for "crds". I would expect "CRD" to come up as an error but it wasn't.

Guaris:
> So far, everything looks good to me. I tested a nav doc, several docs in /src, Konnect docs, and KIC/Mesh docs. But also, I tested 8 different, random docs and got no errors, warnings, or suggestions, which makes me slightly suspicious. I tested the KIC and Mesh docs because I thought for sure there would be spelling errors in there that Vale would flag, but I guess not?
> 
> Ex. "CRD" is listed in kic/guides/upgrade.md and that wasn't flagged even though we only have a rule for "crds". I would expect "CRD" to come up as an error but it wasn't.

Acronyms are now filtered by default. There is room to write a rule that checks if an acronym is used before we introduce the full word if we want.

cloudjumpercat:
> > So far, everything looks good to me. I tested a nav doc, several docs in /src, Konnect docs, and KIC/Mesh docs. But also, I tested 8 different, random docs and got no errors, warnings, or suggestions, which makes me slightly suspicious. I tested the KIC and Mesh docs because I thought for sure there would be spelling errors in there that Vale would flag, but I guess not?
> > Ex. "CRD" is listed in kic/guides/upgrade.md and that wasn't flagged even though we only have a rule for "crds". I would expect "CRD" to come up as an error but it wasn't.
> 
> Acronyms are now filtered by default. There is room to write a rule that checks if an acronym is used before we introduce the full word if we want.

That could be helpful, and it is best practice to define an acronym before using it. But I feel like there will be outlier cases where someone wouldn't land on a doc without knowing what that acronym means, so should we always define what it means? Or the acronym is more popular/understood than the original name (ex. AWS).

Guaris:
> > > So far, everything looks good to me. I tested a nav doc, several docs in /src, Konnect docs, and KIC/Mesh docs. But also, I tested 8 different, random docs and got no errors, warnings, or suggestions, which makes me slightly suspicious. I tested the KIC and Mesh docs because I thought for sure there would be spelling errors in there that Vale would flag, but I guess not?
> > > Ex. "CRD" is listed in kic/guides/upgrade.md and that wasn't flagged even though we only have a rule for "crds". I would expect "CRD" to come up as an error but it wasn't.
> > 
> > 
> > Acronyms are now filtered by default. There is room to write a rule that checks if an acronym is used before we introduce the full word if we want.
> 
> That could be helpful, and it is best practice to define an acronym before using it. But I feel like there will be outlier cases where someone wouldn't land on a doc without knowing what that acronym means, so should we always define what it means? Or the acronym is more popular/understood than the original name (ex. AWS).

I just demo'd this using:

```
extends: conditional
message: "'%s' has no definition"
level: suggestion
scope: text
ignorecase: false
first: '\b([A-Z]{3,5})\b'
second: '(?:\b[A-Z][a-z]+ )+\(([A-Z]{3,5})\)'
exceptions:
  - AWS
  - RBAC
  - API
  - POST
  - PATCH
  - URL
  - URI
  - UUID
  - JSON
  - HTTP
  - HTTPS
  - HTML
  - CSS
  - JS
  - JSON
  - SSL
  - CLI
  - DNS
  - YAML
  - YML
  - SNI
  - OIDC
  - SMTP
  - CSV
  - ASCII
  - UTF
  - FIPS
  - GCP
  - GKE
  - GUI
  - BCC
  - PDK

```

And even with that many exceptions, I'm getting 900+ errors. Im going to consider this out of scope for this PR.

PR #4527, opened 2022-09-27
Updated index.md for OAuth2 Authentication

RagunathInGIT:
Change is to mention various ways of how we can call OAuth2/token endpoint to retrieve access_token for client_credentials grant_type.

Where is the problem?
https://docs.konghq.com/hub/kong-inc/oauth2/

What happened?
I dont see documentation of various ways of calling "/oauth2/token" call for client_credentials flow.

What did you expect to happen?
In Client credentials flow, there is a provision to make "oauth2/token" call by passing the client ID, client secret and grant_type as query param or a JSON body or in a url-encoded form data.

I would like to see if this approved, can be submitted for Kong Hackathon 2022 under documentation category.


RagunathInGIT:
Duplicate

PR #4538, opened 2022-09-30
add filter mode to vale

Guaris:
This adds `filter_mode: added` instead of using `__onlymodified`. 

testing:
This is now:
Screen Shot 2022-09-30 at 11 59 38 AM
It creates an output of:

Screen Shot 2022-09-30 at 12 00 48 PM

While only expressing the errors in the changed files:
image


lena-larionova:
Will every PR include filtered findings? Or will that only happen on a merge commit?

lena-larionova (approved):

PR #4556, opened 2022-10-04
Add Incubator to top nav

mheap:
### Summary
Make Kong Incubator the main CTA in the top nav

### Reason
We need the Incubator projects to be adopted to inform our strategic roadmap

### Testing
Try the link at the top of the screen


lena-larionova (approved):
LGTM, as long as we're not concerned about losing the sales/demo CTA.

PR #4564, opened 2022-10-05
mesh: rbac wildcard tag value

jakubdyszkiewicz:
### Summary
This PR adds docs about wildcard tag value matching feature which is being released in Kong Mesh 1.9.1
We usually do not add features to patch releases, but it's an exception. That's why I added a note about the version.

### Reason
Should I link issues to the private repo?

### Testing
Checked with netlify preview


Guaris:
Brought in the latest branch changes, did a small copy edit.

Guaris (approved):

PR #4586, opened 2022-10-11
Add Segment tracking

mheap:
### Summary
Add segment.io tracking to docs

### Reason
Help understand cross-site journeys

### Testing
The CORS headers are set up to allow `*.konghq.com` only. I tested by copying a blocked request from the network panel as `curl` and executing it in a terminal. The event showed up in Segment as intended.

Based on this, I believe that this will work once deployed to production

### DIFF
--- a/app/_layouts/default.html
+++ b/app/_layouts/default.html
@@ -29,6 +29,16 @@
 
     {% include_cached footer.html %} {% include_cached anchor_links.html %}
 
+    {% if jekyll.environment == "production" %}
+      <script>
+        // Segment
+        window.analytics = window.analytics || {};
+        window.analytics._writeKey = "X7EZTdbdUKQ8M6x42SHHPWiEhjsfs1EQ";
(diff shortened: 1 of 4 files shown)


app/_layouts/default.html
```diff
@@ -29,6 +29,16 @@
 
     {% include_cached footer.html %} {% include_cached anchor_links.html %}
 
+    {% if jekyll.environment == "production" %}
+      <script>
```
fabianrbz, line 33: should we place this script after the `/assets/app.js` is loaded so we know that it executes after the library was loaded?

fabianrbz (approved):
I left a minor comment, but otherwise, it looks good ✨

app/_layouts/default.html
mheap: It has to be placed before to set `writeKey` before the library loads as there are some on-load side effects

PR #4598, opened 2022-10-13
Change Fail to Ignore on failurePolicy for Admission Controller

d19dotca:
### Summary
Changing failurePolicy from Fail to Ignore to match 2.5.0 and newer KIC version defaults.

### Reason
In KIC 2.5.0 and higher, we changed the Admission Controller to use a new default of Ignore instead of Fail, so just bringing the docs into alignment too. Ref: https://github.com/Kong/kubernetes-ingress-controller/issues/2501


mheap:
@d19dotca We've changed how we author docs to ensure that when we fix bugs like this, all versions are updated.

Unfortunately it adds a little bit of complexity to how we make changes that occur from a specific version. I've updated this PR to show you how it works. If you want a hand in the future, let me know and we can run through it together

mheap (approved):

PR #4600, opened 2022-10-14
[fix] Key-auth hybrid mode

Guaris:
Removes text that states that key-auth does not work with hybrid mode. Oauth 2.0 is the only plugin that is not fully compatible with hybrid mode. https://docs.konghq.com/hub/plugins/compatibility/

https://konghq.atlassian.net/browse/DOCU-2618


acgoldsmith (approved):
LGTM

PR #4612, opened 2022-10-18
Fix anchor scroll behaviour

mheap:
### Summary
Fix scroll behaviour when clicking an anchor link by replacing JS implementation with CSS. It previously scrolled to the header, but the header was off-screen

### Reason
Papercut fix - it was annoying

### Testing
Try clicking an anchor link in the right hand sidebar. I've tested with the mobile and full screen layouts
https://deploy-preview-4612--kongdocs.netlify.app/gateway/changelog/#2814

### DIFF
--- a/app/_assets/javascripts/app.js
+++ b/app/_assets/javascripts/app.js
@@ -8,23 +8,6 @@ jQuery(function () {
   var $window = $(window);
   var $docs = $("#documentation");
 
-  // smooth scroll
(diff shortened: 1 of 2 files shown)


Guaris:
That's helpful for linking into big reference documents like:
https://deploy-preview-4612--kongdocs.netlify.app/gateway/latest/admin-api/#target-object

Guaris (approved):

PR #4619, opened 2022-10-19
Add missing .important class in non-root user content

mheap:
### Summary
Followup to #4618 

cc @Guaris 

### Reason
There was a missing class, which meant that the blockquote is not being styled correctly


Guaris (approved):

Guaris:
Thanks Michael

PR #4623, opened 2022-10-20
Add jekyll generator single source

fabianrbz:
### Summary

Add `jekyll-generator-single-source` gem and remove the in-repo implementation. 

### Reason
All the changes have been migrated to the gem and it has been released to rubygems. We can now share the code with Kuma's website.

### Testing
Check that the docs are rendered correctly.


fabianrbz:
@mheap looks like we are running the checks with ruby `2.7.2` and specifying `3.1.2` in the `.ruby-version` file.
It is failing because I added ruby  `>= 3.1.2` as a dependency in the gem.  I could change that, but I think we should update the version in the checks to match the one in `.ruby-version` so that we run them with the right version.

mheap:
@fabianrbz Please do update the checks to match `.ruby-version`. Even better if we can update the workflows to read `.ruby-version` and pass that in to `setup-ruby`

mheap (approved):
Validations done:

* The site builds! Single sourced content appears to render as expected ✅
* Check items using `src:` in a nav data file to specify a different source file (deck defaults) ✅
* Items set to `absolute_url` do not get generated ✅

Things are looking good to me :shipit:

PR #4631, opened 2022-10-21
[FTI-4272] add escape_path to route-transformer-adv plugin

catbro666:
### Summary
Add a new option escape_path to the route-transformer-adv plugin.
The new option will enable the ability to escape the path transformed.

Realted PR: https://github.com/Kong/kong-ee/pull/3760

### Reason
https://konghq.atlassian.net/browse/FTI-4272


cloudjumpercat (approved):
I made a few edits to get Vale to pass and so this would only display for 3.1 or later.

PR #4645, opened 2022-10-25
Set Konnect docs to default to no_version: true

lena-larionova:
### Summary
Add `no_version: true` to all docs under `app/konnect/` by default.
This means that you don't need to manually add `no_version: true` to every single Konnect topic.

Updated Konnect topic metadata
* Removed the `no_version: true` metadata entry from every Konnect topic. 
* Fixed `content-type` to be `content_type`

### Reason
To eliminate redundancy and make Konnect doc writing a tiny bit more efficient. If the entire Konnect doc set is unversioned, there should be no reason to have to set this metadata every since time.

### Testing
Netlify.

Check that the Konnect docs have no version dropdown and no "old version" banners anywhere.


lena-larionova:
Remaining broken links are false positives (links to Okta).

Guaris (approved):

PR #4646, opened 2022-10-26
moved statsd plugins into the analytics and monitoring category

sichvoge:
### Summary
Moved statsd plugins into the analytics and monitoring category 

### Reason
Both statsd plugins are about metrics and monitoring and shouldn't belong to the logging category.


cloudjumpercat (approved):
LGTM!

PR #4677, opened 2022-10-31
doc(security) add notice on file read and write permission for kong

fffonion:
### Summary
Add document on file read and write permissions of Kong.

### Reason

FT-3498, FT-2115


fffonion:
Hi docs team, please let me know if this is the correct location for such doc (do we need to add this to navigation?). We will also need the same section for 2.6.x, 2.7.x and 2.8.x.

Guaris (approved):

Guaris:
Lgtm! Added it to the corresponding sections as well.

PR #4678, opened 2022-10-31
Make Segment events more reliable

mheap:
### Summary
Make Segment events triggering more reliable

### Reason
It previously used a 500ms sleep before triggering a page. We now loop every 30ms until analytics.js is loaded

### Testing 
Will only work in prod due to CORS rules. I've seen the request being attempted and failing in dev tools locally

### DIFF
--- a/app/_layouts/default.html
+++ b/app/_layouts/default.html
@@ -29,17 +29,17 @@
 
     {% include_cached footer.html %} {% include_cached anchor_links.html %}
 
-    {% if jekyll.environment == "production" %}
+    <script src="/assets/app.js?v={{ site.time | date: '%s' }}"></script>
+    {% if jekyll.environment != "production" %}
       <script>
-        // Segment
-        window.analytics = window.analytics || {};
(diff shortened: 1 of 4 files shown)


app/_layouts/default.html
```diff
@@ -29,17 +29,17 @@
 
     {% include_cached footer.html %} {% include_cached anchor_links.html %}
 
-    {% if jekyll.environment == "production" %}
+    <script src="/assets/app.js?v={{ site.time | date: '%s' }}"></script>
+    {% if jekyll.environment != "production" %}
```
fabianrbz, line 33: shouldn't this be
```suggestion
    {% if jekyll.environment == "production" %}
```

fabianrbz (approved):
left a small comment, but otherwise it looks good

app/_layouts/default.html
mheap: This is why we have a review process! That was me testing locally and I forgot to revert the change

PR #4685, opened 2022-11-01
Add beta banner to DCR doc

lena-larionova:
### Summary
Adding the beta banner to doc, release note, and minor edits.

### Reason
DCR is in public beta but there's nothing in the docs to note that.

### Testing
Netlify


acgoldsmith (approved):
LGTM and thanks for catching this release announcement outside of regular request process :)

PR #4694, opened 2022-11-02
Update identity.yaml with latest specs

haydenlam:
### Summary
Update identity.yaml with latest OAS updates.

### Reason

- Spelling/Formatting
- Filtering attribute name -> role
- 200 -> 201 upon create


acgoldsmith (approved):
LGTM

Issue #4647, opened 2022-10-26
KIC Prometheus/Grafana guide has a dead link


pmalek:
### Where is the problem?

https://docs.konghq.com/kubernetes-ingress-controller/2.7.x/guides/prometheus-grafana/#install-services

### What happened?

https://docs.konghq.com/kubernetes-ingress-controller/2.7.x/examples/001_multiple-services.yaml is a dead link.

Most likely the intention was to link to https://github.com/Kong/docs.konghq.com/blob/main/src/kubernetes-ingress-controller/examples/001_multiple-services.yaml but for some reason this is not possible (something changed?).

### What did you expect to happen?

The link should not be dead and should link to https://github.com/Kong/docs.konghq.com/blob/main/src/kubernetes-ingress-controller/examples/001_multiple-services.yaml.

### Code of Conduct and Community Expectations

- [X] I agree to follow this project's Code of Conduct
- [X] I agree to abide by the Community Expectations

randmonkey:
I think we should configure somewhere in the repo to make the link correctly to the file in example. The link is used to download the raw content of the file, but this link https://github.com/Kong/docs.konghq.com/blob/main/src/kubernetes-ingress-controller/examples/001_multiple-services.yaml will return a webpage.

PR #4698, opened 2022-11-03
Add manifests to gulps pipeline

fabianrbz:
### Summary
Include static assets to Gulp's pipeline. This is a workaround until we define a proper way -single source- to handle assets. 
This allows contributors to link to them in the docs.
We might want to change the name of the folder to something more meaningful, e.g. `files`, `static`, etc

cc @mheap @pmalek 

Also, we could try something like [this](https://answers.netlify.com/t/jekyll-site-uses-absolute-urls-deploy-previews-navigate-to-main-site/8441/7) for `site.links.web` so we can test this in preview apps and locally when running it with netlify. WDYT?

### Reason
Looks like we dropped support for this when we moved to single sourced content
Related issue https://github.com/Kong/docs.konghq.com/issues/4647


pmalek:
The link generated in the preview - https://docs.konghq.com/assets/kubernetes-ingress-controller/examples/echo-service.yaml - still returns a 404 😞

fabianrbz:
@pmalek yeah :/.

mheap (approved):

pmalek:
I can confirm that the asset URLs work on prod now (e.g. https://docs.konghq.com/assets/kubernetes-ingress-controller/examples/echo-service.yaml).

I can also confirm that changing the `web` link in `jekyll-dev.yml`

```yaml
links:
  web: http://localhost:3000
```

works locally. Would that be acceptable to get merged or do you foresee any problems with this change?

PR #4710, opened 2022-11-03
Fix QA issues in Networking, Monitor & Manage, Resource Docs

cloudjumpercat:
### Summary

Fixed the following:

- https://deploy-preview-4710--kongdocs.netlify.app/mesh/latest/documentation/configuration/: broken link on Overview text (/mesh/2.0.x/documentation/introduction), but I think that’s linking to the wrong spot in Kuma and should be changed to a different link. I added a redirect to a page that I thought would work rather than fix the link in Kuma right now because they froze docs. 
- Changed nav name to "Kong Mesh data collection" to match Kuma.
- Added this missing page: https://deploy-preview-4710--kongdocs.netlify.app/mesh/2.0.x/generated/cmd/kumactl/kumactl_get_zone

I also found a few issues that were both in Kuma and Kong Mesh (formatting issues, won't impact user's workflows I don't think), that I think could be fixed post 2.0 since Kuma docs are frozen right now.

### Reason

Issues found during QA
### Testing

- Does that link work as a replacement for /documentation/introduction? Is there a better one?
- Are you able to navigate to that kumactl_get_zone page?

### DIFF
--- a/app/_data/docs_nav_mesh_2.0.x.yml
+++ b/app/_data/docs_nav_mesh_2.0.x.yml
@@ -295,7 +295,7 @@ items:
       - text: Annotations and labels in Kubernetes mode
         url: /reference/kubernetes-annotations/
         src: /.repos/kuma/app/docs/dev/reference/kubernetes-annotations/
-      - text: Data collection
+      - text: Kong Mesh data collection
         url: /reference/data-collection/
         src: /.repos/kuma/app/docs/dev/reference/data-collection/
       - text: Resources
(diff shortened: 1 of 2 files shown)


lena-larionova:
(edited the netlify links in your description, they were linking to the wrong netlify preview)

lena-larionova:
> Does that link work as a replacement for /documentation/introduction? Is there a better one?
The page works as the redirect, but I actually got rid of that page in https://github.com/Kong/docs.konghq.com/pull/4712. It was a duplicate of the Kong Mesh intro, but for Kuma. So I set the redirect to point to the Kong Mesh intro page instead. 

> Are you able to navigate to that kumactl_get_zone page?
Yep, it works.

PR #4714, opened 2022-11-04
QA for Policies in Mesh 2.0

cloudjumpercat:
### Summary

- OPA Policy support in sidebar needed a space between OPA and policy
- Found a few links to just "/policies", so I redirected them.
- /mesh/2.0.x/policies/protocol-support-in-kuma was a 404 on the new MeshTrace (beta) page and I don't know why it wasn't transformed to "kong-mesh". I just added a temp redirect.
- Found a 404 to a "Policy matching" page that was unlisted in Kuma, so I added it to the nav in Mesh. 

### Reason

Result of QA 
### Testing

- Test redirects
- From the MeshTrace (beta) page, test the link in "You must explicitly specify the protocol for each service and data plane proxy you want to enable tracing for." and "If you don’t understand this table you should read matching docs."


Guaris (approved):
Both redirects work. lgtm

PR #4729, opened 2022-11-07
Add missing beta docs to Policies section

cloudjumpercat:
### Summary

Charly let me know that we were missing the following two docs in the Mesh nav that currently exist in the Kuma nav:

- https://kuma.io/docs/2.0.x/policies/meshaccesslog/
- https://kuma.io/docs/2.0.x/policies/meshtrafficpermission/

I added these docs back in and double checked that no other docs were missing from the nav.

### Reason

Docs existed in Kuma nav but not Mesh nav.
### Testing

Check that the following pages work in the Netlify preview:

- /policies/meshaccesslog/
- /policies/meshtrafficpermission/

### DIFF
--- a/app/_data/docs_nav_mesh_2.0.x.yml
+++ b/app/_data/docs_nav_mesh_2.0.x.yml
@@ -259,6 +259,12 @@ items:
       - text: MeshTrace (Beta)
         url: /policies/meshtrace/
         src: /.repos/kuma/app/docs/2.0.x/policies/meshtrace/
+      - text: MeshAccessLog (Beta)
+        url: /policies/meshaccesslog/
+        src: /.repos/kuma/app/docs/2.0.x/policies/meshaccesslog/
+      - text: MeshTrafficPermission (Beta)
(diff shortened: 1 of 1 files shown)


lena-larionova (approved):
This update LGTM, but it raises another question: does the Mesh team know that these policies are missing from https://kuma.io/policies/? As a followup, we should add them there and to https://docs.konghq.com/mesh/latest/policies/.

lena-larionova:
Re: failing checks: rebase on `main` to fix the linting check. The smoke test one is flaky, so it'll hopefully pass on the rebase.

cloudjumpercat:
> This update LGTM, but it raises another question: does the Mesh team know that these policies are missing from https://kuma.io/policies/? As a followup, we should add them there and to https://docs.konghq.com/mesh/latest/policies/.

Good catch! I asked Charly about it and he said they shouldn't be listed there because they are "meant to replace existing ones". So once they are out of beta, I think they'll replace the existing ones.

PR #4739, opened 2022-11-08
Add specs to plugins

fabianrbz:
### Summary
Add specs and refactor a bit the `KumaToKongMesh` plugin and a github action that runs them.

### Reason
While we were prepping the release we found a few bugs in the logic, these specs will prevent them from happening again and act a bit as documentaiton


mheap (approved):
Is it work combining the new workflow with https://github.com/Kong/docs.konghq.com/blob/main/.github/workflows/run-smoke-tests.yml and having a single test workflow with multiple jobs?

mheap (approved):

PR #4745, opened 2022-11-08
Konnect: Renaming the default runtime group

lena-larionova:
### Summary
Remove line from doc about not being able to remove the default runtime group.

Added a short release note with today's date, as we don't know when the functionality was pushed live. 

### Reason
The default runtime group can now we renamed. See https://kongstrong.slack.com/archives/C01888Q7PJ7/p1667941368541619.

### Testing
Netlify


acgoldsmith (approved):
LGTM

PR #4758, opened 2022-11-10
CTA to HTTPS

acgoldsmith:
### Summary
Changing CTA link to HTTPS, which Kaitlyn confirms is correct next step.

### Reason
Field pointing to HTTP is flagged as insecure

### Testing
Field pointing to HTTPS should not be flagged as insecure


cloudjumpercat (approved):
LGTM!

PR #4771, opened 2022-11-14
Refactor the Single Source Plugin generator and add specs

fabianrbz:
### Summary
Refactor the `Plugin Single Source` generator and add specs

### Reason
The generator was too complex. This refactor not only add specs but also simplifies the implementation, making future contributions easier and paves the way for the new hub.

### Testing
The plugin hub should look like/work exactly the same


fabianrbz:
@mheap I think this one is ready. I couldn't find any instance for which [this](https://github.com/Kong/docs.konghq.com/blob/main/app/_plugins/generators/plugin_single_source_generator.rb#L28-L33) or [this](https://github.com/Kong/docs.konghq.com/blob/main/app/_plugins/generators/plugin_single_source_generator.rb#L67) was necessary, maybe it was used in the past?

mheap (approved):
@fabianrbz Those lines were used in the past for backwards compatibility. All plugins have now been updated so they're no longer needed

PR #4777, opened 2022-11-14
[DOCU-2790] Retire konnect legacy docs

lena-larionova:
### Summary
* Deleting konnect legacy docs. They don't need to be archived or kept anywhere for reference, as the old environment will no longer be accessible by any users.
* Deleting old Konnect API docs from the archive. They were kept there as reference until we got a new API for Konnect, which we now have.
* Search for and scrub konnect.konghq.com from the rest of the docs. 
* Redirects from old docs to /konnect/

### Reason
Now that Konnect accounts have all been migrated from konnect.konghq.com to cloud.konghq.com, we can remove legacy docs from the site. 

### Testing
https://deploy-preview-4777--kongdocs.netlify.app/konnect/

* Test redirects from /konnect/legacy/ to just /konnect/
* Check that the legacy docs no longer display on the site


Guaris (approved):

lena-larionova:
> With the redirect written this way if a customer has this bookmarked https://docs.konghq.com/konnect/legacy/runtime-manager/gateway-runtime-docker/ it would redirect to `/konnect/` Why not have it point here: https://docs.konghq.com/konnect/runtime-manager/runtime-instances/gateway-runtime-docker/

Based on conversation outside of GH, keeping the redirects more general. We already have redirects for the old locations of the docs, pre-IA rework. For Konnect legacy docs - which were always meant to be temporary - the usage was always extremely small, so maintaining a large list of redirects for docs that we only had for a few months is an unnecessary maintenance burden for us.

Screen Shot 2022-11-17 at 11 19 47 AM

lena-larionova:
Smoke test is being flaky, isn't an actual failure. Merging with that in mind.

PR #4801, opened 2022-11-21
Fix flaky spec

fabianrbz:
### Summary
Improve flaky smoke test.


fabianrbz:
ok, so 10 consecutive 🟢 builds 🎉 🎉 🎉

PR #4799, opened 2022-11-19
doc(kic) deprecate KongIngress for 2.8

rainest:
### Summary
Add deprecation notices for KongIngress if the KIC version is 2.8 or higher.

Add a new guide for routing by header using annotations.

Add new 2.8 annotations to annotations reference.

Kind of in limbo before we have a 2.8 to add. The existing guides are fine to go in as-is, since the new content won't go live until there's a 2.8, but until then the header routing guide is an orphan--no nav content to add it to yet.

### Reason

Related to https://github.com/Kong/kubernetes-ingress-controller/issues/3017 and https://github.com/Kong/kubernetes-ingress-controller/issues/3174

### Testing

Waiting on Netlify for formatting validation.

Partial testing of the new header annotations guide against nightly.

PR #4813, opened 2022-11-22
docs(kic): use generated CRDs reference

czeslavo:

### DIFF
--- a/src/kubernetes-ingress-controller/references/custom-resources.md
+++ b/src/kubernetes-ingress-controller/references/custom-resources.md
@@ -1,458 +1,549 @@
----
-title: Custom Resource Definitions
----
-
-The Ingress Controller can configure Kong specific features
-using several [Custom Resource Definitions(CRDs)][k8s-crd].
-
-Following CRDs enables users to declaratively configure all aspects of Kong:
-
-- [**KongPlugin**](#kongplugin): This resource corresponds to
-  the [Plugin][kong-plugin] entity in Kong.
-- [**KongIngress**](#kongingress): This resource provides fine-grained control
-  over all aspects of proxy behaviour like routing, load-balancing,
-  and health checking. It serves as an "extension" to the Ingress resources
-  in Kubernetes.
-- [**KongConsumer**](#kongconsumer):
-  This resource maps to the [Consumer][kong-consumer] entity in Kong.
-- [**TCPIngress**](#tcpingress):
-  This resource can configure TCP-based routing in Kong for non-HTTP
-  services running inside Kubernetes.
-- [**UDPIngress**](#udpingress):
-  This resource can configure UDP-based routing in Kong.
-- [**KongCredential (Deprecated)**](#kongcredential-deprecated):
-  This resource maps to
-  a credential (key-auth, basic-auth, jwt, hmac-auth) that is associated with
-  a specific KongConsumer.
-
-## KongPlugin
-
-This resource provides an API to configure plugins inside Kong using
-Kubernetes-style resources.
-
-Please see the [concept](/kubernetes-ingress-controller/{{page.kong_version}}/concepts/custom-resources/#KongPlugin)
-document for how the resource should be used.
-
-The following snippet shows the properties available in KongPlugin resource:
-
-```yaml
-apiVersion: configuration.konghq.com/v1
-kind: KongPlugin
-metadata:
-  name: <object name>
-  namespace: <object namespace>
-disabled: <boolean>  # optionally disable the plugin in Kong
-config:              # configuration for the plugin
-  key: value
-configFrom:
-  secretKeyRef:
-     name: <Secret name>
-     key: <Secret key>
-plugin: <name-of-plugin> # like key-auth, rate-limiting etc
-{% if_version gte:2.6.x %}
-ordering:
-  before:
-    <phase>:
-    - <plugin>
-    - <plugin>
-  after:
-    <phase>:
-    - <plugin>
-    - <plugin>
-{% endif_version %}
-```
-
-- `config` contains a list of `key` and `value`
-  required to configure the plugin.
-  All configuration values specific to the type of plugin go in here.
-  Please read the documentation of the plugin being configured to set values
-  in here. For any plugin in Kong, anything that goes in the `config` JSON
-  key in the Admin API request, goes into the  `config` YAML key in this resource.
-  Please use a valid JSON to YAML convertor and place the content under the
-  `config` key in the YAML above.
-- `configFrom` contains a reference to a Secret and key, where the key contains
-  a complete JSON or YAML configuration. This should be used when the plugin
-  configuration contains sensitive information, such as AWS credentials in the
-  Lambda plugin or the client secret in the OIDC plugin. Only one of `config`
-  or `configFrom` may be used in a KongPlugin, not both at once.
-- `plugin` field determines the name of the plugin in Kong.
-  This field was introduced in {{site.kic_product_name}} 0.2.0.
-{% if_version gte:2.6.x %}
-- `ordering` is only available on {{site.ee_product_name}}. `<phase>` is a
-  request processing phase (for example, `access` or `body_filter`) and
-  `<plugin>` is the name of the plugin that will run before or after the
-  KongPlugin. For example, a KongPlugin with `plugin: rate-limiting` and
-  `before.access: ["key-auth"]` will create a rate limiting plugin that limits
-  requests _before_ they are authenticated.
-{% endif_version %}
-
-**Please note:** validation of the configuration fields is left to the user
-by default. It is advised to setup and use the admission validating controller
-to catch user errors.
-
-The plugins can be associated with Ingress
-or Service object in Kubernetes using `konghq.com/plugins` annotation.
-
-### Examples
-
-#### Applying a plugin to a service
-
-Given the following plugin:
-
-```yaml
-apiVersion: configuration.konghq.com/v1
-kind: KongPlugin
-metadata:
-  name: request-id
-config:
-  header_name: my-request-id
-  echo_downstream: true
-plugin: correlation-id
-```
-
-It can be applied to a service by annotating like:
-
-```yaml
-apiVersion: v1
-kind: Service
-metadata:
-  name: myapp-service
-  labels:
-     app: myapp-service
-  annotations:
-     konghq.com/plugins: request-id
-spec:
-  ports:
-  - port: 80
-    targetPort: 80
-    protocol: TCP
-    name: myapp-service
-  selector:
-    app: myapp-service
-```
-
-#### Applying a plugin to an ingress
-
-The KongPlugin above can be applied to a specific ingress (route or routes):
-
-```yaml
-apiVersion: networking.k8s.io/v1
-kind: Ingress
-metadata:
-  name: demo-example-com
-  annotations:
-    konghq.com/plugins: request-id
-spec:
-  ingressClassName: kong
-  rules:
-  - host: example.com
-    http:
-      paths:
-      - path: /bar
-        pathType: ImplementationSpecific
-        backend:
-          service:
-            name: echo
-            port:
-              number: 80
-```
-
-A plugin can also be applied to a specific KongConsumer by adding
-`konghq.com/plugins` annotation to the KongConsumer resource.
-
-Please follow the
-[Using the KongPlugin resource](/kubernetes-ingress-controller/{{page.kong_version}}/guides/using-kongplugin-resource)
-guide for details on how to use this resource.
-
-#### Applying a plugin with a secret configuration
-
-The plugin above can be modified to store its configuration in a secret:
-
-```yaml
-apiVersion: configuration.konghq.com/v1
-kind: KongPlugin
-metadata:
-  name: request-id
-configFrom:
-  secretKeyRef:
-    name: plugin-conf-secret
-    key: request-id
-plugin: correlation-id
-```
-
-```yaml
-apiVersion: v1
-kind: Secret
-metadata:
-  name: plugin-conf-secret
-stringData:
-  request-id: |
-    header_name: my-request-id
-    echo_downstream: true
-type: Opaque
-```
-
-## KongClusterPlugin
-
-A `KongClusterPlugin` is same as `KongPlugin` resource. The only differences
-are that it is a Kubernetes cluster-level resource instead of a namespaced
-resource, and can be applied as a global plugin using labels.
-
-Please consult the [KongPlugin](#kongplugin) section for details.
-
-*Example:*
-
-KongClusterPlugin example:
-
-```yaml
-apiVersion: configuration.konghq.com/v1
-kind: KongClusterPlugin
-metadata:
-  name: request-id
-  annotations:
-    kubernetes.io/ingress.class: <controller ingress class, "kong" by default>
-  labels:
-    global: "true"   # optional, if set, then the plugin will be executed
-                     # for every request that Kong proxies
-                     # please note the quotes around true
-config:
-  header_name: my-request-id
-configFrom:
-    secretKeyRef:
-       name: <Secret name>
-       key: <Secret key>
-       namespace: <Secret namespace>
-plugin: correlation-id
-{% if_version gte:2.6.x %}
-ordering:
-  before:
-    <phase>:
-    - <plugin>
-    - <plugin>
-  after:
-    <phase>:
-    - <plugin>
-    - <plugin>
-{% endif_version %}
-```
-
-As with KongPlugin, only one of `config` or `configFrom` can be used.
-
-Setting the label `global` to `"true"` will apply the plugin globally in Kong,
-meaning it will be executed for every request that is proxied via Kong.
-
-## KongIngress
-
-{:.note}
-> **Note:** Many fields available on KongIngress are also available as
-> [annotations](/kubernetes-ingress-controller/{{page.kong_version}}/references/annotations).
-> You can add these annotations directly to Service and Ingress resources
-> without creating a separate KongIngress resource. When an annotation is
-> available, it is the preferred means of configuring that setting, and the
-> annotation value will take precedence over a KongIngress value if both set
-> the same setting.
-
-Ingress resource spec in Kubernetes can define routing policies
-based on HTTP Host header and paths.
-While this is sufficient in most cases,
-sometimes, users may want more control over routing at the Ingress level.
-`KongIngress` serves as an "extension" to Ingress resource.
-It is not meant as a replacement to the
-`Ingress` resource in Kubernetes.
-
-Please read the [concept](/kubernetes-ingress-controller/{{page.kong_version}}/concepts/custom-resources/#kongingress)
-document for why this resource exists and how it relates to the existing
-Ingress resource.
-
-Using `KongIngress`, all properties of [Upstream][kong-upstream],
-[Service][kong-service], and
-[Route][kong-route] entities in Kong related to an Ingress resource
-can be modified.
-
-Once a `KongIngress` resource is created, it needs to be associated with
-an Ingress or Service resource using the following annotation:
-
-```yaml
-konghq.com/override: kong-ingress-resource-name
-```
-
-Specifically,
-
-- To override any properties related to health-checking, load-balancing,
-  or details specific to a service, add the annotation to the Kubernetes
-  Service that is being exposed via the Ingress API.
-- To override routing configuration (like protocol or method based routing),
-  add the annotation to the Ingress resource.
-
-Please follow the
-[Using the KongIngress resource](/kubernetes-ingress-controller/{{page.kong_version}}/guides/using-kongingress-resource)
-guide for details on how to use this resource.
-
-For reference, the following is a complete spec for KongIngress (for property documentation, see [Upstream](/gateway/latest/admin-api/#upstream-object), [Service](/gateway/latest/admin-api/#service-object) and [Route](/gateway/latest/admin-api/#route-object] entities))
-
-```yaml
-apiVersion: configuration.konghq.com/v1
-kind: KongIngress
-metadata:
-  name: configuration-demo
-upstream:
-  slots: 10000
-  hash_on: none
-  hash_fallback: none
-  healthchecks:
-    threshold: 25
-    active:
-      concurrency: 10
-      healthy:
-        http_statuses:
-        - 200
-        - 302
-        interval: 0
-        successes: 0
-      http_path: "/"
-      timeout: 1
-      unhealthy:
-        http_failures: 0
-        http_statuses:
-        - 429
-        interval: 0
-        tcp_failures: 0
-        timeouts: 0
-    passive:
-      healthy:
-        http_statuses:
-        - 200
-        successes: 0
-      unhealthy:
-        http_failures: 0
-        http_statuses:
-        - 429
-        - 503
-        tcp_failures: 0
-        timeouts: 0
-proxy:
-  protocol: http
-  path: /
-  connect_timeout: 10000
-  retries: 10
-  read_timeout: 10000
-  write_timeout: 10000
-route:
-  methods:
-  - POST
-  - GET
-  regex_priority: 0
-  strip_path: false
-  preserve_host: true
-  protocols:
-  - http
-  - https
-```
-
-## TCPIngress
-
-The Ingress resource in Kubernetes is HTTP-only.
-This custom resource is modeled similar to the Ingress resource but for
-TCP and TLS SNI based routing purposes:
-
-```yaml
-apiVersion: configuration.konghq.com/v1beta1
-kind: TCPIngress
-metadata:
-  name: <object name>
-  namespace: <object namespace>
-  annotations:
-    kubernetes.io/ingress.class: <controller ingress class, "kong" by default>
-spec:
-  rules:
-  - host: <SNI, optional>
-    port: <port on which to expose this service, required>
-    backend:
-      serviceName: <name of the kubernetes service, required>
-      servicePort: <port number to forward on the service, required>
-```
-
-If `host` is not specified, then port-based TCP routing is performed. Kong
-doesn't care about the content of TCP stream in this case.
-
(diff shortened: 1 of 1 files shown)


src/kubernetes-ingress-controller/references/custom-resources.md
```diff
@@ -1,458 +1,429 @@
...
+Package v1 contains API Schema definitions for the konghq.com v1 API group.
+
+- [KongClusterPlugin](#kongclusterplugin)
+- [KongConsumer](#kongconsumer)
+- [KongIngress](#kongingress)
+- [KongPlugin](#kongplugin)
+
+
+
+### ConfigSource
```
randmonkey, line 25: Could we put the internal structures in the lower place of the page?

src/kubernetes-ingress-controller/references/custom-resources.md
```diff
@@ -1,458 +1,429 @@
...
+
+
+
+ConfigSource is a wrapper around SecretValueFromSource.
+
+| Field | Description |
+| --- | --- |
+| `secretKeyRef` _[SecretValueFromSource](#secretvaluefromsource)_ |  |
```
randmonkey, line 33: The empty `Description` column here makes up a strange view of this table. Could we add some contents here to fill up the table? for example 
```
Specifies name and key of secret to refer. namespace is implicitly specified to the same as the object where it is in. 
```

randmonkey (commented):
Genrally :+1:, only  some minor suggestions.

src/kubernetes-ingress-controller/guides/using-kongplugin-resource.md
```diff
@@ -436,6 +470,10 @@ credentials:
 kongconsumer.configuration.konghq.com/harry configured
 ```
 
+**Please note:** validation of the configuration fields is left to the user
+by default. It is advised to setup and use the admission validating controller
+to catch user errors.
+
```
rainest, line 476: This should be earlier, in the introduction, use a [warning admonition](https://docs.konghq.com/contributing/markdown-rules/#admonitions), and include a link to the admission controller guide (`/deployment/admission-webhook/`).

The guide doesn't strictly require this (and the guides are generally simplified for brevity), so this should call out that this is mostly a production concern, e.g.

> The controller does not validate configuration by default. Although not required for this guide, production instances should [enable the admission controller](/deployment/admission-webhook/) to validate plugins.

src/kubernetes-ingress-controller/references/custom-resources.md
```diff
@@ -1,458 +1,429 @@
...
-
-For help configuring credentials for the `KongConsumer` Please refer to the [using the Kong Consumer and Credential resource](/kubernetes-ingress-controller/{{page.kong_version}}/guides/using-consumer-credential-resource) guide.
-
-[k8s-crd]: https://kubernetes.io/docs/tasks/access-kubernetes-api/extend-api-custom-resource-definitions/
-[kong-consumer]: /gateway/latest/admin-api/#consumer-object
-[kong-plugin]: /gateway/latest/admin-api/#plugin-object
-[kong-upstream]: /gateway/latest/admin-api/#upstream-object
-[kong-service]: /gateway/latest/admin-api/#service-object
-[kong-route]: /gateway/latest/admin-api/#route-object
+[This document is generated by KIC's 'generate.docs' make target, DO NOT EDIT]: #
```
rainest, line 6: Any reason this is a link definition instead of a comment? I guess it doesn't render, but I am confused by the current use of Markdown here.

rainest (changes requested):
Is there a way to display nested resources sometimes, similar to how https://gateway-api.sigs.k8s.io/v1alpha2/references/spec/#gateway.networking.k8s.io/v1beta1.Gateway displays GatewaySpec under Gateway?

How will/should this handle version splits? I think, realistically, we have to auto-generate separate files. The update PRs would then use nav links to the specific version. Somehow having the template engine insert `if_versions` correctly instead sounds a bit dubious.

How can this handle content that doesn't quite fit in the struct field comments? For example, the [proposed deprecation warning](https://github.com/Kong/docs.konghq.com/pull/4799/files?show-viewed-files=true&file-filters%5B%5D=#diff-bc8283c8425ac7927024efe717ef16bf95bef9b5a2e47f8e71c01e7e9e2eae7f) is a docs site admonition and shouldn't appear in the schema comments. [includes](https://jekyllrb.com/docs/includes/) are probably a good fit, though we'd need some way to then get the include definitions in the appropriate location within the generated doc. Some of the content originally migrated in https://github.com/Kong/kubernetes-ingress-controller/pull/3196 may be a better fit for those than field comments.

src/kubernetes-ingress-controller/references/custom-resources.md
czeslavo: I used this way to include a comment following the top-voted answer to "[Comments in Markdown](https://stackoverflow.com/questions/4823468/comments-in-markdown)" on SO. I missed the part telling it's the best solution if you want to make the comment not included in the resulting HTML. For us, it shouldn't be a problem even if the comment ends up in the resulting HTML. I will change it to a regular `<!-- -->` style comment.

src/kubernetes-ingress-controller/guides/using-kongplugin-resource.md
czeslavo: 78d3c58

src/kubernetes-ingress-controller/guides/using-kongplugin-resource.md
```diff
@@ -274,6 +277,37 @@ Via: kong/2.8.1
...
+kind: Secret
+metadata:
+  name: plugin-conf-secret
+stringData:
+  add-response-header: |
+    add:
+      headers:
+        - "demo: injected-by-kong"
+type: Opaque
+```
```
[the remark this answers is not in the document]
czeslavo: Fair point, I changed it to better fit the whole guide (also adding the expected outputs): 78d3c58

czeslavo:
> Is there a way to display nested resources sometimes, similar to how https://gateway-api.sigs.k8s.io/v1alpha2/references/spec/#gateway.networking.k8s.io/v1beta1.Gateway displays GatewaySpec under Gateway?

It's not possible out of the box apparently with this tool. [I tried](https://github.com/Kong/kubernetes-ingress-controller/pull/3213/files) to modify our templates so that a nested HTML table is displayed when the field is a type from the same package. It didn't look terrible in my local renderer and [Github's](https://github.com/Kong/kubernetes-ingress-controller/blob/ced336a95bb1de114f07c07630b016215a02a131/docs/api-reference.md) one, but in docs.konghq.com it didn't work:
image
 
The best what I was able to do is [change the order of types being rendered](https://github.com/Kong/kubernetes-ingress-controller/pull/3211). Top-level structs are now displayed first as @randmonkey [was suggesting](https://github.com/Kong/docs.konghq.com/pull/4813#discussion_r1036727462).

src/kubernetes-ingress-controller/references/custom-resources.md
czeslavo: https://github.com/Kong/kubernetes-ingress-controller/pull/3210

czeslavo:
> How will/should this handle version splits? I think, realistically, we have to auto-generate separate files. The update PRs would then use nav links to the specific version. Somehow having the template engine insert if_versions correctly instead sounds a bit dubious.
>
>Edit: per chat with the docs team, the recommended approach for this is indeed to use separate files for each version and link to the appropriate one for nav, so this needs to create a new 2.7-specific document and probably keep the existing doc for <=2.6 (we can't generate older ones because they're missing parts of the reference in code comments).

I think I updated this to work as the docs team has advised: 23d0600

czeslavo:
> How can this handle content that doesn't quite fit in the struct field comments? For example, the [proposed deprecation warning](https://github.com/Kong/docs.konghq.com/pull/4799/files?show-viewed-files=true&file-filters%5B%5D=#diff-bc8283c8425ac7927024efe717ef16bf95bef9b5a2e47f8e71c01e7e9e2eae7f) is a docs site admonition and shouldn't appear in the schema comments. [includes](https://jekyllrb.com/docs/includes/) are probably a good fit, though we'd need some way to then get the include definitions in the appropriate location within the generated doc. Some of the content originally migrated in https://github.com/Kong/kubernetes-ingress-controller/pull/3196 may be a better fit for those than field comments.

If we had a Jekyll function that includes a file if it exists but doesn't fail if it doesn't, we could modify our templates to generate something similar to:
```md
### KongClusterPlugin

{% include_relative_if_exists kong_cluster_plugin.md %}
```

and create files only for the CRDs that we wanna write additional content for.

src/kubernetes-ingress-controller/references/custom-resources.md
czeslavo: PTAL at https://github.com/Kong/kubernetes-ingress-controller/pull/3211

src/kubernetes-ingress-controller/references/custom-resources.md
czeslavo: PTAL at https://github.com/Kong/kubernetes-ingress-controller/pull/3211

rainest:
Re [table nesting](https://github.com/Kong/docs.konghq.com/pull/4813#issuecomment-1335420607) I expect only the docs team can answer if there's a way to do that. If not the order override is probably the best we can get.

Re includes, does including an empty file (maybe with whitespace chomps) cause an issues? I don't know if there's an include-if-exists, but we could just create placeholders and only fill in the ones we need. Probably only for the exported types to avoid a ton that we'd never use.

Minor note since it confused me earlier (I need to add content to the style guide for it): includes require you pass in the page version for them to render `if_version`s correctly, e.g.

```
{% include_cached /md/kic/class.md kong_version=page.kong_version %}
```

czeslavo:
> Re https://github.com/Kong/docs.konghq.com/pull/4813#issuecomment-1335420607 I expect only the docs team can answer if there's a way to do that. If not the order override is probably the best we can get.

I asked the docs team and the only way to make this work is to use HTML for both tables. It becomes problematic as the generator we use is best suited for Markdown generating (it has custom template functions injected which generate e.g. markdown links which won't work in HTML). I suggest skipping this for now as it's not critical. @rainest Are you ok with that?

czeslavo:
> Re includes, does including an empty file (maybe with whitespace chomps) cause an issues? I don't know if there's an include-if-exists, but we could just create placeholders and only fill in the ones we need. Probably only for the exported types to avoid a ton that we'd never use.
> 
> Minor note since it confused me earlier (I need to add content to the style guide for it): includes require you pass in the page version for them to render `if_version`s correctly, e.g.
> 
> ```
> {% include_cached /md/kic/class.md kong_version=page.kong_version %}
> ```

Yeah, seems like including an empty file is not problematic. 👍 PTAL: a42042f

rainest (approved):
Includes/formatting/update strategy LGTM at this point.

app/_includes/md/kubernetes-ingress-controller/kong_ingress_description.md
```diff
@@ -0,0 +1,30 @@
+It serves as an "extension" to Ingress resource. It is not meant as a replacement to the Ingress resource in Kubernetes. 
+Ingress resource spec in Kubernetes can define routing policies based on HTTP Host header and paths.
```
cloudjumpercat, line 2:
```suggestion
The Ingress resource spec in Kubernetes can define routing policies based on HTTP Host header and paths. 
```

app/_includes/md/kubernetes-ingress-controller/kong_plugin_description.md
```diff
@@ -0,0 +1 @@
+Plugins can be associated with Ingress or Service object in Kubernetes using `konghq.com/plugins` annotation.
```
cloudjumpercat, line 1:
```suggestion
Plugins can be associated with the Ingress or Service object in Kubernetes using `konghq.com/plugins` annotation.
```

app/_includes/md/kubernetes-ingress-controller/tcp_ingress_description.md
```diff
@@ -0,0 +1,2 @@
+The Ingress resource in Kubernetes is HTTP-only. This custom resource is modeled similar to the 
+Ingress resource but for TCP and TLS SNI based routing purposes.
```
cloudjumpercat, line 2:
```suggestion
Ingress resource, but for TCP and TLS SNI based routing purposes.
```

app/_includes/md/kubernetes-ingress-controller/udp_ingress_description.md
```diff
@@ -0,0 +1,2 @@
+It makes it possible to route traffic to your UDP services using Kong (e.g. DNS, Game Servers, etc.).
```
cloudjumpercat, line 1:
```suggestion
It makes it possible to route traffic to your UDP services using Kong (for example, DNS or Game Servers). 
```

app/_includes/md/kubernetes-ingress-controller/udp_ingress_description.md
```diff
@@ -0,0 +1,2 @@
+It makes it possible to route traffic to your UDP services using Kong (e.g. DNS, Game Servers, etc.). 
+For each rule provided in the spec the Kong proxy environment must be updated to listen to UDP on that port as well.
```
cloudjumpercat, line 2:
```suggestion
For each rule provided in the spec, the Kong proxy environment must be updated to listen to UDP on that port as well.
```

app/_src/kubernetes-ingress-controller/guides/using-kongclusterplugin-resource.md
```diff
@@ -225,6 +230,69 @@ If you send requests to `PROXY_IP` now, you will see that the header is not
 injected in the responses. The reason being that we have created a
 resource but we have not told Kong when to execute the plugin.
 
+## Storing plugin configuration in a Secret
+
+You can store plugin configuration in a Secret to secure sensitive configuration. A secret can be stored in any
```
cloudjumpercat, line 235:
```suggestion
You can store a plugin configuration in a Secret to secure sensitive configuration. A secret can be stored in any
```

app/_src/kubernetes-ingress-controller/guides/using-kongclusterplugin-resource.md
```diff
@@ -225,6 +230,69 @@ If you send requests to `PROXY_IP` now, you will see that the header is not
...
+namespace.
+
+First, create a namespace in which we will create the secret:
+```sh
+kubectl create namespace secret-ns
+```
+
+The output is similar to the following:
```
cloudjumpercat, line 243:
```suggestion
Example output:
```

app/_src/kubernetes-ingress-controller/guides/using-kongclusterplugin-resource.md
```diff
@@ -225,6 +230,69 @@ If you send requests to `PROXY_IP` now, you will see that the header is not
...
+```
+namespace/secrets-ns created
+```
+
+Next, create a Secret with a key for each field you wish to configure:
```
cloudjumpercat, line 248:
```suggestion
Next, create a Secret with a key for each field you want to configure:
```

app/_src/kubernetes-ingress-controller/guides/using-kongclusterplugin-resource.md
```diff
@@ -225,6 +230,69 @@ If you send requests to `PROXY_IP` now, you will see that the header is not
...
+stringData:
+  add-response-header: |
+    add:
+      headers:
+        - "demo: injected-by-kong"
+type: Opaque
+' | kubectl apply -f -
+```
+
+The output is similar to the following:
```
cloudjumpercat, line 265:
```suggestion
Example output:
```

app/_src/kubernetes-ingress-controller/guides/using-kongclusterplugin-resource.md
```diff
@@ -225,6 +230,69 @@ If you send requests to `PROXY_IP` now, you will see that the header is not
...
+```
+secret/plugin-conf-secret created
+```
+
+{:.note}
+> KongClusterPlugin is a cluster-level resource, therefore specifying a secret's namespace is
+> required while referring it in the `configFrom` field.
+
+Now, create a KongClusterPlugin with a `configFrom` field referring that Secret:
```
cloudjumpercat, line 274:
```suggestion
Now, create a KongClusterPlugin with a `configFrom` field referring to that Secret:
```

app/_src/kubernetes-ingress-controller/guides/using-kongclusterplugin-resource.md
```diff
@@ -225,6 +230,69 @@ If you send requests to `PROXY_IP` now, you will see that the header is not
...
+```
+
+The output is similar to the following:
+```
+secret/plugin-conf-secret created
+```
+
+{:.note}
+> KongClusterPlugin is a cluster-level resource, therefore specifying a secret's namespace is
+> required while referring it in the `configFrom` field.
```
cloudjumpercat, line 272:
```suggestion
> required while referring to it in the `configFrom` field.
```

app/_src/kubernetes-ingress-controller/guides/using-kongclusterplugin-resource.md
```diff
@@ -225,6 +230,69 @@ If you send requests to `PROXY_IP` now, you will see that the header is not
...
+configFrom:
+  secretKeyRef:
+    name: plugin-conf-secret
+    namespace: secret-ns
+    key: add-response-header
+plugin: response-transformer
+' | kubectl apply -f -
+```
+
+The output is similar to the following:
```
cloudjumpercat, line 291:
```suggestion
Example output:
```

app/_src/kubernetes-ingress-controller/guides/using-kongplugin-resource.md
```diff
@@ -6,6 +6,14 @@ This guide walks you through using the {{site.kic_product_name}}
 `KongPlugin` Custom Resource to control proxied requests, including
 restricting paths and transforming requests.
 
+Please see the [concept](/kubernetes-ingress-controller/{{page.kong_version}}/concepts/custom-resources/#KongPlugin)
```
cloudjumpercat, line 9:
```suggestion
See the [concept](/kubernetes-ingress-controller/{{page.kong_version}}/concepts/custom-resources/#KongPlugin)
```

app/_src/kubernetes-ingress-controller/guides/using-kongplugin-resource.md
```diff
@@ -6,6 +6,14 @@ This guide walks you through using the {{site.kic_product_name}}
...
+document for understanding the purpose of `KongPlugin` resource.
```
cloudjumpercat, line 10:
```suggestion
documentation for more information about the purpose of the `KongPlugin` resource.
```

app/_src/kubernetes-ingress-controller/guides/using-kongplugin-resource.md
```diff
@@ -6,6 +6,14 @@ This guide walks you through using the {{site.kic_product_name}}
...
+
+{:.important}
+> The controller does not validate configuration by default. Although not required for this guide, production instances
```
cloudjumpercat, line 13:
```suggestion
> The controller doesn't validate configuration by default. Although not required for this guide, production instances
```

app/_src/kubernetes-ingress-controller/guides/using-kongplugin-resource.md
```diff
@@ -274,6 +282,56 @@ Via: kong/2.8.1
 Here, we have successfully set up a plugin which is executed only when a
 request matches a specific `Ingress` rule.
 
+## Storing plugin configuration in a Secret
+
+You can store plugin configuration in a Secret to secure sensitive configuration. To do so, first create a Secret
```
cloudjumpercat, line 287:
```suggestion
You can store plugin configuration in a Secret to secure the sensitive configuration. To do so, first create a Secret
```

app/_src/kubernetes-ingress-controller/guides/using-kongplugin-resource.md
```diff
@@ -274,6 +282,56 @@ Via: kong/2.8.1
...
+with a key for each field you wish to configure:
```
cloudjumpercat, line 288:
```suggestion
with a key for each field you want to configure:
```

app/_src/kubernetes-ingress-controller/guides/using-kongplugin-resource.md
```diff
@@ -274,6 +282,56 @@ Via: kong/2.8.1
...
+stringData:
+  add-response-header: |
+    add:
+      headers:
+        - "demo: injected-by-kong"
+type: Opaque
+' | kubectl apply -f -
+```
+
+The output is similar to the following:
```
cloudjumpercat, line 305:
```suggestion
Example output:
```

app/_src/kubernetes-ingress-controller/guides/using-kongplugin-resource.md
```diff
@@ -274,6 +282,56 @@ Via: kong/2.8.1
...
+```
+secret/plugin-conf-secret created
+```
+
+Then, create a KongPlugin with a `configFrom` field referring that Secret:
```
cloudjumpercat, line 310:
```suggestion
Then, create a KongPlugin with a `configFrom` field referring to that Secret:
```

app/_src/kubernetes-ingress-controller/guides/using-kongplugin-resource.md
```diff
@@ -274,6 +282,56 @@ Via: kong/2.8.1
...
+  name: add-response-header
+configFrom:
+  secretKeyRef:
+    name: plugin-conf-secret
+    key: add-response-header
+plugin: response-transformer
+' | kubectl apply -f -
+```
+
+The output is similar to the following:
```
cloudjumpercat, line 326:
```suggestion
Example output:
```

cloudjumpercat (changes requested):
I left a few copyedits on the page, if you want to go over those. I also checked how everything renders, and everything looks good to me. 

I did have a question about the generated files. Typically, we limit the number of include files that we have and the ones that we do have are several paragraphs long (and used in multiple files). The include files render correctly in the output, but I just wasn't sure if they were necessary. My thoughts were that it would be better to leave the text in the `custom-resources-2.7.x` file instead of using include files, but I'm not sure how the files are being generated and if the include files are a necessary part of that process.

cloudjumpercat:
@czeslavo When you get the chance, could you change the base branch from `main` to `release/kic-2.8`?

rainest:
Applied suggestions and rebased to the 2.8 branch.

Consolidated my reusable guide snippets and the ref includes under a single `kic` directory.

Added a GWAPI KongIngres support note for https://github.com/Kong/kubernetes-ingress-controller/issues/3020.

Fixed some issues that the rebase didn't properly catch.

Moved the plugins warning into https://github.com/Kong/docs.konghq.com/pull/4894/commits/8e991f972a92e1dafe5c9d99f49c96e77c5a60f2 since that's going to touch a bunch of other content in there, to limit conflicts with this.

czeslavo:
> I did have a question about the generated files. Typically, we limit the number of include files that we have and the ones that we do have are several paragraphs long (and used in multiple files). The include files render correctly in the output, but I just wasn't sure if they were necessary. My thoughts were that it would be better to leave the text in the custom-resources-2.7.x file instead of using include files, but I'm not sure how the files are being generated and if the include files are a necessary part of that process.

The whole point of using includes in the generated reference was to make it possible to add some additional parts that do not exactly fit into the code comments (from which the raw reference is generated). It also gives better control over the formatting of such notes (e.g. using `{:.important}` etc. which are not valid for vanilla markdown).

To provide a wider context, that's how the process of generating the `custom-resources-2.7.x.md` looks like:
1. `make generate.docs` outputs [a "raw" markdown reference](https://github.com/Kong/kubernetes-ingress-controller/blob/main/docs/api-reference.md). In this step, the template also [leaves a "description placeholder"](https://github.com/Kong/kubernetes-ingress-controller/blob/edcc23a7c72bec103248e78e6670533e62129754/scripts/apidocs-gen/template/type.tpl#L12) for every CRD so that it can be used in the next stage.
2. To make it ready for the `docs.konghq.com` repo, we have [a script](https://github.com/Kong/kubernetes-ingress-controller/blob/main/scripts/apidocs-gen/post-process-for-konghq.sh) that post-processes this a little bit: adds things specific to the docs site - turns vale off, replaces description placeholders with `include_cached` directives, adds a title header, etc. 

The code which does all that lives under https://github.com/Kong/kubernetes-ingress-controller/tree/main/scripts/apidocs-gen.

app/_src/kubernetes-ingress-controller/references/custom-resources-2.7.x.md
```diff
@@ -0,0 +1,464 @@
...
+
+- [IngressClassParameters](#ingressclassparameters)
+
+### IngressClassParameters
+
+
+
+IngressClassParameters is the Schema for the IngressClassParameters API.
+
+{% include_cached md/kic/crd-ref/ingress_class_parameters_description.md kong_version=page.kong_version %}
```
cloudjumpercat, line 284: @czeslavo Looks like the include file here is empty. I'm not sure if it should have text in it or if we can just remove this include file and link.

cloudjumpercat (approved):
@czeslavo Thanks for explaining how the generated files works, that makes sense. I just left one comment because it looks like one of the include files is missing content, but other than that, everything LGTM!

app/_src/kubernetes-ingress-controller/references/custom-resources-2.7.x.md
czeslavo: It's added here just because we do generate these hooks for every CRD unconditionally, and in the case of `IngressClassParameters` we did not have any note to add, therefore the file is empty. If we remove the file, the include will fail. It'd be cool if we had a `include_cached_if_exists` for such scenarios.

Will it be ok if we add a comment to the file that it's a placeholder and we intentionally leave it empty for now? 
 34a93c8

PR #4846, opened 2022-12-01
[DOCU-2757] CP/DP forward proxy config

lena-larionova:
### Summary
Doc for setting up a forward proxy connection between CP and DP.

I can't really test this, so I can't verify that this is correct. Need some SME help.

Outstanding questions @dndx:
* From design doc: "Customers need to ensure their proxy does not MITM or mutate the certificate used by Kong DP when connecting to CP" - How do you verify this? 
* What special info do we need for Konnect here?
* From design doc: "We should document roughly how many connections each Kong DP will open to the CP, so that user knows how much proxy servers are needed to support a given number of Kong deployments." - how would you estimate this number?

### Reason
https://konghq.atlassian.net/browse/DOCU-2757

### Testing
https://deploy-preview-4846--kongdocs.netlify.app/gateway/latest/production/networking/cp-dp-proxy/

### DIFF
--- a/src/gateway/production/networking/cp-dp-proxy.md
+++ b/src/gateway/production/networking/cp-dp-proxy.md
@@ -32,7 +32,7 @@ only use this option if any component is explicitly configured to use the proxy.
 `proxy_server` is in HTTPS. Set to `on` if using HTTPS (default), or `off` if
 using HTTP.
 
-* `cluster_use_proxy`: Tell the cluster to use HTTP CONNECT proxy support for
+* `cluster_use_proxy`: Tells the cluster to use HTTP CONNECT proxy support for
 hybrid mode connections. If turned on, {{site.base_gateway}} will use the
 URL defined in `proxy_server` to connect.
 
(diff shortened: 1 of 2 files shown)


src/gateway/production/networking/cp-dp-proxy.md
```diff
@@ -0,0 +1,48 @@
...
+```
+
+* `proxy_server`: Proxy server defined as a URL. {{site.base_gateway}} will
+only use this option if any component is explicitly configured to use the proxy.
+
+* `proxy_server_tls_verify`: Toggles server certificate verification if
+`proxy_server` is in HTTPS. Set to `on` if using HTTPS (default), or `off` if
+using HTTP.
+
+* `cluster_use_proxy`: Tell the cluster to use HTTP CONNECT proxy support for
```
cloudjumpercat, line 35:
```suggestion
* `cluster_use_proxy`: Tells the cluster to use HTTP CONNECT proxy support for
```

cloudjumpercat (approved):
Just one small suggestion, otherwise LGTM!

lena-larionova:
Merging without SME review for now.

rosskukulinski:
@lena-larionova are there any plans to have similar content in the Konnect documention?

lena-larionova:
> @lena-larionova are there any plans to have similar content in the Konnect documention?

There are, but I don't currently have enough info to write the Konnect piece. One of the outstanding questions is still "What special info do we need for Konnect here?"

PR #4848, opened 2022-12-01
DOCU-2818 changelog 2.8.2.2

acgoldsmith:
### Summary
update changelog to reflect the fix the timer error in Core was not part of 2.8.2.1 and instead is releasing as this patch. 

### Reason
DOCU-2818


cloudjumpercat (approved):
LGTM!

PR #4851, opened 2022-12-01
docs(fips) add new fips distros - FT-3310

gszr:
### Summary

This is an attempt at better structuring the FIPS docs; instead of replicating install instructions over at the FIPS docs page, point to the Install page, where FIPS-specific instructions are added.

### Reason

FT-3310.

PR #4853, opened 2022-12-01
FIPS support for ubuntu 22.04

lena-larionova:
### Summary
Adding ubuntu 22.04 to list of supported packages for fips. 

The original PR aims to better structure the install docs to include FIPS packages as an option on the install pages. Not going with this approach as we need to keep FIPS separate for now for legal reasons. (fyi @gszr)

### Reason
Based on https://github.com/Kong/docs.konghq.com/pull/4851/files, which was working with an older version of the doc.

### Testing
Netlify


acgoldsmith (approved):
LGTM

PR #4868, opened 2022-12-05
docs(kic) use support table format in gateway api support page

randmonkey:
### Summary

Use support table format in gateway API support page to provide a clearer view of what gateway API features are supported in which versions of KIC.

### Reason

fixes https://github.com/Kong/kubernetes-ingress-controller/issues/3173

### Testing

https://deploy-preview-4868--kongdocs.netlify.app/kubernetes-ingress-controller/latest/references/gateway-api-support/

### DIFF
--- a/src/kubernetes-ingress-controller/references/gateway-api-support.md
+++ b/src/kubernetes-ingress-controller/references/gateway-api-support.md
@@ -84,6 +84,8 @@ The {{site.kic_product_name}}'s implementation of `UDPRoute` supports multiple `
 |:--------------------------|:---------------------------:|:---------------------------:|:---------------------------:|:---------------------------:|
 | `v1alpha2`                | <i class="fa fa-check"></i> | <i class="fa fa-check"></i> | <i class="fa fa-check"></i> | <i class="fa fa-check"></i> |
 
+### Unsupported
+- Does not support [GEP-957](https://gateway-api.sigs.k8s.io/geps/gep-957/) port matching.
 
 ## TLS Routes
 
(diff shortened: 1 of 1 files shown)


rainest (changes requested):
We probably want to avoid a matrix, at least until we have a docs feature that allows us to provide the matrix as a CSV and have it automatically render into a table appropriate for a given version. We already have issues with the K8S/KIC/Kong matrices having no option other than chopping off old versions, since adding `if_version` tags for columns is a nightmare for manual edits.

We're also generally not removing support for features, so most of the matrices would just be all red until they're all green forever. API versions are a weird case, but I think we can simply omit those and point to the [upstream policy](https://gateway-api.sigs.k8s.io/concepts/versioning/#supported-versions). We don't actually remove support for the older versions, upstream does. The HTTPRoute CRD still supports both v1alpha2 and v1beta1, so you can create a v1alpha2 HTTPRoute and use it on 2.7, since the API server will convert it when we request v1beta1.

A changelog per API resource should then cover what we are responsible for, i.e. adding functionality over time and unsupported core features, so

```
## HTTPRoute

### 2.2

- Added basic support.

### 2.4

- Added support for multiple, weighted entries under `backendRefs`.
- Added support for the `RegularExpression` header match type.

### Not supported

- `queryParam` in HTTPRoute matches.
- `requestRedirect` in HTTPRoute filters.
```

(I didn't actually check the versions in the above, they're just made up to show formatting).

czeslavo:
We've got another PR modifying this doc https://github.com/Kong/docs.konghq.com/pull/4889. I allowed myself to commit 7f0cfd5 to avoid losing it if we decide to go with the table view.

cloudjumpercat:
@randmonkey When you get the chance, could you change the base branch from `main` to `release/kic-2.8`?

rainest:
This was superseded by https://github.com/Kong/docs.konghq.com/pull/4869

PR #4884, opened 2022-12-07
Fix plugins regeneration

fabianrbz:
### Summary
[Related Jira ticket](https://konghq.atlassian.net/browse/DOCU-2828)
Fixes plugin pages re-generation so that they are re-built whenever the corresponding source file changes or its layout/ includes does.

This also removes some conditionals and calculations done in the templates and does them in ruby.
Fixes `Edit this page` link on plugin pages that have a source file for specific versions, e.g. [this one](https://deploy-preview-4884--kongdocs.netlify.app//hub/kong-inc/jwt-signer/2.2.x.html)

### Reason
Plugin pages were being re-generated on every file change, even if the change wasn't related.

### Testing
* Modify one of the plugin source files and check that the corresponding page is re-built
* Modify the layout and check that all the plugin pages are re-built
* Modify any file that is not related and check that the plugins are not re-built 

https://user-images.githubusercontent.com/715229/206397588-ee45ffb5-33ed-4973-9ae2-53b17d4c77e6.mov

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -29,6 +29,7 @@ build: ruby-version-check
 clean:
 	-rm -rf dist
 	-rm -rf app/.jekyll-cache
+	-rm app/.jekyll-metadata
 
 # Runs tests
 test:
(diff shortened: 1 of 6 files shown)


mheap (approved):
Do we have a test case where source_file is not _index.md? I couldn't see one in the diff

fabianrbz:
good call, I'll add one

PR #4892, opened 2022-12-08
Plugins bundled warning

mheap:
### Summary
Add warning for non-bundled plugins

### Reason
Help customers enable shipped, but not enabled plugins

### Testing
Check App Dynamics on review app


acgoldsmith (approved):
LGTM and I like the accordion info block!

PR #4924, opened 2022-12-14
Makes the PDK file structure slightly more consistent

rspurgeon:
### Summary

The table at the bottom of the page lists filenames the plugin developer should use And the descriptions below go into more detail on each file The table and the descriptions are inconsistent regarding the API interface filename

I looked at existing plugins, and there seems to be a standard to use api.lua and not admin-api.lua despite that being the name of the referenced autodoc file.

I suggest we make the filename in the description match the filename in the table so the developer knows what to name their file.

### Reason
To make the PDK more consistent for the reader

### Testing
Built the docs locally and viewed the result manually

### DIFF
--- a/app/_src/gateway/plugin-development/file-structure.md
+++ b/app/_src/gateway/plugin-development/file-structure.md
@@ -97,7 +97,7 @@ master each one of them.
 
 | Module name            | Required   | Description
 |:-----------------------|------------|------------
-| [admin-api.lua](https://github.com/Kong/kong/blob/master/autodoc/admin-api/data/admin-api.lua)          | No         | Defines a list of endpoints to be available in the Admin API to interact with the custom entities handled by your plugin.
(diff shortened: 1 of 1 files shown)


Guaris:
@rspurgeon Sorry I haven't had time to review this yet, do you happen to know if the name change would affect any other docs?

rspurgeon:
> @rspurgeon Hi! Just bumping Angel's comment from before break:
> 
> > @rspurgeon Sorry I haven't had time to review this yet, do you happen to know if the name change would affect any other docs?

No, I'm not sure how the link name would affect other docs.

rspurgeon:
> https://deploy-preview-4924--kongdocs.netlify.app/gateway/latest/plugin-development/entities-cache/#automatic-cache-invalidation

@cloudjumpercat I might be missing something, but this PR and the mention of `admin_api` in those documents do not seem related.

cloudjumpercat:
> > https://deploy-preview-4924--kongdocs.netlify.app/gateway/latest/plugin-development/entities-cache/#automatic-cache-invalidation
> 
> @cloudjumpercat I might be missing something, but this PR and the mention of `admin_api` in those documents do not seem related.

Sounds good! I wasn't sure if they were related or not.

cloudjumpercat (approved):

PR #4927, opened 2022-12-15
Increase test coverage

fabianrbz:
### Summary
Increase test coverage of versions generator and remove unused code. 
This also updates plugins to use relative paths instead of full_paths for re-generation, in the same way, we do for pages.

### Reason
Generators are a crucial part of the system, I wanted to add more coverage in order to make the code mode maintainable/easier to refactor.

### Findings

* We don't set any `aliases` in the versions generator. We had a [method](https://github.com/Kong/docs.konghq.com/blob/main/app/_plugins/generators/versions.rb#L186-L199) for this, but the actual part that set the `alias` to the page was never invoked. Due to how it was implemented, it consistently executed [this](https://github.com/Kong/docs.konghq.com/blob/main/app/_plugins/generators/versions.rb#L193) branch of the code. So all the `aliases` are actually handled by the `latest_version_generator`.  Should we revisit this? Are we missing any?
* We currently set `has_version` to **every** [page](https://github.com/Kong/docs.konghq.com/blob/main/app/_plugins/generators/versions.rb#L84), even if they don't have one. This leads to an error in which the sidebar is rendered for every page, even for the ones that shouldn't display it, e.g. [/enterprise/reference](https://docs.konghq.com/enterprise/references/)
* Pages that don't have a version in their path, end up having either a folder or file name set as the `kong_version`. e.g. [references](https://github.com/Kong/docs.konghq.com/blob/increase-test-coverage/spec/app/_plugins/generators/versions_spec.rb#L118) and [changelog](https://github.com/Kong/docs.konghq.com/blob/increase-test-coverage/spec/app/_plugins/generators/versions_spec.rb#L189). 
This causes a few errors: 
  * `kong_version` is used to populate the [version dropdown ](https://docs.konghq.com/enterprise/references/) on the sidebar (tied to the issue above, the sidebar shouldn't be displayed in the first place)
    Screenshot 2022-12-15 at 10 37 15
  * `kong_version` is also used to generate the [page title](https://github.com/Kong/docs.konghq.com/blob/main/app/_includes/head.html#L3) and `og:title`
     Screenshot 2022-12-15 at 10 39 59
 I definitely think we should not set it for files that don't have a version in their path (single-sourced pages work because the `path` attribute does have the corresponding version set)
* We currently don't set `kong_latest` to pages under `enterprise`, `getting-started-guide`, `konnect`, `contributing`, and `gateway-oss`. This might not necessarily be an issue though. However,
     * We do use `kong_latest` in `archive/gateway-oss` pages, but they can't be accessed. The  `archive` option in the version dropdown points to https://legacy-gateway--kongdocs.netlify.app/ which returns an error
        Screenshot 2022-12-15 at 10 57 30
    * Same for `archive/enterprise`


mheap:
> We don't set any `aliases` in the versions generator. 

This code predates single sourcing + the SEO generator. If you're confident, let's remove it

> We currently set `has_version` to **every** [page](https://github.com/Kong/docs.konghq.com/blob/main/app/_plugins/generators/versions.rb#L84), even if they don't have one. 

This is a bug. It should be `page.data['has_version'] = has_version` after line 99

> Pages that don't have a version in their path, end up having either a folder or file name set as the `kong_version`. e.g. [references](https://github.com/Kong/docs.konghq.com/blob/increase-test-coverage/spec/app/_plugins/generators/versions_spec.rb#L118) and [changelog](https://github.com/Kong/docs.konghq.com/blob/increase-test-coverage/spec/app/_plugins/generators/versions_spec.rb#L189). 

`kong_version` should be `nil` in these instances. We can check if it matches the semver regex and set to `nil` if not

> We currently don't set `kong_latest` to pages under `enterprise`, `getting-started-guide`, `konnect`, `contributing`, and `gateway-oss`. This might not necessarily be an issue though

This is fine. `enterprise`, `gateway-oss` and `getting-started-guide` can never be the latest version (`gateway` is the combined docset which will always contain the latest version). `konnect` and `contributing` aren't versioned, so they're not expected to have a `kong_latest` entry

> The  `archive` option in the version dropdown points to https://legacy-gateway--kongdocs.netlify.app/ which returns an error

This is related to the site move yesterday. The legacy branch has now been redeployed

fabianrbz:
> This code predates single sourcing + the SEO generator. If you're confident, let's remove it
👍 I added some print statements here and built the site a few times, but nothing happened.

I've updated the code to reflect what we discussed. This ended up fixing the sidebar and breadcrumbs for `deck-pre-1.7`

Production
prod

After the fix
fix-versions

mheap (approved):

PR #4951, opened 2022-12-20
Minor grammar update

geocomm-shenningsgard:
"Following figure shows how this deployment looks like:" could be changed to either "...how this deployment looks:" or "...what this deployment looks like". I figured the latter fit better with the tone of the rest of the documentation, but either would be correct.


acgoldsmith (approved):
Thanks, @geocomm-shenningsgard!

PR #4954, opened 2022-12-21
plugin(RLA): custom response

StarlightIbuki:
reflect the change of https://github.com/Kong/kong-ee/pull/3926


lena-larionova (approved):
Thanks for catching this - I know we documented it for the RL community plugin, but we weren't aware of the change also being applicable to RLA.

I added a changelog entry to mirror the RL plugin as well. LGTM!

PR #4973, opened 2022-12-23
Fixed extra http in API call

felderi:
Addressed extra http in API call


Guaris (approved):
lgtm
https://deploy-preview-4973--kongdocs.netlify.app/konnect/api/runtime-groups-config/#supported-content-types

PR #4984, opened 2023-01-03
Minor fixes to konnect deck onboarding

ryanmoore:
### Summary

This PR contains very minor fixes to wording and output that differed from what I saw when following the onboarding guide for Konnect and using declarative config with Konnect.

Namely:

- there is no "Add runtime instance" button. It's "New Runtime Instance". I expect this was a change in Konnect at some point.
- The route is set up at `/mock` not `/mockpath` (this is correctly identified [later in the doc](https://github.com/Kong/docs.konghq.com/blob/634e153fa9111543a407f148ba74eb348ea6c46e/app/konnect/runtime-manager/runtime-groups/declarative-config.md?plain=1))
- deck makes 3 objects - the sample output is missing the keyauth plugin

### Reason

It prevents confusion when the docs are consistent with the experience.

### Testing

I was using decK v1.17.0.

### DIFF
--- a/app/konnect/getting-started/configure-runtime.md
+++ b/app/konnect/getting-started/configure-runtime.md
@@ -40,7 +40,7 @@ is part of the organization admin team by default.
     Every account starts with one group named `default`. If you have an
     Enterprise subscription, you can create additional custom groups.
 
-1. Click **Add runtime instance**.
+1. Click **New Runtime Instance**.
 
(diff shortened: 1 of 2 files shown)


lena-larionova (approved):
Thanks for catching this - the key auth plugin bit was added later but looks like we missed updating some of the examples. LGTM!

PR #4985, opened 2023-01-03
Split changelog from one file into multiple

lena-larionova:
### Summary
Splitting the Gateway changelog from a monolithic file into multiple source files.

Three approaches here, depending on the status of the versions they went into:
* **0.x and 1.x:** moved into archive folder.
* **2.x:** Since the 2.x content is not single-sourced, moved the changelog into separate version files in each 2.x.x version folder. I considered moving them into one 2.x file and building each one conditionally, but then we would still have the issue of the missing left nav.
* **3.x:** Moved both minor versions into a changelog-3.x.md folder and used `{% if version %}` tags to build the content conditionally for each minor version. 

Very much open for feedback on better approaches here.

Also considered using includes for each minor version, but based on team feedback, this would make the user experience for doc editors too complicated.

### Reason

Aiming to resolve the following issues:
* No [left-side versioned navigation in changelog](https://docs.konghq.com/gateway/changelog/), so no way to orient the user in the doc. This is the only doc we have that does this, so it doesn't follow a pattern that users are accustomed to.
* Monolithic file is really hard to edit, especially if you need to work with multiple releases
* Monolithic file is equally hard to read, as page load time and performance is poor
* Old, archived versions were still documented (0.x and 1.x), although the docs for them have been removed from the live site

### Testing

Check the versions, here are a few:
https://deploy-preview-4985--kongdocs.netlify.app/gateway/latest/changelog/
https://deploy-preview-4985--kongdocs.netlify.app/gateway/3.0.x/changelog/
https://deploy-preview-4985--kongdocs.netlify.app/gateway/2.8.x/changelog/
https://deploy-preview-4985--kongdocs.netlify.app/enterprise/2.5.x/changelog/

https://deploy-preview-4985--kongdocs.netlify.app/gateway/changelog/ should redirect to https://deploy-preview-4985--kongdocs.netlify.app/gateway/latest/changelog/


cloudjumpercat (approved):
This all looks good to me! Thanks for doing this!

mheap:
> In the current implementation, customers have the ability to ctrl+f to find an issue they were looking for specifically

This is a big use case. Could we have one changelog per major version, and the changelog link in the sidebar takes you to the correct heading?

lena-larionova:
> This is a big use case. Could we have one changelog per major version, and the changelog link in the sidebar takes you to the correct heading?

Yep, that makes sense. Diana brought this idea up as well, so it's unanimous at this point.

The only concern I have is the technical implementation of that idea. 
* For the single-sourced 3.0 version onward, this is easy - I'll just remove the `{% if_version %}` tags.
* For 2.8.x and earlier, this becomes difficult, unless I use includes - because then we need a changelog.md with the same content in every 2.x version. 

_Or_ we could get away from this entirely and have a changelog/release notes "guide", as it were. This is actually the way that most doc sites handle it. We would then link to that changelog from our main gateway/mesh/etc navs.

It would look something like this: https://docs.apigee.com/release/notes/apigee-release-notes - that is, the left side nav would only include release notes, categorized by product.

lena-larionova:
@Guaris @mheap Here's what I meant by "guide" - similar to our contributing doc. We can split out any release notes or changelogs we maintain like this:

Screenshot 2023-01-06 at 1 38 44 PM

lena-larionova:
Closing, will revisit strategy post-3.2.

PR #4994, opened 2023-01-05
Updated ingress declaration examples to match

geocomm-shenningsgard:
### Summary

The attributes defined for apiVersion: networking.k8s.io/v1 did not match the one defined for apiVersion: networking.k8s.io/v1beta1. I did my best as a kubernetes novice, but please have someone internal review my changes and make any others necessary.

### Reason

We're updating from k8s 1.21 to 1.22 and need these instructions to know exactly what to do wrt kong

### DIFF
--- a/app/_src/kubernetes-ingress-controller/concepts/ingress-versions.md
+++ b/app/_src/kubernetes-ingress-controller/concepts/ingress-versions.md
@@ -121,13 +121,14 @@ metadata:
 spec:
   ingressClassName: kong
   rules:
-  - http:
+  - host: example.com
+    http:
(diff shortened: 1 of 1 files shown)


cloudjumpercat:
@Kong/team-k8s Hi! Could someone review the changes in this PR? Thanks!

rainest (approved):
:+1: nothin major, just makes the two examples use the same values, which they should.

cloudjumpercat:
Thanks @rainest !

cloudjumpercat (approved):
LGTM!

PR #5002, opened 2023-01-06
Add tags to key-auth-enc plugin; update examples

lena-larionova:
Follow-up to https://github.com/Kong/docs.konghq.com/pull/4941. 

Adding `tags` attribute to the `key-auth-enc` plugin as well as `key-auth`. Note that key-auth-enc doesn't yet support `ttl`; see FTI-4066 for detail.

Also updating the config examples with `tags` and `key` values, and with `ttl` for the Key Auth plugin.

### DIFF
--- a/.github/styles/kongplugins/plugin-ignore.txt
+++ b/.github/styles/kongplugins/plugin-ignore.txt
@@ -392,7 +392,9 @@ entityname
 error_code
 error_message
 event_queue_size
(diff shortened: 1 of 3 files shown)


lena-larionova:
@Guaris why does Vale not ignore content inside codeblocks in plugins? The errors it flagged aren't in the YAML section of the doc.

PR #4913, opened 2022-12-13
docs(redis) redis.username cannot be 'default'

outsinre:
### Summary

`redis.username` cannot be *default*. The Doc may confuse customers.

### Reason

Required by Redis 6+.

Fix _[FTI-4151]_

[FTI-4151]: https://konghq.atlassian.net/browse/FTI-4151?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

PR #5003, opened 2023-01-06
Clarify redis.username on all plugins that support it

lena-larionova:
Follow-up to https://github.com/Kong/docs.konghq.com/pull/4913. 
The PR only edited the rate limiting plugins, but according to the linked FTI, the username `default` is not supported in Redis 6.0.0+ in general - this is not specific to Kong, so it should apply anywhere that we allow `redis.username`.

@outsinre can you please verify that this change is correct for all of these plugins?


outsinre:
LGTM.

outsinre (approved):

Guaris (approved):

PR #5007, opened 2023-01-09
Version support pages show a single product

mheap:
### Summary
The support matrix for all products was shown within each product. 

### Reason
It doesn't make sense to show Mesh version support when looking at Gateway or KIC (and vice versa)

### Testing

- /gateway/latest/support-policy/
- /mesh/latest/support-policy/
- /kubernetes-ingress-controller/latest/support-policy/


cloudjumpercat (approved):
LGTM!

PR #5011, opened 2023-01-09
Rate Limiting Advanced plugin: Improve config.namespace description

lena-larionova:
### Summary
Improve description to include DB-less and declarative config info. Also add the value to the autogenerated config examples, since we recommend setting it explicitly for a large portion of users.

### Reason
[Issue raised on Slack](https://kongstrong.slack.com/archives/CDSTDSG9J/p1673022359955879).

Explanation of behavior from FTI ticket:
> As you can see from the discussion on this related PR, we are going to not fix this on purpose: this is currently failing because the namespace field is an auto field, meaning that if the user doesn't provide any input, Kong autogenerates it. Since deck is a tool for declarative configuration, we believe this is a good place for deck to fail since not having a namespace and autogenerate it here would mean that decK is not pushing configuration declaratively: a side-effect of this would be that consecutive deployments with a configuration file not including any namespace would result in misleading diffs, since Kong will keep regenerating a namespace each time.

### Testing
https://deploy-preview-5011--kongdocs.netlify.app/hub/kong-inc/rate-limiting-advanced/

### DIFF
--- a/.github/styles/kongplugins/plugin-ignore.txt
+++ b/.github/styles/kongplugins/plugin-ignore.txt
@@ -393,6 +393,7 @@ error_code
 error_message
 event_queue_size
 example_apikey
+example_namespace
 example_pass
 example_tag
 example_user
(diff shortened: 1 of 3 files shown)


cloudjumpercat (approved):
LGTM!

PR #5012, opened 2023-01-09
[DOCU-2880] Plugin konnect examples

lena-larionova:
### Summary
Searched through the doc for plugins that have `cloud: false` but not `konnect_examples: false` and updated them. The plugins noted in the ticket were the only ones.

Also found that the TLS plugins were supposed to have been checked in Konnect in 3.0 and have the examples enabled if they exist. I checked that, they are there, so I enabled examples.

### Reason
Some plugins had Konnect examples even though they're not available in Konnect.

### Testing
Netlify. 
Konnect examples for the graphql plugins should be gone, and the TLS plugins should have Konnect examples.


cloudjumpercat (approved):
LGTM!

PR #5016, opened 2023-01-10
[DOCU-2373] Key auth encrypted plugin: Re-add ttl when parameter is available

Guaris:
https://konghq.atlassian.net/jira/software/c/projects/DOCU/boards/58?modal=detail&selectedIssue=DOCU-2373

Does this parameter need to be added to the top config parameter section?


lena-larionova:
> Does this parameter need to be added to the top config parameter section?

No, not with the way this doc is structured now. It's not a plugin parameter, exactly.

PR #5017, opened 2023-01-10
Add note about seeding super-admin password

d19dotca:
### Summary
Added a quick note to encourage Enterprise users to skip to the Seeding Super-Admin password section instead of running the initial ```kong migrations``` command without it.

### Reason
A customer highlighted the concern that if following the documentation in-order as currently defined then it's too late to set the Super-Admin password by the time that section is read. To help users avoid this trap, I felt it was best to highlight the area before a user runs the initial ```kong migrations``` command.

### DIFF
--- a/app/_includes/md/gateway/setup.md
+++ b/app/_includes/md/gateway/setup.md
@@ -45,6 +45,8 @@ The following instructions use [PostgreSQL](http://www.postgresql.org/) as a dat
     ```
 
 2. Run the {{site.base_gateway}} migrations using the following command:
+{:.note}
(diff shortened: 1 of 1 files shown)


cloudjumpercat (commented):
@d19dotca Hi! Thanks for making this change! I removed the Seed Super Admin section and instead added the info from that section in the "Using a database" migration step instead. I did that because following the link to the super admin section could distract users from what they were doing and it would be better to have all the info in the migration command steps themselves. What do you think? Would this work?

d19dotca:
Hi @cloudjumpercat - I think that works much better, great idea! :-)

cloudjumpercat (approved):
LGTM! Thanks for suggesting this update!

PR #5029, opened 2023-01-11
Simplify PR template

lena-larionova:
Attempting to make the PR template clearer.

* Combined the summary and reason sections into one
* Added changelog prompt
* Added checklist - is this useful at all? If we want to use it, what else should be part of this checklist?
* Left the info about labels in. It's a big block block of text that I don't think anyone reads, but the info needs to be there to cover why the builds fail. 
  * Or does it? Should it be a bot comment instead, if someone doesn't add a label?

All feedback and ideas welcome.

### DIFF
--- a/.github/PULL_REQUEST_TEMPLATE.md
+++ b/.github/PULL_REQUEST_TEMPLATE.md
@@ -1,11 +1,9 @@
 
-### Summary 
+### Description
 
-Description: <!-- What did you change and why? -->
+What did you change and why?
  
-<!-- Include any supporting resources, e.g. link to a Jira ticket, GH issue, FTI, Slack, Aha, etc. -->
-
-Changelog entry, if applicable:
+Include any supporting resources, e.g. link to a Jira ticket, GH issue, FTI, Slack, Aha, etc.
 
 
 ### Testing instructions
@@ -16,7 +14,9 @@ Netlify link: <!-- Netlify will generate a preview link after PR is opened. Add 
 ### Checklist 
 
 [ ] Review label added
-[ ] Changelog entry (if applicable)
+
+** Release-specific work only **
+[ ] Pointed to correct feature branch (e.g. `release/gateway-3.2`, `release/deck-1.17`)
 
 
 <!-- !!! Only Kong employees can add labels due to a GitHub limitation. If you're an OSS contributor, thank you! The maintainers will label this PR for you !!! -->
@@ -25,7 +25,8 @@ Netlify link: <!-- Netlify will generate a preview link after PR is opened. Add 
 
     review:copyedit: Request for writer review.
     review:general: Review for general accuracy and presentation. Does the doc work? Does it output correctly?
-    review:tech: Request for technical review for a platform change.
+    review:tech: Request for technical review for a docs platform change.
+    review:sme: Request for review from an SME (engineer, PM, etc).
 
 At least one of these labels must be applied to a PR or the build will fail.
 -->
--- a/.github/PULL_REQUEST_TEMPLATE.md
+++ b/.github/PULL_REQUEST_TEMPLATE.md
@@ -1,25 +1,32 @@
-### Summary
-<!-- Description of PR, with any special instructions for your reviewers. -->
 
-### Reason
-<!-- Why are you making this change? Can be a link to a Jira ticket, GH issue, 
-Trello card, etc. -->
+### Summary 
 
-### Testing
-<!-- How can your reviewers test your change? How did you test it? -->
+Description: <!-- What did you change and why? -->
+ 
+<!-- Include any supporting resources, e.g. link to a Jira ticket, GH issue, FTI, Slack, Aha, etc. -->
 
-<!-- (Optional) Include link to topic in Netlify preview after it's generated 
-(around 10mins after PR is created) -->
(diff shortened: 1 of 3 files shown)


mheap:
> Or does it? Should it be a bot comment instead, if someone doesn't add a label?

The label action _just_ got the [ability to add custom comments](https://github.com/mheap/github-action-required-labels#customising-the-failure-message--comment)!

.github/PULL_REQUEST_TEMPLATE.md
```diff
@@ -1,25 +1,32 @@
...
+
+
+<!-- !!! Only Kong employees can add labels due to a GitHub limitation. If you're an OSS contributor, thank you! The maintainers will label this PR for you !!! -->
+
+<!-- When raising a pull request, indicate what type of review you need with one of the following labels:
 
     review:copyedit: Request for writer review.
     review:general: Review for general accuracy and presentation. Does the doc work? Does it output correctly?
-    review:tech: Request for technical review from an SME.
+    review:tech: Request for technical review for a platform change.
```
mheap, line 28: Is this missing `review:sme`?

Guaris:
If the change log entry is in the PR it will conflict with other PR's that add a change log entry like Amy's monthly changelog PR. Not that we should discourage people from adding change log entries but I think we can omit that checkbox. I think the goal is for change log entries to be in the repo for the project's repo and Aha. 

I would like to see a way to ask people if they are pointing to the correct branch and encourage them to ask for help if they don't know. That way if someone decides to follow the template the docs team won't merge something to `main` that is supposed to be for an upcoming release. 

here's a rough example of what I mean: 
** Feature work only **
[ ] Pointed to correct feature branch. (ex: `release/gateway-3.2`, `deck-1.1.2`)

Guaris:
What if we didn't hide this 
```
 <!-- What did you change and why? -->

<!-- Include any supporting resources, e.g. link to a Jira ticket, GH issue, FTI, Slack, Aha, etc. -->
```
If someone doesn't notice it and doesn't add information when they submit a PR the text will appear and serve as a second notification to please fill that out.

.github/PULL_REQUEST_TEMPLATE.md
```diff
@@ -1,25 +1,32 @@
...
 
-### Testing
-<!-- How can your reviewers test your change? How did you test it? -->
+Description: <!-- What did you change and why? -->
+ 
+<!-- Include any supporting resources, e.g. link to a Jira ticket, GH issue, FTI, Slack, Aha, etc. -->
 
-<!-- (Optional) Include link to topic in Netlify preview after it's generated 
-(around 10mins after PR is created) -->
+Changelog entry, if applicable:
```
cloudjumpercat, line 8: Would this be a changelog of their changes in the PR? Or the changelog for the feature?

.github/PULL_REQUEST_TEMPLATE.md
```diff
@@ -1,25 +1,32 @@
...
-!!! Only Kong employees can add labels due to a GitHub limitation. If you're an OSS contributor, thank you! The maintainers will label this PR for you !!!
+### Testing instructions
 
-When raising a pull request, it's useful to indicate what type of review you're looking for from the team. To help with this, we've added three labels that can be applied:
+Netlify link: <!-- Netlify will generate a preview link after PR is opened. Add links to your edited content here. -->
+
+
+### Checklist 
+
+[ ] Review label added
```
cloudjumpercat, line 18:
```suggestion
[ ] Review label added (see below)
```
Maybe? Just so they know which ones they have to add to get the tests to pass?

.github/PULL_REQUEST_TEMPLATE.md
lena-larionova: I _think_ we had said that `review:sme` was a docs-team only label. But I don't remember the details.

lena-larionova:
> If the change log entry is in the PR it will conflict with other PR's that add a change log entry like Amy's monthly changelog PR. Not that we should discourage people from adding change log entries but I think we can omit that checkbox. I think the goal is for change log entries to be in the repo for the project's repo and Aha.

I agree with that, but I'd also love for someone submitting this to add a changelog entry to the description. They don't need to add it to a changelog file. That way, if we have to compile a changelog after the fact, those descriptions are easy to track down.

However, I can also see the argument for saying that the PR summary itself is our changelog entry. 

> I would like to see a way to ask people if they are pointing to the correct branch and encourage them to ask for help if they don't know. That way if someone decides to follow the template the docs team won't merge something to main that is supposed to be for an upcoming release.

> here's a rough example of what I mean:
** Feature work only **
[ ] Pointed to correct feature branch. (ex: release/gateway-3.2, deck-1.1.2)

Love the idea, let's do that.

.github/PULL_REQUEST_TEMPLATE.md
lena-larionova: Changelog for the feature. This seems to be confusing overall, so maybe I'll leave out anything about changelogs for now.

Guaris:
Good point, I think I'd rather deal with the merge conflict potential and have an abundance of changelog entries than the opposite problem.

Guaris (approved):

PR #5037, opened 2023-01-12
Update kuma submodule with `make install`

lena-larionova:
### Description

Adding `git submodule init && git submodule update` to the `make install` command. People (myself included) keep running into issues where the local submodule ends up out of date and the build fails.

### Testing instructions

Test locally by running `make install`. 
You can temporarily delete the `_src/.repos/kuma` folder first before running it to make sure that the command does pull the submodule.

### Checklist 

- [x] Review label added 
- [x] Pointed to correct feature branch

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -15,7 +15,7 @@ install-prerequisites:
 install: ruby-version-check
 	npm ci
 	bundle install
-	git submodule init && git submodule update
+	git submodule update --init
 
 # Using local dependencies, starts a doc site instance on http://localhost:3000.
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -15,6 +15,7 @@ install-prerequisites:
 install: ruby-version-check
 	npm ci
 	bundle install
+	git submodule init && git submodule update
```
mheap, line 18:
```suggestion
	git submodule update --init
```

mheap (approved):

PR #5040, opened 2023-01-13
Refactor generators

fabianrbz:
### Description

Refactor Latest and CanonicalUrl generators.
Added tests and made the code more maintainable/easier to understand.
There are a few improvements that I would like to implement in the future, but it's good enough for :shipit: 

The site should render/perform in the same way, without any noticeable changes. I verified that the sitemap is the same before and after the changes. 

Here's the full list of urls from both sitemaps
* [Original sitemap](https://gist.github.com/fabianrbz/48cf9102767ce9c75b5ee3484c7c6da0)
* [After the refactor ](https://gist.github.com/fabianrbz/c4693da26f297f5ab8a00325f03986e7)

Bear in mind that the order of the urls are not the same.

### Note about the plugin Hub

I removed the Jekyll collection for the Plugin Hub, it wasn't being used and added extra complexity.
It was actually the only reason why the index page of the hub was being generated given that it is under `app/_hub` which is ignored by Jekyll because of the leading `_`.
Because of this, I moved it to `app/hub/`, that way it will be picked up by Jekyll, not as part of a  collection, and will not be treated as a single-sourced page either.

### Testing instructions

[Netlify link](https://deploy-preview-5040--kongdocs.netlify.app/sitemap.xml): 

### Checklist 

- [x] Review label added 

** Release-specific work only **
- [ ] Pointed to correct feature branch (e.g. `release/gateway-3.2`, `release/deck-1.17`)


app/_plugins/generators/latest_version_generator.rb
```diff
@@ -10,54 +10,64 @@ class Page
...
+      product_name, product = PRODUCTS_WITH_LATEST.detect do |k, _|
+        k == product_name_from_path
+      end
+
+      product_name &&
+        release_path == @site.data["kong_latest_#{product}"]['release']
+    end
+
+    def page_index
+      # XXX: this isn't being set on any of the pages
```
mheap, line 61: This is a legacy thing. We used to set `version-index` in `jekyll.yml` but I replaced it with `page.versions.ce` and `page.versions.ee`

lena-larionova (approved):
LGTM, based on limited knowledge. I ran through the sitemaps, loaded it up locally and checked canonicals and latest URLs, and everything looks good. 

Just one question, re:
> I removed the Jekyll collection for the Plugin Hub, it wasn't being used and added extra complexity.
It was actually the only reason why the index page of the hub was being generated given that it is under app/_hub which is ignored by Jekyll because of the leading _.

Will we continue to have two hub folders then, `_hub` and `hub`? Ie, one single-sourced and one not, with `hub` services as the folder for the index page only (and possibly https://docs.konghq.com/hub/plugins/overview/ and https://docs.konghq.com/hub/plugins/compatibility)?

fabianrbz:
@lena-larionova yes, we'll have: `hub` for

app/_plugins/generators/latest_version_generator.rb
fabianrbz: Looks like it is still used in [older pages](https://github.com/Kong/docs.konghq.com/search?q=version-index) that are not single-sourced. It looks like it's also used by some `_includes` so I'm keeping this and get back to it in a follow-up ticket

PR #5079, opened 2023-01-26
Add redirects for mesh

fabianrbz:
### Description

Add some of the redirects we have in Kuma to Kong-Mesh, the new release of Kuma's GUI points to the old urls that's why  we need them. Also, we should definitely automate the generation of this file so that we can update it every time we update kuma, this is a quick and dirty solution we can ship and test before the release.

cc @lahabana 

### Testing instructions

Netlify link: 

### Checklist 

- [x] Review label added 

### DIFF
--- a/app/_mesh_redirects
+++ b/app/_mesh_redirects
@@ -1,6 +1,6 @@
 # TODO: Automate this file's generation
 # Generated redirects
-/mesh/:version/overview/ /mesh/:version/overview/what-is-a-service-mesh 301
+/mesh/:version/overview/ /mesh/:version/introduction/what-is-a-service-mesh 301
 /mesh/:version/installation/ /mesh/:version/installation/kubernetes 301
 /mesh/:version/api/ /mesh/:version/reference/http-api 301
 /mesh/:version/documentation/deployments/ /mesh/:version/introduction/deployments 301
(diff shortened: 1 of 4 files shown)


app/_mesh_redirects
```diff
@@ -0,0 +1,21 @@
+# TODO: Automate this file's generation
+# Generated redirects
+/mesh/:version/overview/ /mesh/:version/overview/what-is-a-service-mesh 301
```
lena-larionova, line 3:
```suggestion
/mesh/:version/overview/ /mesh/:version/introduction/what-is-a-service-mesh 301
```
Tested, this is the topic: https://deploy-preview-5079--kongdocs.netlify.app/mesh/latest/introduction/what-is-a-service-mesh/

app/_mesh_redirects
```diff
@@ -0,0 +1,21 @@
...
+/mesh/:version/installation/ /mesh/:version/installation/kubernetes 301
```
lena-larionova, line 4: Q about this one: why not redirect to the install landing page, ie `/mesh/:version/install/`?

https://deploy-preview-5079--kongdocs.netlify.app/mesh/latest/install/

lena-larionova (changes requested):
One broken redirect, and one that I'm not sure is going to the right place.

lena-larionova (approved):
Approving for the sake of merging this fix asap.

app/_mesh_redirects
fabianrbz: cc @lahabana

lahabana: Yes could work too

lahabana (approved):

PR #5108, opened 2023-02-01
[DOCU-2899] Plugin compatibility table + protocols

lena-larionova:
### Description

* Added a protocols table
  * Chose to do a separate table because there are a lot of protocols.
  * Had to make temporary CSS adjustments to fit the table on a standard laptop screen without scrolling.
  * I think this table needs to have a sticky header - it's super long. Tried but couldn't figure out how to make the table header sticky with just CSS - does this need a JS solution? @fabianrbz 
* Ran through the schema for every plugin using the `/schemas/plugins/{plugin-name}` Admin API endpoint and updated the plugin docs with accurate protocols. 
  * Seems like we added grpc/grpcs support a long time ago (pre-2.1) but the plugins were never updated to include that.
* Added a Konnect support column to to plugin compatibility table.
  * Removed the "network configurations" table column from the plugin compatibility table. The purpose of that column was basically to tell people if the plugin was supported in Konnect, but it was phrased in a confusing way. 
  * Removed unnecessary, on-prem specific info from the Konnect version of the page. 

**Ticket:** https://konghq.atlassian.net/browse/DOCU-2899
Also fixes https://github.com/Kong/docs.konghq.com/issues/5129.

### Testing instructions

Netlify links:
https://deploy-preview-5108--kongdocs.netlify.app/hub/plugins/compatibility/
https://deploy-preview-5108--kongdocs.netlify.app/hub/plugins/license-tiers/

### Checklist 

- [x] Review label added 

### DIFF
--- a/app/hub/plugins/compatibility.md
+++ b/app/hub/plugins/compatibility.md
@@ -127,16 +127,16 @@ simply tools to help you deploy {{ site.base_gateway }} in various environments.
       <a href="{{extn.url}}">{{ extn.name }}</a>
       </td>
       {% if extn.params.protocols == null %}
-        <td>N/A</td>
-        <td>N/A</td>
-        <td>N/A</td>
-        <td>N/A</td>
-        <td>N/A</td>
-        <td>N/A</td>
-        <td>N/A</td>
-        <td>N/A</td>
-        <td>N/A</td>
-        <td>N/A</td>
+        <td style="text-align: center">N/A</td>
+        <td style="text-align: center">N/A</td>
+        <td style="text-align: center">N/A</td>
+        <td style="text-align: center">N/A</td>
+        <td style="text-align: center">N/A</td>
+        <td style="text-align: center">N/A</td>
+        <td style="text-align: center">N/A</td>
+        <td style="text-align: center">N/A</td>
+        <td style="text-align: center">N/A</td>
+        <td style="text-align: center">N/A</td>
       {% else %}
       {% assign protocols = extn.params.protocols | replace: '"', " "%}
       <td style="text-align: center"> 
--- a/app/hub/plugins/compatibility.md
+++ b/app/hub/plugins/compatibility.md
@@ -101,8 +101,8 @@ access to a subset of plugins:
 
 ### Deployment
 
-[Deployment plugins](/hub) are not bundled with any version of Konnect, and are
-simply tools to help you deploy Kong Gateway in various environments.
+[Deployment plugins](/hub) are not bundled with any version of {{site.base_gateway}} or {{site.konnect_short_name}}, and are
+simply tools to help you deploy {{ site.base_gateway }} in various environments.
 
 ## Protocols
 
--- a/app/hub/plugins/compatibility.md
+++ b/app/hub/plugins/compatibility.md
@@ -106,7 +106,7 @@ simply tools to help you deploy {{ site.base_gateway }} in various environments.
 {% assign hub = site.data.ssg_hub %}
 {% assign kong_extns = hub | where: "publisher", "Kong Inc." %}
 
-<table>
+<table class="table-sticky">
   <thead>
       <th>Plugin</th>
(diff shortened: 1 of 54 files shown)


app/hub/plugins/compatibility.md
```diff
@@ -99,4 +102,118 @@ see [{{site.ee_product_name}} for Kubernetes Deployment Options](/gateway/latest
 ### Deployment
 
 [Deployment plugins](/hub) are not bundled with any version of Konnect, and are
```
lena-larionova, line 104:
```suggestion
[Deployment plugins](/hub) are not bundled with any version of {{site.base_gateway}} or {{site.konnect_short_name}}, and are
```

app/hub/plugins/compatibility.md
```diff
@@ -99,4 +102,118 @@ see [{{site.ee_product_name}} for Kubernetes Deployment Options](/gateway/latest
...
-simply tools to help you deploy {{ site.base_gateway }} in various environments.
+simply tools to help you deploy Kong Gateway in various environments.
```
lena-larionova, line 105:
```suggestion
simply tools to help you deploy {{ site.base_gateway }} in various environments.
```

fabianrbz:
@lena-larionova there's a way to make it work without js, I can push a fix to this branch if you want

lena-larionova:
>  I can push a fix to this branch if you want

@fabianrbz Please do!

fabianrbz:
@lena-larionova I added the changes. Unfortunately, I didn't realize that the table was too wide. There's a well-known limitation to `position: sticky` and containers that have `overflow` like in this case, so the only solution (without using JS) is to set a height to the table.

Let me know if this is ok, otherwise, I'll consider adding a JS script that fixes it.

lena-larionova:
@fabianrbz I think it doesn't hurt to do this for one table/tables that need it, as long as we don't make it a habit. But it's not line we have a ton of giant tables like this 😄 . Thank you!

lena-larionova:
Ignore the vale failure, it's erroring on unedited files. I had to do a rebase at one point and it ran the check on the whole repo.

app/hub/plugins/compatibility.md
```diff
@@ -98,5 +96,171 @@ see [{{site.ee_product_name}} for Kubernetes Deployment Options](/gateway/latest
...
+      <th><code>wss</code></th>
+  </thead>
+  <tbody>
+  {% for extn in kong_extns %}
+    <tr>
+      <td>
+      <a href="{{extn.url}}">{{ extn.name }}</a>
+      </td>
+      {% if extn.params.protocols == null %}
+        <td>--</td>
```
cloudjumpercat, line 130: I think this is what is displaying for Vault Auth in the table? If it is, just a nitpick that the spacing looks a bit off with the emoji cross and checkmark, not sure if that's something could be fixed. Also, I think someone in #docs mentioned that they were confused about what "--" meant. So maybe something like N/A would work instead?

app/hub/plugins/compatibility.md
```diff
@@ -98,5 +96,171 @@ see [{{site.ee_product_name}} for Kubernetes Deployment Options](/gateway/latest
...
 simply tools to help you deploy {{ site.base_gateway }} in various environments.
+
+## Protocols
+
+{{site.base_gateway}} and {{site.konnect_short_name}} plugins are compatible with the following protocols:
+
+{% assign hub = site.data.ssg_hub %}
+{% assign kong_extns = hub | where: "publisher", "Kong Inc." %}
+
+<table class="table-sticky">
```
cloudjumpercat, line 109: This works great!

cloudjumpercat (approved):
Just left a small nitpick comment (feel free to ignore it), but everything else looks good to me!

app/hub/plugins/compatibility.md
lena-larionova: Hmm. Good call. Will do N/A.

PR #5130, opened 2023-02-08
change docs keys link keys-entity to keys-object 

raoxiaoyan:
### Summary

INTF-2990 Need to update docs keys link keys-object instead of keys-entity
### Reason

I need to keep the link of `Keys` to the same specification as the links of other entities. Because the docs of the new vaults in KM 3.2 cannot link to the corresponding page correctly


Guaris:
I added the keys and vaults objects from your other PR to this too.

PR #5138, opened 2023-02-09
Corrected hyperlink and added proxy-cache plugin

anup-krai:
Corrected hyperlink and added proxy-cache plugin in the list


cloudjumpercat (approved):
LGTM! Thanks for catching this and updating the link!

PR #5046, opened 2023-01-17
feat(plugins): add datadog tracing plugin

mayocream:
### Description

Add docs for the new Datadog Tracing plugin.

KAG-454

### Checklist 

- [x] Review label added 
- [x] PR pointed to correct branch (`main` for immediate publishing, or a release branch: e.g. `release/gateway-3.2`, `release/deck-1.17`)

[DOCU-2918]: https://konghq.atlassian.net/browse/DOCU-2918?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

PR #5144, opened 2023-02-10
[DOCU-2928] Kong-Debug header + configuration reference update

lena-larionova:
### Description

* Add new parameters to the Configuration Reference.
  * Compared the Enterprise and OSS versions of kong.conf.default and labelled Enterprise features accordingly
  * Had to make a copy of the Postgres table for 3.2.x, couldn't get version conditionals to work right without breaking the table - I think it's because I had to use both `lte` and `gte` one after the other.
* Searched for and updated any instances of the kong-debug header, adding a step to first enable sending debugging info.
  * As part of this, cleaned up the very outdated gRPC service guide, which was mentioning Kong 1.3 everywhere.

Did not add a changelog/breaking change entry. This will be added in a separate changelog PR.
 
Ticket: https://konghq.atlassian.net/browse/DOCU-2928

### Testing instructions

Netlify link: 

### Checklist 

- [x] Review label added 

### DIFF
--- a/app/_src/gateway/how-kong-works/routing-traffic.md
+++ b/app/_src/gateway/how-kong-works/routing-traffic.md
@@ -542,9 +542,20 @@ defined URIs, in this order:
 4. `/version`
 
 Routers with a large number of regexes can consume traffic intended for other rules. Regular expressions are much more expensive to build and execute and can't be optimized easily. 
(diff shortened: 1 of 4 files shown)


Guaris (approved):
Approved, I'll let you merge it in case you decide to add redirects.

lena-larionova:
Not doing redirects for now, as the parameter still exists in the config ref, so the links will continue to work. 
We just need to make sure that we don't link to the parameter from inside our own docs; this is covered in https://github.com/Kong/docs.konghq.com/pull/5046.

PR #5158, opened 2023-02-13
Explain the session_cookie_renew property in more detail

d19dotca:
### Description

Clarified the session_cookie_renew property as it was pointed out by a customer that it wasn't clear this property was how many seconds prior to the lifetime expiration before being renewed.

### Checklist 

- [x] Review label added
- [x] PR pointed to correct branch


cloudjumpercat (approved):
LGTM! Thanks for updating this!

PR #5165, opened 2023-02-14
docs(rate-limiting): remove wrongly placed information

samugi:
### Description

Remove an information that, put that way, is redundant and confusing. The values `service.id` and `service.name` are not related with the configuration of the parameter `limit_by`, which can be set even when the plugin is configured globally.

### Checklist 

- [X] Review label added
- [X] PR pointed to correct branch


cloudjumpercat (approved):

PR #5193, opened 2023-02-21
Support policy URL

lena-larionova:
### Description

Support policy page was renamed from `support-policy.md` to `support/index.md`, which changed the URL to `/support/`. Using the new page as the source, but making it so that we still keep the old `/support-policy/` URL and don't break links.

As part of the support policy rework, two other URLs were removed: OS support and install support. Added them to moved URLs; not sure if they'll work how I want them to, so will need to test in preview.

### Testing instructions

Netlify link: 

### Checklist 

- [x] Review label added 


acgoldsmith (approved):
LGTM and thanks

PR #5209, opened 2023-02-25
[DOCU-2985] Mention document_objects as not managed by decK

lena-larionova:
### Description

Added a note for `document_objects` as being the one sub-entity for services that isn't managed by decK.

**Note:** We also don't have the `document_objects` entity documented in our Admin API doc; in fact, it's purposely not generated, if you look at `admin_api.lua`: https://github.com/Kong/kong-ee/blame/master/autodoc/admin-api/data/admin-api.lua#L1177
It looks like there was a start to a project to add EE endpoints to the Admin API doc, and it stalled.
I think we should wait to revisit that until we have a spec.
 
https://konghq.atlassian.net/browse/DOCU-2985

### Testing instructions

Netlify link: https://deploy-preview-5209--kongdocs.netlify.app/deck/latest/reference/entities/

### Checklist 

- [x] Review label added 

### DIFF
--- a/app/_src/deck/reference/entities.md
+++ b/app/_src/deck/reference/entities.md
@@ -10,7 +10,7 @@ It does not manage {{site.base_gateway}} configuration parameters in `kong.conf`
 
 Entity | Managed by decK?
 -------|-----------------
-Services | <i class="fa fa-check"></i>
+Services | <i class="fa fa-check"></i> <sup>1</sup>
 Routes | <i class="fa fa-check"></i>
 Consumers | <i class="fa fa-check"></i>
 Plugins | <i class="fa fa-check"></i>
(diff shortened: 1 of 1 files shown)


app/_src/deck/reference/entities.md
```diff
@@ -10,7 +10,7 @@ It does not manage {{site.base_gateway}} configuration parameters in `kong.conf`
 
 Entity | Managed by decK?
```
cloudjumpercat, line 11: Not really related to this PR, but would it help to break this table up more into "decK dump" or "decK delete" columns? Since some of them can be managed in one way, but not another?

cloudjumpercat (approved):
LGTM! I left one comment that may or may not be something for a future improvement, I don't understand decK too well. I'll merge this one in though since it looks good to me right now in it's current state.

app/_src/deck/reference/entities.md
lena-larionova: There is no `deck delete` or anything like that. `deck dump`  generates a config file for you, you edit the file however you want, then `deck sync` to your gateway instance to apply configuration. 

There are specific requirements for `workspaces`, and for `document_objects`, which are subsets of the `services` endpoint. With workspaces, it's because a workspace is basically a container for all of your other config; with `document_objects`, it's because they're dev portal entities that happen to be attached to services. 

Does that make sense, and do you think this table still needs some kind of clarification?

cloudjumpercat: That makes sense, I don't think the table needs any additional clarification. I was just trying to see if there was a way to break it into additional columns so we wouldn't need footnotes. But footnotes makes the most sense here I think.

PR #5223, opened 2023-03-01
chore: Add datadog-tracing to konnect plugin priority list

lena-larionova:
### Description

Datadog Tracing plugin is missing from the Konnect plugin priority list.

Should we make an include for this? This content is currently being reused in copy&paste form in Konnect and Gateway docs (this is the Gateway topic: https://docs.konghq.com/gateway/latest/plugin-development/custom-logic/) 

### Testing instructions

Netlify link: https://deploy-preview-5223--kongdocs.netlify.app/konnect/runtime-manager/plugins/

### Checklist 

- [x] Review label added 


cloudjumpercat (commented):
The output looks good to me! I think making this an include makes sense since it's the same as for Gateway. Also, is there anyway to autogenerate this? I think we have another plugin table somewhere (I can't remember where) where the names are pulled from the plugin files and I was wondering if we had these priority orders on the plugins so we could pull those as well? I know I always forget to update this table when I add a new plugin doc.

lena-larionova:
Include done.

>  Also, is there anyway to autogenerate this? I think we have another plugin table somewhere (I can't remember where) where the names are pulled from the plugin files and I was wondering if we had these priority orders on the plugins so we could pull those as well? I know I always forget to update this table when I add a new plugin doc.

There can be, but it's not straightforward. This info isn't included in plugin schemas. I could manually add it to the plugin data file, but the issue would still be the same, I think.

Let me file a ticket for this, because it is a good idea. Another opportunity to automate in the future.

cloudjumpercat (approved):
LGTM!

PR #5233, opened 2023-03-02
fix: align examples to behaviour of add.headers

janv8000:
> Ignored if the header is already set.

Use `v2` to highlight fact that `v2` is ignored because header already set

### Description

Docs


lena-larionova (approved):
Nice catch, thanks for the fix. LGTM

PR #5235, opened 2023-03-02
fix(correlation-id): example log correct header

Tieske:
### Description

the `sent_` prefixed headers are the headers returned by kong to the client. Without that prefix, it will refer to the header sent to the backend. Since the former ones are only available if the "echo to client" checkbox is ticked, this is the safer option for the example.

### Testing instructions

Netlify link: 

### Checklist 

- [x] Review label added 


lena-larionova (approved):
LGTM

PR #5259, opened 2023-03-06
KIC 2.9 version compat

pmalek:
### Description

Update version compatibility matrices for KIC 2.9.x

### Testing instructions

Netlify link: 

### Checklist 

- [x] Review label added 


rainest:
Wait, what made this not get the normal Netlify preview bot comment?

Aside that, @mheap do you know how hard it would be to make a docs widget that takes a YAML compat table input and renders a markdown table with the N most recent versions? Since we now use the single-source docs system it's (more) annoying to have to chop off old table columns and add new ones.

mheap:
@rainest I did a spike on the config driven approach with a sliding window. It’s not trivial with the current setup (liquid). One for @fabianrbz to investigate writing a plugin for

rainest:
Aight, maybe we want to look into handling it out of band then. Maybe Liquid can't do it well, but I bet a slightly scuffed Python script can.

pmalek:
> Wait, what made this not get the normal Netlify preview bot comment?

@rainest its targeting non main branch, that's probably why

pmalek:
Apart from the possible improvements around the process that we can make does that PR look good enough?

rainest (approved):
The table stuff is one of the things where you really need the preview because reading the actual content of the diff is painful, and it's also easy to accidentally break the formatting, so you really want to see the rendered output.

Flipping to main didn't trigger the bot, so I think we just have to review this properly on the final merge to main. AFAIK content-wise we're just copying values from 2.8, so despite not easily being able to confirm they do match from the diff, I wouldn't expect that to be wrong.

PR #5266, opened 2023-03-08
Version the description for the RLA plugin

lena-larionova:
### Description

The Rate Limiting Advanced plugin had some features introduced in 3.1 that are showing up in doc versions before 3.1. 
This is because we can't version the plugin description separately. Currently, the only way to version the description is to split the source files, which is what I'm doing in this PR.

This is a temp measure that won't be needed in the future, as we replace the plugin hub formatting with separate files for each content type.

Issue reported on Slack: https://kongstrong.slack.com/archives/CDSTDSG9J/p1678235473401009

### Testing instructions

Netlify link: 

### Checklist 

- [x] Review label added 


cloudjumpercat (approved):
LGTM!

PR #5273, opened 2023-03-08
chore(*): docker read-only mode docs

samugi:
### Description

Document how to start kong in Docker in `read-only` mode

### Testing instructions

Netlify link: 

### Checklist 

- [X] Review label added 
- [X] PR pointed to correct branch (`main` for immediate publishing, or a release branch: e.g. `release/gateway-3.2`, `release/deck-1.17`)

[KAG-852]

[KAG-852]: https://konghq.atlassian.net/browse/KAG-852?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/app/_src/gateway/install/docker/index.md
+++ b/app/_src/gateway/install/docker/index.md
@@ -441,7 +441,7 @@ docker network rm kong-net
 
 ## Running Kong in read-only mode
 
-Starting with version 3.2.0 of {{site.base_gateway}} it is possible to run the container in read-only mode. To do so, mount a Docker volume to the locations where Kong needs to write data. The default configuration requires write access to `/tmp` and to the prefix path, as provided by the following example:
+Starting with {{site.base_gateway}} 3.2.0, you can run the container in read-only mode. To do so, mount a Docker volume to the locations where Kong needs to write data. The default configuration requires write access to `/tmp` and to the prefix path:
 
 {% navtabs_ee codeblock %}
 {% navtab Kong Gateway %}
--- a/app/_src/gateway/install/docker/index.md
+++ b/app/_src/gateway/install/docker/index.md
@@ -439,6 +439,63 @@ docker container rm kong-dbless
 docker network rm kong-net
 ```
 
+## Running Kong in read-only mode
+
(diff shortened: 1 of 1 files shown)


app/_src/gateway/install/docker/index.md
```diff
@@ -439,6 +439,63 @@ docker container rm kong-dbless
 docker network rm kong-net
 ```
 
+## Running Kong in read-only mode
+
+To run Kong in read-only mode, mount a Docker volume to the locations where Kong needs to write data. The default configuration requires write access to `/tmp` and to the prefix path, as provided by the following example:
```
kikito, line 444: This needs a mention of "starting in 3.2.0". We offer no guarantees for previous versions.

kikito (changes requested):
missing version ingo

app/_src/gateway/install/docker/index.md
samugi: done. @Kong/docs-maintainers should this be handled with some special version tagging?

Hayk-S:
The provided configuration looks correct and I was able to reproduce the read-only issue with this configuration.
Just FYI, in the docs, we provide `read_only` example for `db-less` mode but it is also possible to enable that mode with the classic or hybrid gateway modes.

Hayk-S (approved):

app/_src/gateway/install/docker/index.md
```diff
@@ -439,6 +439,63 @@ docker container rm kong-dbless
 docker network rm kong-net
 ```
 
+## Running Kong in read-only mode
+
+Starting with version 3.2.0 of {{site.base_gateway}} it is possible to run the container in read-only mode. To do so, mount a Docker volume to the locations where Kong needs to write data. The default configuration requires write access to `/tmp` and to the prefix path, as provided by the following example:
```
cloudjumpercat, line 444:
```suggestion
Starting with {{site.base_gateway}} 3.2.0, you can run the container in read-only mode. To do so, mount a Docker volume to the locations where Kong needs to write data. The default configuration requires write access to `/tmp` and to the prefix path:
```

cloudjumpercat (approved):
Just made a small copyedit, but everything else looks good to me! Thanks for making this update!

PR #5277, opened 2023-03-09
Fix absolute links in link checker

mheap:
### Description

Any absolute URL would resolve to `/<product>/<release>` in the previous fix. This inverts the logic so that the URL is used rather than the prefix.

It previously worked as `url == prefix` for index pages (which is what I set out to fix)

### Testing instructions

N/A

### Checklist 

- [x] Review label added 

### DIFF
--- a/tools/broken-link-checker/lib/src-to-urls.js
+++ b/tools/broken-link-checker/lib/src-to-urls.js
@@ -32,10 +32,12 @@ async function srcToUrls(pattern, src) {
 
     urls = urls.concat(
       r.map((u) => {
-        const url = u.is_absolute ? "" : u.url;
+        const prefix = u.is_absolute
+          ? ""
+          : `/${entry.product}/${entry.release}`;
         return {
           source: src,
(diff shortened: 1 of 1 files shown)


tools/broken-link-checker/lib/src-to-urls.js
```diff
@@ -32,10 +32,10 @@ async function srcToUrls(pattern, src) {
 
     urls = urls.concat(
       r.map((u) => {
-        const url = u.is_absolute ? "" : u.url;
+        const prefix = u.is_absolute ? `/${entry.product}/${entry.release}` : u.url;
         return {
           source: src,
-          url: `/${entry.product}/${entry.release}${url}`,
+          url: `${prefix}${u.url}`,
```
fabianrbz, line 38: Does `${u.url}${u.url}` make sense when the url is not absolute?

fabianrbz (approved):
LGTM

fabianrbz:
@mheap can I ship this?

PR #5304, opened 2023-03-17
[Mesh] add community section to 2.2.x

Guaris:
Kuma community section to mesh 2.2


cloudjumpercat (commented):
All the nav changes look correct to me, I think this is just missing the submodule update. I think you tried doing it via the GitHub action, but I'm not sure if that worked. If it didn't, I wonder if the easiest solution would be to update the module in a commit on the `release/mesh-2.2` branch (GitHub action or manually) and then we could rebase our Mesh PRs that are pointing to that branch.

cloudjumpercat (approved):
LGTM! All the files from the submodule look like they were pulled into the nav correctly, so I think this is good to be merged

PR #5321, opened 2023-03-20
[Fix] Update Support policy table

Guaris:
Reported on [slack](https://kongstrong.slack.com/archives/CDSTDSG9J/p1678325245916159)

I changed the nav bars to pull from the same src file being used in 3.2 for 2.8, 3.0, and 3.1.

### DIFF
--- a/app/_data/docs_nav_gateway_2.8.x.yml
+++ b/app/_data/docs_nav_gateway_2.8.x.yml
@@ -1,3 +1,8 @@
+product: gateway
+release: 2.8.x
+generate: true
+assume_generated: false
+items:
 - title: Introduction
   icon: /assets/images/icons/documentation/icn-flag.svg
   items:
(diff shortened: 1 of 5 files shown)


app/_data/docs_nav_gateway_2.8.x.yml
```diff
@@ -6,6 +6,7 @@
       absolute_url: true
     - text: Version Support Policy
       url: /support-policy
+      src: /gateway/support/index
```
lena-larionova, line 9: This doesn't work: https://deploy-preview-5321--kongdocs.netlify.app/gateway/2.8.x/support-policy/

The 2.8 docs isn't set to generate from `_src`. I wonder if you could this this method though:
https://deploy-preview-5321--kongdocs.netlify.app/contributing/single-sourced-versions/#single-source-specific-files (we haven't actually used it anywhere yet).

So  _I think_ it should be like this:

Top of the file:
```
product: gateway
release: 2.8.x
generate: true
assume_generated: false
```

Then set `generate: true` for the one item:
```
- text: Version Support Policy
  url: /support-policy
  src: /gateway/support/index
  generate: true
```

Guaris: That didn't work, I updated the file.

lena-larionova (approved):
LGTM, with one minor edit for links.

PR #5328, opened 2023-03-22
docs(hybrid): config are cached in LMDB instead of JSON after 3.0

dndx:


lena-larionova (approved):
Thanks for the fix, LGTM. Did a quick check for the filename in the rest of the doc, didn't find any more instances.

PR #5400, opened 2023-04-04
Tile grid style fix + konnect runtime instances improvements

lena-larionova:
### Description

* Removed hover & focus pseudoclasses from unlinked install grid tiles. The existence of a hover class was creating the illusion of clickable tiles, when the tiles are just visual. Hover classes are now applied only when there's a link on the tile.
* Added a short description of the runtime instance installation options and a line on how to get to them.
* Based on user testing (John and Amy), removing the "Installation Options" title from the page wasn't the right call, since it's not clear that installation options will appear on the overview. Reverted the title, but kept the URL the same.

### Testing instructions

Netlify link: https://deploy-preview-5400--kongdocs.netlify.app/konnect/runtime-manager/runtime-instances/

### Checklist 

- [x] Review label added 

### DIFF
--- a/app/_assets/stylesheets/page.less
+++ b/app/_assets/stylesheets/page.less
@@ -1404,18 +1404,15 @@ Generic Styling for Desktop
       &.no-description {
         justify-content: center;
       }
-
-      &:hover,
-      &.focus {
-        .install-text {
-          color: @blue-500;
-        }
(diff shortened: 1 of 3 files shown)


acgoldsmith (approved):
LGTM and thank you for the fix

PR #5409, opened 2023-04-06
fix route-transformer-advanced config.path example

gforns:
### Description

config.path is an string. Example value should also be an string, not an array.
 
current declarative example provides a   `(message: "schema violation (config.path: expected a string)")`  error while using deck

### Checklist 

- [ ] Review label added 


cloudjumpercat (approved):

PR #5443, opened 2023-04-17
removed plus badge from various analytics pages

sichvoge:
### Description

Analytics capabilities are available for all tiers and are only restricted by retention today (like described in the overview page). Therefore, I removed all occurrences of the "PLUS" badge across the Analytics pages.


lena-larionova (approved):
LGTM!

lena-larionova:
Broken link is a false positive, as it links to cloud.konghq.com/servicehub, which is behind a login wall.

PR #5464, opened 2023-04-20
mesh: update sidebar to follow IA work for install

lahabana:
### Description

The install page was using older architecture which is not the case anymore according to the IA change.

### Testing instructions

Netlify link: 

### Checklist 

- [ ] Review label added 

### DIFF
--- a/app/_data/docs_nav_mesh_2.2.x.yml
+++ b/app/_data/docs_nav_mesh_2.2.x.yml
@@ -187,7 +187,6 @@ items:
       - text: Control Plane Configuration
         url: /documentation/configuration/
         src: /.repos/kuma/app/_src/documentation/configuration/
-      - text: Requirements
   - title: Policies
     icon: /assets/images/icons/documentation/icn-documentation-small.svg
     items:
@@ -335,6 +334,8 @@ items:
         url: /features/ubi-images
       - text: Windows Support
         url: /features/windows
+      - text: ECS Support
+        url: /installation/ecs
       - text: Auditing
         url: /features/access-audit
   - title: Reference docs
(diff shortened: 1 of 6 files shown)


app/_data/docs_nav_mesh_2.2.x.yml
```diff
@@ -133,35 +139,6 @@ items:
...
-        url: /installation/redhat/
-      - text: CentOS
-        url: /installation/centos/
-      - text: Debian
-        url: /installation/debian/
-      - text: Ubuntu
-        url: /installation/ubuntu/
-      - text: macOS
-        url: /installation/macos/
-      - text: Windows
```
cloudjumpercat, line 163: @lahabana Do we need to keep Windows since that's Kong Mesh-specific?

lahabana: How was this supposed to fit in IA?

cloudjumpercat: @lahabana I think the "Install kumactl" and "Use Kuma" pages were meant to replace all the install pages, but once I got to the Mesh ones, I remembered that those pages weren't shared with the Kuma docs. They were separate. If there isn't a reason why they should be different/their own files, I think we can mirror what's currently in the Kuma nav and remove all the install topics and replace them with "Install kumactl" and "Use Kuma". 

Looking at the Windows one, I think we'd miss some of the Windows specifics in that file, but I think [Install kumactl](https://kuma.io/docs/2.2.x/production/install-kumactl/) does cover the general things like adding kumactl to your path.

lahabana: Ok I've added back a windows install page which seems to work. I've also did minor fixes to remove warnings

cloudjumpercat (approved):
Everything LGTM! Thanks for making these updates!

PR #5518, opened 2023-05-03
Add Konnect DP Certificate renewal with OpenSSL instructions

mheap:
### Description

It's possible to renew DP certs with `openssl` but we didn't document the process publicly. 

### Testing instructions

Netlify link: https://deploy-preview-5518--kongdocs.netlify.app/konnect/runtime-manager/runtime-instances/renew-certificates/#:~:text=Alternatively%2C%20you%20can%20generate%20a%20certificate%20locally

### Checklist 

- [x] Review label added 

### DIFF
--- a/app/konnect/runtime-manager/runtime-instances/renew-certificates.md
+++ b/app/konnect/runtime-manager/runtime-instances/renew-certificates.md
@@ -23,9 +23,8 @@ files.
 
 ## Quick setup
 
-If you originally created your runtime instance container using the
-[quick setup Docker script](/konnect/runtime-manager/runtime-instances/gateway-runtime-docker/#quick-setup),
-we recommend running the script again to create a new instance with renewed
+If you originally created your runtime instance container using one of the
+Docker options in Runtime Manager, we recommend creating a new instance with renewed
 certificates.
 
 1. Stop the runtime instance container.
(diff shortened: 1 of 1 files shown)


mheap:
@Kong/docs-maintainers The failure is legit on this page (docker quickstart link no longer exists). I'm not sure what it should be pointing to

app/konnect/runtime-manager/runtime-instances/renew-certificates.md
```diff
@@ -57,6 +57,19 @@ You can generate a new data plane certificate from the {% konnect_icon runtimes
...
+```bash
+# Generate the certificate + key
+openssl req -new -x509 -nodes -newkey rsa:2048 -subj "/CN=kongdp/C=US" -keyout ./tls.key -out ./tls.crt
+
+# Get the certificate as a single line for the API call
+export CERT=$(awk 'NF {sub(/\r/, ""); printf "%s\\n",$0;}' tls.crt)
+
+# Send the certificate to Konnect
+curl https://us.api.konghq.com/v2/runtime-groups/{runtimeGroupId}/dp-client-certificates --json '{"cert":"'$CERT'"}'
+```
```
lena-larionova, line 71: Can you split this into two tabs, "Using Konnect" and "Generate locally"? Or h4 headings, that would also work.

Also, please split the codeblock into one for each command, using the comment as the step instruction. We get a lot of complaints about codeblocks like this, since someone can't just use the copy button for them.

(If you're fine with it, I'm happy to go ahead and make these changes)

lena-larionova:
> @Kong/docs-maintainers The failure is legit on this page (docker quickstart link no longer exists). I'm not sure what it should be pointing to

I think the answer is that there shouldn't be a link. The reason we even had that line was because the quick start script obfuscated all the config params and it wasn't obvious how to make any changes to it.

How about something like this, instead of the phrasing that's there right now:

> If you originally created your runtime instance container using one of the
Docker options in Runtime Manager, we recommend creating a new instance with renewed
certificates.

mheap:
Sounds good to me. Could you fix this one and merge when you're happy?

lena-larionova (approved):

PR #5520, opened 2023-05-03
[fix] Broken link in Linux installation docs

Guaris:
https://kongstrong.slack.com/archives/CDSTDSG9J/p1683122758154309

- Added a link to https://docs.konghq.com/gateway/latest/support-policy/#supported-versions in all of the installation docs. Replacing the broken link where appropriate. 
- Rewrote redirects that pointed to the broken link. 
- Added redirect for the broken link to https://docs.konghq.com/gateway/latest/support-policy/#supported-versions

### DIFF
--- a/app/_redirects
+++ b/app/_redirects
@@ -1,7 +1,8 @@
 # DEPRECATED OS
-/gateway/3.0.x/install-and-run/centos    /gateway/latest/install-and-run/os-support
-/gateway/latest/install-and-run/centos   /gateway/latest/install-and-run/os-support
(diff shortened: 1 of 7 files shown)


lena-larionova (changes requested):
Can you also fix the url in the moved_urls file? There's an entry at the very bottom of the file but it's not correct.
https://github.com/Kong/docs.konghq.com/blob/fix-linux-link/app/moved_urls.yml

lena-larionova (approved):

PR #5558, opened 2023-05-12
[docu-3186] SOBM page

Guaris:
https://konghq.atlassian.net/browse/DOCU-3186

- I really think this would be better off as a link directly from our side bar to the KB page. I could see this page valuable if we have many links in the future which I imagine we will. 

This link must work to merge the PR
https://support.konghq.com/support/s/article/SBOM-Artifacts-for-Kong-Gateway-3-3-0-0


lena-larionova (changes requested):
Filename has a typo, says "sobm", should be "sbom". 

> I really think this would be better off as a link directly from our side bar to the KB page. I could see this page valuable if we have many links in the future which I imagine we will.

I agree in theory - but it would be just for this release. This would quickly have to change as we add more links though, so I think it's good to set up this page in prep for those releases.

Guaris:
Currently failing on 

```
  {
    "page": "https://deploy-preview-5558--kongdocs.netlify.app/gateway/3.3.x/support/sbom/",
    "source": "app/_src/gateway/support/sbom.md",
    "text": "Kong Gateway 3.3.0.0 SBOM",
    "target": "https://support.konghq.com/support/s/article/SBOM-Artifacts-for-Kong-Gateway-3-3-0-0",
    "reason": "HTTP_404"
  }
]
``` 
but it's expected

lena-larionova (approved):

PR #5581, opened 2023-05-16
Remove centos from 3.x

lena-larionova:
### Description

We don't support centos and haven't supported it since 3.0. The doc shouldn't exist in the single-sourced gateway section.

Looks like an entry for it accidentally got added in our 3.2 support table as well, though it wasn't there for 3.0 and 3.1.

### Testing instructions

Netlify link: 

### Checklist 

- [x] Review label added 


cloudjumpercat (approved):
LGTM! I checked out the Install pages and version support since 3.0 and CentOS was gone as expected.

PR #5586, opened 2023-05-17
docs(expressions): fix incorrect description of priority tie-breaker

dndx:
Expressions does not use `created_at` as a tie-breaker in case two routes have the same `priority` field. UUID is used instead.


app/_src/gateway/key-concepts/routes/expressions.md
```diff
@@ -64,25 +64,25 @@ For a list of all available operators, see the [reference documentation](/gatewa
 
 When the Expressions router is used, available expressions are evaluated using the `priority` field of the corresponding `Route` object
 where the expression is configured. Routes are evaluated in order of priority, where the highest priority integer is evaluated first. If two routes have the same priority, then
-the newest route (that is, routes with the highest `created_at`) will be evaluated first.
+the the route with the higher `id` (UUID) will be evaluated first as a tie-breaker.
```
lena-larionova, line 67:
```suggestion
the route with the higher `id` (UUID) will be evaluated first as a tie-breaker.
```

PR #5613, opened 2023-05-22
Add directory for Insomnia button generation

Guaris:
I need a place to host the specs to generate insomnia buttons with every release. This should be an okay temporary solution.


cloudjumpercat (approved):
LGTM!

PR #5653, opened 2023-05-31
Rename Kubernetes Ingress Controller to Kong Ingress Controller

lena-larionova:
### Description

Renaming Kubernetes Ingress Controller to Kong Ingress Controller to satisfy CNCF requirements.
https://konghq.atlassian.net/browse/DOCU-3164

### Testing instructions

Netlify link: https://kic-rename--kongdocs.netlify.app/kubernetes-ingress-controller/latest/

Had to enable a branch preview, as the previous (overwritten) redirect changes weren't working correctly, even with a deploy cache clear.

### Checklist 

- [x] Review label added 

### DIFF
--- a/app/_src/gateway/production/monitoring/readiness-check.md
+++ b/app/_src/gateway/production/monitoring/readiness-check.md
@@ -132,6 +132,6 @@ readinessProbe:
 
 For more information on Kong and related topics, check out the following resources:
 
-* [Kong Admin API Documentation](https://docs.konghq.com/gateway/latest/admin-api/)
-* [Get Started with Kong Ingress Controller](https://docs.konghq.com/kubernetes-ingress-controller/latest/deployment/)
+* [Kong Admin API Documentation](/gateway/latest/admin-api/)
+* [Get Started with Kong Ingress Controller](/kubernetes-ingress-controller/latest/deployment/overview/)
 * [Kong Helm Chart](https://github.com/Kong/charts/tree/main/charts/kong)
--- a/app/_src/gateway/production/monitoring/readiness-check.md
+++ b/app/_src/gateway/production/monitoring/readiness-check.md
@@ -133,5 +133,5 @@ readinessProbe:
 For more information on Kong and related topics, check out the following resources:
 
 * [Kong Admin API Documentation](/gateway/latest/admin-api/)
-* [Get Started with Kong Ingress Controller](/kubernetes-ingress-controller/latest/deployment/overview/)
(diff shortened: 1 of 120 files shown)


lena-larionova:
@fabianrbz could you take a look at the failing test? I'm not really sure what I missed. It doesn't really pinpoint what failed, it makes it look like everything is broken 😭

mheap:
@lena-larionova I don't think `/kong-ingress-controller/` is the right URL here. I think we should either keep `/kubernetes-ingress-controller/` as it's descriptive (I don't think it runs afoul of the CNCF in a URL) or switch to just `/ingress-controller/` as the URL.

We should refer to Kong Ingress Controller everywhere in text

lena-larionova:
@mheap Do you have a preference on which? If we don't change the URLs then it's probably better for SEO, yeah?

mheap:
Yep, SEO is the main reason to not change for me. I'd opt to keep `/kubernetes-ingress-controller/` unless someone explicitly asks us to remove it

app/_src/gateway/production/monitoring/readiness-check.md
```diff
@@ -132,6 +132,6 @@ readinessProbe:
 
 For more information on Kong and related topics, check out the following resources:
 
-* [Kong Admin API Documentation](https://docs.konghq.com/gateway/latest/admin-api/)
-* [Get Started with Kong Ingress Controller](https://docs.konghq.com/kubernetes-ingress-controller/latest/deployment/)
```
mheap, line 136: Should we add a CI check that disallows this URL (use `{{ site.links.web }}` instead)?

lena-larionova: I would say disallow in general. The instances that we actually need to link to `https://docs.konghq.com` and not use a relative URL are very rare (in fact, I currently can't think of one).

mheap: We do it in k8s manifest URLs that are different per version

lena-larionova: I guess we could just have the error explain that - if you're linking to another doc page, use a relative link; if you have to use the direct link, use `{{ site.links.web }}` instead.

lena-larionova: Wonder if we could do this with Vale @Guaris ?

Guaris: Done https://github.com/Kong/docs.konghq.com/pull/5676

mheap (approved):

Guaris (approved):

PR #5657, opened 2023-06-01
Customize netlify's lighthouse plugin to run on specific pages

fabianrbz:
### Description

Run lighthouse against each product landing page.
We can keep an eye on this and add thresholds to make the build fail if the changes made generate a bad score

### Testing instructions

Netlify link: 

### Checklist 

- [x] Review label added 

### DIFF
--- a/app/_includes/plugins/header.html
+++ b/app/_includes/plugins/header.html
@@ -13,7 +13,7 @@
       <ul class="breadcrumbs install">
         <li class="breadcrumb-item">
           <a href="/hub/">
-            <img src="/assets/images/icons/documentation/hub/icn-breadcrumbs.svg">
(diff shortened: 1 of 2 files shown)


lena-larionova (commented):
Should we maybe throw in some different page types? For example, a plugin config page could be good, as a representative of a generated reference page.

If we really want it to fail us badly, we can also run it on the [Gateway Admin API reference ](https://docs.konghq.com/gateway/latest/admin-api/)😆 .

Maybe also throw in a use case page with a bunch of images, eg https://docs.konghq.com/konnect/analytics/use-cases/?

fabianrbz:
I added a few more and forced lighthouse to run against the actual preview up instead of a local server, so we should get more accurate results.

lena-larionova (approved):

PR #5660, opened 2023-06-01
Plugin API references + splitting ACL and ACME plugins to test

lena-larionova:
### Description

Adding the ability to split plugin content into API reference docs and splitting two plugins as examples.

* **API Reference pages:** I tried implementing this in a couple of different ways and came around to just having one top-level API reference link in the side navigation instead of a Reference section. I did this for a few reasons:
  * Less clicks: people are already complaining about the nested config ref content, and we're also nesting "Using this Plugin"
  * Easy to find critical info
  * Most plugins don't have a lot of reference content beyond API stuff. I wonder if non-API references can be nested under the "Using this plugin" section.

* Split out content from the Overview pages into Using this Plugin and API Reference pages.
* Also added a notice for users looking for the config ref, as we've gotten reports that existing users are confused about where to find it and aren't used to all the clicks (probably should've done this in a separate PR, really)

❗ Reviewers: this needs review from a bunch of angles: technical, information architecture/UX, and general visual cohesiveness. Pinging the team for some different points of view :) 

### Testing instructions

Netlify links: 
* https://deploy-preview-5660--kongdocs.netlify.app/hub/kong-inc/acme/
* https://deploy-preview-5660--kongdocs.netlify.app/hub/kong-inc/acl/

### Checklist 

- [x] Review label added 


cloudjumpercat:
@lena-larionova Overall, I really like the API reference getting split out into it's own section. I also like the separate "Using this Plugin" section. I found myself wondering what the difference was between the pages in "Using this Plugin" and the "Configuration > Example" page. I almost feel like that Example page could be moved as the top page to the "Using this Plugin" section and renamed something like "Basic configuration" since it seems more similar to those pages than the Configuration pages? So something like:

- Overview
- Configuration parameters (maybe?)
- Using the plugin
  - Basic configuration
  - [insert specific use case here]
- API reference
- Changelog

lena-larionova:
> I almost feel like that Example page could be moved as the top page to the "Using this Plugin" section and renamed something like "Basic configuration" since it seems more similar to those pages than the Configuration pages? 

I'm really glad to hear you say this, because this has been bothering me as well ([literally just talked about this morning(https://kongstrong.slack.com/archives/G012HLJ4GJJ/p1685729858788559?thread_ts=1685723589.130179&cid=G012HLJ4GJJ)). So it sounds like you're seeing the same issue with the current layout.

lena-larionova:
@fabianrbz I think this is finally ready to go. I made a couple of small tweaks:
* Made sure that the "Using this plugin" section will always appear in Kong plugins, since we're nesting the examples in it now
* For third-party plugins, made the examples a top-level item. We can always change it in the future, but the aim is to make it as simple as possible for contributors, which means avoiding any extra markdown files.

Any reviewer, check the following:
* Any Kong plugin has a "Using this plugin" section with examples in it, eg [basic auth](https://deploy-preview-5660--kongdocs.netlify.app/hub/kong-inc/hmac-auth/how-to/basic-example/)
* Third-party plugins with examples have a "Basic config examples" nav entry, eg [Imperva](https://deploy-preview-5660--kongdocs.netlify.app/hub/imperva/imp-appsec-connector/how-to/basic-example/)
* Third-party plugins without examples don't have examples in the nav, eg [Wallarm](https://deploy-preview-5660--kongdocs.netlify.app/hub/wallarm/wallarm/)

acgoldsmith (approved):
LGTM

PR #5661, opened 2023-06-02
[DOCU-3203] Distributed deck config: many to one

lena-larionova:
### Description

Turns out there's not a whole lot to say in the doc about this, it's a pretty small doc entry - @zekth , anything else you can think of?

### Testing instructions

Netlify link: 

### Checklist 

- [x] Review label added 

### DIFF
--- a/app/_src/deck/guides/distributed-configuration.md
+++ b/app/_src/deck/guides/distributed-configuration.md
@@ -90,8 +90,8 @@ flag on the `reset` command.
 
 ## Convert distributed configuration into one file
 
-You can convert multiple distributed declarative configuration files into one
-{{site.base_gateway}} declarative configuration file using `deck convert`. 
+You can convert multiple distributed decK configuration files into one
+{{site.base_gateway}} file using `deck convert`. 
 
 For example:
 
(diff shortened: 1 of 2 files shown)


app/_src/deck/guides/distributed-configuration.md
```diff
@@ -88,7 +88,22 @@ configuration.
 You can delete only a subset of entities sharing a tag using the `--select-tag`
 flag on the `reset` command.
 
-## Initial setup problem
+## Convert distributed configuration into one file
+
+You can convert multiple distributed declarative configuration files into one
```
zekth, line 93: into one Gateway file is more accurate don't you think?

cc @GGabriele

GGabriele: Honestly I'd have used `You can convert multiple distributed decK configuration files into one Gateway file`.
The reason for replacing `declarative` with `decK` is that `declarative` is a very specific format in Kong, which is mostly related with the DB-less mode.

That said, I see that `declarative configuration` is consistently used in the docs, so it's better to not diverge from it.

Overall, I think the suggestion from Vincent makes sense.

lena-larionova: > The reason for replacing declarative with decK is that declarative is a very specific format in Kong, which is mostly related with the DB-less mode.
> That said, I see that declarative configuration is consistently used in the docs, so it's better to not diverge from it.

Not always the best reason to keep doing things a certain way 😄 . Ill use your phrasing.

Curious: what _is_ the difference? Is there any different in format? The terms are generally used interchangeably, as you've noticed, and I've never heard anyone mention a difference - I didn't know they weren't referring to the same thing till now, eek.

PR #5669, opened 2023-06-05
Update systemd.md

raylaijh:
### Description

What did you change and why?

I got an error with the default systemd file. The issue is that we need to add on the user and group as root (2 additional lines) to run the system file. I have added an example here. 

The next question of how to run as non-root will be raised. Understand that we do have a doc here: https://docs.konghq.com/gateway/latest/production/running-kong/kong-user/, but the important note is sending mixed signal that Kong can be only run as root. Requesting for engineering team to work on a tested solution to run Kong without root - that will be more palatable to users who are more concerned with security

 
Include any supporting resources, e.g. link to a Jira ticket, GH issue, FTI, Slack, Aha, etc.

### Testing instructions

Netlify link: 

### Checklist 

- [x] Review label added 

### DIFF
--- a/app/_src/gateway/production/running-kong/systemd.md
+++ b/app/_src/gateway/production/running-kong/systemd.md
@@ -81,7 +81,7 @@ sudo systemctl status kong
 The official systemd service is located at `/lib/systemd/system/kong-enterprise-edition.service` for
 {{site.base_gateway}}, or at `/lib/systemd/system/kong.service` for {{site.ce_product_name}}.
 
-## Add a sample systemd service file for reference
+Note that the official Kong systemd service doesn't set `user` or `group` as `root`, which is required to run the system file. See the example below with both `user` and `group` both set to `root`:
 
 ```
 [Unit]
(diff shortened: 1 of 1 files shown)


app/_src/gateway/production/running-kong/systemd.md
```diff
@@ -81,6 +81,35 @@ sudo systemctl status kong
 The official systemd service is located at `/lib/systemd/system/kong-enterprise-edition.service` for
 {{site.base_gateway}}, or at `/lib/systemd/system/kong.service` for {{site.ce_product_name}}.
 
+## Add a sample systemd service file for reference
```
lena-larionova, line 84:
```suggestion
Note that the official Kong systemd service doesn't set `user` or `group` as `root`, which is required to run the system file. See the example below with both `user` and `group` both set to `root`:
```

lena-larionova (changes requested):
Mostly looks good, but I'd add an explanation of why we're providing this sample here.

lena-larionova (approved):
LGTM, thanks for adding this example. I changed the header to an explanation instead though, this way the user knows why this is relevant.

PR #5677, opened 2023-06-06
[DOCU-3157] Portal Client API SDK

cloudjumpercat:
### Description

What did you change and why?
- I added a link to the self-hosted portal SDK in the sidebar. 
- ❗ Reviewers: I considered other locations like "Reference" (but this is more explanation topics/long form) and "API Reference" (but these are APIs, not SDK, so it would need to be renamed) and stuck with this one because it keeps it close to the content that it relates to. What do you think? 

Include any supporting resources, e.g. link to a Jira ticket, GH issue, FTI, Slack, Aha, etc.
DOCU-3157

### Testing instructions

Netlify link: 

### Checklist 

- [x] Review label added 


Guaris:
I think I would make this a top-level item within Dev Portal above troubleshooting.  Why are we calling this "Self-hosted portal SDK" Instead of Dev Portal SDK?

Guaris:
> > I think I would make this a top-level item within Dev Portal above troubleshooting. Why are we calling this "Self-hosted portal SDK" Instead of Dev Portal SDK?
> 
> @Guaris That's a good point. I do think it technically applies to all of Dev Portal, but you'd only use the SDK if you were self-hosting (since the only other option is SaaS hosted, so you could only modify it through the Konnect UI, right?).

I read the brief and it seems like we use `self-hosted` everywhere -- but in the repo https://github.com/kong/sdk-portal-js we don't, or in the spec https://developer.konghq.com/spec/2aad2bcb-8d82-43b3-abdd-1d5e6e84dbd6/b4539157-4ced-4df5-affa-7d790baee356#/applications/list-applications or on the NPM package page https://www.npmjs.com/package/@kong/sdk-portal-js

I don't feel too strongly about it either way, but it feels implied. Maybe leave it as `self-hosted` and see if anyone else feels the way I do about it once it's live? 

I do feel like this should be a top-level item in Dev Portal -- some people may not care about customization unless they know there is an SDK.

Guaris (approved):

PR #5690, opened 2023-06-09
Auto SS 

Guaris:
- Adds screenshot scripts for
   - Konnect dashboard
   - runtime manager dashboard
   - servicehub dashboard
   - Service package dashboard
 - Adds a zoom function to macros.yaml

### DIFF
--- a/tools/screenshots/overview/konnect-dashboard.yaml
+++ b/tools/screenshots/overview/konnect-dashboard.yaml
@@ -1,0 +1,7 @@
+- output: ../../app/_assets/images/docs/konnect/konnect-dashboard2.png
+  url: https://cloud.konghq.com/us/overview/
+  wait: 4000
+  width: 1920
+  height: 1080
+  javascript: |
+    zoomPercentage(-5);
--- a/tools/screenshots/macros.yaml
+++ b/tools/screenshots/macros.yaml
@@ -70,3 +70,11 @@
       node.style.position = 'unset';
       body.appendChild(num);
     }
+- type: function
+  name: zoomPercentage
+  javascript: |
+    function (percentage) {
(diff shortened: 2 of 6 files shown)


tools/screenshots/overview/konnect-dashboard.yaml
```diff
@@ -0,0 +1,7 @@
+- output: ../../app/_assets/images/docs/konnect/konnect-dashboard2.png
```
cloudjumpercat, line 1: @Guaris Where can you see this one (like the URL of the page it's on, I looked at the main Overview page, but didn't see it there)?

cloudjumpercat (commented):
I couldn't find the Konnect dashboard screenshot in the preview URL, but the other ones looked good to me!

I'm not sure what will happen with the Service Hub ones since that page will be overhauled/changed for API Products, but I think it will be good to have them at least in our files in preparation for Service Hub 2.0.

tools/screenshots/overview/konnect-dashboard.yaml
Guaris: Ah, you caught a mistake here. I didn't add the image but I forgot to rename this file back to dashboard2.png

Guaris, line 1:
```suggestion
- output: ../../app/_assets/images/docs/konnect/konnect-dashboard.png
```

Guaris:
> I couldn't find the Konnect dashboard screenshot in the preview URL, but the other ones looked good to me!
> 
> I'm not sure what will happen with the Service Hub ones since that page will be overhauled/changed for API Products, but I think it will be good to have them at least in our files in preparation for Service Hub 2.0.

Hopefully the script is built in such a way that nothing happens? Like once the change goes through we run the script and it takes a screenshot of the latest UI and we can make a PR with that.

PR #5692, opened 2023-06-09
Added Access phase to Response as seen in Konnect API

d19dotca:
### Description

Support case had been filed pointing out that the Konnect API response output in the docs were missing the "access" phase part of the Ordering line. We confirmed in a test that it is indeed in the responses whenever Dynamic Ordering is used on any plugin in the Konnect environment.

### Testing instructions

N/A

### Checklist 

- [Y] Review label added 
- [Y] PR pointed to correct branch (`main` for immediate publishing, or a release branch: e.g. `release/gateway-3.2`, `release/deck-1.17`)


lena-larionova (approved):
Fixed it in a couple of other files too. LGTM!

PR #5695, opened 2023-06-12
Replace gulp with vitejs

fabianrbz:
### Description

What did you change and why?

Replace gulp with vite.js which is the company standard and will make it easier to add vue.js and render the OAS specs in the docs.
A few things changed:
* static assets (images, manifests, keys) now live in `app/assets` - handled via `Jekyll`
* javascripts, fonts,  and stylesheets live in `app/_assets` - handled via `vite.js`, bundled and fingerprinted in production mode
* `make run` runs `vite` + `jekyll build` and netlify's redirect
* as a side effect we might be able to get rid of some of the security warnings

### Testing instructions

Netlify link: 

We need to make sure that the styles and js work as expected.
Some things to note:
* some things will fail due to CORS issues, e.g. tracking requests, etc.
* gtm and recaptcha are not loaded in preview apps, we previously did but I don't think that's necessary

### Checklist 

- [x] Review label added 


exe/build
```diff
@@ -0,0 +1,4 @@
+#!/usr/bin/env bash
+
+set -e
+JEKYLL_ENV=production bundle exec jekyll build --config jekyll.yml --profile
```
mheap, line 4: This will run in production mode locally (and load analytics etc). Should we inherit `JEKYLL_ENV` from the environment?

mheap: It would also result in preview apps being indexed unexpectedly

fabianrbz: 🤦 good catch! yeah, it should inherit it from the environment.

fabianrbz:
🤦 It's an unrelated issue, fix is [up](https://github.com/Kong/docs.konghq.com/pull/5810)

cloudjumpercat (approved):
LGTM! I tested the docs locally and `make install` and `make run` worked for me. I checked the fonts and images on several pages and all of them were rendering as expected.

Guaris:
As far as I can tell everything is working like normal.

lena-larionova (approved):
LGTM as well, everything seems to work as expected.

This is my favourite part: **+4,150 −28,850** 😁

mheap:
It looks like incremental build isn't working on this branch. Any time I change a file it takes a couple of minutes then vite reports a page reload for (what looks like) every page on the site

mheap:
I edited `app/_src/gateway/install/linux/ubuntu.md`. Let me try a full clean + rebuild

mheap:
Full clean + rebuild worked as expected. Page rebuilt in 15s. Something strange on my machine I guess.

mheap:
:shipit:

mheap (approved):

fabianrbz:
It could be that the cache was invalid, probably built using a different branch and Jekyll/vite didn't handle it well 🤷

PR #5706, opened 2023-06-14
docs(cli): db_export now supports Kong Gateway enterprise version

outsinre:
### Description

`db_export` now supports Kong Gateway enterprise version (at least since `3.3.0.0`).

The note on the official page is no longer valid. So, remove it.

image

### Checklist 

- [x] Review label added 

### DIFF
--- a/app/_src/gateway/reference/cli.md
+++ b/app/_src/gateway/reference/cli.md
@@ -69,10 +69,6 @@ Options:
 
 ```
 
-{:.note}
-> **Note:** `db_export` is only supported with open-source
-{{site.base_gateway}} packages.
-
 ---
 
 


lena-larionova:
🤔 I can't find any reference to changes to this command in the past few years at all. I think we're probably safe removing this note from at least all 3.x versions.

RobSerafini:
My recommendation is to leave the notice. Customers tended to use this feature to do declarative config management which is not the intended use case for this command. If we leave it as unsupported we will avoid getting a bunch of bug reports about how it does not correctly update state when doing a db_import on a running kong instance.

outsinre:
@RobSerafini , get it. We leave this as a hidden feature.

I use a lot of local dev environment to initialize my local database with services/routes/plugins.

So, I will close this ticket.

PR #5713, opened 2023-06-14
chore: fix API parameter in wrong version

Guaris:
fix issue with instance_name showing up in 3.1

This is fixed by adding a new API markdown file and referencing it from the 3.1 nav bar.


acgoldsmith (approved):
LGTM

PR #5734, opened 2023-06-20
WIP: Inline plugin example

fabianrbz:
### Description

[JIRA ticket](https://konghq.atlassian.net/browse/DOCU-3259)

What did you change and why?
 
First stab at rendering opinionated plugin examples.
The following snippet renders the same navtabs we use in the plugin hub, but only for the provided targets and formats.
A few things to consider:
* `plugin` must be a combination of `<vendor>/<plugin_name>`
* `example` references the name of a folder under `app/_hub/<vendor>/<plugin_name>/<example>/`. The `plugin_example` is single-sourced, so the folder could contain an `_index.yml` used for all the versions and files for specific versions, e.g. `_2.3.x.yml`.
* The way single-sourcing works, in this case, is by using the page's `kong-version` in which the `plugin_example` is rendered.
* The name of the `targets` and `formats` is TBD. Thoughts?
* The targets are not validated against the schemas but only the ones that are valid will be rendered. This is due to the fact that some plugins can support different targets across different versions.
* Should we raise if the example file does not exist? 

```
{% plugin_example %}
  name: rate-limit-example
  plugin: kong-inc/rate-limiting
  example: complex-example
  targets:
    - service
    - route
    - global
  formats:
    - curl
    - kubernetes
{% endplugin_example %}
```

### Testing instructions

I added a dummy example so we can see how it works in the following link `/gateway/latest/

Netlify link: 

### Checklist 

- [ ] Review label added 


fabianrbz:
TODO: remove dummy example from `app/_src/gateway/index.md`

lena-larionova (approved):
LGTM - leaving to you to do any cleanup and merge when needed.

PR #5761, opened 2023-06-28
[DOCU-3281] Moving plugin descriptions into markdown

lena-larionova:
### Description

Moved all plugin `description` and `extra` content from the `_metadata.yml` files into each plugin's `_index.md` files. 

Why?
* Easier to version-control, since it gives us access to `if_plugin_version` tags
* Easier for any contributor to edit. Before, clicking the "edit this page" button would take you to a partial page with no sign of where the description came from. Depending on the plugin, some of those descriptions could be one sentence, but others are full write-ups with subsections and steps. None of that belongs in metadata.
* `_metadata.yml` still contains the `desc` field, which covers that short description. If the UI team still needs to scrape the docs for UI tooltips for plugins, that's the field that should be used, as it's max 80 characters.
* `extra` was an old field left over from before the config migration into schemas. It never should've existed, because all of that info was generated after the schema tables anyway, and wasn't in context. I pulled that into places where it makes more sense now.

### Testing instructions

https://deploy-preview-5761--kongdocs.netlify.app/hub/ - spot-check some plugins. 
I flipped through every plugin to double-check that everything still displayed correctly, descriptions should all be there.

### Checklist 

- [x] Review label added 


lena-larionova:
No broken links, only 429 errors from GitHub rate limiting.

acgoldsmith:
@lena-larionova found a bug in breadcrumbs: 
* on any basic config examples page, click the preceding breadcrumb link (ie to go up one level)
* hits 404 b/c is trying to go from /how-to/basic-example/ back to /how-to/

acgoldsmith (approved):
LGTM and thanks for clarifying the known issue

lena-larionova:
Will look into the breadcrumb issue in a followup, it's not related to this PR.

PR #5771, opened 2023-07-03
Update _metadata.yml

mod42:
added some more search aliases


cloudjumpercat (approved):
LGTM! Tried the aliases out and they all worked. Thanks for adding those!

PR #5772, opened 2023-07-04
fix(kic): correct curl --resolve option

steveatkonghq:
### Description

The [`--resolve`](https://curl.se/docs/manpage.html#--resolve) option for `curl` only accepts an argument in the form `{SOURCE IP}:{SOURCE PORT}:{TARGET IP}`; a target port cannot be specified. The option [`--connect-to`](https://curl.se/docs/manpage.html#--connect-to) should be used instead as this can also accept a target port.

Additionally, if users have followed the KIC deployment instructions for minikube, the `PROXY_IP` variable will contain a URL including a scheme (e.g. `http://127.0.0.1:12345`); the scheme should be stripped from the variable before being passed to `curl` to avoid a parsing error.

### Testing instructions

Netlify link: 
https://deploy-preview-5772--kongdocs.netlify.app/

### Checklist 

- [x] Review label added 

### DIFF
--- a/app/_includes/md/kic/http-test-routing.md
+++ b/app/_includes/md/kic/http-test-routing.md
@@ -12,7 +12,7 @@ Create routing configuration to proxy `{{ path }}` requests to the echo server:
 Test the routing rule:
 
 ```bash
-curl -i http://{{ hostname }}{{ path }} --resolve {{ hostname }}:80:$PROXY_IP
+curl -i http://{{ hostname }}{{ path }} --connect-to kong.example:80:${PROXY_IP##http://}
 ```
 Response:
 ```text
--- a/app/_src/kubernetes-ingress-controller/guides/configure-acl-plugin.md
+++ b/app/_src/kubernetes-ingress-controller/guides/configure-acl-plugin.md
@@ -55,7 +55,7 @@ right ACL.
 Requests without credentials are now rejected:
 
 ```bash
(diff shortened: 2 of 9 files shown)


Rajakavitha1:
@mheap could you please let me know your preference between  `--connect-to` and `--resolve` option with curl.

steveatkonghq:
> Hi @steveatkonghq . I noticed that the issue you mentioned about Minikube deployment has been addressed in #5844. Before I validate the commands in the rest of the guides please let me know if they would still hold good.

@Rajakavitha1 Thanks, I'm happy that this PR has been addressed in #5844. I'm in favour of sticking with `--resolve` providing that the documentation is using this correctly, as it's in more common usage and more closely aligned to the intent of the command (overriding a local DNS resolution).

Rajakavitha1:
> > Hi @steveatkonghq . I noticed that the issue you mentioned about Minikube deployment has been addressed in #5844. Before I validate the commands in the rest of the guides please let me know if they would still hold good.
> 
> @Rajakavitha1 Thanks, I'm happy that this PR has been addressed in #5844. FWIW I'm in favour of sticking with `--resolve` providing that the documentation is using this correctly, as it's in more common usage and more closely aligned to the intent of the command (overriding a local DNS resolution).

Thank you @steveatkonghq  can we close this PR and yes, we will ensure that the usage is consistent across the documentation.

Rajakavitha1:
Closing this PR for now. Please feel free to reopen if any other issue needs to be addressed.

PR #5778, opened 2023-07-06
Fix performance issue caused by plugin pages without `source_file`

fabianrbz:
### Description

What did you change and why?
 
We use the `source_file` as the page's `relative_path` which is used internally by Jekyll to konw which pages should be re-generated whenver a file changes.

Configuration and plugin hub example pages were setting it to `nil` causing all those pages to be re-generated whenever any file changed. This fixes the issue by setting the right `source_file` to those pages.

This only affects local build times.

### Testing instructions

Netlify link: 

### Checklist 

- [x] Review label added 

### DIFF
--- a/app/_plugins/generators/plugin_single_source/pages/configuration.rb
+++ b/app/_plugins/generators/plugin_single_source/pages/configuration.rb
@@ -25,7 +25,9 @@ module PluginSingleSource
         @dropdown_url ||= "#{base_url}VERSION/configuration/"
       end
 
-      def source_file; end
+      def source_file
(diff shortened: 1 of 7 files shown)


lena-larionova (approved):
LGTM, though build time didn't seem to go down for me 🤔

lena-larionova:
> That's weird, although it is true that a complete build doesn't benefit a lot from it there should be an improvement. It's more straightforward for edits, e.g. (after editing app/_src/gateway/_index.md)

Ah! I was running a complete build. Yep, much better for rebuilds when already running. Nice work 🎉

PR #5830, opened 2023-07-20
Update enable-rbac.md

raylaijh:
negative=true flag should be indicated for the curl command for these examples here as negative=false is set by default without specific indication

### Description

--data negative=true flag was left out in this example

### Testing instructions

Netlify link: 

### Checklist 

- [x] Review label added 


lena-larionova (approved):
Nice catch! LGTM

PR #5878, opened 2023-08-01
Set `kong_version` to hub pages.

fabianrbz:
### Description

What did you change and why?

Inline plugin examples rely on pages having a `kong_version` set so that they can select the right version of the plugin's schema.

### Testing instructions

Netlify link: 

### Checklist 

- [ ] Review label added 


lena-larionova (approved):
LGTM, this actually solves two problems - we can now use the `{{page.version}}` variable in plugins to link to Gateway docs, if we wanted to. 

Tested locally with opinionated examples, everything works!

PR #5880, opened 2023-08-01
feat: Guide for using the HTTP Log plugin with Splunk

lena-larionova:
### Description

There's two related changes in this PR:
* New guide for using the HTTP Log plugin with Splunk, based on a [KB article](https://support.konghq.com/support/s/article/how-to-send-kong-log-to-splunk-by-using-HTTP-Log-plugin). 
  * Tested instructions in article, they only apply to 2.8 and earlier - this is why there are two examples in the guide, depending on the version.
* Split out content from the HTTP Log landing page into a how-to section + some cleanup.
  * Custom fields by Lua applies to 2.4.x onward, but we don't currently have a clean way of specifying a minimum version per file - that's why there's a note in earlier versions saying this feature isn't available. 

### Testing instructions

New guide:
https://deploy-preview-5880--kongdocs.netlify.app/hub/kong-inc/http-log/how-to/splunk/

Changes to existing content:
https://deploy-preview-5880--kongdocs.netlify.app/hub/kong-inc/http-log/
https://deploy-preview-5880--kongdocs.netlify.app/hub/kong-inc/http-log/log-format/
https://deploy-preview-5880--kongdocs.netlify.app/hub/kong-inc/http-log/using-custom-fields/

### Checklist 

- [x] Review label added 


app/_hub/kong-inc/http-log/_index.md
```diff
@@ -1,7 +1,11 @@
-HTTP Log plugin lets you send request and response logs to an HTTP server.
+The HTTP Log plugin lets you send request and response logs to an HTTP server.
 
-{% if_version gte:3.3.x %}
-## Queueing
+It also supports stream data (TCP, TLS, and UDP).
+
+## Features
```
lena-larionova, line 5: If anyone has a better title to group this info, I'm very open to it.

mheap:
CleanShot 2023-08-11 at 13 32 23

None of the following are blocking.

* Do we still need (1) if we have a concrete Splunk example? I'd expect the default example to disappear when we start adding opinionated guides
* Should (2) and (3) be under the `Overview` section? They're more "How this plugin works" than "How to use this plugin"?

lena-larionova:
> Do we still need (1) if we have a concrete Splunk example? I'd expect the default example to disappear when we start adding opinionated guides

Two concerns with that idea:
* This would create an inconsistent experience, if we just remove it for random plugins.
* If we remove that and leave only the splunk example, we're leading them to configure the plugin for splunk and nothing else. Is that really what we want to do? I'd be more comfortable removing if we had a generic "Get started with HTTP Log" guide that contained an example as part of the flow.

> Should (2) and (3) be under the Overview section? They're more "How this plugin works" than "How to use this plugin"?

Are you thinking nested as individual topics, or on the overview page itself?

app/_includes/md/plugins-hub/kong-process-errors.md
```diff
@@ -1,9 +1,12 @@
 <!---shared with logging plugins --->
 
-This logging plugin logs HTTP request and response data, and also supports streams
+This logging plugin logs HTTP request and response data, and also supports stream
 data (TCP, TLS, and UDP).
 
-If you are looking for the Kong process error file (which is the nginx error file),
+If you are looking for the Kong process error file (which is the Nginx error file),
```
Guaris, line 6:
```suggestion
The Kong process error file is synonymous with the Nginx error file and you can find it at the following path:
```

Guaris: Not the best, but I tried to reword this a bit.

Guaris (changes requested):

app/_hub/kong-inc/http-log/how-to/_splunk.md
```diff
@@ -0,0 +1,77 @@
...
+
+You can use the HTTP Log plugin to send {{site.base_gateway}} logs to Splunk.
+
+{:.note}
+> **Note**: The following example uses Splunk 9.0.2. If you are using a different version of Splunk,
+check the Splunk documentation for the appropriate method.
+
+## Prerequisites
+
+You have a Splunk authorization token.
```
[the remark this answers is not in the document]
lena-larionova: I disagree with this, since shortening it this way would make it inconsistent with the rest of our docs. If you think we should be dropping the "you have xyz" or "your X is set up in Y way" etc, we should bring that up as a style guide topic.

app/_includes/md/plugins-hub/kong-process-errors.md
lena-larionova: I think it's even more straightforward than that: Kong uses the Nginx error file as its process error file.
Our own config ref says this about log level settings:

> Log level of the Nginx server. Logs are found at `<prefix>/logs/error.log`.
> See http://nginx.org/en/docs/ngx_core_module.html#error_log for a list of
accepted values.

mheap:
> If we remove that and leave only the splunk example, we're leading them to configure the plugin for splunk and nothing else. Is that really what we want to do? I'd be more comfortable removing if we had a generic "Get started with HTTP Log" guide that contained an example as part of the flow.

Agree. I meant to say add a generic "send to a webhook" example in addition to Splunk. The generated examples are OK, but if we're adding examples with more effort the first one we should do is a generic guide imo.

> Are you thinking nested as individual topics, or on the overview page itself?

I was thinking individual topics + _Overview_ becomes a folder

lena-larionova:
> Agree. I meant to say add a generic "send to a webhook" example in addition to Splunk. The generated examples are OK, but if we're adding examples with more effort the first one we should do is a generic guide imo.

That may be true, but I don't have a generic guide handy, and I do have a Splunk guide. If this is a prerequisite, then do we hold off adding any specific examples until each plugin has a generic long-form example? That would put us further back than we already are.

Guaris (approved):

PR #5895, opened 2023-08-04
Chore: Fix vale, Update Github action

Guaris:
Updated Github action to use Reviewdog

```
steps:
      - uses: actions/checkout@v3
      - uses: errata-ai/vale-action@reviewdog
```

image

Vale is failing, I suspect it has something to do with pinning the version, removing the parameter `-fail-on-error` makes this work while still preserving the error reporting ability: 
image

However, this makes the PR pass even with errors reported 

image


fabianrbz (approved):
LGMT 🎉

PR #5903, opened 2023-08-07
Update plugins-kubernetes.md: Specify alpine so docker does not defau…

bharkreader:
…lt to ubuntu

### Description

Dockerfile did not work as-is because the commands relied on the alpine version, but docker downloaded ubuntu by default for me.


lena-larionova:
As of our latest release yesterday (3.4.0), we have actually deprecated Alpine, and are no longer building any more Alpine packages.
Building on Ubuntu is desired, in this case. You can manually set Alpine if you want to pick up older versions, but we won't be defaulting to it.

PR #5909, opened 2023-08-08
fix(konnect-sso): both auth types are allowed

kinman:
Administrators can allow login via Konnect built-in methods (username and password) or via their Idp. These are not mutually exclusive.

### Description

Removed the statement that only one auth type is allowed, because that is no longer true.

### Testing instructions

Netlify link: 

### Checklist 

- [ ] Review label added 


lena-larionova (approved):
LGTM!

PR #5924, opened 2023-08-09
Rework KIC version compatibility into a template sourced from YAML

rainest:
### Description

Sources KIC version compatibility with a table similar to gateway's https://docs.konghq.com/gateway/3.3.x/support/third-party/ page, but simpler.

Editing the current raw Markdown table is painful, and we'll soon run out of old versions we can discard (which makes it less painful) without omitting some of the versions covered by the single source page.

### Checklist 

- [ ] Review label added 


mheap:
I took a look at this as I love the idea of these pages being data driven, but I don't think we can get the same look and feel. 

I had a go at formatting without the version tabs so that everything's shown on a single page per KIC version, but it didn't work out (and I forgot to commit 🤦 )

On balance, I think the cross/check UI is better for consumers even if it's more painful for us to maintain. If maintenance is too painful, we should look at investing in tooling to generate the same view from data

## Current

CleanShot 2023-09-27 at 12 48 20

## Travis

CleanShot 2023-09-27 at 12 48 50

mheap:
I didn't lose it! I just did in on #5989 

Moving my above comment there and closing this PR as if we move forward we want the most automation

PR #5934, opened 2023-08-11
[DOCU 3362/3349] homepage refresh

fabianrbz:
### Description

What did you change and why?
 
Related [Jira ticket](https://konghq.atlassian.net/browse/DOCU-3349)

WIP. Copy needed

### Testing instructions

Netlify link: 

### Checklist 

- [x] Review label added 


acgoldsmith (commented):
@fabianrbz, looks good! Is it possible to style the feature icons so they're the same blue as the banner? it looks like we're doing some modification with them in the left menu in the Konnect docs; color there is what I'd like to replicate on homepage use if possible.

lena-larionova:
I think the icons need to be smaller + having more whitespace around them. They're not really designed to be that big in relation to the other elements. How about something like this:

Screenshot 2023-08-14 at 9 21 48 AM

Guaris:
Agree with the icons being too large, but I also think there is something off about the highlighting would it make sense for the highlighting to reach all the way to the edges of the container?

image

Can these align some how?
image

fabianrbz:
@Guaris I'll align the items to the left.
The highlighting is the same one we have in prod right now, the cards have padding so links/buttons stretch to the container's width minus the padding.

fabianrbz:
@acgoldsmith the icon's color in konnect is gray, and the sidebar is blue, like this

Screenshot 2023-08-15 at 09 38 59

we can make the icons have the same gray color or a similar blue to the one the banner has (it's a gradient so it technically isn't just one color).

Another option would be to add a blue banner to the cards and make the icons gray, something like this
Screenshot 2023-08-15 at 10 25 18

fabianrbz:
Gray icon example
Screenshot 2023-08-15 at 11 10 36

acgoldsmith:
@fabianrbz, thanks for suggestion, I'd meant the docs nav: 
Screenshot 2023-08-14 at 9 21 28 AM

The styling there is what I'm hoping we can recreate on the homepage

fabianrbz:
I've updated the icons

app/_data/landing_page.yml
```diff
@@ -3,98 +3,106 @@ features:
     items:
```
cloudjumpercat, line 3: @fabianrbz I think the "Key features" heading text is kind of small in comparison to the heading text on a tile ("Service Hub") for example. I think maybe the "Key features" or section title header text should be larger than the heading of a tile.

cloudjumpercat:
I'm not sure where it's located in the homepage file, but the gorilla graphic below the Plugin Hub tile looks a bit odd since it's not centered. I'd probably be in favor of just removing it?

lena-larionova:
> I'm not sure where it's located in the homepage file, but the gorilla graphic below the Plugin Hub tile looks a bit odd since it's not centered. I'd probably be in favor of just removing it?

+1 on this.

lena-larionova:
I think this one is my fault. The header text for the Kong Gateway and Kong Mesh tiles isn't styled the same way as the header text for Kong Konnect (different colour, grey vs dark blue). We should make them all dark blue.

One more thing: it feels like the swirly headers for Konnect, Gateway, and Mesh don't really go with anything anymore. Anyone have alternative ideas? I realize that just removing them will make the cards probably look really empty.

fabianrbz:
Yeah, here's what it looks like without them

Screenshot 2023-08-16 at 09 14 23

And different background colors from the design system:
Screenshot 2023-08-16 at 09 22 03
Screenshot 2023-08-16 at 09 23 41


Unfortunately, the [card compontent](https://www.figma.com/file/Yze0SWXl5nKjR0rFdilljK/Components?type=design&node-id=1152-19831&mode=design&t=upB54hCFfjIiQySY-0) is still being developed 🚧 and it's currently empty :/

lena-larionova (approved):
LGTM!

PR #5946, opened 2023-08-14
Konnect support for Gateway 3.4

Guaris:
This PR skips changelog. 

Adds a changelog note that support for Konnect 3.4 and all compatible plugins is in. 

https://konghq.atlassian.net/browse/DOCU-3333


acgoldsmith (approved):
LGTM

PR #5949, opened 2023-08-15
Fix Ubuntu 3.4 package install instructions

mheap:
We no longer build Ubuntu `bionic` packages for Kong Gateway 3.4+.

Update the conditional rendering to use Ubuntu `jammy` for all releases except 3.0, which provides a `bionic` package but no `jammy` package


fabianrbz (approved):
LGTM

PR #5953, opened 2023-08-16
Rename: Konnect introduction section

Guaris:
This PR skips the changelog. 

Renames the elements in the introduction section of the Konnect docs. 

```
Runtime Manager to Gateway Manager
Runtime Groups to Control Plane
Composite Runtime Groups to Control Plane Group
Runtime Instances to Data Plane Nodes
```

https://konghq.atlassian.net/browse/DOCU-3369

PR #5954, opened 2023-08-16
Change directories in Konnect section as a foundation for renaming

Guaris:
Cherry-picked the redirects from #5953, merging them to the branch separately


app/_redirects
```diff
@@ -3,6 +3,14 @@
 # More general/sweeping redirects should be at the bottom.
 
 
+
+
+# Konnect rename
+
+/konnect/runtime-manager/* /konnect/gateway-manager/:splat
+/konnect/runtime-manager/runtime-instances/* /konnect/gateway-manager/data-plane-nodes/:splat
```
cloudjumpercat, line 11: Not sure if this one is working correctly. I tried `/konnect/runtime-manager/runtime-instances/renew-certificates/` and `/konnect/runtime-manager/runtime-instances/upgrade/` and they both resulted in 404 instead of redirecting to `data-plane-nodes`. The `gateway-manager` part redirected correctly though, just the `data-plane-nodes` part didn't.

app/_redirects
```diff
@@ -3,6 +3,14 @@
...
+/konnect/runtime-manager/composite-runtime-groups/*     /konnect/gateway-manager/control-plane-groups/:splat
```
cloudjumpercat, line 12: Same issue with this one as the one above. `gateway-manager` part redirects, but it doesn't change to `control-plane-groups` as well, so there's a 404.

app/_redirects
lena-larionova: ^ These two redirects are in the wrong order. A more specific redirect should come before a general redirect. So 
`/konnect/runtime-manager/*` should be the last entry in this list.

cloudjumpercat (approved):
LGTM! Redirects are working now

lena-larionova (approved):

PR #5956, opened 2023-08-16
chore: Remove support for editable codeblocks

lena-larionova:
### Description

Editable codeblocks were an attempt at a feature for the docs site, where we made it possible to edit variables in line with the text. However, after implementing the feature, we ran into issues:
* Can't use it for anything that might be considered personal information, such as hostnames and passwords
* This creates inconsistency - a codeblock might have some values as editable and others not, which was misleading
* The implementation wasn't great. It required us to use HTML tags, and didn't ignore spaces. This made it difficult to maintain.

We already cleared out most of these during our migration to single-sourced content in 3.0. This PR fully clears out support and all of the remains of this functionality from the docs.

### Testing instructions

Netlify link: 

### Checklist 

- [ ] Review label added 


app/contributing/markdown-rules.md
```diff
@@ -444,18 +375,22 @@ you can use the indent filter on a capture of your tabs.
 {% capture the_code %}
 {% navtabs codeblock %}
 {% navtab cURL %}
-<div class="copy-code-snippet"><pre><code>curl -i -X POST http://<div contenteditable="true">{HOSTNAME}</div>:8001/event-hooks \
+```sh
```
fabianrbz, line 378: something is off with these navtabs, they are not being rendered right

fabianrbz: I found out what the problem is
now that each tab has a code block in it, the parser doesn't know where they start/end.
we need to add another ` to the `{% raw %}` block, [here](https://github.com/Kong/docs.konghq.com/pull/5956/files#diff-531b7ce7eac5adee79702a5d9130a418ef63b625d41e3134e4e83837b8669648L443) and [here](https://github.com/Kong/docs.konghq.com/pull/5956/files#diff-531b7ce7eac5adee79702a5d9130a418ef63b625d41e3134e4e83837b8669648L463) so that the parser knows that it needs to treat everything inside it as different code blocks

fabianrbz (changes requested):
This is awesome 🎉 :thanks:

I left a comment in one of the files because the content wasn't being rendered properly

app/contributing/markdown-rules.md
lena-larionova: Fixed!

fabianrbz (approved):

PR #5962, opened 2023-08-17
Fix: Rendering issue on clusering reference page.

Guaris:
This is not on the changelog

User reported issues with the rendering in the `/gateway/3.0.x/production/clustering/` doc. The problem was that the headings used code tags and relative linking. Removed that in place of direct linking at the bottom of each section. Made a more info section at the bottom as well.


lena-larionova (approved):
lgtm

PR #5969, opened 2023-08-17
Changelog: specify which version to write using github labels. 

Guaris:
**I don't think this is the right solution for the problem**

This PR adds a function `getVersionFromLabels` that extracts version labels from PRs and uses it to write specific versions to changelog.md. 

Where to go from here: 
* Descending order
* Add a function to check for duplicates. 
* Discussed at end of hackathon: Only print the latest version or potentially latest + lts only.

### DIFF
--- a/tools/changelog-generator/changelog.js
+++ b/tools/changelog-generator/changelog.js
@@ -4,7 +4,7 @@ const groupBy = require("lodash.groupby");
 
 const convertFilePathsToUrls = require("../_utilities/path-to-url");
 const now = startOfDay(new Date());
-const earliestDate = subDays(now, 4);
+const earliestDate = subDays(now, 7);
 
 (async function () {
   // Create an octokit instance
--- a/tools/changelog-generator/changelog.js
+++ b/tools/changelog-generator/changelog.js
@@ -4,7 +4,7 @@ const groupBy = require("lodash.groupby");
 
 const convertFilePathsToUrls = require("../_utilities/path-to-url");
 const now = startOfDay(new Date());
-const earliestDate = subDays(now, 7);
+const earliestDate = subDays(now, 4);
 
 (async function () {
   // Create an octokit instance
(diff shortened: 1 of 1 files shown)


tools/changelog-generator/changelog.js
```diff
@@ -4,7 +4,7 @@ const groupBy = require("lodash.groupby");
 
 const convertFilePathsToUrls = require("../_utilities/path-to-url");
 const now = startOfDay(new Date());
-const earliestDate = subDays(now, 7);
+const earliestDate = subDays(now, 4);
```
Guaris, line 7:
```suggestion
const earliestDate = subDays(now, 7);
```

mheap:
> I don't think this is the right solution for the problem

Tell us more! What doesn’t feel right, and what other ideas do you have?

lena-larionova:
> Tell us more! What doesn’t feel right, and what other ideas do you have?

Based on our team chat in one of our syncs today, here's the summary (while it's fresh, and since it's past Angel's end of day):
* A PR might have a mix of files that apply to one version or another. If you apply versions 3.3.x and 3.4.x as labels to a PR, how does the generator know which file to generate as 3.3.x and which as 3.4.x? 
* If we have one label per version, we're going to have a _lot_ of labels. 
  * Gateway alone has 6 currently maintained versions, plus occasional backports/fixes that require editing file before 2.8.
  * KIC has a massive list of versions
  * There might be overlap between versions of different products, so then labels would need to be longer, eg gw-28.x instead of just 2.8.x.
* Our audience is _mostly_ ourselves. We know how to choose a different doc version. If we just link to latest instead of linking to specific versions, plus optionally 2.8.x as the LTS, we cover 99% of situations. 
  * Going along the same lines, it's easier if we just see /latest/ in most situations, and we don't have to wade through all the versions to find the most important/relevant one.

I _think_ I got most of it, but @Guaris can expand tomorrow.

Guaris:
Lena captured all of my thoughts. Overnight I was wondering if this functionality would be good to keep for use cases we haven't uncovered yet. Currently, if there is no label it writes every link, but maybe this is a step towards a better auto-generated changelog in the future.

Guaris:
Closing to reopen later

PR #5970, opened 2023-08-17
feat: Konnect API examples for plugins

lena-larionova:
### Description

For any plugin that is supported in Konnect, it’s now possible to configure it using the Konnect runtime groups API, via `core-entities`. This PR adds support for generated Konnect API examples in the plugin docs, which makes the Konnect API more visible and positions it as a parallel option to the Kong Admin API.

A simple service example for the Basic auth plugin:
Screenshot 2023-08-17 at 1 54 20 PM

A global example for the Rate Limiting Advanced plugin:
Screenshot 2023-08-17 at 2 57 19 PM

Along with that, there were several fixes/improvements that I had to make:
* Placeholders in curl examples: after team discussion, settled on `{ }` as the placeholder characters for URLs. This format is heavily used in our API specs, and is the OpenAPI standard: https://swagger.io/docs/specification/describing-parameters/. Going forward, we should be using `{ }` to enclose variables **in specs and API examples**. 
  * Placeholders in other situations still need revisiting. Out of scope here.
 
* The global example now explains what "global" means in the context of Konnect, Kong Enterprise, and Kong OSS. This is a long-running painpoint, and it seems timely to add this as we add Konnect examples.

* Removed the duplicate `konnect_examples` and `manager_examples` metadata keys from plugin docs. We don't have Kong Manager examples, and we don't need a separate `konnect_examples` key. `konnect: false/true` already performs the same function.

https://konghq.atlassian.net/browse/DOCU-2949

### Testing instructions

Check any plugin doc. For example:
* [Rate Limiting Advanced plugin](https://deploy-preview-5970--kongdocs.netlify.app/hub/kong-inc/rate-limiting-advanced/how-to/basic-example/) - has all example types, good way to see everything in one place
* [OAuth2 plugin](https://deploy-preview-5970--kongdocs.netlify.app/hub/kong-inc/oauth2/how-to/basic-example/) - isn't supported in Konnect so the Konnect API tab doesn't show up
* [Update to style guide](https://deploy-preview-5970--kongdocs.netlify.app/contributing/style-guide/#placeholder-and-example-values)

### Checklist 

- [x] Review label added 


fabianrbz (approved):
LGTM 🎉 
I ❤️ the new placeholders!

I left a small comment about a typo, but other than that it looks great!

app/_plugins/drops/plugins/examples/konnect.rb
```diff
@@ -0,0 +1,48 @@
+# frozen_string_literal: true
+
+module Jekyll
+  module Drops
+    module Plugins
+      module Examples
+        class Konnect < Base
+          URLS = {
+            'consumer' => 'https://global.api.konghq.com/v2/{RUNTIME-GROUP-ID}/core-entities/consumers/{CONSUMER_NAME|ID}/plugins',
```
mheap, line 9: The global endpoint won’t work here. It needs to be us.api or eu.api

lena-larionova: Is that just for consumers, or consumer groups as well?

mheap: For all APIs that aren't kauth related

lena-larionova:
Finally got around to testing this and realized that the Konnect API doesn't accept the same format as the Kong Admin API. It has to be json data, eg:

```
  curl -i -X POST \
  https://us.api.konghq.com/v2/runtime-groups/{runtime-group-id}/core-entities/services/{service-id}/plugins \
  --header 'accept: application/json' \
  --header 'Content-Type: application/json' \
  --header 'Authorization: Bearer kpat_abcd' \
  --data '{"name":"basic-auth","config":{"anonymous":null,"hide_credentials":false}}'
```

So now I'm trying to figure out how to generate the config part, and struggling.

fabianrbz:
@lena-larionova how can I help?

fabianrbz:
@lena-larionova I pushed something, would that work?

lena-larionova:
Tested out the changes you made, I think we're good now. The examples work!

PR #5977, opened 2023-08-18
Chore: Add path parameter style note to style guide

Guaris:
From hackathon conversation. Path parameters are represented with curly braces, this is the standard when writing about API path parameters.


acgoldsmith (approved):
lgtm

PR #5989, opened 2023-08-22
Kic version loop

rainest:
### Description

Like https://github.com/Kong/docs.konghq.com/pull/5924 but with more automation

### Testing instructions

Netlify link: https://deploy-preview-5989--kongdocs.netlify.app/kubernetes-ingress-controller/latest/references/version-compatibility/

### Checklist 

- [ ] Review label added 

### DIFF
--- a/app/_data/tables/support/kic/base.yml
+++ b/app/_data/tables/support/kic/base.yml
@@ -1,0 +1,21 @@
+_base:
+  ce: &ce
+    min: 2.0
+    max: 3.4
+    name: Kong Gateway (OSS)
+  ee: &ee
+    min: 2.1
+    max: 3.4
+    name: Kong Enterprise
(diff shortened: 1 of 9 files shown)


mheap:
I took a look at this as I love the idea of these pages being data driven, but I don't think we can get the same look and feel. 

I had a go at formatting without the version tabs so that everything's shown on a single page per KIC version, but it didn't work out (see below)

On balance, I think the cross/check UI is better for consumers even if it's more painful for us to maintain. If maintenance is too painful, we should look at investing in tooling to generate the same view from data

## Current

CleanShot 2023-09-27 at 12 48 20

## Travis

CleanShot 2023-09-27 at 12 48 50

## Michael table

CleanShot 2023-09-27 at 12 56 36

PR #6010, opened 2023-08-25
[Fix] URLS that weren't linked

acgoldsmith:
### Description

Papercut ticket DOCU-3008

### Testing instructions

Netlify link: 

### Checklist 

- [x] Review label added 

### DIFF
--- a/app/_src/gateway/reference/configuration.md
+++ b/app/_src/gateway/reference/configuration.md
@@ -684,10 +684,10 @@ datastore.
 
 Example: `proxy_listen = 0.0.0.0:443 ssl, 0.0.0.0:444 http2 ssl`
 
-See http://nginx.org/en/docs/http/ngx_http_core_module.html#listen for a
+See [http://nginx.org/en/docs/http/ngx_http_core_module.html#listen](http://nginx.org/en/docs/http/ngx_http_core_module.html#listen) for a
 description of the accepted formats for this and other `*_listen` values.
 
-See https://www.nginx.com/resources/admin-guide/proxy-protocol/ for more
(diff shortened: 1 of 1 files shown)


app/_src/gateway/reference/configuration.md
```diff
@@ -4281,6 +4280,6 @@ Notes:
 **Default:** none
 
 
-[Penlight]: http://stevedonovan.github.io/Penlight/api/index.html
-[pl.template]: http://stevedonovan.github.io/Penlight/api/libraries/pl.template.html
-[templates]: https://github.com/kong/kong/tree/master/kong/templates
+[Penlight](http://stevedonovan.github.io/Penlight/api/index.html)
+[pl.template](http://stevedonovan.github.io/Penlight/api/libraries/pl.template.html)
+[templates](https://github.com/kong/kong/tree/master/kong/templates)
```
cloudjumpercat, line 4285: @acgoldsmith I think these ones are a different type of link, don't know what the name is, but basically they don't appear in the output docs, but they reference a link, so all you have to do is something like "For more info, see [Penlight](Penlight)." for example and it will pull in the right link. So you can kind of reuse links on a page and won't have to keep defining them. So I think this one was correct the way it originally was.

cloudjumpercat (commented):
@acgoldsmith I left one comment at the bottom about some links that I think are reference reused links, but all the rest of the links look correct to me! I think my only concern is that these changes will be overwritten when we autogenerate this file again. And I'm not sure if the source file for this supports the kind of markdown we are using for links. I wondered if that's why the links weren't hyperlinks in the original.

PR #6023, opened 2023-08-28
Konnect rename: Renames the analytics section.

Guaris:
https://konghq.atlassian.net/browse/DOCU-3375

There is a link to the runtime manager section. I left that behind. I think when someone tackles the runtime manager section they should change the link based on what they rename those files to.


acgoldsmith (approved):
LGTM

PR #6059, opened 2023-09-05
Fix(plugin scopes): Consumer groups scope and application registration plugin

lena-larionova:
### Description

The [scopes compatibility table](https://docs.konghq.com/hub/plugins/compatibility/#scopes) was missing a column for consumer groups. This PR enables generating the column and all entries from the plugins' schemas.

Additionally, the Application Registration plugin was listed as incompatible with all scopes, which is wrong - it's compatible with services only. Turns out we were selecting for the wrong key. Fixing this also fixes the display of the plugin's [config](https://docs.konghq.com/hub/kong-inc/application-registration/configuration/) and [basic examples](https://docs.konghq.com/hub/kong-inc/application-registration/how-to/basic-example/), which were both missing services.

### Testing instructions

Netlify link: 

### Checklist 

- [x] Review label added 

### DIFF
--- a/app/_plugins/generators/plugin_single_source/plugin/schemas/base.rb
+++ b/app/_plugins/generators/plugin_single_source/plugin/schemas/base.rb
@@ -88,10 +88,10 @@ module PluginSingleSource
         end
 
         def enable_on_service?
-          field = fields.detect { |f| f.key?('service') }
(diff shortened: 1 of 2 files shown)


acgoldsmith (approved):
LGTM

PR #6060, opened 2023-09-05
fix: Partial revert of plugin scopes changes

lena-larionova:
### Description

Partial revert of https://github.com/Kong/docs.konghq.com/pull/6059, undoing the app registration plugin part. That plugin is an exception and needs to be handled that way, and I can't figure out how to do it right. 

Reverting for now to prevent spec tests from failing in every PR. That just brings us back to the initial problem with the app reg plugin, which still needs fixing, but requires more Ruby knowledge than I have at the moment.

### Testing instructions

Netlify link: 

### Checklist 

- [x] Review label added 


acgoldsmith (approved):
lgtm

PR #6063, opened 2023-09-06
Update vale to use the default dictionary with plugins docs

Guaris:
### Description

Edits the config so that the plugins section uses normal rules.

Removes the `we` warning.


lena-larionova:
Should we delete the whole plugins library section of the vale config? This directory: https://github.com/Kong/docs.konghq.com/tree/vale-standardize-on-plugins/.github/styles/kongplugins

.github/styles/kong/We.yml
```diff
@@ -4,8 +4,6 @@ link: 'https://developers.google.com/style/pronouns#personal-pronouns'
 level: warning
 ignorecase: true
 tokens:
-  - we
-  - we'(?:ve|re)
   - ours?
   - us
```
lena-larionova, line 8: Honestly, I think we should get rid of this rule entirely, not just "we". There's still a lot of exceptions where "our/ours", "us" are valid.

Can we also get rid of the "I" rule? https://github.com/Kong/docs.konghq.com/blob/vale-standardize-on-plugins/.github/styles/kong/First-person.yml

lena-larionova (approved):

PR #6072, opened 2023-09-07
fix: Algolia config for specs landing page

lena-larionova:
### Description

Minor adjustments to our algolia configuration.

Open API spec landing page wasn't appearing in search, and basic-examples pages for plugins were causing indexing problems. As well, we were running the indexer too often (daily). This PR lowers that to weekly, while still leaving the option to run ad hoc builds manually at any time.

As part of this update, deleting some very old algolia config files that haven't been in use for the past two years, since @acgoldsmith just deleted the extra indices on the algolia side.

Need to merge this to get the API spec selector live, so that we can select on the right thing.

https://konghq.atlassian.net/browse/DOCU-3463

### Testing instructions

Nothing to preview, will need to merge to apply everything. Tested locally, the index runs with no issues.

### Checklist 

- [x] Review label added 


acgoldsmith (approved):
LGTM and thank you!

PR #6095, opened 2023-09-13
feat: Use real "edit" links for autogenerated reference docs

lena-larionova:
### Description

We've had a few complaints from support and other users that it is difficult to find the source for our autogenerated docs: Gateway config reference, PDK references, and the CLI reference.

All of these references already have a `source_url` tag that links off to the source file that the doc is generated from. This update uses the existing `source_url` for "Edit this page" links, so that any page with a custom `source_url` set in its front matter will now link to the real source.

### Testing instructions

Preview link: 

### Checklist 

- [x] Review label added 


app/_layouts/docs-v2.html
```diff
@@ -41,7 +41,9 @@
           {% assign default_edit_link = page.path | replace: "latest", page.kong_version %}
           {% assign edit_link_no_https = page.edit_link | remove: "https://" %}
 
-          {% if page.edit_link and edit_link_no_https == page.edit_link %}
+          {% if page.source_url %}
+            {{ page.source_url }}
```
mheap, line 45: Should this be the edit link? We'd have to do some manipulation of `source_url` but I think it would be better if the experience was consistent

mheap: The `source_url` is the view link:

https://github.com/Kong/kong/tree/master/kong/pdk/client.lua

vs:

https://github.com/Kong/kong/edit/master/kong/pdk/client.lua

Replace `tree` with `edit` and it's good to go

```suggestion
            {{ page.source_url | replace: "/tree/", "/edit/" }}
```

lena-larionova: Ah! I didn't even think of that 😬 yep let's do that.

cloudjumpercat (approved):
LGTM! I tested out several of the autogenerated docs and the links went to the correct pages. I'll let you merge this in case there was anything else you wanted to add.

PR #6103, opened 2023-09-14
fix(docs): correct the influxdb - gateway startup order

AndyZhang0707:
### Description

- Update the startup order and startup InfluxDB before starting up kong-gateway.

- From 3.4.0.0 Kong doesn't release the alpine docker image anymore thus correct the doc.

 

### Testing instructions

Preview link: 

### Checklist 

- [x] Review label added 

### DIFF
--- a/app/_src/gateway/kong-enterprise/analytics/influx-strategy.md
+++ b/app/_src/gateway/kong-enterprise/analytics/influx-strategy.md
@@ -56,7 +56,7 @@ will work for the purposes of this guide.
 1. Pull the following Docker image.
 
     ```bash
-    docker pull kong/kong-gateway:{{page.versions.ee}}-alpine
+    docker pull kong/kong-gateway:{{page.versions.ee}}
     ```
 
     {:.important}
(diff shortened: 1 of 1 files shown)


lena-larionova:
> Another option to correct this is to move the steps of starting an influxdb instance before the gateway start steps. This requires some rewording of the steps so I'd leave the change to the docs team.

I think this would probably be the better option, unless you think those environment variables that you removed aren't necessary at all.

AndyZhang0707:
> > Another option to correct this is to move the steps of starting an influxdb instance before the gateway start steps. This requires some rewording of the steps so I'd leave the change to the docs team.
> 
> I think this would probably be the better option, unless you think those environment variables that you removed aren't necessary at all.

Yes, I agree it's a better option. I may draft an initial version later and send it for review.

AndyZhang0707:
Hi @lena-larionova, I've updated the doc to reorder the startup steps. Please help and take a look. Thanks.

lena-larionova (approved):
LGTM

PR #6117, opened 2023-09-18
Update KIC EOL dates

mheap:
### Description

KIC EOL dates were incorrect. Updated to match 2.11 EOL
 
### Testing instructions

Preview link: [/kubernetes-ingress-controller/latest/support-policy/](https://deploy-preview-6117--kongdocs.netlify.app//kubernetes-ingress-controller/latest/support-policy/)

### Checklist 

- [x] Review label added 

### DIFF
--- a/app/_src/kubernetes-ingress-controller/support-policy.md
+++ b/app/_src/kubernetes-ingress-controller/support-policy.md
@@ -74,7 +74,7 @@ LTS versions of {{site.kic_product_name}} are supported for 3 years after releas
 
   <tr>
     <td>1.22</td>
-    <td>4.8</td>
+    <td>4.9</td>
     <td>Oct 2022</td>
     <td>2.8 LTS</td>
     <td>2.5 LTS</td>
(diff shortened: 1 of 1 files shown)


app/_src/kubernetes-ingress-controller/support-policy.md
```diff
@@ -24,7 +27,7 @@ LTS versions of {{site.kic_product_name}} are supported for 3 years after releas
   <th>Upstream K8s EOL</th>
   <th>Supported {{site.base_gateway}} Versions</th>
   <th>Recommended KIC Version</th>
-  <th>EOL</th>
+  <th>KIC EOL</th>
```
lena-larionova, line 30: The KIC EOL date is in descending order while all the other columns are in ascending order:
https://deploy-preview-6117--kongdocs.netlify.app//kubernetes-ingress-controller/latest/support-policy/

Since you have to pick one, I think it would be better to have all columns in descending order.

mheap: This is accurate. The LTS version support ends after the later releases. 

This table is ordered by _Kubernetes version_ - are you suggesting to sort it in descending order by k8s version?

lena-larionova: Ah, I didn't see that it was LTS so the 2025 date was confusing when the release was so much earlier than 2.11.x.

But yes, I do think that it should be sorted in descending order of the k8s version, since that's the first column. It would match the sorting of the table below it (which is KIC versions, yes, but the first column is in descending order).

That's just a gut feeling though - take that feedback as you will.

lena-larionova (approved):
Approving, and leaving it up to you to decide what to do with the ordering of the K8s versions table.

PR #6120, opened 2023-09-18
Retire handwritten control plane config API docs

Guaris:
https://konghq.atlassian.net/browse/DOCU-3462

Not in this PR: 
- Add an insomnia button to the spec 
This has to be addressed in the spec itself and will be addressed once we merge everything to main. 
It will look like this: 

image


acgoldsmith (approved):
LGTM

PR #6123, opened 2023-09-18
Update: Rename runtime-groups in decK and bump version

lena-larionova:
### Description

As part of the Konnect rename, decK gets the following changes:

* `--konnect-runtime-group-name` to `--konnect-control-plane-name`
* `_konnect.runtime_group_name` to `_konnect.control_plane_name`

This will come out with a new version - my guess is 1.27, but it could be 1.26.2, or 1.30. We'll know for sure when they tag the new version on Wednesday.

We were also missing deck 1.26.x, which is now also included (metadata update only).

While updating these flags in the CLI reference topics, I realized that it would make way more sense to reuse the global flags as an include. They're written in one deck config file, `root.go`, so they're always the same for every single CLI command (hence global). This way, when there's a change to global CLI flags, we only need to update one file with version tags, not 15.

### Testing instructions

**Explanation of edits**
❗ Reviewers, you might notice that I used version tags in some places and just straight up changed the names in others. My logic was that:
* If it's a flag or a parameter (`-konnect-runtime-group-name` or `konnect.runtime_group_name`), then it's version-specific. The new flags are introduced in deck version 1.27, and there's no getting around that. These got `if_version` tags.
* If it's a mention of runtime groups or runtime instances directly in the text, then I changed them without setting version tags. This is because the _terms_ "control plane" and "data plane node" are _not_ version-specific. Since Konnect is unversioned, anyone using decK with konnect will be using the new terms.

Hope that made sense.

https://deploy-preview-6123--kongdocs.netlify.app/deck/latest/guides/konnect/
https://deploy-preview-6123--kongdocs.netlify.app/deck/latest/reference/deck/

### Checklist 

- [x] Review label added 


acgoldsmith (approved):
LGTM and good call on switching to use include for global flags

PR #6153, opened 2023-09-21
fix: Layout issues with menus

lena-larionova:
### Description

Fixing some issues with navbar formatting:
* On smaller screens, the top nav is misaligned and the buttons on the far right get cut off
     old:
  
  Screenshot 2023-09-21 at 11 11 23 AM

  new:
  Screenshot 2023-09-21 at 10 19 29 AM
  
* The margins of the top nav menu items overlap with some clickable areas of the docs, making them unusable
* In the API docs, the left nav scrolls together with the content body, making it difficult to use the left menu when scrolled down too far.
  * Scroll was already set to `sticky` but with no `top` setting, there was nothing for it to stick to.

Issues 2 and 3 reported on slack.

### Testing instructions

You can test the menu resizing, but the other two fixes can't be tested via netlify preview since they're fixes to the API docs.
I tested them via dev tools in chrome, they _should_ work.

### Checklist 

- [x] Review label added 

### DIFF
--- a/app/_assets/javascripts/views/SpecView.vue
+++ b/app/_assets/javascripts/views/SpecView.vue
@@ -122,6 +122,7 @@ function initActiveProductVersionId () {
   position: sticky;
   height: calc(100vh - 60px);
   margin-top: 60px;
+  top: 60px;
   border-right: 1px solid var(--section_colors-stroke);
 }
 .sidebar {
--- a/app/_assets/stylesheets/header-v2.less
+++ b/app/_assets/stylesheets/header-v2.less
@@ -317,6 +317,11 @@ header.navbar-v2 {
       display: flex;
       align-items: center;
       transition: transform 0.5s;
+      max-height: 60px;
+
+      > ul {
+        max-height: 40px;
+      }
 
       .navbar-item {
         position: relative;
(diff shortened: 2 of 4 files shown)


cloudjumpercat (commented):
For the "free Trial", the button still got cut off when I resized Chrome (I'm on a laptop as well). Looks like it does it at 1379x845 resolution. The Free Trial button disappears altogether at 1348x845, not sure if that's expected.

I didn't encounter any overlap issues when I tested Netlify, so I think that's fixed.

I tried testing the API docs in dev tools, but all of the specs were generating an error.

lena-larionova:
> For the "free Trial", the button still got cut off when I resized Chrome (I'm on a laptop as well). Looks like it does it at 1379x845 resolution. The Free Trial button disappears altogether at 1348x845, not sure if that's expected.

That is indeed expected. Our nav is only so big, so they have to be hidden at a certain size. 

I think I might hide them at a wider resolution, then. 


> I tried testing the API docs in dev tools, but all of the specs were generating an error.

You won't be able to test them, they don't generate in netlufy preview.

cloudjumpercat (approved):
LGTM!

PR #6193, opened 2023-09-28
Fix issue with Vale reporting incorrect things as broken

Guaris:
### Description

Update Vale Github Action so that it only runs against added/copied/modified files.
This also prevents Vale from running on `2.6.x` and `2.7.x` Gateway file changes.

### Note

Because of an [API limitation](https://github.com/reviewdog/reviewdog#filter-mode-support-table) two types of `comments` might appear as a result of the check. Excerpt from the docs:
> Note that not all reporters provide full support of filter mode due to API limitation. e.g. github-pr-review reporter uses [GitHub Review API](https://docs.github.com/en/rest/pulls/reviews) and [GitHub Review Comment API](https://docs.github.com/en/rest/pulls/comments) but these APIs don't support posting comment outside diff file, so reviewdog will use [Check annotation](https://docs.github.com/en/rest/checks/runs) as fallback to post those comments [1].

* Github Annotations - for pre-existing errors (not part of the diff) on files, this will no longer appear once we fix all the errors. They look like this
image (2)
* Github Comments - for newly introduced errors. They look like this
Screenshot 2023-10-05 at 17 41 28

### DIFF
--- a/.github/workflows/linting.yml
+++ b/.github/workflows/linting.yml
@@ -84,13 +84,24 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v4
+      - name: Get changed files
+        id: changed-files
+        uses: tj-actions/changed-files@v38
+        with:
+          sha: ${{ github.sha }}
+          files: |
+            app/
+            docs/
+          files_ignore: |
+            app/_src/.repos/kuma/**
+          json: true
+          quotepath: false
+          escape_json: false
       - uses: errata-ai/vale-action@reviewdog
         with:
-          #version: 2.24.0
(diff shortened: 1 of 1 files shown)


.github/workflows/linting.yml
```diff
@@ -84,11 +84,12 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v3
-      - uses: errata-ai/vale-action@reviewdog
+      - uses: errata-ai/vale-action@v2.0.1
```
Guaris, line 87:
```suggestion
      - uses: errata-ai/vale-action@v1.5.0
```

fabianrbz (approved):

PR #6204, opened 2023-09-29
Fix: Add missing info endpoints to Gateway Admin API EE spec [skip ci]

lena-larionova:
### Description

The spec was missing the following information endpoints:

* `GET /endpoints`
* `GET /timers`
* `GET /status`

https://konghq.atlassian.net/browse/DOCU-235

### Testing instructions

Preview link: N/A, have to preview in Dev Portal or a spec renderer.

### Checklist 

- [x] Review label added 


lena-larionova:
@Guaris can you please take a look? Also, this adds 5k lines to the spec - is there any way around that, or is that just what we have to live with?

Guaris:
It's hard to see a diff in a spec like this, but here's what I did. 

- pulled the response in `/endpoints` into a parameter variable and then referenced it. This improves readability in line. 
- Fixed an issue with a tags example object not actually being an array object in the example. 
- Fixed an issue with this workspace config property that was written like: 

```
 portal_developer_meta_fields:
              default:
                - label: Full Name
                  title: full_name
                  validator:
                    required: true
                    type: string
```
and needed to be turned into an object by adding a `properties` field.                     


This is everything that had to do with your initial changes.. i think. 

In the process: 
-  I applied the `consumer_groups` and `licenses` tags to the appropriate endpoints that were originally rendered under the default category in Swagger. (Here's a link the live site where they are rendered there https://docs.konghq.com/gateway/api/admin-ee/latest/#/default)

- Standardized to no quotes in `description:` fields, which meant I had to reword some descriptions that contained a colon because they caused the rendered version to break unless quotes were present. I didn't mean to touch this in this PR but when I made the first commit I realized some had quotes and some didn't so I opted for none.  (sorry about putting this in this PR)

Guaris (approved):

PR #6225, opened 2023-10-03
Fix broken variable in Gateway overview doc

Guaris:
Variable was broken and not rendering properly, fixed it.


cloudjumpercat (approved):
LGTM

PR #6235, opened 2023-10-04
fix: Plugin tiers for konnect-incompatible plugins

lena-larionova:
### Description

Some konnect-incompatible plugins were displaying Konnect Paid or Konnect Premium badges. This was happening because the metadata was incorrectly set to include either `paid: true` or `premium: true` for these plugins.

Fixing that, and adding a banner to konnect-incompatible plugin landing pages to call that out.

### Testing instructions

Preview link: 

### Checklist 

- [x] Review label added 

### DIFF
--- a/app/_assets/stylesheets/page.less
+++ b/app/_assets/stylesheets/page.less
@@ -613,6 +613,13 @@ Generic Styling for Desktop
       border: none;
     }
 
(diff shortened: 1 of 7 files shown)


fabianrbz (approved):
LGTM

PR #6236, opened 2023-10-04
update: Third-party categorization

lena-larionova:
### Description

Third party plugins are now categorized as:
* Partner plugins: third party contributors that have a close relationship with Kong, usually meaning contracts, rigorous testing, etc.
* Unsupported community plugins: Plugins from contributors that haven't gone through the above process.

This update makes it clearer which is which by using better filtering, banners, and categorization.

https://konghq.atlassian.net/browse/DOCU-3509

### Testing instructions

https://deploy-preview-6236--kongdocs.netlify.app/hub/
https://deploy-preview-6236--kongdocs.netlify.app/hub/?support=third-party

Partner plugin: https://deploy-preview-6236--kongdocs.netlify.app/hub/imperva/imp-appsec-connector/
Community plugin: https://deploy-preview-6236--kongdocs.netlify.app/hub/amberflo/kong-plugin-amberflo/

### Checklist 

- [x] Review label added 


mheap:
In the URL `support=third-party` still implies some level of support. Can we change to `support=none`?

The “Not Supported” phrasing makes me uneasy. It implies that they won’t work, when we actually mean “functional, but no support contract”. Can we rephrase to something like “No Support Available”?

lena-larionova:
> In the URL support=third-party still implies some level of support. Can we change to support=none?
Yep, I can do that.

> The “Not Supported” phrasing makes me uneasy. It implies that they won’t work, when we actually mean “functional, but no support contract”. Can we rephrase to something like “No Support Available”?

Maybe "not supported by Kong"? "No support available" is also not quite right, since all of these plugins have support links, eg https://deploy-preview-6236--kongdocs.netlify.app/hub/moesif/kong-plugin-moesif/

mheap (approved):
The “no support from Kong” phrasing is ok for me too. I’ll leave it to you to make the call

lena-larionova:
The more I think about it, the more it sounds to me like "Not supported by Kong" is going to sound like "Not supported by Kong Gateway", since Kong is still an unofficial short-form for Gateway in... well, everywhere. Going with your suggestion as the lesser of two evils here :)

PR #6252, opened 2023-10-06
Fix: curl examples for custom plugins schemas in Konnect

lena-larionova:
### Description

Turns out that you can't pass a file directly via cURL to our Konnect CP configuration API. The contents of the file don't get escaped correctly. When testing this out originally, I used a mix of cURL and HTTPie, and HTTPie had no issues - but our default cURL instructions have problems here.

Issue was reported on [Slack](https://kongstrong.slack.com/archives/C03NRECFJPM/p1696622470557839).

Updating the doc with Fero's help, tested it all with curl this time around.

The jq instructions are in a "tip", since we can't tell them to use jq if they don't want to or don't have it.

### Testing instructions

Preview link: 

### Checklist 

- [x] Review label added 


fabianrbz (approved):
LGTM!

PR #6261, opened 2023-10-09
Generate OAS Overlays

mheap:
### Description

Proof of concept OAS Overlay generation from source files. This needs more work if we want to explore further, but it shows what the source files look like and how the output file could be generated

**DO NOT MERGE**

## Path to overlay adoption

* Jekyll ignores overlay markdown files
* Ability to apply multiple overlays and generate one file. 
* **Optional** Automatic upload to Dev Portal 
* Define workflow for everything for specs that live in the platform-api repo and the Kong Admin API specs that live in the docs repo. 
    * This is because we currently own the Gateway admin API specs but eventually they will live in another repo, we also need to figure out how we will interact with the specs in platform-api. 

### 

Preview link: N/A

### Checklist 

- [x] Review label added 

### DIFF
--- a/docs/overlays.md
+++ b/docs/overlays.md
@@ -1,0 +1,43 @@
+# Generating OAS Overlays
+
+1. Add all the entries to the gateway nav file in the `unlisted` section
+2. Create entries in `app/_src/gateway/overlays` with the following format:
+
+```yaml
+---
+layout: overlay
+target: '$.paths["/{runtimeGroupId}/core-entities/certificates/{certificate_id}"].put'
+---
+
+This is a description.
+
+It can be as long as you like.
+
+{% if_version gte:3.3.x %}
+And even use conditionals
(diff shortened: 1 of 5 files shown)


docs/overlays.md
```diff
@@ -0,0 +1,43 @@
...
+{% if_version gte:3.3.x %}
+And even use conditionals
+{% endif_version %}
+```
+
+Build the site with `make run`:
+
+```bash
+export VERSION=3.4.x;
+for i in dist/gateway/$VERSION/overlays/*/index.html; do cat $i >> $VERSION.yaml; echo "" >> $VERSION.yamldone;
```
Guaris, line 25:
```suggestion
for i in dist/gateway/$VERSION/overlays/*/index.html; do cat $i >> $VERSION.yaml; echo "" >> $VERSION.yaml; done;
```

PR #6354, opened 2023-10-23
Automated OAS Update

kong-apiops:
Generated OAS files from platform-api


lena-larionova (approved):

Guaris (approved):
Reviewed control planes config, different format is because new one gets built in platform-api

PR #6358, opened 2023-10-24
fix(portal) remove examples with path in portal gui host

nateslo:
### Summary
Removes examples of adding a path when configuring `portal_gui_host`

Related: https://github.com/Kong/docs.konghq.com/pull/4179

### Reason
Adding a path to `portal_gui_host` is not supported


lena-larionova (approved):
LGTM, thanks for updating it both here and in source :)

PR #6365, opened 2023-10-25
Fix: Use placeholder token in TBAC examples instead of hashed value

jamesgkong:
### Description

As per case: 00038584 the changed examples try to use the token value returned in Step 3 here: https://docs.konghq.com/gateway/latest/production/access-control/enable-rbac/#super-admin-creates-one-admin-for-each-team However this is a hashed value: https://docs.konghq.com/gateway/latest/admin-api/rbac/reference/ and not the actual user token which was set for adminA in Step 1 here: https://docs.konghq.com/gateway/latest/production/access-control/enable-rbac/#super-admin-creates-one-admin-for-each-team

### Testing instructions

Preview link: 

### Checklist 

- [x] Review label added 

### DIFF
--- a/app/_src/gateway/production/access-control/enable-rbac.md
+++ b/app/_src/gateway/production/access-control/enable-rbac.md
@@ -357,7 +357,7 @@ user token:
 
     ```sh
     curl -i -X GET http://localhost:8001/teamB/rbac/users \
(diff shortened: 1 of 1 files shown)


lena-larionova (approved):
Ah, nice catch! LGTM

PR #6395, opened 2023-10-29
fix: Custom Resource Definitions

Rajakavitha1:
### Description

ported content to 3.0
 

### Testing instructions

Preview link: 

### Checklist 

- [x] Review label added 


czeslavo:
Do we want to split the reference into one file per CRD? If so, I think this should not be done manually as the CRDs reference is generated in KIC. If want to go with one file per CRD we should change the way it's generated in KIC. @mheap

mheap:
Let's keep them as a single file. I'll push a page that we can overwrite with new CRDs at release

mheap (approved):

PR #6406, opened 2023-10-30
fix(opentelemetry): address invalid request to set up plugin

samugi:
### Description

The request documented to configure the Opentelemetry plugin is invalid because it uses a dot `.` in the name of the attribute, which is interpreted as a field separator in the curl form data.

This updates the documentation without changing the example, using JSON instead of form data to allow passing a name that includes a dot.

### Testing instructions

Preview link: 

### Checklist 

- [x] Review label added 


lena-larionova (approved):
LGTM - tested a bunch of formats, your way is the only one that worked here.

PR #6436, opened 2023-11-02
Automated OAS Update

kong-apiops:
Generated OAS files from platform-api

### DIFF
--- a/api-specs/Konnect/v2/yaml/portal-management.yaml
+++ b/api-specs/Konnect/v2/yaml/portal-management.yaml
@@ -11,6 +11,8 @@ servers:
     description: US Region Base URL
   - url: https://eu.api.konghq.com/v2
     description: EU Region Base URL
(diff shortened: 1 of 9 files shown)


api-specs/Konnect/v2/yaml/portal-management.yaml
```diff
@@ -2667,6 +2667,7 @@ components:
           enum:
             - us
             - eu
+            - au
```
lena-larionova, line 2670: Is the au region publicly available yet? If not, we might have some issues with using these automated updates to pull specs into the docs, since we're getting some non-public changes. @mheap @acgoldsmith

cloudjumpercat: @lena-larionova It's only available to new customers. It will be available to all customers next week.

cloudjumpercat (approved):
LGTM!

PR #6464, opened 2023-11-06
Updating link to IETF Draft

rLitto:
RateLimit header fields for HTTP -> https://datatracker.ietf.org/doc/draft-ietf-httpapi-ratelimit-headers/

### Testing instructions

Preview link: 

### Checklist 

- [x] Review label added 


tysoekong:
After PR, the hyperlink points directly back at the approved IETF RFC, which seems correct to me.

lena-larionova (approved):
Nice catch, thanks for the fix! Agree with you, that looks like a more accurate & evergreen doc; the old one redirects to an outdated doc that's in "Replaced" status.

LGTM

lena-larionova:
Merging with failed smoke test; it's only failing because the build was blocked because it's from a fork. We'll get that check to skip in the future :)

PR #6545, opened 2023-11-21
fix(saml): adjust priority to 1010

tzssangglass:
### Description

 

sister PR: https://github.com/Kong/kong-ee/pull/7267

### Testing instructions

Preview link: 

### Checklist 

- [x] Review label added 

### DIFF
--- a/app/_includes/md/plugin-priority.md
+++ b/app/_includes/md/plugin-priority.md
@@ -74,6 +74,7 @@ basic-auth                  | 1100
 openid-connect              | 1050
 hmac-auth                   | 1030
 jwt-signer                  | 1020
+saml                        | 1010
 request-validator           | 999
 websocket-size-limit        | 999
 websocket-validator         | 999
@@ -90,7 +91,6 @@ rate-limiting               | 910
 rate-limiting-advanced      | 910
 graphql-rate-limiting-advanced | 902
 response-ratelimiting       | 900
-saml                        | 900
 oas-validation              | 850
 route-by-header             | 850
 jq                          | 811


windmgc (approved):

mheap:
Comment to @Kong/team-docs: This include needs an `if_version` tag one the change is released. The old priority was `900`, and the new one is `1010`. Each version of Gateway should show the correct priority

tzssangglass:
ping this

lena-larionova:
@mheap the renderer doesn't seem to want to display `if_version` correctly in the tables here. I always end up with either a break in the table (ie, breaks into two tables), or the entry appears on the same line as the previous.

Screenshot 2024-01-23 at 3 32 04 PM

Screenshot 2024-01-23 at 3 34 07 PM

Screenshot 2024-01-23 at 3 34 14 PM


I don't know which combination of things is messing with it: is it the tabs? The fact that this is an include? 

I've tried a bunch of approaches, none of these have worked:
* using `inline:true`
* leaving newlines around the table rows
* leaving no space
* leaving a newline either before or after the row
* using `{%- if_version -%}` syntax, with or without `inline:true`

fabianrbz:
@lena-larionova I _think_ something like this should work (versions might be wrong though)
```
saml                        | {% if_version lte:3.4.x inline: true %} 900 {% endif_version %}{% if_version gte:3.5.x inline: true %} 1010 {% endif_version %}
```

lena-larionova:
> @lena-larionova I _think_ something like this should work (versions might be wrong though)
> 
> ```
> saml                        | {% if_version lte:3.4.x inline: true %} 900 {% endif_version %}{% if_version gte:3.5.x inline: true %} 1010 {% endif_version %}
> ```

I thought about that - that _would_ work in most situations, but in this case the table row needs to move (ie be in two different positions for different versions), since the rows are in descending order.

fabianrbz:
🤦 I'll give it another shot

fabianrbz:
closing this one, this should be fixed  by #6866

PR #6586, opened 2023-11-30
feat(plugin): Rate Limiting Advanced window types explanation

Rajakavitha1:
### Description

https://konghq.atlassian.net/browse/DOCU-3397
Turn KB into doc: https://support.konghq.com/support/s/article/What-is-the-expected-behaviour-for-the-Rate-Limiting-Advanced-plugin-with-sliding-window-strategy

### Testing instructions

Preview link: https://deploy-preview-6586--kongdocs.netlify.app/hub/kong-inc/rate-limiting-advanced/#other-startegies

### Checklist 

- [x] Review label added 


app/_hub/kong-inc/rate-limiting-advanced/overview/_index.md
```diff
@@ -115,6 +115,23 @@ Two common use cases are:
    only to protect backend services from overloading that's caused either by specific
    users or by attacks.
 
+#### Other startegies   
+
```
lena-larionova, line 119: Feels like this needs an intro:
```suggestion
The Rate Limiting Advanced plugin supports the following window types:

* **Fixed window**: Fixed windows consist of buckets that are statically assigned to a definitive time range. Each request is mapped to only one fixed window based on its timestamp and will affect only that window’s counters.
* **Sliding window** (default): A sliding window tracks the number of hits assigned to a specific key (such as an IP address, consumer, credential) within a given time window, taking into account previous hit rates to create a dynamically calculated rate.
```

app/_hub/kong-inc/rate-limiting-advanced/overview/_index.md
```diff
@@ -115,6 +115,23 @@ Two common use cases are:
...
+The _sliding window_ strategy ensures a resource is not consumed at a higher rate than the configured.
+
+For example, consider this configuration:
+
+Limit size = 10
+Window size = 60 seconds
+
+In a _fixed window_ strategy you can predict when the window is going to be reset and if the client sends a burst of traffic. For example, 12 request per minute, then 10 requests are accepted with a `response 200` and 2 requests are rejected with `response 429` in a minute.
+
+If you configure a _sliding window_ strategy and the client sends a burst of 12 requests per minute, the first 10 requests are accepted with `response 200` and the rest of the requests are rejected with `response 429`. In this case, it appears to the client that the window is never reset. The algorithm counts the `response 429` and the "API is blocked forever". This happens because the burst of traffic rate of 12 requests per minute is higher than the rate configured in the plugin which is 10 requests per minute. If the client reduces the number of requests, then you get the `response 200` again.
```
lena-larionova, line 129:
```suggestion
If you use a sliding window, the first instance is the same: the client sends a burst of 12 requests per minute, 10 requests are accepted with a `200` response and two requests are rejected with a `429` response. 
In this case, it appears to the client that the window is never reset.
The algorithm counts the response `429` and the API is blocked indefinitely.
This happens because the burst of traffic rate of 12 requests per minute is higher than the rate configured in the plugin, which is 10 requests per minute. 
If the client reduces the number of requests, then you get the `response 200` again.
```

app/_hub/kong-inc/rate-limiting-advanced/overview/_index.md
```diff
@@ -115,6 +115,23 @@ Two common use cases are:
...
+
+When the client receives a `429 response`, it also receives a `Retry-After:<seconds>` header. Which means the client has to wait <seconds> before making a new request. If the client makes another request in less than this time, you get the `response 429` again. Otherwise, the window is reset.
```
lena-larionova, line 131:
```suggestion
When the client receives a `429` response, it also receives a `Retry-After:<seconds>` header. This means the client has to wait some number of seconds before making a new request. 
If the client makes another request in less than this time, you get the `429` response again. Otherwise, the window is reset.
```

app/_hub/kong-inc/rate-limiting-advanced/overview/_index.md
```diff
@@ -115,6 +115,23 @@ Two common use cases are:
...
+
+The _sliding window_ strategy ensures the API is consumed in the configured requests per second rate. This is not always true for the _fixed window_ strategy. Consider the same example with 10 requests per minute, if the client sends 10 requests at the 59th second of the window, then the window is reset and then it sends other 10 requests at the 1st second of the window, all the requests are accepted, making the acceptance rate higher than the configured in that time lapse. This does not happen with a _sliding window_ strategy, because the window "moves" during the last 60 seconds or the window size to ensure it meets the configured rate.
```
lena-larionova, line 133:
```suggestion
The sliding window type ensures the API is consumed in the configured requests per second rate. 
This is not always true for the fixed window strategy. 

Consider the same example with 10 requests per minute instead of 12. 
Let's say the client sends all 10 requests in the 59th second of the window:
* In a fixed window, the window resets a second later, and the client can send another 10 requests in the first second of the following window. All of the requests are accepted, making the acceptance rate higher than the configured rate in that two-second time period.
* In a sliding window, the window moves during the last 60 seconds to ensure it meets the configured rate.
```

lena-larionova (changes requested):
Left some comments on organizing this a bit better, and some general cleanup.

app/_hub/kong-inc/rate-limiting-advanced/overview/_index.md
```diff
@@ -88,6 +88,37 @@ otherwise, an error occurs:
 ```plaintext
 You must provide the same number of windows and limits
 ```
+### Window sizes
```
lena-larionova, line 91:
```suggestion
## Window types
```

PR #6657, opened 2023-12-13
Refactor sidebar - make it more a11y friendly

fabianrbz:
### Description

Refactor the sidebar to be accessible by following the `Treeview`.
There are a few things we might want to tweak, e.g.: 
* what should happen when there's one open menu, and we try to open another one? Should we close the first one?
* what should happen with all the nested open menus, when another top-level menu is closed? Should we keep its state? Collapse all the nested submodules?

### Testing instructions

Preview link: 

### Checklist 

- [x] Review label added 


cloudjumpercat:
> what should happen when there's one open menu, and we try to open another one? Should we close the first one?
I think closing the first one is a good idea otherwise it would be harder to scroll the nav.

> what should happen with all the nested open menus, when another top-level menu is closed? Should we keep its state? Collapse all the nested submodules?
I like the idea of collapsing the nested submodules, both for scroll reasons and presumably if someone is going to a different part of the nav, they no longer need the part of the nav they were just on.

mheap:
+1 to what @cloudjumpercat said.

I've just tested, it's looking good with both mouse + keyboard nav

mheap:
Vimium can't see collapsed sections

CleanShot 2023-12-14 at 14 00 17

fabianrbz:
@mheap that's because the menu-items aren't links. Let me see what I can do about it.

mheap:
```
Array.from(document.querySelectorAll("[role='treeitem']")).map(n => n.classList.add('menubutton'))
```

Worked for me based on https://github.com/philc/vimium/blob/881a6fdc3644f55fc02ad56454203f654cc76618/content_scripts/link_hints.coffee#L720-L721

fabianrbz:
That's an interesting hack :)
I was thinking about turning the arrows into proper buttons, but I might consider it

acgoldsmith:
also +1 to @cloudjumpercat 's response on behavior.

mheap:
Out of interest, why use `TreeView` here rather than `Menu`, `Group` and `MenuItem`? The latter would be accessible by default without the button class

fabianrbz:
I felt that it reflected more the type of structure that we have, I can easily change it to `menu`, `group` and `menuitem` though

mheap:
I have no strong feelings either way. I was asking out of curiosity as I’ve not done much ARIA work :shipit:

mheap (approved):

fabianrbz:
@mheap I ended up adding buttons and removing the CSS class, it's the right implementation

mheap:
LGTM

PR #6733, opened 2024-01-09
DOCU-3511: Refactor version is

fabianrbz:
### Description

Related [Jira ticket](https://konghq.atlassian.net/browse/DOCU-3511)

* Point jekyll-generator-single-source to the feature branch
* Set `latest: true` to the corresponding versions in `kong_versions.yml`
* Refactor Versions generators, we now have specific classes that handle
  setting the releases and versions information.
* Rename `app/_data/docs_nav_mesh_dev.yml` to
  `app/_data/docs_nav_mesh_2.6.x.yml`.
* Refactor Versions dropdown, we no longer need to maintain the routes
  and version info in the template.
* Don't include URLs from versions that are labeled to the sitemap
* Add `label: dev` to the unreleased version of mesh

### How it works

* We need to explicitly set `latest: true` to the corresponding versions in `app/_data/kong_versions.yml`
* For adding a new unreleased version, we need to add a new entry in `app/_data/kong_versions.yml`  with a label (it could be any string, e.g. `dev`, `next`, etc) and a new nav file, e.g. for gateway 

In `app/_data/kong_versions.yml`
```yaml 
- release: "3.5.x"
  ee-version: "3.5.0.2"
  ce-version: "3.5.0"
  edition: "gateway"
  latest: true # <- Set latest to the right version
  ...
- release: "3.6.x"
  ee-version: "3.6.0.0"
  ce-version: "3.6.0"
  edition: "gateway"
  label: dev # <- label the new unreleased version 
  ...
```
and we need to create the file `app/_data/docs_nav_gateway_3.6.x.yml`
The URLs generated for this version would be: `/gateway/dev/.../`, i.e. `/<edition>/<label or release>/...`.

Note that the file includes the actual release name in its name and not the label. 
This simplifies the release process. We just need to remove the `label` and set `latest: true` to it. 

### `page.release`

`page.release` is a [Liquid Drop](https://github.com/Shopify/liquid/wiki/Introduction-to-Drops) which is essentially an object that can be used in the templates (code [here](https://github.com/Kong/jekyll-generator-single-source/blob/82966bb101a62400839d35c06c0d406d5a1439d5/lib/jekyll/generator-single-source/liquid/drops/release.rb)). 

It's a replacement for `page.kong_version`, and whenever it is used in a template the [to_s](https://github.com/Kong/jekyll-generator-single-source/blob/82966bb101a62400839d35c06c0d406d5a1439d5/lib/jekyll/generator-single-source/product/release.rb#L45-L52) method gets called. It returns the release's label if it has one, or the release's value, e.g `3.4.x`.

It has a few other methods defined that should be useful:
* `value`: the actual release defined in `kong_versions.yml`, e.g. `3.4.x`.
* `label`: the name of the label.
* `latest?`: returns true if the release is marked as `latest` in `kong_versions.yml`.
* `versions`:  returns a hash with the corresponding versions defined in `kong_versions.yml`. For `gateway` it returns the versions without the suffix, i.e.  `{ 'ee' => '3.0.0.0', 'ce' => '3.0.0'  }`, for products that only have one version it returns a has with `default` as key, e.g. `{ 'default' => '2.6.x' }`.

The only time when we can't rely on the `to_s` method and we need to use `page.release.value` is when it is used as key when accessing hashes, like the [compatibility table](https://github.com/Kong/docs.konghq.com/blob/8508a4d9479b73a40390af8eeae0ba65598f73c8/app/gateway/2.6.x/compatibility.md#L12).

### Testing instructions

For now, I've only updated `mesh` to work, but we can try it with other products, bear in mind that gateway requires the plugin schemas to be available.

### TODO:

* [x] Replace `page.kong_version` with `page.release`
* [x] Replace `page.kong_versions` with `page.releases_hash` 
* [x] Do something about [hardcoded index URLs](https://github.com/Kong/docs.konghq.com/blob/main/app/_data/docs_nav_gateway_3.5.x.yml#L10)  in docs nav files for unreleased versions, I need to replace the `release` segment with the `label`.
* [x] Add a ton of tests
* [x] Add docs

### Checklist 

- [x] Review label added 


app/_layouts/docs-v2.html
```diff
@@ -83,7 +91,7 @@
 
       <div class="page-content">
         <div class="content show-anchor-links">
-          {% if page.kong_version != page.kong_latest.release and page.no_version != true %}
+          {% unless page.release.latest? or page.no_version == true %}
               <blockquote id="version-notice" class="important">
                 {% if page.edition == 'mesh' and page.version_data['releaseDate'] == nil %}
```
mheap, line 96: Does this need updating to be generic?

app/_plugins/generators/seo/index_entry/non_product_page.rb
```diff
@@ -0,0 +1,21 @@
...
+  module IndexEntry
+    class NonProductPage < Base
+      def key
+        @key ||= @page.url
+      end
+
+      def process!(index)
+        super(index)
+
+        @page.data['canonical_url'] = @page.url
```
mheap, line 13: Canonical URL should always have a trailing slash. Based on the link in the notification on https://deploy-preview-6733--kongdocs.netlify.app/mesh/dev/introduction/architecture/ we might need to enforce this here?

app/_plugins/generators/seo/index_entry/unversioned_product_page.rb
```diff
@@ -12,6 +12,10 @@ def process!(index)
         @page.data['canonical_url'] = @page.url
       end
 
+      def indexable?(_pages_index)
+        super && url_segments[1] != 'pre-1.7'
```
mheap, line 16: Worth adding a comment about what pre-1.7 is?

mheap (approved):
Some huge refactorings in here 🙌

app/_plugins/generators/seo/index_entry/non_product_page.rb
fabianrbz: hmmm, weird. both the meta tag and the link in the notification have a trailing slash, but it could be that  netlify is adding those, I'll take a look 

Screenshot 2024-01-12 at 09 19 35

app/_layouts/docs-v2.html
fabianrbz: 💯

lena-larionova (approved):
LGTM! This is awesome work, can't wait to start using the new process for gateway 😄 

Tested version switching locally, messed around with Gateway and plugins mainly. Couldn't find any issues.

PR #6797, opened 2024-01-22
Add WebAssembly filter chains to Gateway OSS 3.4 and 3.5 OpenAPI specs

hishamhm:
### Description

Add WebAssembly filter chains to Gateway OSS 3.4 and 3.5 OpenAPI specs

JIRA: KAG-1885

### Checklist 

- [x] Review label added 

### DIFF
--- a/api-specs/Gateway-OSS/3.5/kong-oss-3.5.yaml
+++ b/api-specs/Gateway-OSS/3.5/kong-oss-3.5.yaml
@@ -296,14 +296,13 @@ components:
         route: null
         service: "20487393-41ed-47f6-93a8-3407cade2002"
         filters:
-          - name: "go-rate-limiting",
-            enabled: true,
-            config: '{ \"minute\": 30 }'
-          - name: "rust-response-transformer",
-            enabled: true,
-            config: '{ \"remove_header\": \"X-Example\" }'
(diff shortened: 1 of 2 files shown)


Guaris:
@hishamhm I had to make some changes to the indentation and the way some of the datatypes were modeled. Thanks for contributing.

Guaris (approved):

api-specs/Gateway-OSS/3.5/kong-oss-3.5.yaml
```diff
@@ -7123,10 +7125,10 @@ paths:
                         properties:
                           headers:
                             type: "array"
-                            elements:
+                            items:
                               type: "string"
-                        required: "headers"
-                    required: "add"
+                        required: [headers]
+                    required: [add]
```
hishamhm, line 7131: do these need to be `["headers"]` and `["add"]` or YAML auto-stringifies those?

hishamhm:
@Guaris Thank you for the review and merge! FYI, the specs for filter chains in 3.6 should be the same as 3.5.

(Main difference between 3.4 and 3.5 is that 3.5 accepts filter `config` as an object, while 3.4 always expects a string. 3.6 is the same as 3.5.)

PR #6821, opened 2024-01-24
Fix: Use config.scopes instead of config.scopes_required in OIDC okta guide

lena-larionova:
### Description

The OIDC Okta guide incorrectly references `config.scopes_required` when it's actually configuring `config.scopes`.
Fixing that + some broken/outdated links + Dev Portal info that should be behind a conditional tag.
Also fixing the same links in the rest of the OIDC 3rd party guides.

https://konghq.atlassian.net/browse/DOCU-2277

### Testing instructions

Preview link: 

### Checklist 

- [x] Review label added 

### DIFF
--- a/app/_hub/kong-inc/openid-connect/how-to/third-party/_auth0.md
+++ b/app/_hub/kong-inc/openid-connect/how-to/third-party/_auth0.md
@@ -46,5 +46,5 @@ For basic authentication, use your client ID as the username and your client sec
 [client-credentials-grant]: https://auth0.com/docs/api-auth/tutorials/client-credentials
 [create-auth0-api]: https://auth0.com/docs/apis#how-to-configure-an-api-in-auth0
 [non-interactive-client]: https://auth0.com/docs/clients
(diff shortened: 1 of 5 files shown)


fabianrbz:
@lena-larionova I don't think those links at the bottom are not being rendered at all, they don't look like valid markdown (you can check the preview in GitHub). 
Should we convert them into bullet points and add a `See also` section or just remove them?

lena-larionova:
> @lena-larionova I don't think those links at the bottom are not being rendered at all, they don't look like valid markdown (you can check the preview in GitHub). Should we convert them into bullet points and add a `See also` section or just remove them?

They shouldn't be rendered. They're defining shortlinks for the pages, so that elsewhere on the page someone can use the `[Service][add-service]` syntax for a link. We don't really use that method anymore normally, but it doesn't really hurt for it to exist in some of these older topics? Or, I can update further and just write the links in normally, like we do everywhere else 😄

cloudjumpercat (approved):
LGTM! I'll let you merge this because I know Fabian had a question on it

PR #6859, opened 2024-01-30
Update: Change dev label and banner color

lena-larionova:
### Description

Changing the "dev" label to "unreleased" to make it clearer what this doc version is supposed to be. 

Also changing the banner colour to red, as the previous banner looked too similar to the "outdated version" banner.

❗ Reviewers (@cloudjumpercat @fabianrbz) I'm running into warnings from Kuma here - do we also need to update the dev label in the Kuma website repo?

Screenshot 2024-01-30 at 11 10 03 AM

### Testing instructions

Preview link: 

### Checklist 

- [x] Review label added 
- [x] For unreleased versions: [conditional version tags](https://docs.konghq.com/contributing/conditional-rendering/#conditionally-render-content-by-version) added, if applicable.

    For example, if this change is for an upcoming 3.6 release, surround your change in `{% if_version gte:3.6.x %} <content> {% endif_version %}` tags (or `if_plugin_version` tags for plugins).


app/_layouts/docs-v2.html
```diff
@@ -92,12 +92,13 @@
       <div class="page-content">
         <div class="content show-anchor-links">
           {% unless page.release.latest? or page.no_version == true %}
+            {% if page.release.label != nil %}
```
mheap, line 95: This was an issue before, but I could create a label called “legacy” and it would show “unreleased”. Should we use the value for `label` in the banner message?

docs/product-releases.md
```diff
@@ -4,14 +4,14 @@
...
+Labels provide flexibility in handling releases. Specifying `label: unreleased` on a release, alters the URL structure from `/<product>/<release number>/**/` to `/<product>/<label>/**/`, such as `/<product>/unreleased/**/`. Labels are particularly useful for working on unreleased versions of products. Once a label is added to a release, the content can be edited using [conditional rendering](https://docs.konghq.com/contributing/single-sourced-versions/) and merged it into the `main` branch without the need for long-standing release branches. The content will be rendered only in `/<product>/<label>/`.
```
lena-larionova, line 7: Ignoring the vale error here, it's not accurate - we shouldn't be using relative URLs in internal docs, since these aren't published to prod. 

@Guaris is there a way for certain Vale checks to ignore directories, or is that defined for all checks only? In this case, we'd want Vale to _not_ run the kong.Relativeurls check on anything in the /docs/ directory.

app/_layouts/docs-v2.html
lena-larionova: Hmm, didn't even think of that as a potential problem. Let me test.

cloudjumpercat:
> I'm running into warnings from Kuma here - do we also need to update the dev label in the Kuma website repo?

@lena-larionova That could be since they use "dev" there too. But they've been using it for a long time, so not sure if they'd want it to change.

lena-larionova:
> > I'm running into warnings from Kuma here - do we also need to update the dev label in the Kuma website repo?
> 
>  That could be since they use "dev" there too. But they've been using it for a long time, so not sure if they'd want it to change.

@cloudjumpercat Seems like it would be silly to write a rule just for Mesh, to convert "dev" to "unreleased". Maybe we leave it as "dev" for them in the dropdown, but then leave the banner as "unreleased" as the explanation? Ah, then we still have to write a rule, but at least just for the banner (which is simple).

mheap:
We can call it unreleased on docs.konghq.com. We shouldn’t change it on Kuma

cloudjumpercat:
> Seems like it would be silly to write a rule just for Mesh, to convert "dev" to "unreleased". Maybe we leave it as "dev" for them in the dropdown, but then leave the banner as "unreleased" as the explanation? Ah, then we still have to write a rule, but at least just for the banner (which is simple).

@lena-larionova That sounds good to me!

cloudjumpercat (approved):
LGTM! Tested by checking the unreleased pages of both Gateway and a plugin.

PR #6866, opened 2024-01-31
Plugin priorities

fabianrbz:
### Description

To generate the table, use the priorities defined in `app/_src/.repos/kong-plugins/app/data/priorities/`.
Add a new liquid tag `plugins_priority_table`, that takes `enterprise` or `oss` as an argument and uses the page's release (or latest for Konnect) to fetch the data. 

@lena-larionova I wasn't sure how we wanted to handle the case for `correlation-id`, so I added a separate table for it 🤷. From what I understand `free mode` uses the enterprise value and `oss` well, the oss one.

WDYT?
We could try to replicate the current behaviour too, it's a bit more complex though, but definitely doable.

### Testing instructions

[Preview link](https://deploy-preview-6866--kongdocs.netlify.app/gateway/latest/plugin-development/custom-logic/)

### Checklist 

- [ ] Review label added 
- [ ] For unreleased versions: [conditional version tags](https://docs.konghq.com/contributing/conditional-rendering/#conditionally-render-content-by-version) added, if applicable.

    For example, if this change is for an upcoming 3.6 release, surround your change in `{% if_version gte:3.6.x %} <content> {% endif_version %}` tags (or `if_plugin_version` tags for plugins).

### DIFF
--- a/app/_includes/plugins/priority_table.html
+++ b/app/_includes/plugins/priority_table.html
@@ -1,22 +1,3 @@
-{% if include.type == 'oss' %}
-  <table>
-    <thead>
-      <tr>
-        <th>correlation-id</th>
-        <th>Priority</th>
-      </tr>
-    </thead>
-    <tbody>
-      {% for type in include.correlation_id %}
-        <tr>
-          <td><span class="badge {{ type[0] }}"></span></td>
-          <td>{{ type[1] }}</td>
-        </tr>
-      {% endfor %}
-    </tbody>
-  </table>
-{% endif %}
-
 <table>
   <thead>
     <tr>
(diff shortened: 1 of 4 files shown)


lena-larionova:
>  I wasn't sure how we wanted to handle the case for correlation-id, so I added a separate table for it 🤷. From what I understand free mode uses the enterprise value and oss well, the oss one.

I wonder if it makes more sense to just remove Correlation ID from the OSS/Free tab, and change the tabs to "OSS package"/"Enterprise package". Because "free mode" is just the Enterprise gateway package without a license, so there's no access to certain plugins. 

So we could do something like:

Tab 1: OSS (or OSS package? Honestly, maybe just OSS)
Note: "If you're running Kong Gateway Enterprise in Free mode, see the Enterprise tab".

Tab 2: Enterprise (same thought as above)


This way, we can add Correlation ID into the main tables and just ignore the "free mode" difference.

fabianrbz:
@lena-larionova I've updated it so that:
* we render the plugins and their priorities as they are defined in the json file
* changed the tab to `OSS`

Can you update the [note](https://github.com/Kong/docs.konghq.com/blob/main/app/_src/gateway/plugin-development/custom-logic.md?plain=1#L450-L453) ?

lena-larionova:
@fabianrbz made those updates, but oddly enough, the list is missing the saml plugin 🤔 The plugin that kicked all of this off in the first place.

lena-larionova (approved):
pre-approving, just needs a rebase i think. the plugin priority changes work great though!

PR #6900, opened 2024-02-05
fix(mesh): remove duplicated getting started page

lahabana:
We have one directly integrated inside kuma-website

### Testing instructions

Preview link: 

### Checklist 

- [x] Review label added 


lena-larionova (approved):

lahabana:
@lena-larionova the tests don't pass but I don't see how this is related. Do you have any idea? (updating the branch just in case)

fabianrbz:
@lahabana you can ignore the test failures, it's because they don't work in forks

PR #6901, opened 2024-02-05
Fix: grammatical error

acgoldsmith:
### Description

"we recommends" to "we recommend" 

### Testing instructions

Preview link: 

### Checklist 

- [x] Review label added 


lena-larionova (approved):

lena-larionova:
FYI, you can append `[skip ci]` to the title of any PR to skip netlify builds, if there's ever anything that feels unnecessary. We do it all the time for readme or internal doc updates, but I'd say a one-letter type qualifies too 😄

PR #6912, opened 2024-02-06
chore: Set explicit timeouts for all workflows [skip ci]

lena-larionova:
### Description

Setting timeouts for all jobs in our workflows that don't have a timeout. This is to prevent CI from setting the `${{ fromJSON(vars.GHA_DEFAULT_TIMEOUT) }}` timeout variable in our workflows (for example, see https://github.com/Kong/docs.konghq.com/pull/6888).

I set 10 minutes for any label checking jobs, and any jobs that normally take under 1 minute to run (see https://github.com/Kong/docs.konghq.com/actions). Anything longer, I set 30 minutes as the timeout.

### Testing instructions

Preview link: N/A

### Checklist 

- [x] Review label added 

### DIFF
--- a/.github/workflows/check-links.yml
+++ b/.github/workflows/check-links.yml
@@ -11,6 +11,7 @@ env:
 jobs:
   check:
     runs-on: ubuntu-latest
+    timeout-minutes: 10
     steps:
       - name: Check if manual review has been performed
         uses: actions/github-script@v7
--- a/.github/workflows/auto-close-kong-issues.yml
+++ b/.github/workflows/auto-close-kong-issues.yml
@@ -7,6 +7,7 @@ jobs:
   auto-close-org-issues:
     name: Auto Close Org Issues
     runs-on: ubuntu-latest
+    timeout-minutes: 10
     steps:
       - name: Generate app token
         id: generate-app-token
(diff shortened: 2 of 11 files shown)


.github/workflows/check-links.yml
```diff
@@ -11,6 +11,7 @@ env:
 jobs:
   check:
     runs-on: ubuntu-latest
+    timeout-minutes: 10
```
Guaris, line 14: I think this should be longer it looks like sometimes it takes closer to 11 minutes
https://github.com/Kong/docs.konghq.com/actions/runs/7791282104
https://github.com/Kong/docs.konghq.com/actions/runs/7795979570

lena-larionova: This is the timeout for the check label job

Guaris:
I don't understand why we **have** to set this. Is it just so that the eng-enablement bot doesn't open PRs in our repos about it?

lena-larionova:
> I don't understand why we **have** to set this. Is it just so that the eng-enablement bot doesn't open PRs in our repos about it?

Yep, that's basically it. If we set it, we avoid having those PRs open against our repo every so often. It also doesn't hurt us to set this explicitly.

Guaris (approved):

PR #6921, opened 2024-02-07
Fix: ACL doc appearing in 3.5 when it should be in 3.6

Guaris:
ACL Doc was showing [Using ACLs with Consumer Groups](https://docs.konghq.com/hub/kong-inc/acl/how-to/consumer-groups/). in 3.5 it should only be accessible in 3.6+


app/_hub/kong-inc/acl/overview/_index.md
lena-larionova, line 1: This is the wrong file, it should be the one in the 3.5.x folder: `/app/_hub/kong-inc/acl/_3.5.x/overview/_index.md`

Which means that instead of a conditional, you can just move the link from that file.

lena-larionova (approved):

PR #6934, opened 2024-02-09
Chore: spec endpoint links open in new tab

acgoldsmith:
### Description

experimenting w/ link formatting to see if we can have the links to specs from table open in new tab, similar to nav behavior.

### Testing instructions

Preview link: https://deploy-preview-6934--kongdocs.netlify.app/gateway/latest/admin-api/

### Checklist 

- [x] Review label added 


lena-larionova:
FYI, the kramdown syntax for link attributes is this:

```
[link](url){:target="_blank"}
```
That might work for what you're trying to do. Things get weird markdown tables though, so if it doesn't work, you'll have you use html tags instead.

lena-larionova (approved):

PR #6952, opened 2024-02-15
release deck 1.34.0

AntoineJac:
### Description

Add deck 1.34 release information with new` deck file namespace` command
 

### Testing instructions

Preview link: https://deploy-preview-6952--kongdocs.netlify.app/deck/latest/reference/deck_file_namespace/

### Checklist 

- [x] Review label added 


app/_src/deck/reference/deck_file_namespace.md
```diff
@@ -0,0 +1,66 @@
...
+content_type: reference
+---
+
+Apply a namespace to routes in a decK file by prefixing the path.
+
+By prefixing paths with a specific segment, colliding paths to services can be
+namespaced to prevent collisions. For example, two API definitions that both expose a
+'/list' path. By prefixing one with '/addressbook' and the other with '/cookbook',
+the resulting paths '/addressbook/list' and '/cookbook/list' can be exposed without
+colliding.
```
lena-larionova, line 13:
```suggestion
namespaced to prevent collisions. For example, two API definitions that both expose a
`/list` path. By prefixing one with `/addressbook` and the other with `/cookbook`,
the resulting paths `/addressbook/list` and `/cookbook/list` can be exposed without
colliding.
```

app/_src/deck/reference/deck_file_namespace.md
```diff
@@ -0,0 +1,66 @@
...
+
+To remove the prefix from the path before the request is routed to the service, the
+following approaches are used:
+- If the route has 'strip_path=true,' then the added prefix will already be stripped.
+- If the related service has a 'path' property that matches the prefix, then the
+  'service.path' property is updated to remove the prefix.
+- A "pre-function" plugin will be added to remove the prefix from the path.
```
lena-larionova, line 20:
```suggestion
- If the route has `strip_path=true`, then the added prefix will already be stripped.
- If the related service has a `path` property that matches the prefix, then the
  `service.path` property is updated to remove the prefix.
- A `pre-function` plugin will be added to remove the prefix from the path.
```

lena-larionova (approved):
LGTM with some minor edits for formatting

PR #6974, opened 2024-02-19
fix: Add 3.6 conditional rendering for single backup nodes

cloudjumpercat:
### Description

We missed some conditional rendering for a note about single backup nodes. It should only apply to 3.6 and later? (waiting on confirmation for the correct version, so this PR might change a little)
 

https://kongstrong.slack.com/archives/CDSTDSG9J/p1708369514356029?thread_ts=1708368658.991799&cid=CDSTDSG9J
### Testing instructions

Preview link: 

### Checklist 

- [x] Review label added 


lena-larionova (approved):
LGTM based on what our conditional rendering supports. We can change it once we support `neq` tags.

Issue #6972, opened 2024-02-19
The "OR" operator in expression router reference page is not correctly displayed


randmonkey:
### Where is the problem?

https://docs.konghq.com/gateway/latest/reference/expressions-language/language-references/#operators

### What happened?

The `||` (Or) operator is displayed as `\|\|` in the table of operators. I searched in different versions of docs, the `OR` operator is displayed in `\|\|` format since version 3.6.0 of docs. While in the changelog of Kong gateway 3.6,0, I did not see if the format of `OR` operator changed. (If it is really changed, it should be a major BREAKING change). 

### What did you expect to happen?

The `||` (OR) Operator is displayed correctly.

### Code of Conduct and Community Expectations

- [X] I agree to follow this project's Code of Conduct
- [X] I agree to abide by the Community Expectations

PR #6977, opened 2024-02-20
Fix: OR operator hasn't changed in 3.6

lena-larionova:
### Description

The OR operator (`||`) hasn't changed in 3.6. This change was an accident, as GH-flavored markdown expects `|` characters to be escaped to properly display them, but kramdown/our docs renderer doesn't, so it prints the syntax literally as `\|\|`. Erring in favor of the prod docs (which will mean that it displays incorrectly in GH markdown, but that shouldn't matter).

Fixes https://github.com/Kong/docs.konghq.com/issues/6972.

### Testing instructions

Preview link: 

### Checklist 

- [x] Review label added 


lena-larionova:
Closing as duplicate.

PR #6981, opened 2024-02-21
Experiment release

lahabana:
### Testing instructions

Preview link: 

### Checklist 

- [ ] Review label added 

### DIFF
--- a/app/_data/kong_versions.yml
+++ b/app/_data/kong_versions.yml
@@ -408,6 +408,6 @@
   branch: release-2.6
 - edition: mesh
   version: preview
-  release: 2.7.0
+  release: 2.7.x
   branch: master
   label: dev


mheap:
Almost. `version` should be 2.7.0 rather than preview. This allows if_version to work on prerelease docs

lahabana:
What I'm afraid of is that we use `page.version` to retrieve the `preview` version in install commands (.e.g: https://deploy-preview-6981--kongdocs.netlify.app/mesh/dev/production/install-kumactl/).

If I use 2.7.0 there it will just not work anymore and we'll have not working install commands...

mheap:
You're right 🤦 

This PR will work as expected

fabianrbz (approved):

PR #6983, opened 2024-02-21
Fix: replace benchmark redirect with moved url entry

lena-larionova:
### Description

This URL got flagged as 404ing at /latest/ by google console. Did some digging, realized that it should probably be in moved URLs instead anyway, based on the following rule:
* When a topic is moved and not removed or renamed, we should add an entry to moved URLs instead of redirects.

This _should_ generate the correct redirects automatically.

### Testing instructions

Preview link: 
https://deploy-preview-6983--kongdocs.netlify.app/gateway/3.6.x/production/benchmark/
https://deploy-preview-6983--kongdocs.netlify.app/gateway/latest/production/benchmark/

### Checklist 

- [x] Review label added 

### DIFF
--- a/app/_redirects
+++ b/app/_redirects
@@ -83,7 +83,6 @@
 
 # 3.6
 /gateway/3.6.x/production/benchmark/  /gateway/3.6.x/production/performance/benchmark/
-/gateway/latest/production/benchmark/ /gateway/latest/production/performance/benchmark/
 
 # 3.5
 /gateway/latest/kong-enterprise/analytics/* /gateway/3.4.x/kong-enterprise/analytics/:splat
--- a/app/_redirects
+++ b/app/_redirects
@@ -81,6 +81,10 @@
 /gateway/3.5.x/admin-api/admins/reference/   /gateway/api/admin-ee/latest/#/admins/get-admins/
 /gateway/3.5.x/admin-api/*                   /gateway/api/admin-ee/latest/
 
+# 3.6
+/gateway/3.6.x/production/benchmark/  /gateway/3.6.x/production/performance/benchmark/
+/gateway/latest/production/benchmark/ /gateway/latest/production/performance/benchmark/
+
 # 3.5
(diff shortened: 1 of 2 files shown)


lena-larionova:
This doesn't seem to actually work 🤔 The URL just 404s:
https://deploy-preview-6983--kongdocs.netlify.app/gateway/3.6.x/production/benchmark/
https://deploy-preview-6983--kongdocs.netlify.app/gateway/latest/production/benchmark/

It only redirects if you click the "latest documentation here" link in the banner of v3.5.x: https://deploy-preview-6983--kongdocs.netlify.app/gateway/3.5.x/production/benchmark/

fabianrbz:
The `latest documentation here` works because it points to the right URL.

This one is working for me:
https://deploy-preview-6983--kongdocs.netlify.app/gateway/latest/production/benchmark/

This one doesn't 
https://deploy-preview-6983--kongdocs.netlify.app/gateway/3.6.x/production/benchmark/
because of the way moved URLs work, we [just replace](https://github.com/Kong/docs.konghq.com/blob/main/app/_plugins/generators/alias_generator.rb#L78-L82) `VERSION` with latest.

So we probably need to add the redirect for 3.6.x.

app/_redirects
```diff
@@ -82,7 +82,8 @@
 /gateway/3.5.x/admin-api/*                   /gateway/api/admin-ee/latest/
 
 # 3.6
-/gateway/3.6.x/production/benchmark /gateway/3.6.x/production/performance/benchmark
+/gateway/3.6.x/production/benchmark/  /gateway/3.6.x/production/performance/benchmark/
+/gateway/latest/production/benchmark/ /gateway/latest/production/performance/benchmark/
```
fabianrbz, line 86: I think you can drop this one
```
/gateway/latest/production/benchmark/ /gateway/latest/production/performance/benchmark/
```
it should be handled by the moved_urls

fabianrbz (approved):

PR #6984, opened 2024-02-21
Add documentation on higher ulimit requirement for 3.6.0.0

Achiel:
### Description

Document missing entry for ulimit breaking change.

### Testing instructions

Preview link: https://deploy-preview-6984--kongdocs.netlify.app/gateway/latest/breaking-changes/

### Checklist 

- [x] Review label added 

### DIFF
--- a/app/_src/gateway/breaking-changes/36x.md
+++ b/app/_src/gateway/breaking-changes/36x.md
@@ -16,6 +16,9 @@ Review the [changelog](/gateway/changelog/#3600) for all the changes in this rel
 ## General
 If you are using `ngx.var.http_*` in custom code to access HTTP headers, the behavior of that variable changed slightly when the same header is used multiple times in a single request. Previously, it would return the first value only; now it returns all the values, separated by commas. {{site.base_gateway}}'s PDK header getters and setters work as before.
 
+## Operating system requirements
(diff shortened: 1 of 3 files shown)


lena-larionova (approved):
I did some minor cleanup and applied it directly, since the changes are small. LGTM and thanks for making the update!

app/_src/gateway/breaking-changes/36x.md
```diff
@@ -16,6 +16,11 @@ Review the [changelog](/gateway/changelog/#3600) for all the changes in this rel
 ## General
 If you are using `ngx.var.http_*` in custom code to access HTTP headers, the behavior of that variable changed slightly when the same header is used multiple times in a single request. Previously, it would return the first value only; now it returns all the values, separated by commas. {{site.base_gateway}}'s PDK header getters and setters work as before.
 
+## Operating system requirements
+{{site.base_gateway}} 3.6 requires a higher [ulimit](https://linux.die.net/man/3/ulimit) to function properly. It will not start properly with a ulimit set to 1024 or lower. We recommend setting the ulimit for your operating system to at least 4096.
```
gszr, line 20: This is inaccurate; `ulimit` is a tool to get and set limits -- it is not a limit per se. We want to say something like
```suggestion
{{site.base_gateway}} 3.6 requires a higher limit on the number of file descriptions to function properly. It will not start properly with a limit set to 1024 or lower. We recommend using `ulimit` on your operating system to set it to at least 4096 using `ulimit -n 4096`.
```

gszr: same applies to other places below

gszr: cc @lena-larionova

PR #6996, opened 2024-02-26
Updated example of Auth0 configuration _auth0.md

motaparthipavankumar:
Auth0 configurations example is misleading in docs, updated the doc to remove properties that are not needed and added the important property that is needed.

### Description

Auth0 configurations example is misleading in docs, updated the doc to remove unwanted properties and added the important property that is needed.

### Testing instructions

Preview link: 

### Checklist 

- [ ] Review label added 


lena-larionova (approved):
Nice catch! LGTM

PR #7014, opened 2024-02-29
Fix: Add workspace RBAC role creation [skip-ci]

Guaris:
The API documentation for creating RBAC roles here:
https://konghq.atlassian.net/browse/DOCU-3344
 [RBAC Reference - Kong Gateway - v3.3.x | Kong Docs](https://docs.konghq.com/gateway/3.3.x/admin-api/rbac/reference/#add-a-role) only provides details on how to create RBAC roles in the default workspace:

`POST /rbac/roles`
however it would be more correct to confirm that roles for workspaces can be created with:

`POST /{workspace_name_or_id}/rbac/roles`
As currently this is not clear without experimentation or inferring this from the later methods describe on this page.

### DIFF
--- a/api-specs/Gateway-EE/latest/kong-ee.yaml
+++ b/api-specs/Gateway-EE/latest/kong-ee.yaml
@@ -14388,6 +14388,140 @@ paths:
       operationId: get-event-hooks-event-hook-id-ping
       description: |
         Ping a webhook event hook.
+  '/{workspace_name_or_id}/rbac/roles':
+    parameters:
+      - schema:
+          type: string
+        name: workspace_name_or_id
+        in: path
+        required: true
+        description: The workspace name or UUID.
+    get:
+      summary: List roles for a workspace
+      tags: 
+        - Workspaces
+      responses:
+        '200': 
+          description: OK
+          content:
+            application/json:
+              schema:
+                type: object
+                properties:
+                  data:
+                    type: array
+                    items:
+                      type: object
+                      properties:
+                        comment:
+                          type: string
+                        created_at:
+                          type: integer
+                        id:
+                          type: string
+                        name:
+                          type: string
+                  next:
+                    type: string
+                x-examples:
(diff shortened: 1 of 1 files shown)


api-specs/Gateway-EE/latest/kong-ee.yaml
```diff
@@ -14388,6 +14388,140 @@ paths:
       operationId: get-event-hooks-event-hook-id-ping
       description: |
         Ping a webhook event hook.
+  '/{workspace_name_or_id}/rbac/roles':
```
lena-larionova, line 14391: Similarly, for consistency with the rest of the spec, should these endpoints be under RBAC instead of under Workspaces? I don't know which one is optimal organization, but currently we do things like `/{workspace}/certificates` under Certificates, and so on.

lena-larionova:
I found another ticket that is _almost_ the same, but goes a bit deeper:
https://konghq.atlassian.net/browse/DOCU-751 

I believe that translates to 
`/{workspace}/rbac/roles/{role}/endpoints/{workspace}/`

api-specs/Gateway-EE/latest/kong-ee.yaml
```diff
@@ -14388,6 +14388,204 @@ paths:
...
+                        created_at: 1557506249
+                        id: 888117e0-f2b3-404d-823b-dee595423505
+                        name: super-admin
+                      - comment: 'null'
+                        created_at: 1557532241
+                        id: b5c5cfd4-3330-4796-9b7b-6026e91e3ad6
+                        name: doc_lord
+                    next: 'null'
+      operationId: get-rbac-roles-by-workspace
+      description: Listr all roles by workspace
```
lena-larionova, line 14468:
```suggestion
      description: List all roles by workspace
```

lena-larionova (approved):
LGTM, just one typo.

PR #7019, opened 2024-02-29
fix: Typos in oidc for kong manager doc

lena-larionova:
### Description

Noticed a double `:` in the doc --> fixed it --> saw some formatting issues in the doc while fixing --> fixed those too.

### Testing instructions

Preview link: 

### Checklist 

- [x] Review label added 


acgoldsmith (approved):
LGTM

PR #7022, opened 2024-03-01
Feat: Convert navbar to dark theme

lena-larionova:
### Description

Converting navbar to dark theme and using new logos. 

I left the blue headers as-is mostly, and just had them fade into the navbar - I think that looks pretty good for the in-between stage.

I have asked the design team for what they expect from the expanded menu as well - leaving it as white for now, until they respond.

@fabianrbzg there's couple of things I couldn't figure out here, if you could take a look:
* Why the hover/interact state for the search bar looks... not so great. But I don't know how to find the source of the variable it's using for the hover state - looks like we're pulling it from somewhere else?
* Some issues with the mobile nav (see video) - can't figure out why the slideout overlaps with the nav while it's sliding:

https://github.com/Kong/docs.konghq.com/assets/54370747/29846a70-9f93-47b5-ac91-15d2d094df3d

https://konghq.atlassian.net/browse/DOCU-3681

### Testing instructions

Preview link: check a bunch of pages. For example:

https://deploy-preview-7022--kongdocs.netlify.app/
https://deploy-preview-7022--kongdocs.netlify.app/hub/
https://deploy-preview-7022--kongdocs.netlify.app/api/
https://deploy-preview-7022--kongdocs.netlify.app/konnect/

### Checklist 

- [x] Review label added 


fabianrbz:
@lena-larionova I fixed the mobile issues, and yes, the styles of the search button come from Algolia.
@mheap I changed the styles of the two buttons, we might want to change the "hover" of the blue button though, unfortunately, it's not specified in Figma. Should we treat them like the ones in konghq?


https://github.com/Kong/docs.konghq.com/assets/715229/2f060fac-22cc-4ef9-9fa9-9f5835d1987a

lena-larionova:
@fabianrbz I think we might be good here, if you are.

fabianrbz (approved):

PR #7029, opened 2024-03-04
fix: Updated HTTP method for setting targets health

dascole:
### Description

Updated the HTTP method from POST to PUT for setting targets health. The current doc references a POST method which results in a 405 - Method not allowed error.
 
### Testing instructions

Preview link: 

### Checklist 

- [x] Review label added 


cloudjumpercat (approved):
Looks good to me! Thanks for making this update!

PR #7034, opened 2024-03-04
Fix: Plugin name and instance_name descriptions

lena-larionova:
### Description

Fixing some minor issues with plugin schema & example docs:
* Description for the `name` field
* Description for the `instance_name` field
* Outdated links to the Admin API from the plugin config descriptions and plugin examples
* Standardize name variables in API URLs (`{serviceName}` etc)

https://konghq.atlassian.net/browse/DOCU-1498
https://konghq.atlassian.net/browse/DOCU-3412
https://konghq.atlassian.net/browse/DOCU-3706

### Testing instructions

Preview link: 

### Checklist 

- [x] Review label added 

### DIFF
--- a/app/_includes/plugins/hub-examples/consumer_groups.html
+++ b/app/_includes/plugins/hub-examples/consumer_groups.html
@@ -4,7 +4,7 @@
 {% if include.render_intro_text %}
 The following examples provide some typical configurations for enabling
 the <code>{{ hub_example.plugin_name }}</code> plugin on a
-[consumer group](/gateway/latest/admin-api/consumer-groups/reference/).
+[consumer group](/gateway/api/admin-ee/latest/#/consumer_groups).
 {% endif %}
 
 {% navtabs %}
--- a/api-specs/Gateway-EE/latest/kong-ee.yaml
+++ b/api-specs/Gateway-EE/latest/kong-ee.yaml
@@ -1432,7 +1432,13 @@ components:
               instance_name:
                 type: string
                 description: |
-                  The plugin instance name.
+                  An optional custom name to identify an instance of the plugin, for example `basic-auth_example-service`.
+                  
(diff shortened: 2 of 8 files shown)


app/_includes/plugins/hub-examples/consumer_groups.html
```diff
@@ -68,7 +68,7 @@
     {% if hub_example.render_yaml? %}
     {% navtab Declarative (YAML) %}
 
-Add this section to your declarative configuration file:
+Add this section to your declarative configuration or decK file:
```
mheap, line 71: Is this true? Consumer groups are different between declarative config + deck

lena-larionova: Hmm. What does it looks like with declarative config? I think all of these examples are more accurate written for decK.

I wonder if it would make more sense to just label them "yaml" like the UI does.

lena-larionova: Rolling this change back, will fix separately. There's a few other places in the docs where we conflate decK/declarative config, so I'd rather handle them together.

lena-larionova: I don't currently know how to set up a db-less instance (we don't document it with docker, but I'm sure there's a way 🤔) so I wasn't able to test originally. But I'll give it a shot and try to figure it out separately from this PR.

mheap:
```
curl -Ls https://get.konghq.com/quickstart | bash -s -- -D
curl -X POST http://localhost:8001/config -d "$(cat config.yaml)"
```

The `-D` flag in quickstart means DBless

acgoldsmith (approved):
LGTM

PR #7036, opened 2024-03-04
Fix: rephrase 'sugar parameter' into more common terms

lena-larionova:
### Description

Adjusting the phrasing to remove confusing term.

https://konghq.atlassian.net/browse/DOCU-3707

### Testing instructions

Preview link: 

### Checklist 

- [x] Review label added 


acgoldsmith (approved):
LGTM - sweet papercut to fix ;)

PR #7044, opened 2024-03-05
KGO managed gateways

mlavacca:
### Testing instructions

Preview link: 

### Checklist 

- [x] Review label added 

### DIFF
--- a/app/_src/gateway-operator/concepts/gateway-configuration.md
+++ b/app/_src/gateway-operator/concepts/gateway-configuration.md
@@ -8,6 +8,32 @@ These customizations are primarily used to set the container image and any envir
 
 Here is an example of the `GatewayConfiguration` that provides an enterprise license for {{ site.base_gateway }}.
 
+{% if_version lte: 1.1.x %}
+
+```yaml
+kind: GatewayConfiguration
+apiVersion: gateway-operator.konghq.com/v1alpha1
+metadata:
+  name: kong
+  namespace: <your-namespace>
+spec:
+  dataPlaneOptions:
+    deployment:
+      podTemplateSpec:
+        spec:
+          containers:
+          - name: proxy
+            image: kong/kong-gateway:{{ site.data.kong_latest_gateway.ee-version }}
+          env:
+          - name: KONG_LICENSE_DATA
+            valueFrom:
+              secretKeyRef:
+                key: license
+                name: kong-enterprise-license
+```
+
+{% else %}
+
 ```yaml
 kind: GatewayConfiguration
 apiVersion: gateway-operator.konghq.com/v1beta1
@@ -29,5 +55,7 @@ spec:
                 key: license
                 name: kong-enterprise-license
 ```
+
+{% endif_version %}
 
 For more information about `GatewayConfiguration` see the [GatewayConfiguration CRD reference](/gateway-operator/{{ page.release }}/reference/custom-resources#gatewayconfiguration).
--- a/app/_src/gateway-operator/concepts/gateway-configuration.md
+++ b/app/_src/gateway-operator/concepts/gateway-configuration.md
@@ -10,7 +10,7 @@ Here is an example of the `GatewayConfiguration` that provides an enterprise lic
 
 ```yaml
 kind: GatewayConfiguration
-apiVersion: gateway-operator.konghq.com/v1alpha1
+apiVersion: gateway-operator.konghq.com/v1beta1
 metadata:
   name: kong
   namespace: <your-namespace>
--- a/app/_src/gateway-operator/concepts/gateway-configuration.md
+++ b/app/_src/gateway-operator/concepts/gateway-configuration.md
@@ -2,17 +2,20 @@
 title: Gateway Configuration
 ---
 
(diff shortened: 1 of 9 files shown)


app/_src/gateway-operator/concepts/gateway-configuration.md
```diff
@@ -10,7 +10,7 @@ Here is an example of the `GatewayConfiguration` that provides an enterprise lic
 
 ```yaml
 kind: GatewayConfiguration
-apiVersion: gateway-operator.konghq.com/v1alpha1
+apiVersion: gateway-operator.konghq.com/v1beta1
```
pmalek, line 13: This has to have a version conditional around it right?

app/_src/gateway-operator/concepts/managed-gateways.md
```diff
@@ -0,0 +1,14 @@
+---
+title: Managed Gateways
+---
+
+{{ site.kgo_product_name }} handles `Gateway`s in the so-called managed mode, which implies the following characteristics:
+
+## 1:1 relationship between `Gateway`s and `ControlPlane`s
+
+When {{ site.kgo_product_name }} detects a new `Gateway`, it creates a `ControlPlane` and a `DataPlane`. The `ControlPlane` is a {{ site.kic_product_name }} instance configured with a specific flag that allows it to reconcile and manage one `Gateway` only, which is bound to.
```
pmalek, line 9:
```suggestion
When {{ site.kgo_product_name }} detects a new `Gateway`, it creates a `ControlPlane` and a `DataPlane`. The `ControlPlane` is a {{ site.kic_product_name }} instance configured with a specific flag that allows it to reconcile and manage one `Gateway` only, which it is bound to.
```

app/_src/gateway-operator/customization/data-plane-image.md
```diff
@@ -37,7 +37,7 @@ You can customize both the container image and version.
 1.  Define the image in the `GatewayConfiguration`.
     ```yaml
     kind: GatewayConfiguration
-    apiVersion: gateway-operator.konghq.com/v1alpha1
+    apiVersion: gateway-operator.konghq.com/v1beta1
```
pmalek, line 40: Also a version conditional here.

app/_src/gateway-operator/production/upgrade/data-plane/rolling.md
```diff
@@ -24,7 +24,7 @@ The `GatewayConfiguration` API can be used to provide the image and the image ve
 
 ```yaml
 kind: GatewayConfiguration
-apiVersion: gateway-operator.konghq.com/v1alpha1
+apiVersion: gateway-operator.konghq.com/v1beta1
```
pmalek, line 27: Ditto.

pmalek (commented):
I'm not sure if the version conditionals are required but I'd add them so that we're explicitly using the API version that we used in associated KGO verison. WDYT?

mlavacca:
> I'm not sure if the version conditionals are required but I'd add them so that we're explicitly using the API version that we used in associated KGO verison. WDYT?

Yes, I think it makes sense. I'll add them

app/_src/gateway-operator/customization/data-plane-image.md
```diff
@@ -2,6 +2,11 @@
 title: Customizing the Data Plane image
 ---
 
+{% assign gatewayConfigApiVersion = "v1beta1" %}
+{% if_version lte:1.1.x %}
+{% assign gatewayConfigApiVersion = "v1alpha1" %}
+{% endif_version %}
```
pmalek, line 8: Nit: can this be placed in a reusable import somewhere?

pmalek (approved):

app/_src/gateway-operator/customization/data-plane-image.md
mheap: A proposal that I sent to docs earlier today:

I'm using the following pattern more and more in the KIC docs as we migrate from alpha to beta versions of CRDs:
```
{% assign gatewayConfigApiVersion = "v1beta1" %}
{% if_version lte:1.1.x %}
{% assign gatewayConfigApiVersion = "v1alpha1" %}
{% endif_version %}
```

I've just had to add it to three different pages which feels wrong to me, so I've been thinking of alternatives.
What if we had variable configs somewhere in the site? Maybe something like this:

```
# app/_data/product_vars/gateway-operator.yaml

- name: gatewayConfigApiVersion
  default: v1beta1
  overrides:
    - constraint: "lte:1.1.x"
      value: v1alpha1
```

Then I could call `page.vars.gatewayConfigApiVersion` in the template (single source plugin would populate `page.vars` based on constraints) and know that it would just work.

mlavacca: I've applied the change you suggested, but the version is not rendered at all. I suppose there is some piece missing, do you have any clue @mheap?

mheap: Ah! This doesn't actually work. It's a proposal to solve the duplication. For now we have to do the `assign` in every file

mlavacca:
@mheap, I'd need your explicit approval for this PR to be merged

PR #7045, opened 2024-03-05
fix: decK file merge example

lena-larionova:
### Description

`deck file merge` example uses `deck file patch`, where it should be `merge`. 

Parallel PR in decK repo: https://github.com/Kong/deck/pull/1237

### Testing instructions

Preview link: 

### Checklist 

- [x] Review label added 


cloudjumpercat (approved):
LGTM! I'll let you merge this one since I think it might depend on the ticket in the decK repo

lena-larionova:
decK repo ticket is also merged already, so this is going in.

PR #7050, opened 2024-03-06
Fix: Removes Proxy and test a service doc

Guaris:
Removes [proxy and test a service](https://docs.konghq.com/konnect/gateway-manager/deploy-service/). 

The information available in that doc is now better served by [add-api](https://docs.konghq.com/konnect/getting-started/add-api/).There is no clear purpose for for the proxy and test a service doc.


cloudjumpercat (approved):
LGTM!

PR #7052, opened 2024-03-06
chore: Adjust landing page card styling

lena-larionova:
### Description

Adjust font colours and sizing to fit with the layout better (following figma). 
Since we adjusted some things but not others, the font colours and sizing now stand out awkwardly, not looking quite right.
One exception: switched the header subtitle to white instead of the text colour in Figma, since we've kept the blue header.

Other things:
* One of the tickets that got closed with the landing page update was to fix header order for accessibility. That wasn't quite completed in the previous PR, so fixing that. 
* Set the heading subtitle to h2 to match plugin hub, and the subtitle text to do the same.
* Using the script icon.

### Testing instructions

@fabianrbz We've recently received a bunch of feedback around findability of various things - what do you think about adding two links below the subtitle here: [Install](/install/) and [Advanced Search](/search/)? The formatting would be the same as the plugin hub links to Overview, Compatibility, and License Tiers. Don't know if we'd need to adjust the header formatting overall to make that work though.

Preview link: https://deploy-preview-7052--kongdocs.netlify.app/

### Checklist 

- [x] Review label added 


fabianrbz (approved):

fabianrbz:
thanks @lena-larionova ! We can add the links to the subtitle, it doesn't require a lot of changes though, I can do that after we ship the install page.

lena-larionova:
> We can add the links to the subtitle, it doesn't require a lot of changes though, I can do that after we ship the install page.

Let me throw that into a ticket for tracking purposes, then.

PR #7053, opened 2024-03-07
Update default-ports.md

akoshok1:
### Description

Added the default status listen port.  It is 8007 now, but used to be 8100 in previous releases.  Might be worthwhile to find out when this change was made.

### Testing instructions

Preview link: https://deploy-preview-7053--kongdocs.netlify.app/gateway/latest/production/networking/default-ports/

### Checklist 

- [x] Review label added 


lena-larionova:
Looks like it was just introduced in 3.6, and was actually nothing in earlier releases (default was `off`): https://github.com/Kong/kong/pull/12254

lena-larionova (approved):
Had to fix the table formatting, it was very broken - and we had missed the Dev Portal ports, looks like. Should be good to go now, LGTM! Thanks for catching the missing port.

PR #6721, opened 2024-01-05
(Feat) add support for Azure Linux to 3.5 matrix

Guaris:
We now support Azure Linux for Gateway 3.5

reference: 
https://github.com/Kong/docs.konghq.com/pull/6587
https://github.com/Kong/docs.konghq.com/pull/6720

### Testing instructions

Preview link: https://deploy-preview-6587--kongdocs.netlify.app/gateway/latest/support-policy/

### Checklist 

- [x] Review label added 
- [x] PR pointed to correct branch (`main` for immediate publishing, or a release branch: e.g. `release/gateway-3.2`, `release/deck-1.17`)

PR #7055, opened 2024-03-07
(fix) remove package tag from Azure Linux

acgoldsmith:
There is no expectation of having an Azure Linux download package, was erroneously added as part of initial DOCU-3595 update. 

### Testing instructions

Preview link: 

### Checklist 

- [ ] Review label added 

### DIFF
--- a/app/_data/tables/support/gateway/packages.yml
+++ b/app/_data/tables/support/gateway/packages.yml
@@ -31,7 +31,7 @@ _packages:
   azurelinux2xx: &azurelinux2xx
     os: Azure Linux
     version: 2.x.x
-    package: true
+    package: false
     arm: false
     fips: false
     graviton: false


app/_data/tables/support/gateway/packages.yml
```diff
@@ -31,7 +31,7 @@ _packages:
   azurelinux2xx: &azurelinux2xx
```
lena-larionova, line 31: If I understand this correctly, "Azure Linux" support means that Kong was added to the Azure marketplace, which is something that was driven by the partneships team. There should not be an entry in our packages table, so we should remove the whole entry + the entries in the other data files (see original change here: https://github.com/Kong/docs.konghq.com/pull/6721/files) 

We might actually be able to just revert https://github.com/Kong/docs.konghq.com/pull/6721, more accurately.

acgoldsmith: Updating based on our convo that we're going to hold this PR until we've identified new location for the info, tied to a wider conversation about partnership support content. Thanks for great catch and discussion!

lena-larionova (approved):

PR #7107, opened 2024-03-20
docs(api): clarify behaviour of uuid as names

nowNick:
### Description

This PR adds a clarification after fixing a bug in admin API: 

https://github.com/Kong/kong/pull/12734

### Testing instructions

Preview link: 

Visit the API docs and verify if the text is displayed correctly.

### Checklist 

- [ ] Review label added 

### DIFF
--- a/api-specs/Gateway-EE/latest/kong-ee.yaml
+++ b/api-specs/Gateway-EE/latest/kong-ee.yaml
@@ -6981,6 +6981,11 @@ paths:
     put:
       description: |
         Create or update a key using ID or name.
+        Please note that if you used uuid as key name and would like to
+        update it you need to use such key ID. Using this uuid-name will
+        result in an upsert of a new key with this uuid set as ID (or fail with
+        collision error). If the name is a regular string you can use either
+        name or ID.
       operationId: upsert-key
       responses:
         '200':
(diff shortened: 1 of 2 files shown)


api-specs/Gateway-EE/latest/kong-ee.yaml
```diff
@@ -9804,7 +9869,13 @@ paths:
...
       operationId: update-route-with-service-workspace
     put:
-      description: Create or Update a route associated with a Service using ID or name in a workspace.
+      description: |
+        Create or Update a route associated with a Service using ID or name in a workspace.
+        Please note that if you used uuid as route name and would like to
+        update it you need to use such route ID. Using this uuid-name will
+        result in an upsert of a new route with this uuid set as ID (or fail with
+        collision error). If the name is a regular string you can use either
+        name or ID.
```
Guaris, line 9878:
```suggestion
      description: |
        Create or Update a route associated with a Service using ID or name in a workspace.
        Please note that if you used uuid as route name and would like to
        update it you need to use such route ID. Using this uuid-name will
        result in an upsert of a new route with this uuid set as the ID (or fail with
        collision error). If the name is a regular string you can use either
        name or ID.
```

Guaris (approved):

nowNick:
We decided to drop the changes in the linked PR. The corner case it tries to cover brings unnecessary complexity and we have not heard of a single user hitting this case.

PR #7115, opened 2024-03-21
Fix: Generate 3.6 Gateway kong.conf reference

lena-larionova:
### Description

Regenerating the configuration reference for Gateway based on latest 3.6.1.1 patch. 
We didn't generate it with the release and there were significant changes to the file in that version. 
Now that these files are split out by version, this is easier to see. 

Also making some tweaks to run.lua and data.lua to use the same text that we have in the page description, and to remove a badge that doesn't need to exist.

Fixes https://konghq.atlassian.net/browse/DOCU-2889.

Will do the same for 3.5 with upcoming patch release (should be end of next week).

### Testing instructions

Preview link: https://deploy-preview-7115--kongdocs.netlify.app/gateway/latest/reference/configuration/

### Checklist 

- [x] Review label added 


app/_src/gateway/reference/configuration/configuration-3.6.x.md
```diff
@@ -972,6 +972,8 @@ Defines the TLS ciphers served by Nginx.
 
 Accepted values are `modern`, `intermediate`, `old`, `fips` or `custom`.
 
+If you want to enable TLSv1.1, this value has to be `old`.
```
cloudjumpercat, line 975:
```suggestion
If you want to enable TLSv1.1, this value must be `old`.
```

app/_src/gateway/reference/configuration/configuration-3.6.x.md
```diff
@@ -997,9 +999,11 @@ Enables the specified protocols for client-side connections. The set of
 supported protocol versions also depends on the version of OpenSSL Kong was
 built with. This value is ignored if `ssl_cipher_suite` is not `custom`.
 
+If you want to enable TLSv1.1, you should set `ssl_cipher_suite` to `old`.
```
cloudjumpercat, line 1002:
```suggestion
If you want to enable TLSv1.1, you must set `ssl_cipher_suite` to `old`.
```
Maybe? Just to match the previous wording where it was a requirement.

cloudjumpercat (approved):
Left a few small suggestions, but everything LGTM!

app/_src/gateway/reference/configuration/configuration-3.6.x.md
lena-larionova: I have to ignore these unfortunately, since this is a generated file :(

PR #7122, opened 2024-03-22
(fix) Update auth0 instructions to include warning on trailing forward slash in URL

motaparthipavankumar:
Updated the Issuer URL information, as providing a trailing slash will cause issues.

### Description

When we have trailing slash for issuer URL, it get concatenated which will result in issuer as this https://dev-26uur0y8i4n8x3ez.us.auth0.com//api/v2/". So, informing users via docs will be valuable until if this can be gracefully handled.

### Testing instructions

Following steps mentioned in this docs and just have additional change of including / at the end of issuer url. And try to create an application in Dev Portal, which will fail with "Unable to create one time access token" error.


acgoldsmith (approved):
LGTM

PR #7160, opened 2024-04-03
plugins(jwt): support more algorithms

jschmid1:
### Description

Update docs to reflect recent changes in the JWT plugins.

Added support for:
* ES512
* PS256
* PS384
* PS512
* EdDSA

@kong-docs Should we restructure the docs a bit to use `Conditional version tags` more effectively?

### Checklist 

- [x] Review label added 

Jira: https://konghq.atlassian.net/browse/KAG-4029

### DIFF
--- a/app/_hub/kong-inc/jwt/overview/_index.md
+++ b/app/_hub/kong-inc/jwt/overview/_index.md
@@ -101,8 +101,8 @@ field/parameter                | default         | description
 ---                            | ---             | ---
 `consumer`                   |                 | The `id` or `username` property of the [consumer][consumer-object] entity to associate the credentials to.
 `key`<br>*optional*            |                 | A unique string identifying the credential. If left out, it will be auto-generated.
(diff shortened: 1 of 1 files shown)


fabianrbz:
@jschmid1 we definitely can, what do you have in mind? What issues are you facing?

jschmid1:
@fabianrbz The list of supported algorithms is dependent on the kong version. Currently the list is maintained as an inline string which makes it awkward to conditionally render this list. How about maintaining the list somewhere and render the string containing the supported algorithms? Alternatively, we could pull this information from the schema description?

fabianrbz:
oh! Yeah, that would be great, I didn't realize we were updating that table manually.
Unfortunately, I don't think those fields are returned in the API response.
It looks like they are not defined in the schema.lua  but in a [daos.lua](https://github.com/Kong/kong/blob/bff6515a7d8922bf0cc6345f2e1b14c201e098a1/kong/plugins/jwt/daos.lua#L22-L47)

jschmid1:
This can expose it `http :8001/schemas/jwt_secrets` Would that be possible?

fabianrbz:
I've created a Jira ticket to track that work, we should review/merge this PR in the meantime though

lena-larionova (approved):
Added some conditional version tagging, LGTM.

PR #7165, opened 2024-04-03
Chore: Browser support rework

lena-larionova:
### Description

Update the browser version support policy page to a more general statement, as we don't test _specific_ versions, we only guarantee latest.

@veenaraja please let me know if this is sufficient. We're required to list support for at least the two latest versions of each browser.

https://konghq.atlassian.net/browse/DOCU-3689

### Testing instructions

Preview link: 

### Checklist 

- [x] Review label added 


app/_src/gateway/support/browser.md
```diff
@@ -3,26 +3,12 @@ title: Browser Support
...
-[Edge](https://blogs.windows.com/msedgedev/2021/07/15/opt-in-extended-stable-release-cycle/), [Chrome](https://www.chromium.org/chrome-release-channels/) and [Firefox](https://support.mozilla.org/en-US/kb/switch-to-firefox-extended-support-release-esr) release a new major version every 4 weeks, with extended support available for 8 weeks.
+Kong supports the two latest stable versions and any extended support versions of the following "evergreen" desktop browsers:
 
-Kong supports N-1 versions of Edge, Chrome, Firefox and Safari on desktop plus any extended support versions.
+{% include_cached gateway-support-browsers.html %}
+
+[Chrome](https://www.chromium.org/chrome-release-channels/), 
+[Firefox](https://support.mozilla.org/en-US/kb/switch-to-firefox-extended-support-release-esr), and 
+[Edge](https://blogs.windows.com/msedgedev/2021/07/15/opt-in-extended-stable-release-cycle/) release a new major version every 4 weeks, 
+with extended support available for 8 weeks.
```
cloudjumpercat, line 13: Is the extended support on the browser's end or ours? Ex. is it Kong that offers extended support for 8 weeks?

lena-larionova: It's the browser's support.

lena-larionova: I looked at this some more and considered flipping this sentence back up to the top, but it doesn't mention Safari so that part is confusing. There is no doc for Safari fyi, because Apple.

PR #7169, opened 2024-04-04
Fix: Yellow -> Green in control plane group diagram description 

Guaris:
Diagram says green, text says yellow.


mheap:
Closing in favour of 7171

PR #7199, opened 2024-04-10
Fix {:.warning} in Kong mesh

fabianrbz:
### Description

The markup we used in the tag was outdated. Change it so that a note is render and update the type of note so that it matches the one in Kuma. (the use a different system)

### Testing instructions

Preview link: https://deploy-preview-7199--kongdocs.netlify.app/mesh/latest/reference/kubernetes-annotations/#kumaiodirect-access-services

### Checklist 

- [x] Review label added 


lahabana (approved):
Thx for the quick fix here!

slonka (approved):

PR #7205, opened 2024-04-10
kic: fix version compat table for v3 versions

pmalek:
### Description

For some reason the version compatibility matrix at https://docs.konghq.com/kubernetes-ingress-controller/3.1.x/reference/version-compatibility/#general

doesn't show up properly.

image

This PR fixes that. I'm not 100% sure of the reason for the broken rendering but extracting the old version compat matrix from v2 tree did help.

It also cleans up the file a bit and removes old k8s versions which are way out of support as of now.

### Testing instructions

https://deploy-preview-7205--kongdocs.netlify.app/kubernetes-ingress-controller/latest/reference/version-compatibility/#kubernetes

### Checklist 

- [x] Review label added 

### DIFF
--- a/app/_data/docs_nav_kic_3.0.x.yml
+++ b/app/_data/docs_nav_kic_3.0.x.yml
@@ -189,7 +189,6 @@ items:
         url: /reference/troubleshooting
       - text: Version Compatibility
         url: /reference/version-compatibility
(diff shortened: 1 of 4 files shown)


lena-larionova (approved):
Truly strange, but looks like it worked. lgtm!

PR #7234, opened 2024-04-15
fix kong mesh navigation

Automaat:
### Description
Fix to showing correct Kong Mesh versions in docs

### Testing instructions

Preview link: 

### Checklist 

- [x] Review label added 

### DIFF
--- a/app/_data/docs_nav_mesh_dev.yml
+++ b/app/_data/docs_nav_mesh_dev.yml
@@ -4,7 +4,7 @@ release: 2.7.x
 max_depth: 3
 inherit:
   path: /.repos/kuma/app/_src
-  nav: ../_src/.repos/kuma/app/_data/docs_nav_kuma_2.7.x.yml
+  nav: ../_src/.repos/kuma/app/_data/docs_nav_kuma_dev.yml
   patches:
     - path: [ Introduction ]
       action: modify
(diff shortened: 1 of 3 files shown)


fabianrbz (approved):

app/_data/docs_nav_mesh_dev.yml
fabianrbz, line 1: this file should be renamed to `app/_data/docs_nav_mesh_2.8.x.yml` (assuming 2.8.x is the next version)
also, `release: 2.7.x` should be updated

Automaat: changed

app/_data/kong_versions.yml
```diff
@@ -389,7 +389,11 @@
   endOfLifeDate: "2025-02-01"
   branch: release-2.6
 - edition: mesh
-  version: preview
+  version: 2.7.0
   release: 2.7.x
+  branch: release-2.7
```
fabianrbz, line 394: Is 2.7.x the new latest? If so, we should set latest: true to it, and remove that bit from 2.6.x

Automaat: 2.7.x is not yet released, I will update it in a separate PR, after we release this version

fabianrbz (approved):

Automaat:
needed for: https://github.com/Kong/docs.konghq.com/pull/7243

PR #7240, opened 2024-04-15
Cloud gateway: Add Insomnia link, and set max-width to overview page

Guaris:
Add Insomnia link, and set max-width to overview page


acgoldsmith (approved):
LGTM

PR #7242, opened 2024-04-16
Fix: Upstream URL for llama2 completion example

dascole:
### Description

The example provided in the curl command references the ollama endpoint, http://ollama-server.local:11434/v1/chat, which returns a 404.

Per the docs (referenced earlier on this page), the correct endpoint should be /api/chat which I can confirm works.

https://github.com/ollama/ollama/blob/main/docs/api.md#generate-a-chat-completion

example:

$ curl localhost:11434/api/generate -d '{ "model": "llama2", "prompt":"Why does Kong make the best gateway?", "stream":false }' -i
HTTP/1.1 200 OK

$ curl localhost:11434/v1/generate -d '{ "model": "llama2", "prompt":"Why does Kong make the best gateway?", stream":false  }' -i
HTTP/1.1 404 Not Found
Content-Type: text/plain
Date: Tue, 16 Apr 2024 01:56:56 GMT

404 page not found

### Testing instructions

Preview link: 

### Checklist 

- [x] Review label added 

### DIFF
--- a/app/_hub/kong-inc/ai-proxy/how-to/_llama2.md
+++ b/app/_hub/kong-inc/ai-proxy/how-to/_llama2.md
@@ -108,7 +108,7 @@ plugins:
         provider: "llama2"
         name: "llama2"
         llama2_format: "ollama"
-        upstream_url: "http://llama2-server.local:11434/v1/chat"
+        upstream_url: "http://llama2-server.local:11434/api/chat"
 ```
 {% endnavtab %}
 {% endnavtabs %}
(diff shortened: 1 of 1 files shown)


app/_hub/kong-inc/ai-proxy/how-to/_llama2.md
```diff
@@ -108,7 +108,7 @@ plugins:
         provider: "llama2"
         name: "llama2"
         llama2_format: "ollama"
-        upstream_url: "http://llama2-server.local:11434/v1/chat"
+        upstream_url: "http://llama2-server.local:11434/api/chat"
```
lena-larionova, line 111: Looks like the config on line 92 needs updating as well (same thing, but in cURL format).

lena-larionova (approved):
Applied the same fix to the curl-format example as well. LGTM, thanks!

PR #7259, opened 2024-04-18
Feat: shot scraper macro for mermaid diagrams

Guaris:
A new macro that supports taking screenshots of mermaid diagrams

Todo: 
- Rename poc.yaml
- Write script for all mermaid containing pages
- Docs that explain the `--retina` command
- GHA that runs and creates a diagram for every mermaid thing and creates a package. 
- Test multiple diagrams on the same page


tools/screenshots/mermaid/poc.yaml
```diff
@@ -0,0 +1,8 @@
+- output: ../../app/assets/images/transit-gateways.png
```
lena-larionova, line 1: Let's put them in a specific directory so that they're easier to keep track of, and so that we don't dump a bunch of unsorted content directly into `images`. What about:
```suggestion
- output: ../../app/assets/images/mermaid/transit-gateways.png
```

tools/screenshots/mermaid/poc.yaml
```diff
@@ -0,0 +1,8 @@
...
+  url: https://docs.konghq.com/konnect/gateway-manager/data-plane-nodes/transit-gateways/
+  wait: 6000
+  width: 9000
+  height: 9000
+  selector: "pre.mermaid svg"
```
mheap, line 6: How does this work with multiple diagrams on a single page?

mheap: Before I answer, more questions.

Why are we taking screenshots of something that's rendered on-demand in the docs?

mheap: Do they want the diagrams for presentations, or do they want the source files so that they can make their own edits?

If it's the former, I'd recommend letting them take screenshots as needed. If it's the latter, screenshots won't help them

Guaris:
Closing in favor of expanding our docs on how to create diagrams

PR #7274, opened 2024-04-19
Chore: Split AI Gateway into its own section

lena-larionova:
### Description

Splitting AI gateway into its own nav section to make it more prominent. 

This is a 3.7 ticket, but it really also applies to 3.6, so making the change in both.

https://konghq.atlassian.net/browse/DOCU-3792

### Testing instructions

Preview link: 
https://deploy-preview-7274--kongdocs.netlify.app/gateway/latest/ai-gateway/

### Checklist 

- [x] Review label added 


acgoldsmith (approved):
LGTM - this is a nice solve!

PR #7290, opened 2024-04-23
Plugins troubleshooting

fabianrbz:
### Testing instructions

Preview link: 

### Checklist 

- [ ] Review label added 


lena-larionova (approved):

lena-larionova:
I wonder - if we have troubleshooting content that applies to multiple plugins, I guess we'd just use an include, then have two markdown files where the include is used? For example, something that applies to Rate Limiting and Rate Limiting Advanced.

fabianrbz:
that should work

fabianrbz:
I removed the dummy content.

PR #7319, opened 2024-04-28
Tries to fix the broken CI workflows

n0vad3v:
### Description

Tries to fix the broken CI workflows in PR.

### Testing instructions

Preview link: 

### Checklist 

- [ ] Review label added 

### DIFF
--- a/.github/workflows/autogenerated-warning.yml
+++ b/.github/workflows/autogenerated-warning.yml
@@ -24,7 +24,10 @@ jobs:
     if: ${{ github.event.action == 'opened' || github.event.action == 'synchronize' }}
     runs-on: ubuntu-latest
     timeout-minutes: 30
-    permissions: write-all
+    permissions:
+      contents: write
+      issues: write
+      pull-requests: write
     steps:
       - uses: actions/checkout@v4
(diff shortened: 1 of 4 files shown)


n0vad3v:
Hmmm, maybe it's needed to adjust permissions at https://github.com/organizations/Kong/settings/actions to allow GITHUB_TOKEN for more privileges

Screenshot from 2024-04-28 23-27-24

Otherwise, even manually set:

```
    permissions:
      contents: write
      issues: write
      pull-requests: write
```

theses permissions at workflow will still result in 
```
GITHUB_TOKEN Permissions
  Contents: read
  Issues: read
  Metadata: read
  PullRequests: read
```

in actual run.

Screenshot from 2024-04-28 23-26-03

mheap:
The issue here is that the action triggers on `pull_request` rather than `pull_request_target`, which doesn't have access to secrets when the contributor is submitting from a fork.

Switching to `pull_request_target` is ok in this instance as executing the check from `main` is preferred 99% of the time

n0vad3v:
@mheap Got it, 😃, so maybe it's save to close this PR now.

PR #7327, opened 2024-05-01
Feat: Plugin examples

lena-larionova:
### Description

Add hub examples directory with some existing examples for openId connect, HTTP Log, and AI Proxy (three plugins that I knew had useable examples). 

https://konghq.atlassian.net/browse/DOCU-3835

Notes:
* I put them into `app/assets/hub/kong-inc` to mimic the pattern that KIC and KGO use for their examples. Does that work?
* Added a "description" field to explain what the config is doing.
  * Included prereqs that are as short as possible
  * Included values that the user needs to customize

@fabianrbz For the prereqs and customizable values, is it fine to just have them manually written in the description? Or would it be better to do something like this, in case the various bits have to get pulled into a layout separately:

```yaml
description: |
  Configure the plugin to do a thing.
prerequisites: |
  Do this thing first.
custom_values:
  - client_id
  - client_secret
  - redirect_uri
config:
  ...
```

Maybe that's a question for Carl & whoever implements the frontend eventually.

### Testing instructions

Preview link: 

### Checklist 

- [x] Review label added 


mheap:
Can we keep `custom_values` in a separate list? We can generate templated output from that and it makes the data useful for the UI e.g. a "move to next field" button to cycle through required fields

lena-larionova:
> Can we keep custom_values in a separate list? We can generate templated output from that and it makes the data useful for the UI e.g. a "move to next field" button to cycle through required fields

That makes sense to me.

What about prereqs?

fabianrbz:
I would keep pre_reqs in a separate list too, it gives us more flexibility

lena-larionova:
@fabianrbz done. Kept the label short as `prereqs` because `prerequisites` is prone to spelling errors 😄

app/assets/hub/kong-inc/ai-proxy/anthropic.yaml
```diff
@@ -0,0 +1,21 @@
+description: |
+  This configuration enables AI Proxy with Anthropic.
+
+prereqs: 
+  - A route with `paths[]=~/anthropic-chat$` configured. Apply the plugin to this route.
+
+custom_values: 
+  - config.auth.header_value
```
mheap, line 8: Do we need the `config.` prefix? Are all custom values implicitly in the config?

app/assets/hub/kong-inc/rate-limiting-advanced/fixed-window-1.yaml
```diff
@@ -0,0 +1,9 @@
+description: |
```
mheap, line 1: Can we use a representative filename? It's likely that the filename will form part of the API e.g. `GET /api/plugin-examples/FILENAME`

`fixed-window-500-per-hour` makes more sense than `fixed-window-1` to me

app/assets/hub/kong-inc/rate-limiting-advanced/silding-window-1.yaml
```diff
@@ -0,0 +1,9 @@
+description: |
```
mheap, line 1: Typo in the filename. `silding` -> `sliding`

app/assets/hub/kong-inc/rate-limiting-advanced/sliding-window-2.yaml
```diff
@@ -0,0 +1,9 @@
+description: |
+  Each user can make up to 300 requests in any rolling 30 minute period, with older requests dropping off as new requests are made.
```
mheap, line 2: What does "dropping off" mean?

app/assets/hub/kong-inc/ai-proxy/anthropic.yaml
lena-larionova: I think that's a question for @fabianrbz - what would be better?

fabianrbz: I don't think we need it, they should be implicitly in the config.
However, I don't know who and how this API will be consumed, as long as it's clear to everyone I'm ok with dropping it.

mheap (approved):

PR #7389, opened 2024-05-16
update the perf stats for 3.7

tonydongkong:
### Updating the Perf stats for public benchmark

The raw data is posted [here](https://docs.google.com/spreadsheets/d/1knu87tczpuTQ4a5si3WSkOoFHWf0pS8vqNpC44-Oxmw/edit#gid=499697946)


lena-larionova (approved):
Switched it so that we add a version-specific table instead of deleting the old one. 

LGTM, thanks for updating

PR #7494, opened 2024-06-10
Fix support for x-errors in OAS

kong-apiops:
@mheap made additional changes to the x-error rendering in this PR. Please review

### DIFF
--- a/app/_plugins/generators/oas_definition/product.rb
+++ b/app/_plugins/generators/oas_definition/product.rb
@@ -4,28 +4,54 @@ module OasDefinition
   class Product
     attr_reader :site, :file, :product
 
-    def initialize(product:, file:, site:)
+    def initialize(product:, file:, site:, frontmatter:)
       @product = product
       @file = file.gsub("#{site.source}/", '')
       @site = site
+      @frontmatter = frontmatter
     end
 
     def generate_pages!
+      oas_errors = oas_x_errors(@frontmatter)
+
       product['versions'].map do |version|
         data = OasDefinition::PageData.generate(product:, version:, file:, site:)
 
-        generate_product_pages!(data)
+        generate_product_pages!(data, oas_errors)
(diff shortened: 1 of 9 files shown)


app/_plugins/generators/oas_definition/product.rb
```diff
@@ -4,28 +4,54 @@ module OasDefinition
...
 
     private
 
-    def generate_product_pages!(product, data)
+    def oas_x_errors(frontmatter)
+      spec_path = api_spec_path(frontmatter)
+      return nil unless spec_path
+
+      oas = load_api_file(spec_path)
+      raise "Could not load #{spec_path}" unless oas
```
fabianrbz, line 34: @mheap do we need all these implementation changes?
I think we could achieve the same thing by adding this check before [this](https://github.com/Kong/docs.konghq.com/pull/7494/files#diff-86ae4e7231f5c58a4b243dc4ff98788251b4259f8a064acee8752bdb08554049L80) line.

mheap: At that point the `site.pages << generate_error_page` has run. Is there a way to abort page generation from within `ErrorPage`?

fabianrbz: got it, we no longer support `generateErrorsPage` then?

fabianrbz (approved):

fabianrbz:
@mheap looks like the branch was force-pushed by the bot :(

app/_plugins/generators/oas_definition/product.rb
mheap: Nope. I realised that it was manual metadata in a file that's autogenerated from a tool

mheap:
@fabianrbz It was just for testing purposes in this repo. I need to fix the upstream build to push it down

PR #7542, opened 2024-06-17
Remove -alpine in the image tag

xibuka:
The '-alpine' tag does not exist after 3.4. This will lead to an error when the user copies and executes the command.

### Testing instructions

Preview link: 

### Checklist 

- [ ] Review label added 


lena-larionova (approved):
LGTM, thanks for the fix!

PR #7556, opened 2024-06-20
Fix: instead of Update Cluster Config should be Edit or Resize Cluster

a-gabriella:
Documentation does not line up with the product: instead of Update Cluster Config should be Edit or Resize Cluster

### Testing instructions

Preview link: 

### Checklist 

- [ ] Review label added 


Guaris (approved):

Guaris:
LGTM

PR #7590, opened 2024-06-27
Fix: GraphQL RLA example needs a warning to not run in prod

lena-larionova:
### Description

Adding a disclaimer to plugin basic examples about not using the examples in prod, and adding some extra language for the GraphQL RLA examples to caution against `sync_rate = -1` with the `cluster` strategy.

We can use this same templating to add extra example descriptions to any of the plugins in the future.

https://konghq.atlassian.net/browse/DOCU-3489

### Testing instructions

Additional note for GraphQL RLA:  https://deploy-preview-7590--kongdocs.netlify.app/hub/kong-inc/graphql-rate-limiting-advanced/how-to/basic-example/

### Checklist 

- [x] Review label added 

### DIFF
--- a/app/_layouts/plugins/configuration_examples.html
+++ b/app/_layouts/plugins/configuration_examples.html
@@ -5,10 +5,10 @@ type: plugin
 
 <h2 id="configuration">Basic configuration examples</h2>
 
+{% if page.examples_description %}
 The following configuration examples are for illustrative purposes only, and shouldn't be used as-is in production.
 Adjust the configuration for your own purposes.
 
-{% if page.examples_description %}
   {{ page.examples_description | markdownify }}
 {% endif %}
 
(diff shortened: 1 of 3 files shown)


app/_layouts/plugins/configuration_examples.html
```diff
@@ -5,6 +5,13 @@
 
 <h2 id="configuration">Basic configuration examples</h2>
 
+The following configuration examples are for illustrative purposes only, and shouldn't be used as-is in production.
```
mheap, line 8: Shouldn't we be updating the examples to be production ready?

lena-larionova: We should, but these examples won't be production ready for some time. We just don't have all the info. Also, I don't think that this basic examples page will ever serve that purpose, since it's currently limited to only one example with little context.

mheap: I'm worried that this will cause concern for others. Can we show this message inside `{% if page.examples_description %}`? That allows us to only show it on specific plugins?

lena-larionova: Done.

mheap (approved):

PR #7598, opened 2024-07-01
Update default-ports.md

duncdeane:
### Description

The status listener section (8007) is incorrect. The Status API is available in OSS, so it should be changed from "Kong Gateway Enterprise tier" to "All Tiers and Modes". Also, the description does not sound right. It does not only listen for calls from the command line. The purpose of this API is to expose metrics to monitoring tools/clients

### Testing instructions

Preview link: 

### Checklist 

- [ ] Review label added 


lena-larionova (approved):
LGTM!

PR #7646, opened 2024-07-16
Update _local-testing-development.md

jamesgkong:
### Description

Removed the "path" value from route created in step 2, as this would cause request in step 4 "Trigger certificate creation" to fail, and will not work with the "/.well-known/acme-challenge/x" path which would also need to be associated to this test route for the purpose of this walkthrough.
 
Noted in Support case: 00047179


cloudjumpercat (approved):
Looks good to me! Thanks for fixing this!

PR #7675, opened 2024-07-24
fix(gateway/request-validator): update document

vm-001:
### Description

The previous sentences already clarified the strictly validation behavior. Hence remove a sentence that may confuse users.

> The parameter is strictly validated, which means a request with a parameter (for example, `application/json; charset=UTF-8`) is NOT considered valid for one without the same parameter (for example, `application/json`). The type, subtype, parameter names, and the value of the charset parameter are not case sensitive based on the RFC explanation.

 

### Testing instructions

Preview link: 

### Checklist 

- [x] Review label added 

FTI-6102


lena-larionova (approved):
lgtm, thanks!

PR #7692, opened 2024-07-26
Feat: Brotli support

lena-larionova:
### Description

We added Brotli compression support in 3.6 but it was never documented. Adding a short doc on how to enable it.

Notes:
* I considered adding it to the Nginx directives doc, but there's nothing in there about configuring specific groups of directives for a goal - that topic is a more general thing. 
* I'm also not sure about putting this under "reference", but honestly, we don't really have a place for content like this at the moment.

Addresses https://konghq.atlassian.net/browse/DOCU-3911.

### Testing instructions

Preview link: https://deploy-preview-7692--kongdocs.netlify.app/gateway/latest/production/performance/brotli/

### Checklist 

- [x] Review label added 


app/_src/gateway/reference/brotli.md
```diff
@@ -0,0 +1,29 @@
+---
+title: Brotli compression
```
cloudjumpercat, line 2:
```suggestion
title: Improve performance with Brotli compression
```
Just to give users an indication of what Brotli compression can do for them here

app/_data/docs_nav_gateway_3.8.x.yml
```diff
@@ -737,6 +737,8 @@ items:
         src: /gateway/reference/configuration/configuration-3.7.x
       - text: Injecting Nginx Directives
         url: /reference/nginx-directives
+      - text: Brotli Compression
```
cloudjumpercat, line 740:
```suggestion
      - text: Improve performance with Brotli compression
```
Just to give users an indication of what Brotli compression can do for them here

cloudjumpercat: For the location in the nav, maybe this could go in Kong in Production? Maybe we could rename "Performance testing" to "Performance" and add it there?

cloudjumpercat (approved):
Left a few suggestions, but I'll approve this for now so you're unblocked :)

PR #7703, opened 2024-07-30
Fix: Hardcode gateway version in centos doc

lena-larionova:
### Description

As of 2.8.4.12, we are no longer building Centos packages for Gateway 2.8. Changing the variables on the 2.8 page to hardcoded so that they don't pick up a version of a package that doesn't exist. The version no longer needs to change dynamically with releases.

### Testing instructions

Preview link: 

### Checklist 

- [x] Review label added 


Guaris (approved):

Guaris:
Looks good on centos7 evaluates to 
https://packages.konghq.com/public/gateway-28/rpm/el/7/x86_64/kong-enterprise-edition-2.8.4.11.el7.noarch.rpm

PR #7708, opened 2024-07-31
Add clarifying statement about private dev portals

timodwhit:
### Description

The prior sentence states that `and are discoverable on the internet` for Public Dev Portals. It seems like it would be good to  explicit that private are not (which is just an assumption I made)

### Testing instructions

Preview link: 

### Checklist 

- [x] Review label added 


lena-larionova (approved):
LGTM, that's a good distinction to point out. Thanks!

lena-larionova:
Smoke test always fails on fork, the failure is expected.

PR #7737, opened 2024-08-08
Update key-auth request behaviour matrix

gravitystorm:
I've read and re-read this section a few times - and I'm pretty sure there's a typo! But my apologies in advance if I've misunderstood the situation.

Kong will return a 401 when the API key is **not** known.

### Checklist 

- [ ] Review label added 


lena-larionova (approved):
LGTM, nice catch. Thanks for the fix!

PR #7755, opened 2024-08-14
Fix: Control planes can be deleted.

Guaris:
https://konghq.atlassian.net/browse/DOCU-4017

### DIFF
--- a/app/konnect/gateway-manager/index.md
+++ b/app/konnect/gateway-manager/index.md
@@ -56,9 +56,6 @@ entity-specific permissions. For more information, see [Administer teams](/konne
 
 ### {{site.base_gateway}} control planes
 
-Every region in every organization starts with one default control plane.
-This control plane can't be deleted, and its status as the default can't be changed.
-
 With {{site.konnect_short_name}} you can configure additional {{site.base_gateway}}
 control planes, each of which will have isolated configuration.
 Use multiple control planes in one {{site.konnect_short_name}} organization to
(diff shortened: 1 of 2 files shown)


lena-larionova (changes requested):
Can you please add a changelog entry as well?

app/konnect/gateway-manager/index.md
```diff
@@ -56,9 +56,6 @@ entity-specific permissions. For more information, see [Administer teams](/konne
 
 ### {{site.base_gateway}} control planes
 
-Every region in every organization starts with one default control plane.
```
lena-larionova, line 59: Why delete the first line? It's still accurate.

lena-larionova (commented):
Please also remove the line here: https://docs.konghq.com/konnect/gateway-manager/#deleting-a-control-plane

"You cannot delete the default control plane."

app/konnect/gateway-manager/index.md
mheap: I think I saw that we're going to stop provisioning the default CP too

lena-larionova: Looks like that hasn't gone live yet: 

> we will also stop provisioning the default control plane in each region soon 

https://kongstrong.slack.com/archives/C04RXLGNB6K/p1723651386893509?thread_ts=1723650140.813759&cid=C04RXLGNB6K

lena-larionova (approved):

PR #7915, opened 2024-09-11
Enable Terraform examples for all plugins

mheap:
### Description

Enabled TF example generation for all plugins.

### Testing instructions

Preview link: [/hub/kong-inc/ai-semantic-prompt-guard/how-to/basic-example/](https://deploy-preview-7915--kongdocs.netlify.app/hub/kong-inc/ai-semantic-prompt-guard/how-to/basic-example/)

### Checklist 

- [x] Review label added

### DIFF
--- a/app/_plugins/drops/plugins/hub_examples.rb
+++ b/app/_plugins/drops/plugins/hub_examples.rb
@@ -2,37 +2,6 @@
 
 # There are also oauth2, acl and mtls-auth but I don't have examples of how to use them yet
 AUTH_PLUGINS = %w[basic-auth hmac-auth jwt key-auth].freeze
-TERRAFORM_ENABLED_PLUGINS = %w[
-  key-auth
-  cors
-  rate-limiting
-  basic-auth
-  rate-limiting-advanced
-  openid-connect
-  jwt
-  prometheus
-  acl
-  request-termination
-  file-log
-  request-transformer
-  correlation-id
-  proxy-cache
-  request-transformer-advanced
-  response-transformer
-  response-transformer-advanced
-  ip-restriction
-  pre-function
-  oauth2
-  opentelemetry
-  ai-proxy
-  ai-prompt-guard
-  ai-prompt-template
-  ai-prompt-decorator
-  aws-lambda
-  jq
-  jwt-signer
-  saml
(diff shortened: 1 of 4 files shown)


app/_plugins/drops/plugins/hub_examples.rb
```diff
@@ -63,11 +32,7 @@ def render_kubernetes?
         end
 
         def render_terraform?
-          terraform_implemented? && @formats.include?(:terraform)
-        end
-
-        def terraform_implemented?
-          TERRAFORM_ENABLED_PLUGINS.include?(plugin_name)
+          @example.fetch('config', nil) && @formats.include?(:terraform)
```
mheap, line 35: @fabianrbz This renders if there is a sample config provided (which is false for partners in most cases). _Really_ I'd like to add a condition that only renders if the maintainer is `kong-inc`. Any ideas?

lena-larionova:
Any chance that we could get terraform added as a config option for the opinionated examples as well (`{% plugin_examples %}`, or is there any reason not to?

Eg here:
https://docs.konghq.com/hub/kong-inc/http-log/how-to/splunk/

mheap:
It should work if you add Terraform to the formats list for each custom example. Shout if it doesn’t

lena-larionova:
> It should work if you add Terraform to the formats list for each custom example. Shout if it doesn’t

It fails but looks like you just need to adjust this to add `terraform` as an option: https://github.com/Kong/docs.konghq.com/blob/main/app/_plugins/blocks/inline_plugin_example/config.rb#L7

app/_plugins/drops/plugins/hub_examples.rb
fabianrbz: @mheap one quick  way would be to pass the `vendor` to the initialiizer.
It's available [here](https://github.com/Kong/docs.konghq.com/blob/main/app/_plugins/drops/plugins/hub_examples.rb#L118) in `schema.vendor` it returns a string that matches the folder name of the vendor, so you'll need to check if it matches `kong-inc` in the code

fabianrbz (approved):
@mheap  LGTM!
Looks like we need to update the specs though.

PR #7957, opened 2024-09-23
Update deck_file_kong2tf.md

battlebyte:
References to kubernetes removed, should be referencing terraform resources.

### Checklist 

- [X] Review label added


lena-larionova (approved):
nice catch, lgtm

Issue #7533, opened 2024-06-15
Confusing terms and concepts: "directives" and "phases"


lfavreli:
### Where is the problem?

https://docs.konghq.com/gateway/3.7.x/plugin-development/custom-logic/

### What happened?

(Kong Gateway > Develop Custom Plugins > Implementing Custom Logic)

I'm not an Ningx/OpenResty® expert, but I find the use of the "Phase" term on this page confusing.

In the [Available contexts](https://docs.konghq.com/gateway/3.7.x/plugin-development/custom-logic/#available-contexts) table, all the links refer to "**directives**" rather than "**phases**", if we adopt the [lua-nginx-module](https://github.com/openresty/lua-nginx-module) terminology. For example, according to the _lua-nginx-module_ documentation: 
- "phase init_worker" in Kong -> stands for "starting-worker" **phase** and "init_worker_by_lua_*" **directive**
- "phase rewrite" in Kong -> stands for "rewrite tail" **phase** and "rewrite_by_lua_*" **directive**
- "phase access" in Kong -> stands for "access tail" **phase** and "access_by_lua_*" **directive**
- ...

An other issue concerns the "reponse" function, for which the phase is "access". However, according to the description, it replaces "header_filter()" and "body_filter()". As shown in the diagram in the [Directives](https://github.com/openresty/lua-nginx-module?tab=readme-ov-file#directives) section, they are part of "Content Phase", not "Access Phase". This seems to be a mistake.

### What did you expect to happen?

What could be preferred:
1. Standardize terminology and retain in Kong's documentation only the four main phases described in the _lua-nginx-module_ diagram: Initialization Phase, Rewrite/Access Phase, Content Phase, Log Phase.
2. Replace the term "Phase" by "Directive", and continue to link to each of the _lua-nginx-module_ directives.
3. Another thing?

In any case, I think there's something to clear up! 😀

Hope this helps! 🙏

### Code of Conduct and Community Expectations

- [X] I agree to follow this project's Code of Conduct
- [X] I agree to abide by the Community Expectations

lena-larionova:
@lfavreli You have a very good point here, but looks like Kong is using the term "phases" in a custom way. 
I did a bunch of digging into this, and it looks like the term "phases" is referring to Kong-specific phases, not phases as defined by nginx. Here's the list of Kong phases: https://github.com/Kong/kong/blob/master/kong/pdk/private/phases.lua

[...]

PR #8001, opened 2024-10-02
bug-bash: Change phase to directive in custom logic doc

Guaris:
Fixes https://github.com/Kong/docs.konghq.com/issues/7533

### DIFF
--- a/app/_src/gateway/plugin-development/custom-logic.md
+++ b/app/_src/gateway/plugin-development/custom-logic.md
@@ -45,17 +45,17 @@ of {{site.base_gateway}}'s execution life-cycle:
 | Function name       | Kong Phase        | Nginx Directives         | Request Protocol              | Description
 |---------------------|-------------------|--------------------------|-------------------------------|------------
 | `init_worker`       | `init_worker`     | [`init_worker_by_*`]     | *                             | Executed upon every Nginx worker process's startup.
-| `certificate`       | `ssl_certificate` | [`ssl_certificate_by_*`] | `https`, `grpcs`, `wss`       | Executed during the SSL certificate serving phase of the SSL handshake.
+| `certificate`       | `certificate`     | [`ssl_certificate_by_*`] | `https`, `grpcs`, `wss`       | Executed during the SSL certificate serving phase of the SSL handshake.
 | `rewrite`           | `rewrite`         | [`rewrite_by_*`]         | *                             | Executed for every request upon its reception from a client as a rewrite phase handler. <br> In this phase, neither the `Service` nor the `Consumer` have been identified, hence this handler will only be executed if the plugin was configured as a global plugin.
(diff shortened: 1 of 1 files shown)


app/_src/gateway/plugin-development/custom-logic.md
```diff
@@ -42,7 +42,7 @@ of {{site.base_gateway}}'s execution life-cycle:
 - **[HTTP Module]** *is used for plugins written for HTTP/HTTPS requests*
 {% if_version lte: 3.3.x %}
 
-| Function name       | Phase             | Request Protocol        | Description
+| Function name       | Directive             | Request Protocol        | Description
```
lena-larionova, line 45: I think it's a bit more complicated than this. The items in this column cover both a phase and a directive. Check out the diagram here: https://github.com/openresty/lua-nginx-module?tab=readme-ov-file#directives.



As the reporter points out in the GH issue,
* "rewrite" in Kong -> stands for the "rewrite tail" phase and the "rewrite_by_lua_*" directive: https://github.com/openresty/lua-nginx-module#rewrite_by_lua_block

It looks like what we're actually missing is another column, which details what the directive is. So for rewrite, you get:

Function: rewrite
Phase: rewrite tail
Directives: `rewrite_by_lua_*`

lena-larionova (changes requested):
This is missing the second part of the issue as well, which is:

> An other issue concerns the "reponse" function, for which the phase is "access". However, according to the description, it replaces "header_filter()" and "body_filter()". As shown in the diagram in the [Directives](https://github.com/openresty/lua-nginx-module?tab=readme-ov-file#directives) section, they are part of "Content Phase", not "Access Phase". This seems to be a mistake.

Edit: After doing more digging, I found that the `response` function runs in the Kong `response` phase, as a whole separate thing: https://github.com/Kong/kong/blob/master/spec/fixtures/custom_plugins/kong/plugins/enable-buffering-response/handler.lua. The links to the directives are correct though (header_filter and body_filter). 

See all the Kong phases here: https://github.com/Kong/kong/blob/master/kong/pdk/private/phases.lua

app/_src/gateway/plugin-development/custom-logic.md
lena-larionova: Did a bunch more digging, and this isn't quite right either. It seems that Kong has come up with its own terminology here, where "access phase" or "init_worker phase" are accurate _only for Kong_. So - technically this column _is_ correctly named "phase", but the links to directives make it confusing.

lena-larionova (approved):
tl;dr for the changes I made to this PR:
* marked ws phases as enterprise only
* added a directives column and renamed the original phases column to "Kong phases", which are not the same thing as the nginx phases
* checked each function against its real phase in Kong; the only one that actually differs from its function name is the `configuration` function

PR #8029, opened 2024-10-08
chore: Remove backport lines from Gateway changelog

lena-larionova:
### Description

Removing any "backported from" lines from the Gateway changelog, as they were causing many issues:
* Manually maintained, so the editor needs to have knowledge that an entry already exists, or else look for it in the changelog - this is difficult because the phrasing can change
* Inconsistently added - since this is manually maintained, it's hard to keep this content accurate, and difficult to tell which version the original change came from. Items get missed all the time.
* Doesn't add enough value (or much value at all) to warrant the maintenance

### Testing instructions

Preview link: 

### Checklist 

- [x] Review label added 

### DIFF
--- a/app/gateway/changelog.md
+++ b/app/gateway/changelog.md
@@ -4039,7 +4039,7 @@ images or packages, and Kong will not test package installation on Ubuntu 18.04.
 
 #### Core 
 * Fixed the OpenResty `ngx.print` chunk encoding duplicate free buffer issue that
-  lead to the corruption of chunk-encoded response data.
+  led to the corruption of chunk-encoded response data.
   [#10816](https://github.com/Kong/kong/pull/10816)
   [#10824](https://github.com/Kong/kong/pull/10824)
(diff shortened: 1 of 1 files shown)


app/gateway/changelog.md
cloudjumpercat, line 1: On line 4423 there's a "Backported the openresty ngx.print chunk encoding buffer double free bug fix that was leading to the corruption of chunk-encoded response data. [#10816](https://github.com/Kong/kong/pull/10816)[#10824](https://github.com/Kong/kong/pull/10824)". Not sure if we want to fix that one as well.

cloudjumpercat (approved):
Just left one comment about another item that lists a backport, but I'll approve this PR for now so you're not blocked!

app/gateway/changelog.md
lena-larionova: Nice catch, I adjusted it to the phrasing that we use for the same issue elsewhere in the changelog.

PR #8099, opened 2024-10-29
docs(api): add minimum length to tags pagination parameter

nowNick:
### Description
When quering entities by tags a user could sent a request `tags=''` and it used to make Kong crash. Such request is invalid as either the tags should be specified or they should not be in the query params.

Documents the fix in: https://github.com/Kong/kong/pull/13723

KAG-5496

### Testing instructions

Preview link: 

### Checklist 

- [x] Review label added 


lmilan:
Closing this because I included these changes in the PR for 3.9 API specs updates: #8234

PR #8173, opened 2024-11-26
Fix: Konnect supported geos

lena-larionova:
### Description

We are missing a table of the supported Konnect geos. 

Fix requested on Slack: https://kongstrong.slack.com/archives/C04RXLGNB6K/p1732410184765539

Reza mentions in the thread that this is mostly an implementation detail, as users can't choose their specific regions for the CP, they only choose the larger geo. However, it's a talking point for compliance, and is useful to our support teams.

I also reformated the other lists of geos for dedicated CWGs into tables to make them easier to read.

### Testing instructions

Preview link: 

### Checklist 

- [x] Review label added 

### DIFF
--- a/app/konnect/geo.md
+++ b/app/konnect/geo.md
@@ -13,11 +13,13 @@ Geographic regions allow you to also operate {{site.konnect_saas}} in a similar 
 
 ### Control planes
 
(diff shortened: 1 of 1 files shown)


lena-larionova:
Closing as this level of detail is not desired.

PR #8271, opened 2024-12-16
Update changelog.md

jamesgkong:
As per the detail in the pull here: https://github.com/Kong/kong/pull/10355 this value is added to the ngx.ctx table NOT kong.ctx

### Description

Changed reference of location of KONG_UPSTREAM_DNS_TIME to correct internal table as per pull request this change is in respect of.

See also: https://konghq.atlassian.net/browse/FTI-5243

### Testing instructions

N/A typo

### Checklist 

N/A typo


lena-larionova (approved):
Nice catch, thanks for the fix! Lgtm

PR #8321, opened 2025-01-07
remove -o from deck dump command

Guaris:
I can only dump multiple workspaces using 
`deck gateway dump --all-workspaces`

Which dumps them in yaml files sorted by name.


lena-larionova (approved):
Nice catch, lgtm. Tested to make sure.

PR #8418, opened 2025-02-10
Fix versioning issues with OAS and insomnia buttons

lena-larionova:
### Description

Fixing two issues with our Gateway Admin EE and OSS specs:
1. The version number in the specs for 3.9 and latest is incorrect, both should be 3.9 but are showing 3.8. The content of the files are correct for 3.9 though, I tracked down the file histories and it just looks like the version number update was missed.
2. The insomnia buttons to the versioned specs are incorrect. Easiest solution, added a shortcode for the version from 3.4 onward that we can access via `{{page.release.versions.short}}`.

Specs are already updated in Konnect Dev Portal.

Tested the buttons, they all import the correct specs. Note that the filename format changed in 3.8, so that's why I had to version tag the URLs.

Issue reported on Slack.

### Testing instructions

Preview link: 

### Checklist 

- [x] Review label added 

### DIFF
--- a/api-specs/Gateway-EE/3.9/kong-ee.yaml
+++ b/api-specs/Gateway-EE/3.9/kong-ee.yaml
@@ -6236,7 +6236,7 @@ info:
     name: Apache 2.0
     url: 'https://www.apache.org/licenses/LICENSE-2.0.html'
   title: Enterprise Kong Admin API
-  version: 3.8.0
+  version: 3.9.0
 openapi: 3.0.0
 paths:
   /:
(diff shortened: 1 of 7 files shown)


lena-larionova:
> You can dynamically add the short version entry if you prefer
> 
> e.g.
> 
> https://github.com/Kong/docs.konghq.com/blob/dd680e0187c3ef94bdfa4ad3941755f6b0201f7b/app/_plugins/generators/pages/version_data.rb#L51

❗ Didn't know this existed; couldn't find it it in the versions file. Good thing this didn't take me very long 😄 I'll replace with that.

mheap (approved):

PR #8562, opened 2025-03-14
Add cli and pdk docs

fabianrbz:
### Description

https://konghq.atlassian.net/browse/DOCU-4207
https://konghq.atlassian.net/browse/DOCU-4206

### Testing instructions

Preview link: 

### Checklist 

- [ ] Review label added 

### DIFF
--- a/app/_data/docs_nav_gateway_3.10.x.yml
+++ b/app/_data/docs_nav_gateway_3.10.x.yml
@@ -578,6 +578,12 @@ items:
         - text: kong.vault
           url: /plugin-development/pdk/kong.vault
 
+        - text: kong.websocket.client
+          url: /plugin-development/pdk/kong.websocket.client
+
+        - text: kong.websocket.upstream
+          url: /plugin-development/pdk/kong.websocket.upstream
+
       - text: Plugins in Other Languages
         items:
           - text: Go
--- a/app/_data/docs_nav_gateway_3.10.x.yml
+++ b/app/_data/docs_nav_gateway_3.10.x.yml
@@ -578,12 +578,6 @@ items:
         - text: kong.vault
           url: /plugin-development/pdk/kong.vault
 
-        - text: kong.websocket.client
-          url: /plugin-development/pdk/kong.websocket.client
-
-        - text: kong.websocket.upstream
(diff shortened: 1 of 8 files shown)


app/_data/docs_nav_gateway_3.10.x.yml
lena-larionova, line 1: This shouldn't be getting removed from the nav, both of these PDK modules still exist - they're just in a different folder, but the generator doesn't know that: https://github.com/Kong/kong-ee/blob/master/kong/enterprise_edition/pdk/websocket.lua

app/_src/gateway/reference/cli.md
```diff
@@ -314,8 +319,10 @@ Options:
 
  -p,--prefix      (optional string)   Override prefix directory.
 
+{% if_version lte:3.9.x %}
  --v              verbose
  --vv             debug
+{% endif_version %}
```
lena-larionova, line 325:
```suggestion
 --v              verbose
 --vv             debug
```
Why would this be removed? I checked the CLI in 3.10.0.0-rc.2, both of these flags are still here.

app/_src/gateway/reference/cli.md
```diff
@@ -435,13 +442,22 @@ Options:
...
 {:.badge .enterprise}
+{% endif_version %}
+
 
 ```
+{% if_version gte:3.10.x %}
+Usage: kong runner file.lua [args]
+{% else %}
 Usage: kong runner [file] [args]
+{% endif_version %}
```
lena-larionova, line 458:
```suggestion
Usage: kong runner file.lua [args]
```
This just looks like a clarification fix, doesn't seem like it needs to be version-tagged.

lena-larionova (changes requested):

app/_src/gateway/reference/cli.md
fabianrbz: weird, I didn't get them in the output. I'll add them back

app/_data/docs_nav_gateway_3.10.x.yml
fabianrbz: added them back

Issue #8425, opened 2025-02-12
[KIC] Fix `TLSRoute` documentation


mlavacca:
### Where is the problem?

https://docs.konghq.com/kubernetes-ingress-controller/latest/guides/services/tls/

### What happened?

The linked page has two different problems that need to be fixed:

1. `TLSRoute` requires the Gateway API experimental channel and the `GatewayAlpha` feature gate enabled to properly work. We need to add them as prerequisites, similarly to what we do for `TCPRoute` in https://docs.konghq.com/kubernetes-ingress-controller/latest/guides/services/tcp/
2. The `Gateway` using `TLSRoute` for TLS passthrough in the documentation uses port 443, which is inconsistent with the default value of the proxy port for TLS traffic (8899) we need to fix the `Gateway` port accordingly.

### Code of Conduct and Community Expectations

- [x] I agree to follow this project's Code of Conduct
- [x] I agree to abide by the Community Expectations

PR #8598, opened 2025-03-20
docs(TLS&TCP): update examples to reflect new port configuration and incl…

tao12345666333:
…ude prerequisites

### Description

fix: https://github.com/Kong/docs.konghq.com/issues/8425
 

### Testing instructions

Preview link: https://deploy-preview-8598--kongdocs.netlify.app/kubernetes-ingress-controller/latest/guides/services/tls/

### Checklist 

- [x] Review label added 

### DIFF
--- a/app/_src/kubernetes-ingress-controller/guides/services/tcp.md
+++ b/app/_src/kubernetes-ingress-controller/guides/services/tcp.md
@@ -88,7 +88,7 @@ There are two modes available:
 1.  Configure TCPRoute (Gateway API Only)
 
     {:.note}
-    > If you are using the Gateway APIs (TCPRoute), your Gateway needs additional configuration under `listeners`.
+    > If you are using the Gateway APIs (`TCPRoute`), your Gateway needs additional configuration under `listeners`.
 
     ```bash
     kubectl patch --type=json gateway kong -p='[
--- a/app/_src/kubernetes-ingress-controller/guides/services/tcp.md
+++ b/app/_src/kubernetes-ingress-controller/guides/services/tcp.md
@@ -191,9 +191,8 @@ receives on port 9000 to `echo` service on port 1025.
 {% navtabs codeblock %}
 {% navtab Gateway API %}
 ```bash
-# Wait a few seconds for the controller to process the route
-sleep 5
-kubectl get tcproute echo-plaintext -ojsonpath='{.status.parents[0].conditions[?(@.reason=="Accepted")]}'
+# Wait for the controller to process the route
+kubectl wait tcproute echo-plaintext --for='jsonpath={.status.parents[0].conditions[?(@.reason=="Accepted")]}=True'
 ```
 {% endnavtab %}
 {% navtab Ingress %}
@@ -253,11 +252,11 @@ echo-plaintext   192.0.2.3   3m18s
 ## Route based on SNI
 
 {:.note}
-> For Gateway API, we use TLSRoute for SNI-based routing because it's specifically designed to handle TLS traffic with SNI information, while still proxying the underlying TCP stream. This is still considered TCP proxying, just with the added capability of routing based on the SNI in the TLS handshake.
+> For Gateway API, we use `TLSRoute` for SNI-based routing because it's specifically designed to handle TLS traffic with SNI information, while still proxying the underlying TCP stream. This is still considered TCP proxying, just with the added capability of routing based on the SNI in the TLS handshake.
 
 {% include /md/kic/add-certificate.md hostname='tls9443.kong.example' release=page.release %}
 
-1. Create the TLSRoute (for Gateway API) or TCPIngress (for Ingress) resource to route TLS-encrypted traffic to the `echo` service.
+1. Create the `TLSRoute` (for Gateway API) or `TCPIngress` (for `Ingress`) resource to route TLS-encrypted traffic to the `echo` service.
   {% capture the_code %}
 {% navtabs codeblock %}
 {% navtab Gateway API %}
--- a/app/_src/kubernetes-ingress-controller/guides/services/tcp.md
+++ b/app/_src/kubernetes-ingress-controller/guides/services/tcp.md
@@ -5,7 +5,7 @@ purpose: |
   How to proxy TCP requests
 ---
 
-Create TCP routing configuration for {{site.base_gateway}} in Kubernetes using either the `TCPIngress` custom resource or `TCPRoute` and `TLSRoute` Gateway APIs resource.
+Create TCP routing configuration for {{site.base_gateway}} in Kubernetes using either the `TCPIngress` custom resource or Gateway APIs resources (`TCPRoute` for plain TCP and `TLSRoute` for TLS-encrypted TCP with SNI-based routing).
 TCP-based Ingress means that {{site.base_gateway}} forwards the TCP stream to a Pod of a Service that's running inside Kubernetes. {{site.base_gateway}} does not perform any sort of transformations.
 
 There are two modes available:
@@ -88,7 +88,7 @@ There are two modes available:
 1.  Configure TCPRoute (Gateway API Only)
 
     {:.note}
-    > If you are using the Gateway APIs (TCPRoute), your Gateway needs additional configuration under `listeners`. 
+    > If you are using the Gateway APIs (TCPRoute), your Gateway needs additional configuration under `listeners`.
 
     ```bash
     kubectl patch --type=json gateway kong -p='[
@@ -186,11 +186,13 @@ receives on port 9000 to `echo` service on port 1025.
 
 ### Test the configuration
 
-1. Check if the Service is ready on the route.
+1. Check if the Service is ready on the route. Note that it may take a few moments for the Kong controller to process the route.
     {% capture the_code %}
 {% navtabs codeblock %}
 {% navtab Gateway API %}
 ```bash
+# Wait a few seconds for the controller to process the route
+sleep 5
 kubectl get tcproute echo-plaintext -ojsonpath='{.status.parents[0].conditions[?(@.reason=="Accepted")]}'
 ```
 {% endnavtab %}
@@ -224,12 +226,13 @@ echo-plaintext   192.0.2.3   3m18s
 {% endcapture %}
 {{ the_code | indent }}
 
-1. Connect to this service using `telnet`.
+1. Get the proxy IP address and connect to this service using `telnet`.
     ```shell
+    $ PROXY_IP=$(kubectl get svc -n kong kong-gateway-proxy -o jsonpath='{.status.loadBalancer.ingress[0].ip}')
     $ telnet $PROXY_IP 9000
     ```
 
-    After you  connect, type some text that you want as a response from the echo Service. 
+    After you  connect, type some text that you want as a response from the echo Service.
     ```
     Trying 192.0.2.3...
     Connected to 192.0.2.3.
@@ -249,9 +252,12 @@ echo-plaintext   192.0.2.3   3m18s
 
 ## Route based on SNI
 
+{:.note}
+> For Gateway API, we use TLSRoute for SNI-based routing because it's specifically designed to handle TLS traffic with SNI information, while still proxying the underlying TCP stream. This is still considered TCP proxying, just with the added capability of routing based on the SNI in the TLS handshake.
+
 {% include /md/kic/add-certificate.md hostname='tls9443.kong.example' release=page.release %}
 
-1. Create the TCPIngress resource to route TLS-encrypted traffic to the `echo` service.
+1. Create the TLSRoute (for Gateway API) or TCPIngress (for Ingress) resource to route TLS-encrypted traffic to the `echo` service.
   {% capture the_code %}
 {% navtabs codeblock %}
 {% navtab Gateway API %}
@@ -305,7 +311,7 @@ spec:
 {% navtabs codeblock %}
 {% navtab Gateway API %}
 ```text
-tcproute.gateway.networking.k8s.io/echo-tls created
+tlsroute.gateway.networking.k8s.io/echo-tls created
 ```
 {% endnavtab %}
 {% navtab Ingress %}
@@ -325,6 +331,9 @@ You can now access the `echo` service on port 9443 with SNI `tls9443.kong.exampl
 In real-world usage, you would create a DNS record for `tls9443.kong.example`pointing to your proxy Service's public IP address, which causes TLS clients to add SNI automatically. For this demo, add it manually using the OpenSSL CLI.
 
 ```bash
+# If you haven't set the PROXY_IP variable yet
+PROXY_IP=$(kubectl get svc -n kong kong-gateway-proxy -o jsonpath='{.status.loadBalancer.ingress[0].ip}')
+
 echo "hello" | openssl s_client -connect $PROXY_IP:9443 -servername tls9443.kong.example -quiet 2>/dev/null
 ```
 Press Ctrl+C to exit.
--- a/app/_src/kubernetes-ingress-controller/guides/services/tcp.md
+++ b/app/_src/kubernetes-ingress-controller/guides/services/tcp.md
@@ -192,7 +192,7 @@ receives on port 9000 to `echo` service on port 1025.
 {% navtab Gateway API %}
 ```bash
 # Wait for the controller to process the route
-kubectl wait tcproute echo-plaintext --for='jsonpath={.status.parents[0].conditions[?(@.reason=="Accepted")]}=True'
+kubectl wait tcproute echo-plaintext --for='jsonpath={.status.parents[0].conditions[?(@.type=="Accepted")].status}'=True
 ```
 {% endnavtab %}
 {% navtab Ingress %}
--- a/app/_src/kubernetes-ingress-controller/guides/services/tls.md
+++ b/app/_src/kubernetes-ingress-controller/guides/services/tls.md
@@ -2,7 +2,7 @@
 title: TLS Termination / Passthrough
 type: how-to
 purpose: |
-  How to do TLS Termination and TLS Passthrough with KIC and Kong Gateway
+  How to do TLS Termination and TLS Passthrough with KIC and {{site.base_gateway}}
 ---
 
 
--- a/app/_src/kubernetes-ingress-controller/guides/services/tls.md
+++ b/app/_src/kubernetes-ingress-controller/guides/services/tls.md
@@ -24,6 +24,7 @@ The Ingress API supports TLS termination using the `.spec.tls` field. To termina
 
 ## Examples
 
+{% include /md/kic/prerequisites.md release=page.release disable_gateway_api=false gateway_api_experimental=true %}
 
 ### TLS Termination
 {% assign gwapi_version = "v1" %}
@@ -45,7 +46,7 @@ The Ingress API supports TLS termination using the `.spec.tls` field. To termina
       gatewayClassName: kong 
       listeners:
       - name: https
-        port: 443
+        port: 8899
         protocol: HTTPS
         hostname: "demo.example.com"
         tls:
@@ -132,7 +133,7 @@ The Ingress API supports TLS termination using the `.spec.tls` field. To termina
       gatewayClassName: kong 
       listeners:
       - name: https
-        port: 443
+        port: 8899
         protocol: TLS
         hostname: "demo.example.com"
         tls:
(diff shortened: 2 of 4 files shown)


app/_src/kubernetes-ingress-controller/guides/services/tls.md
```diff
@@ -37,15 +38,15 @@ The Ingress API supports TLS termination using the `.spec.tls` field. To termina
 1. Create a `Gateway` resource.
 
     ```yaml
-    apiVersion: gateway.networking.k8s.io/{{ gwapi_version }}
```
pmalek, line 40: Why are we changing this to a hardcoded value? Was the version incorrect previously? ( I don't think so but in case that was the case can't we tweak the conditions above to set the correct `gwapi_version`?)

app/_src/kubernetes-ingress-controller/guides/services/tcp.md
```diff
@@ -88,7 +88,7 @@ There are two modes available:
 1.  Configure TCPRoute (Gateway API Only)
 
     {:.note}
-    > If you are using the Gateway APIs (TCPRoute), your Gateway needs additional configuration under `listeners`. 
+    > If you are using the Gateway APIs (TCPRoute), your Gateway needs additional configuration under `listeners`.
```
pmalek, line 91:
```suggestion
    > If you are using the Gateway APIs (`TCPRoute`), your Gateway needs additional configuration under `listeners`.
```

app/_src/kubernetes-ingress-controller/guides/services/tls.md
tao12345666333: This is my mistake; I forgot that this document can switch between different KIC versions. It has been corrected. Thank you!

app/_src/kubernetes-ingress-controller/guides/services/tcp.md
```diff
@@ -186,11 +186,13 @@ receives on port 9000 to `echo` service on port 1025.
...
 
-1. Check if the Service is ready on the route.
+1. Check if the Service is ready on the route. Note that it may take a few moments for the Kong controller to process the route.
     {% capture the_code %}
 {% navtabs codeblock %}
 {% navtab Gateway API %}
 ```bash
+# Wait a few seconds for the controller to process the route
+sleep 5
 kubectl get tcproute echo-plaintext -ojsonpath='{.status.parents[0].conditions[?(@.reason=="Accepted")]}'
```
pmalek, line 196: We can you `kubectl wait` here:

```suggestion
# Wait for the controller to process the route
kubectl wait tcproute echo-plaintext --for='jsonpath={.status.parents[0].conditions[?(@.reason=="Accepted")]}=True'
```

app/_src/kubernetes-ingress-controller/guides/services/tcp.md
```diff
@@ -249,9 +252,12 @@ echo-plaintext   192.0.2.3   3m18s
 
 ## Route based on SNI
 
+{:.note}
+> For Gateway API, we use TLSRoute for SNI-based routing because it's specifically designed to handle TLS traffic with SNI information, while still proxying the underlying TCP stream. This is still considered TCP proxying, just with the added capability of routing based on the SNI in the TLS handshake.
+
 {% include /md/kic/add-certificate.md hostname='tls9443.kong.example' release=page.release %}
 
-1. Create the TCPIngress resource to route TLS-encrypted traffic to the `echo` service.
+1. Create the TLSRoute (for Gateway API) or TCPIngress (for Ingress) resource to route TLS-encrypted traffic to the `echo` service.
```
pmalek, line 260:
```suggestion
1. Create the `TLSRoute` (for Gateway API) or `TCPIngress` (for `Ingress`) resource to route TLS-encrypted traffic to the `echo` service.
```

app/_src/kubernetes-ingress-controller/guides/services/tcp.md
```diff
@@ -249,9 +252,12 @@ echo-plaintext   192.0.2.3   3m18s
 
 ## Route based on SNI
 
+{:.note}
+> For Gateway API, we use TLSRoute for SNI-based routing because it's specifically designed to handle TLS traffic with SNI information, while still proxying the underlying TCP stream. This is still considered TCP proxying, just with the added capability of routing based on the SNI in the TLS handshake.
```
pmalek, line 256:
```suggestion
> For Gateway API, we use `TLSRoute` for SNI-based routing because it's specifically designed to handle TLS traffic with SNI information, while still proxying the underlying TCP stream. This is still considered TCP proxying, just with the added capability of routing based on the SNI in the TLS handshake.
```

app/_src/kubernetes-ingress-controller/guides/services/tls.md
pmalek, line 1: Following this guide, from https://deploy-preview-8598--kongdocs.netlify.app/kubernetes-ingress-controller/latest/guides/services/tls/#tls-termination, I'm getting the following status on the `Gateway` (and not traffic being routed):

```
status:
  addresses:
  - type: IPAddress
    value: 172.18.0.101
  - type: IPAddress
    value: 172.18.0.101
  conditions:
  - lastTransitionTime: "2025-04-08T16:25:04Z"
    message: this unmanaged gateway has been picked up by the controller and will
      be processed
    observedGeneration: 1
    reason: Accepted
    status: "True"
    type: Accepted
  - lastTransitionTime: "2025-04-08T16:25:04Z"
    message: ""
    observedGeneration: 1
    reason: Programmed
    status: "True"
    type: Programmed
  listeners:
  - attachedRoutes: 1
    conditions:
    - lastTransitionTime: "2025-04-08T16:25:04Z"
      message: no Kong listen with the requested protocol is configured
      observedGeneration: 1
      reason: UnsupportedProtocol
      status: "False"
      type: Accepted
    - lastTransitionTime: "2025-04-08T16:25:04Z"
      message: ""
      observedGeneration: 1
      reason: Invalid
      status: "False"
      type: Programmed
    - lastTransitionTime: "2025-04-08T16:25:04Z"
      message: ""
      observedGeneration: 1
      reason: InvalidCertificateRef
      status: "False"
      type: ResolvedRefs
    name: https
    supportedKinds:
    - group: gateway.networking.k8s.io
      kind: HTTPRoute
    - group: gateway.networking.k8s.io
      kind: GRPCRoute
```

The default installation using the `ingress` chart seems to yield http and tls ports (with the following default `KONG_PROXY_LISTEN`: `0.0.0.0:8000, [::]:8000, 0.0.0.0:8443 http2 ssl, [::]:8443 http2 ssl`). When Gateway ports are reconciled in https://github.com/Kong/kubernetes-ingress-controller/blob/2c43b76ff7e764a043f623d84ffc5f0683d99288/internal/controllers/gateway/gateway_utils.go#L364 the Gateway listener protocol is set to HTTPS in the examples from this doc and hence it does not match and yields `no Kong listen with the requested protocol is configured`.

I think we may want to fix that. One way to do this would be to change the `Gateway` listener protocol to `TLS` but that would change how the hostname matching works (https://github.com/kubernetes-sigs/gateway-api/blob/a8fe5c8732a37ef471d86afaf570ff8ad0ef0221/apis/v1/gateway_types.go#L256-L264). Another way to fix this is to revisit the protocol and listener matching in KIC (but that requires a release to make this work).

Using the stream listen with `kong` chart also doesn't seem to do the trick 

```
proxy:
  stream:
    - containerPort: 8900
      servicePort: 8900
      protocol: TCP
      parameters:
      - ssl
      - reuseport
```

because that fails to match kong protocol to listener protocol in [the same place](https://github.com/Kong/kubernetes-ingress-controller/blob/2c43b76ff7e764a043f623d84ffc5f0683d99288/internal/controllers/gateway/gateway_utils.go#L363-L364).

All in all I'd say we handle this and fix the guide.

app/_src/kubernetes-ingress-controller/guides/services/tls.md
```diff
@@ -93,7 +94,7 @@ The Ingress API supports TLS termination using the `.spec.tls` field. To termina
     spec:
      ingressClassName: kong
      tls:
-     - secretName: demo-example-com
+     - secretName: demo-example-com-cert
```
pmalek, line 97: I don't see us defining an example `Secret` with a cert that can be used in this guide. WDYT about adding it so that this guide can be following step by step (include the certificate creation)?

app/_src/kubernetes-ingress-controller/guides/services/tls.md
```diff
@@ -106,13 +107,13 @@ The Ingress API supports TLS termination using the `.spec.tls` field. To termina
              service:
                name: echo
                port:
-                 number: 80
+                 number: 1027
```
pmalek, line 110: Can we make this `echo` deployment also part of the guide? Currently the guide expects users to have it defined in the cluster but doesn't instruct how to deploy it.

randmonkey: Added it in the guide.

app/_src/kubernetes-ingress-controller/guides/services/tls.md
randmonkey: @pmalek This should be fixed by using port 443 in the listener. The [determineListenersFromDataPlane](https://github.com/Kong/kubernetes-ingress-controller/blob/b6b1ff06a38b0289d5e9784c0a5891da63f1667a/internal/controllers/gateway/gateway_controller.go#L796) filters the listeners which satisfies:
- The port number is in the ports of proxy service
- The target port is listened by Kong gateways proxy
- The listening protocol on Kong gateway supports the protocols of the listener
So setting the port of the listener to `443` should fix.
I tried with the port 443 and see the traffic is forwarded.

randmonkey (approved):

tao12345666333:
@mheap If you have time could you take a look? Thank you!

app/_src/kubernetes-ingress-controller/guides/services/tls.md
```diff
@@ -20,29 +20,92 @@ For `HTTPRoute` or `GRPCRoute`, the route's `hostname` must match the listener h
...
+  - port: 1027
+    protocol: TCP
+    targetPort: 1027
+  selector:
+    app: echo
+  type: ClusterIP
+
+  ```
+
+For TLS termination, you need to configure a `Secret` for listener certificate on the `Gateway`. or for the certificate on `spec.tls` of the `Ingress`. This certificate will be used in setting up TLS connection between your client and {{site.base_gateway}}.
```
pmalek, line 82:
```suggestion
For TLS termination, you need to configure a `Secret` for listener certificate on the `Gateway` or for the certificate on `spec.tls` of the `Ingress`. This certificate will be used in setting up TLS connection between your client and {{site.base_gateway}}.
```

app/_src/kubernetes-ingress-controller/guides/services/tls.md
pmalek: I've tried to follow the TLS guide now (via https://deploy-preview-8598--kongdocs.netlify.app/kubernetes-ingress-controller/latest/guides/services/tls/) and the termination guide works for me but TLS passthrough does not. The `Gateway` still has the status: 

```
apiVersion: gateway.networking.k8s.io/v1
kind: Gateway
metadata:
  annotations:
    konghq.com/publish-service: kong/kong-gateway-proxy
  name: example-gateway-passthrough
  namespace: kong
spec:
  gatewayClassName: kong
  listeners:
  - allowedRoutes:
      namespaces:
        from: Same
    hostname: demo.example.com
    name: https
    port: 8899
    protocol: TLS
    tls:
      mode: Passthrough
status:
  addresses:
  - type: IPAddress
    value: 172.18.0.100
  conditions:
  - lastTransitionTime: "2025-04-22T11:38:15Z"
    message: this unmanaged gateway has been picked up by the controller and will
      be processed
    observedGeneration: 3
    reason: Accepted
    status: "True"
    type: Accepted
  - lastTransitionTime: "2025-04-22T11:38:15Z"
    message: ""
    observedGeneration: 3
    reason: Programmed
    status: "True"
    type: Programmed
  listeners:
  - attachedRoutes: 0
    conditions:
    - lastTransitionTime: "2025-04-22T11:38:15Z"
      message: no Kong listen with the requested protocol is configured
      observedGeneration: 3
      reason: UnsupportedProtocol
      status: "False"
      type: Accepted
    - lastTransitionTime: "2025-04-22T11:38:15Z"
      message: ""
      observedGeneration: 3
      reason: Invalid
      status: "False"
      type: Programmed
    name: https
    supportedKinds:
    - group: gateway.networking.k8s.io
      kind: TLSRoute
```

I've tried changing the listener port to 443 and that didn't help.

> I tried with the port 443 and see the traffic is forwarded.

Can you be more precise what have you changed @randmonkey? If the guide need a change to work I believe this should be applied in this PR.

app/_src/kubernetes-ingress-controller/guides/services/tls.md
```diff
@@ -55,13 +118,13 @@ The Ingress API supports TLS termination using the `.spec.tls` field. To termina
...
 
     ```yaml
     apiVersion: gateway.networking.k8s.io/{{ gwapi_version }}
     kind: HTTPRoute
     metadata:
-      name: demo-example 
+      name: demo-example
     spec:
       parentRefs:
       - name: example-gateway
```
pmalek, line 130: Can we add a `curl` command that will verify that the connectivity is achieved?

Something like:

```
curl -i -v --cacert example.com-tls.crt -H "Host: demo.example.com" https://$PROXY_IP/echo
```

(I for some reason couldn't make this work but have verified that `curl -i -v -k -H "Host: demo.example.com" https://$PROXY_IP/echo` works. This will need to be verified so that the provided CA actually works)

mheap:
I can take a look at this on Thursday

app/_src/kubernetes-ingress-controller/guides/services/tls.md
randmonkey: This requires configuring `stream_listen` for Kong gateways and enable `TLSRoute` on KIC by feature gates. I'll add the instructions to make it work.

randmonkey: Added the instructions in the PR.

app/_src/kubernetes-ingress-controller/guides/services/tls.md
randmonkey: Added instructions to both modes.

app/_src/kubernetes-ingress-controller/guides/services/tcp.md
```diff
@@ -186,12 +186,13 @@ receives on port 9000 to `echo` service on port 1025.
...
 
-1. Check if the Service is ready on the route.
+1. Check if the Service is ready on the route. Note that it may take a few moments for the Kong controller to process the route.
     {% capture the_code %}
 {% navtabs codeblock %}
 {% navtab Gateway API %}
 ```bash
-kubectl get tcproute echo-plaintext -ojsonpath='{.status.parents[0].conditions[?(@.reason=="Accepted")]}'
+# Wait for the controller to process the route
+kubectl wait tcproute echo-plaintext --for='jsonpath={.status.parents[0].conditions[?(@.reason=="Accepted")]}=True'
```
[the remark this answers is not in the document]
pmalek:
```suggestion
# Wait for the controller to process the route
kubectl wait tcproute echo-plaintext --for='jsonpath={.status.parents[0].conditions[?(@.type=="Accepted")].status}'=True
```

app/_src/kubernetes-ingress-controller/guides/services/tls.md
```diff
@@ -167,3 +339,24 @@ The Ingress API supports TLS termination using the `.spec.tls` field. To termina
...
+You need to let the domain `demo.example.com` resolving to the IP of the load balancer IP of {{ site.base_gateway }}'s proxy service by DNS or `/etc/hosts` file when verifying the connection. The IP can be fetched by: 
+
+```bash
+kubectl get svc -n kong kong-gateway-proxy -o jsonpath={.status.loadBalancer.ingress[0].ip}
+```
+
+Then, you can verify that the TLS passthrough is configured correctly (for example, by `openssl`'s TLS client):
+
+```bash
+ openssl s_client -connect demo.example.com:8899
```
pmalek, line 354:
```suggestion
To verify that the TLS passthrough is configured correctly (for example, by `openssl`'s TLS client) use the following commands:

```bash
export PROXY_IP=$(kubectl get svc -n kong kong-gateway-proxy -o jsonpath={.status.loadBalancer.ingress[0].ip})
openssl s_client -connect ${PROXY_IP}:8899 -servername demo.example.com
```

app/_src/kubernetes-ingress-controller/guides/services/tls.md
```diff
@@ -167,3 +339,24 @@ The Ingress API supports TLS termination using the `.spec.tls` field. To termina
 > The Ingress API does not support TLS passthrough
 {% endnavtab %}
 {% endnavtabs %}
+
+#### Verification
```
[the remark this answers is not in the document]
pmalek: Applied the suggestion in https://github.com/Kong/docs.konghq.com/pull/8598/commits/1832370078048add8e08c249e8826819cdb24894

pmalek (commented):
This looks good. I've left 2 comments that I think need some investigation:

- `TLSRoute` not accepted https://github.com/Kong/docs.konghq.com/pull/8598/files#r2058043889. This should be answered in this PR as we want full guide to work for users.
- gateway serving its own self generated cert instead of the provided `Gateway` listener cert: https://github.com/Kong/docs.konghq.com/pull/8598/files#r2057956856. This I believe could be answered in a separate issue or PR but is definitely puzzling.

app/_src/kubernetes-ingress-controller/guides/services/tcp.md
```diff
@@ -325,6 +330,9 @@ You can now access the `echo` service on port 9443 with SNI `tls9443.kong.exampl
 In real-world usage, you would create a DNS record for `tls9443.kong.example`pointing to your proxy Service's public IP address, which causes TLS clients to add SNI automatically. For this demo, add it manually using the OpenSSL CLI.
 
 ```bash
+# If you haven't set the PROXY_IP variable yet
+PROXY_IP=$(kubectl get svc -n kong kong-gateway-proxy -o jsonpath='{.status.loadBalancer.ingress[0].ip}')
+
 echo "hello" | openssl s_client -connect $PROXY_IP:9443 -servername tls9443.kong.example -quiet 2>/dev/null
```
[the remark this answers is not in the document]
randmonkey: The backendRef should use `1030` or the service's port that pointing to the `tlsecho` container's `1030` port. The `go-echo` image listens on:
- port 1025 for TCP (without SSL)
- port 1026 for UDP
- port 1027 for HTTP
- port 1030 for TLS.
I'll update the manifests here to introduce it with more details.

app/_src/kubernetes-ingress-controller/guides/services/tls.md
```diff
@@ -106,33 +169,142 @@ The Ingress API supports TLS termination using the `.spec.tls` field. To termina
...
 {% endnavtab %}
 {% endnavtabs %}
 
+#### Verification
+
+You can run the verification by using `curl`:
+
+```bash
+  export PROXY_IP=$(kubectl get svc -n kong kong-gateway-proxy -o jsonpath={.status.loadBalancer.ingress[0].ip})
+  curl  --cacert ./example.com-tls.crt -i -k -v -H"Host:demo.example.com"  https://{PROXY_IP}/echo
```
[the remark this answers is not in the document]
randmonkey: done.

randmonkey:
> This looks good. I've left 2 comments that I think need some investigation:
> 
> * `TLSRoute` not accepted https://github.com/Kong/docs.konghq.com/pull/8598/files#r2058043889. This should be answered in this PR as we want full guide to work for users.
> * gateway serving its own self generated cert instead of the provided `Gateway` listener cert: https://github.com/Kong/docs.konghq.com/pull/8598/files#r2057956856. This I believe could be answered in a separate issue or PR but is definitely puzzling.

- For TLSRoute: The `go-echo` listens on different ports for TCP(Without SSL) and TLS connections. TLS uses port 1030. So following guides in the `tcp.md` does not work.
- For gateway listener certificates: Looks like we should do something on both KIC/KO side and gateway side. Let's keep the `-k` option in the PR and investigate it later.

pmalek (commented):
Other than the last comment it all works for me.

I've created https://github.com/Kong/kubernetes-ingress-controller/issues/7376 to track the issue where Kong returns its own self signed cert rather than the one configured in `Gateway` listener configuration.

app/_src/kubernetes-ingress-controller/guides/services/tcp.md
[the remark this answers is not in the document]
pmalek: @tao12345666333 @randmonkey the service port listen above is still an issue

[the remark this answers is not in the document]
randmonkey: @pmalek updated the `example/echo-service.yaml` in assets. Please see if the service port is present now.

PR #8661, opened 2025-04-03
docs(plugin): only one plugin instance is executed for the current request

outsinre:
### Description

 

docs(plugin): only one instance of a plugin is executed for the current request

The multiple instances of the same plugin are active for the current request context, the very first instance in the order of precedence is executed, and the rest are abandoned.

https://konghq.atlassian.net/browse/FTI-6582

### Testing instructions

Preview link: 

### Checklist 

- [x] Review label added 

### DIFF
--- a/app/_includes/md/about-plugins.md
+++ b/app/_includes/md/about-plugins.md
@@ -2,7 +2,7 @@
 ## What are plugins?
 
 {{site.base_gateway}} is a Lua application designed to load and execute modules, which
-we commonly refer to as _plugins_. Plugins provide advanced functionality and extend the 
+we commonly refer to as _plugins_. Plugins provide advanced functionality and extend the
 use of the {{site.base_gateway}}, allowing you to add more features to your implementation.
 
 Kong provides a set of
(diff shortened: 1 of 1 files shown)


app/_includes/md/about-plugins.md
```diff
@@ -59,13 +59,15 @@ The complete order of precedence for plugins configured to multiple entities is:
 1. **Consumer**: Applies to all requests from a specific, authenticated consumer across all routes and services.
 1. **Consumer group**: Affects all routes and services for a designated group of authenticated users.
 1. **Route**: Specific to given route.
-1. **Service**: Specific to given service. 
+1. **Service**: Specific to given service.
 1. **Globally configured plugins**: Lowest precedence, applies to all requests across all services and routes regardless of consumer status.
 
+The very first matched instance is selected, and all other instances of the same plugin are ignored for the current request context. For example, if you configure the `rate-limiting-advanced` plugin for both the service *foo* and the route *bar* of that service, the instance associated with the route *bar* is executed due to the order of precedence. To work around this limitation, you can create two routes for the service *foo* and bind a separate `rate-limiting-advanced` plugin to each route. If you need to run the plugin twice, you can point the service *foo* to another route (*baz*) within the same {{site.base_gateway}}.
```
lena-larionova, line 65:
```suggestion
Only the first matched instance is selected, and all other instances of the same plugin are ignored for the current request context. For example, if you configure the `rate-limiting-advanced` plugin for both the service *foo* and the route *bar* of that service, the instance associated with the route *bar* is executed due to the order of precedence. 

To work around this limitation, you can create two routes for the service *foo* and bind a separate `rate-limiting-advanced` plugin to each route. If you need to run the plugin twice, you can point the service *foo* to another route (*baz*) within the same {{site.base_gateway}}.
```

